PR #535, opened 2014-07-08
replace `condition` in `conditionalPanel()` with `disable` and `hide`

r2evans:
I wanted to be able to keep an element within view (both for spacing and for the visual effect of it being immutable), so I added two optional arguments to `conditionalPanel()`: `disable` and `hide`.

`hide` is merely a reversal logically from `condition`. In fact, if you accept that having disable/hide as options is relevant, I suggest that deprecating `condition` might be appropriate.

`disable` was a little more challenging and admittedly not as "global" in its capabilities. Namely, it currently only affects text and input fields. For text, it just adds `.shiny-text-disabled` (a new class in shiny.css that lightens the font color). For input elements, it also disables the elements two ways (using `.ui-state-disabled` from jQuery ui and the `disabled` HTML5 property).

There might be a way to broaden this to other types of elements (e.g., images and plots), but I'm not certain how to do that with plots that shiny touches. My vision has an output plot graying out and not updating with input changes, but this would probably require a bit more JS <--> Shiny interaction than I'm able to tackle at the moment.

Tested using the kmeans-example on win7, R-3.1.0, chrome 35.0.1916.153 m. (Perhaps this time the travis-CI test will pass.)

### DIFF
--- a/R/bootstrap.R
+++ b/R/bootstrap.R
@@ -463,11 +463,12 @@ mainPanel <- function(..., width = 8) {
 #'     selectInput('xcol', 'X Variable', names(iris)),
 #'     selectInput('ycol', 'Y Variable', names(iris),
 #'                 selected=names(iris)[[2]]),
-#'     conditionalPanel(disable='input.xcol == "Sepal.Width"',
-#'                      hide='input.ycol == "Species"',
-#'                      numericInput('clusters', 'Cluster count', 3,
-#'                                   min = 1, max = 9)
-#'                      )
+#'     conditionalPanel(
+#'         disable='input.xcol == "Sepal.Width"',
+#'         hide='input.ycol == "Species"',
+#'         numericInput('clusters', 'Cluster count', 3,
+#'                      min = 1, max = 9)
+#'         )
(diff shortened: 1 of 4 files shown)


daattali:
Not sure if this helps, but my package [shinyjs](https://github.com/daattali/shinyjs) might be able to do what you're looking for - conditional disabling of input elements (not disabling panels, but just input elements). If this will be accepted into shiny then disregard my message, but until then hopefully this is useful.

Example: enable a submit button only when a name has been provided and a checkbox has been checked, otherwise disable the button.

```
runApp(shinyApp(
  ui = fluidPage(
    shinyjs::useShinyjs(),
    textInput("name", "Name", ""),
    checkboxInput("agree", "I agree to the terms"),
    actionButton("submit", "Submit")
  ),
    observe({
      shinyjs::toggleState("submit", nchar(input$name) > 0 && input$agree)
    })
  }
))
```

Hope this helps

r2evans:
I had seen it a little bit ago and intend to look at it next opportunity I get. Thanks Dean!

PR #688, opened 2015-01-14
Specifying destination file for runURL, runGist, and runGitHub

xbhsu:
Currently, when users run an application from a URL, gist, or GitHub repository, the source files are downloaded to a temporary file and immediately removed when the application is exited. It'd be useful to change this behavior and download the source files to a destination directory so that users can make modifications locally.

This pull request adds options to runUrl, runGist, and runGitHub so that users can save the downloaded application source files to a specific destination directory. This is done through the additional option destDir = ... in runUrl, runGist, and runGitHub. Default is NULL and gives the original behavior.

### DIFF
--- a/R/run-url.R
+++ b/R/run-url.R
@@ -25,7 +25,7 @@
 #' runUrl("https://github.com/rstudio/shiny_example/archive/master.zip",
 #'  subdir = "inst/shinyapp/")
 #' }
-runUrl <- function(url, filetype = NULL, subdir = NULL, ...) {
+runUrl <- function(url, filetype = NULL, subdir = NULL, destFile = NULL, ...) {
 
   if (!is.null(subdir) && ".." %in% strsplit(subdir, '/')[[1]])
     stop("'..' not allowed in subdir")
@@ -43,8 +43,14 @@ runUrl <- function(url, filetype = NULL, subdir = NULL, ...) {
     stop("Unknown file extension.")
 
   message("Downloading ", url)
-  filePath <- tempfile('shinyapp', fileext=fileext)
-  fileDir  <- tempfile('shinyapp')
+  if (is.null(destFile)) {
+    filePath <- tempfile('shinyapp', fileext = fileext)
+    fileDir  <- tempfile('shinyapp')
+  } else {
+    fileDir <- destFile
+    filePath <- paste(destFile, fileext)
+  }
+  message("App data files in ", fileDir)
   dir.create(fileDir, showWarnings = FALSE)
   if (download(url, filePath, mode = "wb", quiet = TRUE) != 0)
     stop("Failed to download URL ", url)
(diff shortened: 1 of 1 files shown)


R/run-url.R
```diff
@@ -64,7 +71,7 @@ runUrl <- function(url, filetype = NULL, subdir = NULL, ...) {
     first <- as.character(unzip(filePath, list=TRUE)$Name)[1]
     unzip(filePath, exdir = fileDir)
   }
-  on.exit(unlink(fileDir, recursive = TRUE), add = TRUE)
+  on.exit(unlink(fileDir, recursive = is.null(destDir)), add = TRUE)
```
jcheng5: Is there a reason you did it like that and not like this?

```r
if (is.null(destDir)) {
  on.exit(unlink(fileDir, recursive = TRUE), add = TRUE)
}
```

xbhsu:
I was thinking of having destDir control whether or not folders were deleted recursively, but you're right, this doesn't make much sense and is harder to read. I've edited it so that destDir is now explicitly checked. Thanks for the input!

yihui:
This PR looks good to me. I guess you need to send the contributor agreement: https://github.com/rstudio/shiny/blob/master/CONTRIBUTING.md Thanks!

xbhsu:
Glad to be of some help! I've sent the form to JJ.

R/run-url.R
```diff
@@ -14,6 +14,7 @@
 #' @param subdir A subdirectory in the repository that contains the app. By
 #'   default, this function will run an app from the top level of the repo, but
 #'   you can use a path such as `\code{"inst/shinyapp"}.
+#' @param destDir Directory to store the downloaded application files
```
wch: It would be helpful to add, 

"If \code{NULL} (the default), the application files will be stored in a temporary directory and removed when the app exits."

R/run-url.R
```diff
@@ -43,8 +44,14 @@ runUrl <- function(url, filetype = NULL, subdir = NULL, ...) {
...
-  filePath <- tempfile('shinyapp', fileext=fileext)
-  fileDir  <- tempfile('shinyapp')
+  if (is.null(destDir)) {
+    filePath <- tempfile('shinyapp', fileext = fileext)
+    fileDir  <- tempfile('shinyapp')
+  } else {
+    fileDir <- destDir
+    filePath <- paste(destDir, fileext)
+  }
+  message("App data files in ", fileDir)
```
wch: I don't think it's necessary to have a `message()` here, since the user presumably would know that they've specified a destination directory.

R/run-url.R
```diff
@@ -25,7 +26,7 @@
 #' runUrl("https://github.com/rstudio/shiny_example/archive/master.zip",
 #'  subdir = "inst/shinyapp/")
 #' }
-runUrl <- function(url, filetype = NULL, subdir = NULL, ...) {
+runUrl <- function(url, filetype = NULL, subdir = NULL, destDir = NULL, ...) {
```
wch: For consistency with the other arguments, I'd please use `destdir` instead of `destDir`.

xbhsu:
Thanks for taking a look at this @wch! I've fixed the formatting issues, removed the extraneous ```message``` and updated the documentation

wch:
Excellent, thanks!

PR #694, opened 2015-01-21
Allow DOM element replacement in htmlOutput

wch:
This PR adds a `replace` option to `htmlOutput()`. When set to TRUE, it tells Shiny that the output from `renderUI` should replace the container DOM element, instead of being placed inside of the container. This is useful in cases where an intervening `<div>` or `<span>` element isn't allowed, or will interfere with CSS selectors.

When replacement happens, the new element gets the old element's ID, and also gets assigned the classes ` "shiny-html-output shiny-html-replace-output"`. If `renderUI` returns something other than a single tag (like a `tagList` or bare text), it's not possible to replace the container element with the new content, preserving ID and and keeping those classes; in this case, the content will be placed inside the container element, as though `replace=FALSE`.

This example doesn't show the proper styling for the `li` (red text), because of the extra `div` in the DOM structure:

```R
shinyApp(
  ui = bootstrapPage(
    tags$head(tags$style(HTML("ul > li { color: red; }"))),
    tags$ul(
      tags$li("static item"),
      uiOutput("item")
    )
  ),
    output$item <- renderUI(tags$li("dynamic item"))
  }
)
```

With `replace=TRUE`, the `li` comes up with the correct styling:

```R
shinyApp(
  ui = bootstrapPage(
    tags$head(tags$style(HTML("ul > li { color: red; }"))),
    tags$ul(
      tags$li("static item"),
      uiOutput("item", replace = TRUE)
    )
  ),
    output$item <- renderUI(tags$li("dynamic item"))
  }
)
```

This addresses rstudio/shinydashboard#1.


wch:
Closing because:

* Items updated with this code can only be updated once.
* To allow further updates, it requires changes to the output binding behavior.
* Changing the binding behavior is too risky for the relatively small benefit.

Issue #675, opened 2014-12-14
Bug (?) in version 0.10.2.9006 - odd alignment of checkboxGroupInput


smartinsightsfromdata:
My UI.r has the following:
        fluidRow(
          checkboxGroupInput("monthIdppu", "Spread value(s) over the following months:", month.abb[1:12]
                             , selected = c("Jan","Apr") , inline = TRUE)
          ),
Normally as it is a fluidRow the checkboxGroupInput presents itself as a one single line of tick boxes.

With 0.10.2.9006 the checkboxGroupInput extends only to the extent of the label ("Spread value(s) over the following months:"), then breaks in 3 other lines, as in
Column(x, checkboxGroupInput
With x = length("Spread value(s) over the following months:").

gjoneski:
Hi Winston.

One issue we've come across is: due to this hard-coded container width, selectInput, for example, no longer behaves the same with respect to its `width` parameter.  In particular, passing width='100%', expands the input to the limits of the hard-coded container width.  This is a change in behavior against bootstrap 2, when this would have been expanded within the limits of the next enclosing element - often something like:

column(width=x,selectInput(,...width='100%'))

At any rate, just wanted to give you a heads up about this - if nothing can be done within shiny >=0.11, we are aware of the shinybootstrap2 alternative.

Thanks.

wch:
@gjoneski, I just pushed a fix in the dev version. Could you try it out and let me know if it behaves as expected?

gjoneski:
Thanks for looking into this Winston.

That does work for sliderInput.  Hopefully all of [these guys](https://github.com/rstudio/shiny/commit/7b3e9ab0315353ab8e25ec83319d5399941cc46e)  will get that facelift as well.

Thanks again.

wch:
It's a little tricky since most of those don't offer a `width` option to begin with. Perhaps it would be better to add some custom css to your app?

gjoneski:
Sure - makes sense, we've already done this.

I do think select(ize)Input, as widely used as it is, should probably respect its `width` parameter.

At any rate, thanks for looking into this.

PR #702, opened 2015-01-26
Allow control over selectize width

wch:
Fixes one of the issues raised in the comments of #675.

Previously, you could set the width of a selectInput with a pixel value like `"500px"`, but it wouldn't respect percent widths like `"100%"`. This commit makes `width` control the wrapper container's width, instead of setting the `data-width`, so it will respect percent values.

This commit also allows control over non-selectize select inputs.

All of these work now:

```R
library(shiny)
shinyApp(
  ui = fluidPage(
    selectInput("n", "default", c(5, 10, 100), selectize = FALSE),
    selectInput("n2", "100%", c(5, 10, 100), selectize = FALSE, width="100%"),
    selectInput("n3", "500px", c(5, 10, 100), selectize = FALSE, width="500px"),

    selectInput("n4", "default", c(5, 10, 100)),
    selectInput("n5", "100%", c(5, 10, 100), width="100%"),
    selectInput("n6", "500px", c(5, 10, 100), width="500px"),

    selectizeInput("n7", "default", c(5, 10, 100)),
    selectizeInput("n8", "100%", c(5, 10, 100), width="100%"),
    selectizeInput("n9", "500px", c(5, 10, 100), width="500px"),
    plotOutput("plot")
  ),
    output$plot <- renderPlot( plot(head(cars, as.numeric(input$n))) )
  }
)
```

@yihui Do you see potential pitfalls with doing it this way? It seems to me that the reason the previous version used `data-width` was because it didn't have a wrapper div around the select.

### DIFF
--- a/R/bootstrap.R
+++ b/R/bootstrap.R
@@ -800,13 +800,14 @@ selectInput <- function(inputId, label, choices, selected = NULL,
   # return label and select tag
   res <- div(
     class = "form-group shiny-input-container",
+    style = if (!is.null(width)) paste0("width: ", validateCssUnit(width)),
     controlLabel(inputId, label),
     div(selectTag)
   )
 
   if (!selectize) return(res)
 
-  selectizeIt(inputId, res, NULL, width, nonempty = !multiple && !("" %in% choices))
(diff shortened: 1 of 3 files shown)


yihui:
I do not see potential pitfalls. Thanks!

PR #704, opened 2015-01-27
htmlOutput: allow custom container function

wch:
This PR allows using a custom container for `uiOutput`, instead of just `div` or `span`. It also allows adding custom classes to the container.

This is for rstudio/shinydashboard#1.

### DIFF
--- a/R/bootstrap.R
+++ b/R/bootstrap.R
@@ -1877,6 +1877,9 @@ dataTableOutput <- function(outputId) {
 htmlOutput <- function(outputId, inline = FALSE,
   container = if (inline) span else div, ...)
 {
+  if (anyUnnamed(list(...))) {
+    warning("Unnamed elements in ... will be replaced with dynamic UI.")
+  }
   container(id = outputId, class="shiny-html-output", ...)
 }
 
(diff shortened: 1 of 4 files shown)


R/bootstrap.R
```diff
@@ -1858,18 +1858,26 @@ dataTableOutput <- function(outputId) {
...
+#'
+#' # Using a custom container and class
+#' tags$ul(
+#'   htmlOutput("summary", container = tags$li, class = "custom-li-output")
+#' )
 #' @export
-htmlOutput <- function(outputId, inline = FALSE) {
-  container <- if (inline) span else div
-  container(id = outputId, class="shiny-html-output")
+htmlOutput <- function(outputId, inline = FALSE,
```
jcheng5: If we really wanted to have the arguments be `id, container, inline` (both for consistency and because we think container is more likely?) then we could just do that and treat `container=TRUE` as `container=span` and `container=FALSE` as `container=div`.

I'm also 100% fine with taking this PR as-is. This suggestion is only if it bothers *you* that the order is different.

jcheng5: Should it be a warning if you pass unnamed arguments to `...`? (meaning contents that will be wiped out when a value is rendered)

I could see people using that to display loading messages. I'm not sure if that's how we want to solve that problem though.

wch:
Good idea with checking for unnamed args in `...`. I've added it.

Issue #709, opened 2015-01-28
[fixedPanel] match.arg(cursor) : 'arg' must be of length 1


unikum:
Hi.

```r
> absolutePanel()
<div style="position:absolute;cursor:inherit;"></div> 
> fixedPanel()
Error in match.arg(cursor) : 'arg' must be of length 1
Calls: fixedPanel -> absolutePanel -> match.arg
> formals(fixedPanel)$cursor
c("move", "default", "inherit")
> formals(absolutePanel)$cursor
c("auto", "move", "default", "inherit")
```

jcheng5:
Thanks! Should be merged soon:

https://github.com/rstudio/shiny/pull/710

On Tue, Jan 27, 2015 at 6:26 PM, Artem A. Klevtsov <notifications@github.com

PR #710, opened 2015-01-28
Fix fixedPanel default cursor handling

jcheng5:
Previous to this fix, calling fixedPanel without an
explicit cursor argument would cause an error due to
the vector of possible values being passed as the
actual cursor argument to absolutePanel.


wch:
Looks good to me.

wch:
Closes #709.

PR #715, opened 2015-02-01
downloadHandler temp files should have same extension as final download

jcheng5:


wch:
Looks good to me. I think tthe `parseQueryString.rd` is probably left over from a documentation change (563b986591d4fd498f9f53f0978ad6f2b5e4599a) that previously wasn't re-run through roxygen, though.

PR #721, opened 2015-02-03
Bugfix/conditional events

wch:
Fixes #717 and fixes #693.

@jcheng5 This is without `$updateConditionals` being debounced.

PR #723, opened 2015-02-03
Fix oversubscription of shown/hidden events

jcheng5:
This is an alternative to #721. That fix allows sendImageSize[0] to be called many times but debounces the many calls into a single one that is invoked on the next tick. This fix causes the number of times sendImageSize is called to be reduced, by fixing an overly-inclusive set of event subscriptions and not firing show/shown events for conditional panels that are already showing (or hide/hidden for ones that are already hidden). The advantage of this fix is that it preserves the possibility of using a single websocket message to transmit the fact that a value has changed plus the changes to shown/hidden state and sizes; in some cases this might save some unnecessary calculation on the server side.

We are still paying for sendImageSize to be called 10 times if 10 conditionalPanels change state for one call of $updateConditionals; we could avoid this by suspending the sendImageSize calls until all the conditionalPanels are updated.

[0] *Everything in this description for sendImageSize also applies to sendOutputHiddenState*


inst/www/shared/shiny.js
```diff
@@ -3296,6 +3299,14 @@
         sendOutputHiddenStateDebouncer.immediateCall();
     });
 
+    function filterEventsByNamespace(namespace, handler) {
+      return function(e) {
+        if (e.namespace === namespace) {
```
wch: Does this exact comparison do the right thing? I would have expected a substring match to be needed, splitting on `.`. I have a little test in a  jsfiddle here: http://jsfiddle.net/winstonchang/b4n8ccyc/

jcheng5: No, it doesn't. Sorry, I must have pushed the wrong commit as I was experimenting.

inst/www/shared/shiny.js
```diff
@@ -3296,6 +3299,14 @@
         sendOutputHiddenStateDebouncer.immediateCall();
     });
 
+    function filterEventsByNamespace(namespace, handler) {
+      return function(e) {
+        if (e.namespace.indexOf(namespace) === 0) {
```
wch: I think this comparison isn't quite right. It happens to work for the `bs` events, but I don't think it will work in general because the components of the namespace can be re-ordered. From the jQuery [event docs](http://learn.jquery.com/events/event-extensions/):

> When multiple namespaces are given, they are separated by periods and sorted in ascending alphabetical order.

wch:
I might be missing something really basic, but wouldn't it be more efficient if `sendImageSize` and `sendOutputHiddenState` were scoped to the element, instead of searching over the entire document?

For example, suppose my `<li>` gets the event `shown.bs.tab`, right now that ends up invoking `sendImageSize`, which searches for all images and plots in the document, but it seems to me that it could just search for those things inside the `<li>`. In the event handler, it's accessible via `e.currentTarget`.

jcheng5:
In the general case a "shown" event can cause resizing all over the document. Certainly a child being hidden can affect the size of its ancestors, and thus the size of its siblings. Now the same might not be true for sendOutputHiddenState so I guess it's possible we could scope that one.

jcheng5:
BTW I noticed that sendOutputHiddenState is debounced, but there's also a `lastChanceCallback` array on inputBatchSender that avoids the downside of debouncing (moving the operation to a different websocket frame). That's a really nice solution and I don't know why I didn't also do that for sendImageSize, other than maybe I thought sendImageSize was really cheap. So we could add that fix to this PR as well, if we wanted to.

wch:
That makes sense about the resizing all over the place.

I think it would be worthwhile to debounce `sendImageSize` with that `lastChanceCallback`. In the radiant app, I see it getting called a few dozen times when switching to Regression->OLS for the first time, and about a dozen times when switching to again.

Guiui21:
Hi,
I'm trying to use shiny to create google charts like the one describe here: http://shiny.rstudio.com/gallery/google-charts.html. My problem is that i'm using RmySQL to retrieve my data from my server and fill the chart with those data. I works fine on R but when i try to plug my chart into a webpage (by using shinyapps.io) it doesn't work. I think that's because the app can not connect to my server.
What do you think ? Do you have any solution to my problem ?
Thank you,

Guillaume

jcheng5:
@Guiui21 Please try the [shiny-discuss group](https://groups.google.com/group/shiny-discuss) for general Shiny questions. Thanks.

Guiui21:
Thank you, i just did.

PR #728, opened 2015-02-05
Fix #725: Tabs with icons give incorrect values

### LINKED ISSUE #725: Conditional panels and tabs with icons
Hello,
when I am using names of tabs in conditions of conditional panels and the tabs have icons, condition is never solved as true. Is there any way how to solve this problem?
I'm using Shiny 0.10.2.2 and fontawsome icons.

Thank you in advance!

jcheng5:
When no explicit value argument was present on tabPanel,
the inner text of the HTML anchor was used. This gave
weird results when an icon was used, since it appears in
the anchor.

The fix is to always use the value parameter, defaulting
it to the title.

### DIFF
--- a/R/bootstrap.R
+++ b/R/bootstrap.R
@@ -1404,7 +1404,7 @@ dateRangeInput <- function(inputId, label, start = NULL, end = NULL,
 #'   )
 #' )
 #' @export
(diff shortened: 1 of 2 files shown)


wch:
Looks good to me.

PR #729, opened 2015-02-05
selectInput: add control for size (height)

wch:
This addresses https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/shiny-discuss/-gFgWdMLkvg/MM1UdM6R9kkJ

This adds control over the height of select inputs -- for both multiple and single inputs. It's not surprising that we haven't supported control over the height of single-select inputs, but I'm surprised we've never had the ability to control the height of multiple select boxes before.


jcheng5:
LGTM

PR #747, opened 2015-03-01
byte serving capability in shiny:::staticHandler

Przemol:
Make use of HTTP_RANGE header to byte serve files defined with `shiny::addResourcePath`.
Only using defined range, e.g. `bytes=500-999`, is implemented.
See more: http://tools.ietf.org/html/rfc2616#section-14.35

I need byte serving to use Biodalliance (https://github.com/dasmoth/dalliance) with my Shiny apps.
Biodalliance uses bite serving to fetch big binary indexed files, like BigWig (http://genome.ucsc.edu/goldenpath/help/bigWig.html).
Byte serving in Shiny might be a useful feature for community. It making Shiny file serving compatible with an array of client applications.

### DIFF
--- a/R/middleware.R
+++ b/R/middleware.R
@@ -205,7 +205,19 @@ staticHandler <- function(root) {
 
     ext <- tools::file_ext(abs.path)
     content.type <- getContentType(ext)
-    response.content <- readBin(abs.path, 'raw', n=file.info(abs.path)$size)
+        if(length(req$HTTP_RANGE) && grepl('^bytes=.+', req$HTTP_RANGE)) {
+        rng <- as.integer(
+            strsplit(gsub('^bytes=', '', req$HTTP_RANGE), '-')[[1]]
+        )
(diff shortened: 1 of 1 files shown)


bwlewis:
I would like to see something like this merged if possible. I also need HTTP_RANGE, in my case for serving mp4 files. Most client browsers use byte ranges for such files.

MarWoes:
I would like to see this feature as well, I'm serving `.bam` files. You might however want to consider returning a `206` status for `HTTP_RANGE` requests.

MarWoes:
Is there any update on this? This feature would be very convenient for one of our apps that serves BAM files.

sounkou-bioinfo:
> Is there any update on this? This feature would be very convenient for one of our apps that serves BAM files.
@MarWoes 
FWIW, facing the same issue i made this package to serve go around this limitation of httpuv https://sounkou-bioinfo.r-universe.dev/goserveR (only work on unix)

PR #753, opened 2015-03-04
Fix downloading of files with no extension. Fixes #575

### LINKED ISSUE #575: Download failure due to MIME lookup
Reported here: https://groups.google.com/d/msgid/shiny-discuss/929074a7-9903-4373-bddf-a65062f5ce8e%40googlegroups.com?utm_medium=email&utm_source=footer

A minimal reproducible example:

```r
x = 1:10
library(shiny)
runApp(list(
  ui = fluidPage(
    downloadButton('try', 'Save')
  ),
    output$try <- downloadHandler(
      filename = 'foo',
      content = function(file) {
        save(x, file = file)
      }
    )
  }
))
```

Shiny may use my **mime** package, although it is not necessary if we only want to solve this particular issue.

wch:
Previously, attempting to download a file with this example resulted in an error.

```R
x = 1:10
library(shiny)
runApp(list(
  ui = fluidPage(
    downloadButton('try', 'Save')
  ),
    output$try <- downloadHandler(
      filename = 'foo',
      content = function(file) {
        save(x, file = file)
      }
    )
  }
))
```

The error was:

```
Error in knownContentTypes$get(tolower(ext)) : 
  attempt to use zero-length variable name
```

The cause was that the file had no extension, and `knownContentTypes` was a `Map` object, where it wasn't valid to access a member with the name `""`.

In this PR, I just changed it from a `Map` to a `list`, which simplifies the object and makes this problem go away because it's valid to access `""` in a list.

@jcheng5 if there's a reason to keep it as a `Map`, please let me know.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -290,8 +290,7 @@ download <- function(url, ...) {
   }
 }
 
-knownContentTypes <- Map$new()
-knownContentTypes$mset(
+knownContentTypes <- list(
   html='text/html; charset=UTF-8',
   htm='text/html; charset=UTF-8',
   js='text/javascript',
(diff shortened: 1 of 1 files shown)


yihui:
I was hoping to just remove `knownContentTypes` from shiny and use `mime::guess_type()` instead.

PR #756, opened 2015-03-05
Add shiny.port option

wch:
This adds an option, `shiny.port`, so that the user can set the default port to use. This makes it easier to run apps and examples on a predetermined port, especially when a `shiny.appobj` object is used.


yihui:
I think it is a good idea.

jcheng5:
Does it help that a shiny.appobj can itself request a particular port?

`shinyApp(ui, server, options = list(port = 8101))`

wch:
A little bit, but it's still a little annoying to have to add that every time you want to run a shiny.appobj. We make the host IP configurable via an option, and it makes sense to me to also make the port configurable.

PR #759, opened 2015-03-06
Fix checkbox input binding (closes #206)

### LINKED ISSUE #206: checkboxes reactive after Shiny.unbindAll()
checkboxes don't automatically get assigned shiny-bound-input class and don't seem to have a binding: .data('shiny-input-binding')

I haven't looked deeper into the code, but the result is that calling Shiny.unbindAll() still reacts to changes for checkboxes

wch:
This fixes the bug introduced by #754.

The root of the problem was that the `checkboxInputBinding` didn't have a `subscribe` method; the reason they worked (prior to the merging of #754) was because they were found by the multiInput code.

This app now works:

```R
library(shiny)
shinyApp(
  ui = fluidPage(checkboxInput('foo', 'Option')),
    observe({
      print(input$foo)
    })
  })
)
```


wch:
@yihui I just realized that this might break the row checkboxes that you added in DT.

During our call, I wondered how those checkboxes worked. I think they're added to the DOM without a subsequent call to `bindAll()`, and I was surprised that the values would get sent to the server without `bindAll()`. Now I realize that checkbox inputs were never bound via `bindAll()` -- originally, the `bindMultiInput` function attached event listeners to the *document*, not the checkboxes, to listen for changes to checkbox values:
https://github.com/rstudio/shiny/blob/a0c6feb386ad0279420ec1dfabb15edcb75f9dda/inst/www/shared/shiny.js#L3086-L3104

Now that checkboxes behave like all other inputs, with the event listener bound to the input itself by `bindAll()` (rather than being bound to the document when the page is loaded), this may cause problems to your DT row checkbox implementation. But I do think this is the correct way for checkboxes to work.

wch:
I'm going to merge this because it fixes checkbox inputs; it's certainly no worse than prior to this commit (though we can discuss going back to the original behavior if needed).

yihui:
That makes sense. Thanks!

PR #769, opened 2015-03-17
Add more interactions to images

wch:
This PR adds support for double-clicking and brushing to plotOutputs. It also adds all four types of mouse interactions (clicking, double-clicking, hovering, brushing) to imageOutputs.

There are a couple more features that would be useful in the future, but I'd like to get this branch merged first because it's already huge. The other features are:

* Handles for resizing brush
* ggplot2 support

Demo app: https://winston.shinyapps.io/interactive_graphic/
Source: https://gist.github.com/wch/bf20e5b86f06b1d968e2


timelyportfolio:
wow, very nice work

wch:
Thanks, I think I set a new personal record for largest pull request with this one. :)

timelyportfolio:
Had the thought that maybe color-picking might also be a nice additional behavior, such as in http://timelyportfolio.github.io/rCharts_color_thief/index_jsfeat.html.  Here is the little bit of code
https://github.com/timelyportfolio/rCharts_color_thief/blob/gh-pages/index_jsfeat.html#L211-L226
 from http://stackoverflow.com/questions/6735470/get-pixel-color-from-canvas-on-mouseover.

R/bootstrap.R
```diff
@@ -1695,90 +1695,342 @@ verbatimTextOutput <- function(outputId) {
...
+#'   (the default), a string, or an object created by the
+#'   \code{\link{hoverOpts}} function. If you use a value like
+#'   \code{"plot_hover"} (or equivalently, \code{hoverOpts(id="plot_hover")}),
+#'   the plot will send coordinates to the server pauses on the plot, and the
+#'   value will be accessible via \code{input$plot_hover}. The value will be a
+#'   named list with \code{x} and \code{y} elements indicating the mouse
+#'   position. To control the hover time or hover delay type, you must use
+#'   \code{\link{hoverOpts}}.
+#' @param clickId Deprecated; use \code{click} instead. Also see the
+#'   \code{\link{hoverOpts}} function.
```
jcheng5: Should the see-also be to `clickOpts`?

R/bootstrap.R
```diff
@@ -1695,90 +1695,342 @@ verbatimTextOutput <- function(outputId) {
...
+    ))
+  }
+
+  if (!is.null(brush)) {
+    if (is.character(brush)) {
+      brush <- brushOpts(id = brush)
+    }
+
+    args <- c(args, list(
+      `data-brush-id` = brush$id,
```
jcheng5: Rather than manually manage all these mappings between the opts objects and these lists, what if the opts functions themselves created these lists? If you don't want them to be called `data-brush-*` initially, at least `list(delay-type=delayType)` and then just munge the names here?

If you do want the "native" opts object to look different than these lists here, then maybe have the code that do the translating between the former and the latter live in dedicated (non-exported) functions right next to the opt functions in the source code.

R/image-interact-opts.R
```diff
@@ -0,0 +1,126 @@
...
+}
+
+
+#' Create an object representing double-click options
+#'
+#' This generates an object representing dobule-click options, to be passed as
+#' the \code{dblclick} argument of \code{\link{imageOutput}} or
+#' \code{\link{plotOutput}}.
+#'
+#' @inheritParams clickOpts
```
jcheng5: Do you need to override the `@param id` to refer to dblclick instead of click?

R/image-interact-opts.R
```diff
@@ -0,0 +1,126 @@
...
+#' @param direction The direction for brushing. If \code{"xy"}, the brush can be
+#'   drawn and moved in both x and y directions. If \code{"x"}, or \code{"y"},
+#'   the brush wil work horizontally or vertically.
+#' @param resetOnNew When a new image is sent to the browser (via
+#'   \code{\link{renderImage}}), should the brush be reset? The default,
+#'   \code{FALSE}, is useful if you want to update the plot while keeping the
+#'   brush. Using \code{TRUE} is useful if you want to clear the brush whenever
+#'   the plot is updated.
+#' @export
+brushOpts <- function(id = NULL, color = "#666", outline = "#000",
```
jcheng5: Maybe "stroke" instead of "outline"? Maaaybe "fill" instead of "color"?

Allow independent opacity for stroke vs. fill?

R/shinywrappers.R
```diff
@@ -140,20 +141,20 @@ renderPlot <- function(expr, width='auto', height='auto', res=72, ...,
...
           top = usrCoords[4]
         ),
         # The bounds of the plot area, in DOM pixels
-        bounds = c(
-          left = grconvertX(usrBounds[1], 'user', 'nfc') * width,
-          right = grconvertX(usrBounds[2], 'user', 'nfc') * width,
-          bottom = (1-grconvertY(usrBounds[3], 'user', 'nfc')) * height,
-          top = (1-grconvertY(usrBounds[4], 'user', 'nfc')) * height
+        bounds = list(
+          left = grconvertX(usrBounds[1], 'user', 'nfc') * width - 1,
```
jcheng5: Have we had an off-by-1 bug all this time?

R/shinywrappers.R
```diff
@@ -140,20 +141,20 @@ renderPlot <- function(expr, width='auto', height='auto', res=72, ...,
       }
 
       coordmap <<- list(
-        usr = c(
+        usr = list(
```
jcheng5: I assume this is the jsonlite issue?

inst/www/shared/shiny.js
```diff
@@ -1260,106 +1279,801 @@
...
+        return;
+      }
 
-        img = document.createElement('img');
-        // Copy items from data to img. This should include 'src'
-        $.each(data, function(key, value) {
-          if (value !== null)
-            img[key] = value;
-        });
+      function strToBool(str) {
```
jcheng5: Seems like this should be hoisted up to the top level

inst/www/shared/shiny.js
```diff
@@ -1260,106 +1279,801 @@
...
+      });
+
+      var $img = $(img);
+
+      // Firefox doesn't have offsetX/Y, so we need to use an alternate
+      // method of calculation for it. Even though other browsers do have
+      // offsetX/Y, we need to calculate relative to $el, because sometimes the
+      // mouse event can come with offset relative to other elements on the
+      // page. This happens when the event listener is bound to, say, window.
+      function mouseOffset(mouseEvent) {
```
jcheng5: That's nice

R/bootstrap.R
wch: Yup, good catch.

R/bootstrap.R
wch: How about creating S3 methods that translate from the option object to a list?

R/image-interact-opts.R
wch: Sure, I'll add all the params individually instead of inheriting them.  `clip` could also use a little editing.

R/image-interact-opts.R
wch: Sure, stroke and fill sound good - it's also consistent with ggvis terminology.

As for independent opacity of stroke, it's a little tricky since it requires using `rgba()` syntax, and this will require translating from color names to RGB values. Also, IE8 apparently doesn't support `rgba()`, so we'll have to add a workaround for that.

R/shinywrappers.R
wch: Yup, working on a patch for this in jsonlite now.

R/shinywrappers.R
wch: I think so - I believe R gives pixel coords starting from 1, while in JS the positions start from 0. It is hard to be really precise testing this, but it seems more accurate with the `- 1`.

inst/www/shared/shiny.js
wch: OK, sounds good.

inst/www/shared/shiny.js
```diff
@@ -1260,106 +1279,801 @@
...
+          // mousedown.
+          if (brush.brushing || brush.dragging) return;
+
+          // Listen for left mouse button only
+          if (e.which !== 1) return;
+
+          var offset = mouseOffset(e);
+
+          // Ignore mousedown events outside of plotting region
+          if (opts.brushClip && !isInPlottingRegion(offset)) return;
```
jcheng5: If this is doing what I think it is, I think it would be better if the offset was moved into the plotting region (instead of flat out ignored). So if you start a brushing operation above and to the left of the top-left corner, the brush would start at the top-left corner.

inst/www/shared/shiny.js
```diff
@@ -1260,106 +1279,801 @@
...
+
+          // Because the x and y directions of the pixel space may differ from
+          // the x and y directions of the data space, we need to recalculate
+          // the min and max.
+          var coords = {
+            xmin: Math.min(min.x, max.x),
+            xmax: Math.max(min.x, max.x),
+            ymin: Math.min(min.y, max.y),
+            ymax: Math.max(min.y, max.y),
+            '.nonce': Math.random()
```
jcheng5: Maybe brush shouldn't get a nonce? This doesn't feel "eventy" to me, whereas click/dblclick and arguably hover do.

inst/www/shared/shiny.js
wch: Hm, I'll have to try it out to see if that works, or if it seems weird. It might be that allowing clicks in a, say, 5-pixel boundary around the plotting region would seem more natural.

inst/www/shared/shiny.js
wch: That makes sense.

inst/www/shared/shiny.js
```diff
@@ -1260,106 +1279,801 @@
...
+
+        // If there's a dblclick listener, make sure not to count this as a
+        // click on the first mousedown; we need to wait for the dblclick
+        // delay before we can be sure this click was a single-click.
+        clickCount++;
+        if (clickCount === 1) {
+          clickTimer = setTimeout(function() {
+            e2 = $.Event('mousedown2', eInfo);
+            $el.trigger(e2);
+            clickCount = 0;
```
jcheng5: I feel like the `clickCount = 0` should be two lines above, just because `$el.trigger(e2)` could execute unbounded amounts of event-handling code.

inst/www/shared/shiny.js
```diff
@@ -1260,106 +1279,801 @@
...
+          }, opts.dblclickDelay);
+
+        } else {
+          // If second click happens within specified delay, trigger our
+          // custom 'dblclick2' event.
+          clearTimeout(clickTimer);
+          e2 = $.Event('dblclick2', eInfo);
+
+          $el.trigger(e2);
+          clickCount = 0;
```
jcheng5: Same for this `clickCount`.

inst/www/shared/shiny.js
```diff
@@ -1260,106 +1279,801 @@
...
+          clickTimer = setTimeout(function() {
+            e2 = $.Event('mousedown2', eInfo);
+            $el.trigger(e2);
+            clickCount = 0;
+
+          }, opts.dblclickDelay);
+
+        } else {
+          // If second click happens within specified delay, trigger our
+          // custom 'dblclick2' event.
```
jcheng5: I would make sure that the mouse has not moved (or not beyond some very small number of pixels). If the user is quickly clicking lots of distinct data points you certainly don't want to interpret that as double-click.

inst/www/shared/shiny.js
```diff
@@ -1260,106 +1279,801 @@
...
+
+            if (this.$div)
+              this.$div.remove();
+
+            return this;
+          },
+
+          // If there's an existing brush div, use that div to set the new
+          // brush's settings.
+          importOldBrush: function() {
```
jcheng5: What if the plot margins or user coordinates change?

jcheng5:
I've finished reviewing the code, I haven't reviewed the docs/examples.

The JS mouse handling code in particular is beautifully written. Impressive work! Maybe could use an additional code comment to sketch out the high-level overview of how everything is organized.

I hesitate to even bring this up, but I have a feeling that resizing the window will cause plots to lose their brush (unless brush reset is disabled)? We might have to live with that.

wch:
Thanks! I can add in some info explaining how it works.

Yes, resizing will make the brush go away (if brush reset is on) or make the brush be positioned strangely (if brush reset is off).

For the case where brush reset is on, I couldn't think of a good way to fix it. We would have to know that the new image was sent because of a width/height change (not for some other reason), and I don't think there's a straightforward way to do that.

For the case where brush reset is off, if there is a coordmap, it would be possible to reposition the div to match the previous coordinates. If there's no coordmap (and the brush is sending pixel coordinates), then I don't think there's a sensible way to reposition the div.

So there's one case that we could improve. That said, I think our energy would probabl be better spent on other work, like ggplot2 support.

R/bootstrap.R
jcheng5: I don't particularly have a preference of what the implementation is. My only two desires here are 1) less code, 2) make functions that strongly couple with each other colocated in the source code.

R/image-interact-opts.R
jcheng5: Alright never mind then.

R/shinywrappers.R
jcheng5: Whoops, sorry.

inst/www/shared/shiny.js
jcheng5: Yeah, you might be right.

(I know it's a bunch of code to add for what might seem a pretty nitpicky concern, but sweating these kinds of details is often the difference between a UI that just feels right and one that works but feels indefinably clunky.)

inst/www/shared/shiny.js
wch: Implemented in 8d662f4.

inst/www/shared/shiny.js
wch: The brush still gets imported to the new image. This is because, although we can detect that a new image has been sent, we can't tell *why* it was sent, as mentioned previously.

wch:
OK, I've made all the changes. If there's nothing else, it should be good to merge now.

wch:
@timelyportfolio That does seem useful. Do  you know how much code would it be to fetch the RGB value from a ``?

PR #773, opened 2015-03-19
use ISO8601 and UTC for POSIX[c|l]t time by default

yihui:
I want to change two default arguments related to date/time in jsonlite::toJSON(): basically I want to make sure the default conversion is to a predictable standard representation, instead of a timezone-specific one, otherwise users will not be able to faithfully parse the original date/time back, e.g. the conversion in R might be using the server's timezone, and JS might parse the time using the browser's timezone info when a shiny app is published to a server.

### DIFF
--- a/R/shiny.R
+++ b/R/shiny.R
@@ -90,7 +90,7 @@ toJSON <- function(x, ..., digits = getOption("shiny.json.digits", 16)) {
 
   jsonlite::toJSON(x, dataframe = "columns", null = "null", na = "null",
                    auto_unbox = unbox, digits = digits, use_signif = TRUE,
-                   force = TRUE, ...)
+                   force = TRUE, POSIXt = "ISO8601", UTC = TRUE, ...)
 }
 
 # Call the workerId func with no args to get the worker id, and with an arg to


wch:
I think this would definitely be an improvement, but I think we need to think over the consequences carefully of using jsonlite's default time conversion behavior at all. Some example code comparing output:

```R
time <- Sys.time()
RJSONIO::toJSON(time)
# "[ 1.4268e+09 ]"
RJSONIO::toJSON(time, digits = 16)
# [1] "[ 1426800574.407632 ]"
jsonlite::toJSON(time)
# ["2015-03-19 16:29:34"] 
jsonlite::toJSON(POSIXt = "ISO8601", time)
# ["2015-03-19T16:29:34"] 
jsonlite::toJSON(POSIXt = "ISO8601", UTC = TRUE, time)
# ["2015-03-19T21:29:34Z"]
```

RJSONIO by default doesn't keep enough significant digits to do anything useful with the value. But with the `digits` we use in Shiny, it is useful.

wch:
@yihui, where are you encountering code that does times? It would be useful to look at how complicated the change to the new behavior would be. If it ends up being too much of a roadblock for users, we might need to copy the RJSONIO behavior.

yihui:
I'm working on date/time filters in DT. Each time column in the data corresponds to a slider that reads the range from the min/max of the time converted from as.numeric(POSIXct). The time column written to the table is rendered via toJSON(), and the default `POSIXt = "string"` makes the time values ambiguous in the table, e.g. the filter will work differently for users in different timezones.

I do not like the behavior of RJSONIO here. It is basically as.numeric(), and in my case, I will have to format the dates/times in the table manually, e.g. via `(new Date(value * 1000)).toWhateverString()`.

wch:
If there was no existing date/time filtering in DT that was broken by the switch from RJSONIO to jsonlite, then I think we're OK for your case.

I just want to make sure that if there is any existing code that depends on the RJSONIO conversion behavior, that we can provide a smooth migration path.

yihui:
Yeah, understood. Date/time filtering is completely new in DT; it was not available in shiny before, either.

wch:
It looks like dygraphs already does an explicit conversion to ISO 8601 time, so we won't have to worry incompatibilities there:
https://github.com/rstudio/dygraphs/blob/fec885fe/R/utils.R#L25-L29

wch:
I'm going to merge this, since ISO8601 is certainly no worse than the current time format. There is a small chance that we'll have to go back to epoch time, but I'm optimistic that we'll be able to stick with this.

PR #775, opened 2015-03-23
Split up shiny.js into multiple files

wch:
This splits up shiny.js into multiple source files in a directory called /srcjs.

It adds a `grunt concat` task, which concatenates all the files into a single shiny.js file, along with a sourcemap file, so that you can see the original .js file when debugging in the browser. The `grunt uglify` task also uses these sourcemaps so that even when minified version is used, the original source files can still be viewed.

To use, you'll need to make sure the grunt packages are installed and updated first:

```
cd tools
npm install
```

Then you can just run `grunt` to run all the tasks. The information in `/tools/README.md` goes into more detail.


jcheng5:
I'm curious, what were the tradeoffs of this approach (concat with user-provided ordered list of files) vs. other approaches you were looking at?

wch:
I looked at D3 and Vega for some ideas:

* D3 uses [smash](https://github.com/mbostock/smash), which lets you specify dependencies in the files. For example, this file pulls in all the imported files: https://github.com/mbostock/d3/blob/master/src/d3.js
* Vega has a [Makefile](https://github.com/trifacta/vega/blob/master/Makefile), with an explicit list of files, which it `cat`s together.

I ended up using  `grunt-contrib-concat` with an explicit list of files. It supports source maps, which is a plus for debugging. And it works easily with `grunt`, which we already have set up.

I don't think there are any major differences between the approaches, at least for a project of this scale; they all get the job done.

PR #778, opened 2015-03-24
Allow updateSelectInput(choices=character(0)) to clear select choices

jcheng5:
This is just a strawman proposal. I feel like we have a general problem with ambiguous `NULL` values in update*Input functions.

### DIFF
--- a/R/update-input.R
+++ b/R/update-input.R
@@ -403,10 +403,10 @@ updateRadioButtons <- function(session, inputId, label = NULL, choices = NULL,
 #' @export
 updateSelectInput <- function(session, inputId, label = NULL, choices = NULL,
                               selected = NULL) {
-  choices <- choicesWithNames(choices)
+  choices <- if (!is.null(choices)) choicesWithNames(choices)
   if (!is.null(selected))
     selected <- validateSelected(selected, choices, inputId)
-  options <- if (length(choices)) selectOptions(choices, selected)
+  options <- if (!is.null(choices)) selectOptions(choices, selected)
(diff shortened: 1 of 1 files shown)


wch:
Seems like a good idea to me.

jcheng5:
We could also do `missing()` as the criteria. If you pass `choices=NULL` wouldn't you kind of expect it to be cleared?

wch:
For a short while I used `if(missing())` in some of my code, but I stopped doing it. It works OK for functions that are called interactively at the console, but not so much when your function is called from other functions.

For example:

```R
f <- function(x = NULL) {
  g(x)
}

g <- function(x) {
  if (missing(x)) "x missing"
  else "x present"
}

g(1)
# [1] "x present"
f(1)
# [1] "x present"

g()
# [1] "x missing"
f()
# [1] "x present"
```

yihui:
Using `character(0)` sounds better than `missing()` to me.

mcfrank:
Hi all, just wanted to mention that we are using Shiny - which we love - in a selector-intensive application (wordbank.stanford.edu) and would really like to see this feature merged! Thank you for all your work on this!

kevinushey:
FWIW, I think there was an R bug where `missing` did not return the correct value when called within deeply nested contexts, but I think that should (?) be fixed in current releases of R.

wch:
@mcfrank Glad you guys are making use of Shiny!

@jcheng5 I think this is good to merge, unless you see other issues. We could also implement this for other inputs like radio buttons and checkbox groups.

jcheng5:
Thanks for the tip @kevinushey. Scary.

PR #795, opened 2015-04-09
Treat the output of toJSON() as a verbatim JSON string

yihui:
https://github.com/jeroenooms/jsonlite/commit/6fd201b3ad

also see ramnathv/htmlwidgets#28


wch:
Looks good to me - we should bump the jsonlite dependency when it's released.

yihui:
Great. Will do. Thanks!

PR #796, opened 2015-04-14
flowLayout's man page: % needs to be escaped

mikelove:
or else it turns into a commend in the Rd file


jcheng5:
Thanks!

wch:
Thanks - I also fixed a few other formatting issues in  02caf05.

PR #823, opened 2015-05-10
Load selectize options after initialization in the server mode

### LINKED ISSUE #814: R Shiny selectizeInput server side not initializing correctly
When trying to embed the selectize input widget in an Shiny application I ran into the following problem:

The item list isn't initialized with the complete list of items when using `server = TRUE`. Only after entering a character in the searchbox will the items appear. This is contrary to the behaviour when `server = FALSE`. In this case all items appear in the box initially (and this is the behaviour that I want). The reason that I want to use the `server = TRUE` option is that I want to be able to apply formatting to the items and this is not possible when generating the items client side. The example below reproduces the phenomenon with one selectize input using `server = TRUE` and the other with `server = FALSE`. In my opinion the behaviour should be the same in both cases.

Example app.R to reproduce the issue:

```r
library(shiny)

  updateSelectizeInput(session, 'cars1', server = TRUE, choices = rownames(mtcars))  
  updateSelectizeInput(session, 'cars2', server = FALSE, choices = rownames(mtcars))
}

ui <- fluidPage(
  selectizeInput('cars1', label = "server side", choices = NULL),
  selectizeInput('cars2', label = "client side", choices = NULL)  
  )

```
Is this a bug or am I using the widget incorrectly? I am running R version 3.1.3 and Shiny 0.11.1 on Windows 7 64-bit.

I've also posted this issue as a question on stackoverflow: http://stackoverflow.com/questions/29966936/r-shiny-selectizeinput-server-side-not-initializing-correctly/29970532#29970532

yihui:
This PR intends to fix #814.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -50,6 +50,9 @@ shiny 0.11.1.9xxx
 * Fixed bug where checkboxes were still active even after `Shiny.unbindAll()`
   was called. (#206)
 
+* The server side selectize input will load the first 1000 options by default
+  before users start to type and search in the box. (#823)
+
 shiny 0.11.1
 --------------------------------------------------------------------------------
 
--- a/R/update-input.R
+++ b/R/update-input.R
@@ -434,9 +434,6 @@ updateSelectizeInput <- function(session, inputId, label = NULL, choices = NULL,
   if (!server) {
     return(updateSelectInput(session, inputId, label, choices, selected))
   }
-  # in the server mode, the choices are not available before we type, so we
-  # cannot really pre-select any options, but here we insert the `selected`
-  # options into selectize forcibly
   value <- unname(selected)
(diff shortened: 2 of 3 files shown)


yihui:
I removed the commit for grunt so I don't have to rebase against master and solve conflicts every time another person's minified js files have been committed to master.

This is a pretty simple PR. It will be nice if it can go to the next CRAN release.

yihui:
Thanks @wch. You can run grunt later and commit the .min.js/.map files. I can do it as well, but I'm a little concerned about committing things that I cannot clearly see, so I will leave this to you.

wch:
Sure, I can do that. It would be nice if we had an easier way of dealing with generated files, but I don't know that there is.

PR #837, opened 2015-05-20
Ensure that callbacks fire in a FIFO order

jcheng5:
Version bump required so Leaflet can detect this fix

cc @wch

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -580,7 +580,11 @@ Callbacks <- R6Class(
       })
     },
     invoke = function(..., onError=NULL) {
-      for (callback in .callbacks$values()) {
+      # Ensure that calls are invoked in the order that they were registered
+      keys <- as.character(sort(as.integer(.callbacks$keys()), decreasing = TRUE))
+      callbacks <- .callbacks$mget(keys)
+
+      for (callback in callbacks) {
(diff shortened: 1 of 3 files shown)


R/utils.R
```diff
@@ -580,7 +580,11 @@ Callbacks <- R6Class(
       })
     },
     invoke = function(..., onError=NULL) {
-      for (callback in .callbacks$values()) {
+      # Ensure that calls are invoked in the order that they were registered
+      keys <- as.character(sort(as.integer(.callbacks$keys()), decreasing = TRUE))
```
wch: Yuck... but I don't see a better way to do this.

jcheng5: Yep. FWIW, ordered lists with fast random-access removal always seem to be pretty gross.

wch:
Would it make sense to use a queue instead of a map to store the callbacks? I have a pretty good implementation here (see Queue2 in the README and R code):
  https://github.com/wch/qstack/

wch:
I should mention that the Stack in shiny is basically Stack2 from that package, adapted to be an R6 class.

jcheng5:
No, because random-access deletes are important.

wch:
OK, looks good then.

PR #840, opened 2015-05-21
Depend on methods so Rscript doesn't fail

jcheng5:


wch:
IIRC, Rscript can still fail with this change if you call `rscript -e shiny::runApp()`, because that doesn't cause methods to be attached. In order for methods to be attached, you need to do `library(shiny)`. Let me dig up the relevant mailing list threads.

yihui:
That is what I have not finished typing in shiny-discuss...

wch:
Here's the message I posted to r-devel about it:
  http://r.789695.n4.nabble.com/new-cannot-be-accessed-when-running-from-Rscript-and-methods-package-is-not-loaded-tp4685079p4685130.html

jcheng5:
Yeah, but we call `library(shiny)` in `shiny::runApp()`. It seems to work...?

wch:
Oh, I suppose that's true. Do you mind putting a comment in there somewhere that explains why methods is needed, and why this works? Otherwise we'll keep getting confused about it.

jcheng5:
OK done.

wch:
Awesome. Looks good to me.

yihui:
IMHO the best solution is to use a hack in `.onAttach`/`.onLoad`: `x <- 'methods'; library(x, character.only = TRUE)` This will shut up R CMD check, and makes sure **methods** is always attached.

I vaguely remember even `Depends: methods` + `import(methods)` won't work in certain cases. Let me see if I can still find such a case.

jcheng5:
Fine but let's try not to bikeshed too hard on this, it doesn't seem like an issue that's worth much more of our time/energy.

yihui:
Okay I'll let it go then :)

Issue #142, opened 2013-04-06
sliderInput and animation


leonawicz:
This is more a suggestion than an issue. I apologize if this is not where I should be posting it.

When starting from some selected range of values on the slider, there does not appear to be a possibility of anchoring the lower boundary (or upper) when using animate = T or animate = [insert list of animationOptions] so that the animation may proceed by increasing the selected range instead of just shifting it.

For instance, unveiling a time series plot (using a fixed maximum xlim and ylim of course) as opposed to only being able to view a moving window of the series. This would be a nice feature if possible.

alandipert:
@wch implemented this in https://github.com/wch/shiny/commit/7a5e0e4 but it was not included with #852.

Presumably it would still be easy to add, if we decided to.

leonawicz:
Just got the notification- wow I forgot I even requested this in 2013! 😄 Suffice it to say I must have never really needed the feature that badly because I forgot all about it. This could probably be closed unless it's important to others to keep open.

alandipert:
@leonawicz 😄

PR #852, opened 2015-06-02
Slider fixes

wch:
This fixes some some issues in the behavior of slider animation:

* Single-handle sliders would loop, even when `loop=FALSE`.
* With range sliders, only the lower handle would move when animated. Now both of them move.

I've also implemented the feature requested in #142 (for range sliders, add option to control which handle moves when animated) on a separate branch, but I'm not sure it's worth it to add the complexity and future maintenance work, given that no one else has asked about it. The change is here:
https://github.com/wch/shiny/commit/7a5e0e4


jcheng5:
Was slider.R -> input-slider.R merely a rename?

wch:
Pretty much. Some code was moved from bootstrap.R and slider.R into input-slider.R. I guess it wasn't necessary for this PR -- it was actually a step that helped clear things up before adding the feature allowing control of which slider handle is animated -- but I didn't include that commit in this PR.

I have the goal of splitting out all the input and render functions into separate files as I touch them. I did the same for the JS code and it's now much easier to navigate.

jcheng5:
OK, that's fine with me, as long as the code movement is in separate
commits from actual code changes and the commit message indicates that that
is the case. Thanks!

PR #857, opened 2015-06-08
Fix #856: Outputs can not be unbound and re-bound

### LINKED ISSUE #856: Outputs can not be unbound and re-bound
This is a regression in Shiny 0.12.

https://groups.google.com/forum/#!msg/shiny-discuss/czXoYWnXGe8/CjO_S5swG2kJ
https://groups.google.com/forum/#!msg/shiny-discuss/kHUI5E9ZDvE/1aZR0mdlVNUJ

```r
library(shiny)

ui <- shinyUI(fluidPage(
  mainPanel(
    verbatimTextOutput('test'),
    tags$script(
      "setTimeout(function(){Shiny.unbindAll();Shiny.bindAll();},1000)"
    )
  )
))

server <- shinyServer(function(input, output, session) {
  output$test <- renderPrint({
    invalidateLater(500, session)
    runif(1)
  })
})

runApp(list(ui=ui, server=server))
```

If the bug is present, the numbers will stop changing about a second after the app loads. If the bug is not present, then the numbers will keep changing forever.

This was introduced in v0.12 while I was fixing https://github.com/rstudio/shiny/issues/749. The actual commit is https://github.com/rstudio/shiny/commit/6c36ea5753b39ea1204c2ff68474c1a50c7f4424.

The problem with that commit is that in order to avoid binding an output that's currently bound, it checks `$el.data("shiny-output-binding")`. When a binding is unbound, that data was not removed (which arguably is also a bug). A better test is to check for the CSS class `shiny-bound-output`.

jcheng5:


jcheng5:
@yihui @wch This really sucks, possibly bad enough to do an immediate patch release.

srcjs/init_shiny.js
```diff
@@ -47,12 +47,14 @@ function initShiny() {
...
-      var bindingAdapter = $(outputs[i]).data('shiny-output-binding');
+      var $el = $(outputs[i]);
+      var bindingAdapter = $el.data('shiny-output-binding');
       if (!bindingAdapter)
         continue;
       var id = bindingAdapter.binding.getId(outputs[i]);
       shinyapp.unbindOutput(id, bindingAdapter);
-      $(outputs[i]).removeClass('shiny-bound-output');
+      $el.removeClass('shiny-bound-output');
+      $el.removeData('shiny-output-binding');
```
yihui: LGTM. So the key problem is we forgot to `.removeData()`?

jcheng5: Yes. I could have just added that one line I suppose. I changed the test to check the class instead of the data, because the former is more obviously visible when debugging using the Chrome inspector (and therefore less likely to be buggy).

jcheng5:
Ref: https://github.com/rstudio/shiny/issues/856

PR #874, opened 2015-06-18
Remove uploaded files when session ends. Fixes #798

### LINKED ISSUE #798: Delete uploaded files (in temp directory) when they are no longer used
I think uploaded files aren't deleted until the R process exits. That's too late. We should remove files:

1. When the session responsible for uploading them ends
2. When a file upload is replaced by another upload to the same control

1 is simple, possibly we could use a finalizer to implement 2.

wch:
This does part (1) of #798.

### DIFF
--- a/R/fileupload.R
+++ b/R/fileupload.R
@@ -114,6 +114,12 @@ FileUploadContext <- R6Class(
     # Remove the directories containing file uploads; this is to be called when
     # a session ends.
     rmUploadDirs = function() {
+      # Make sure all_paths is underneath the tempdir()
+      if (!grepl(normalizePath(tempdir()), normalizePath(private$basedir), fixed = TRUE)) {
+        stop("Won't remove upload path ", private$basedir,
+          "because it is not under tempdir(): ", tempdir())
+      }
+
       all_paths <- file.path(private$basedir, private$ids)
       unlink(all_paths, recursive = TRUE)
     }
--- a/R/fileupload.R
+++ b/R/fileupload.R
@@ -84,7 +84,8 @@ FileUploadContext <- R6Class(
   class = FALSE,
   private = list(
     basedir = character(0),
-    operations = 'Map'
+    operations = 'Map',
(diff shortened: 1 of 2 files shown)


wch:
Test app:

```R
shinyApp(
  ui = bootstrapPage(
    fileInput("file", "fileInput", multiple = TRUE),
    verbatimTextOutput("fileinfo")
  ),
    vals <- reactiveValues(files = character(0))

    output$fileinfo <- renderPrint({
      files <- vals$files
      print(row.names = FALSE,
        data.frame(
          filename = files,
          exists = vapply(files, file.exists, logical(1)),
          stringsAsFactors = FALSE
        )
      )
    })
    observe({
      inFile <- input$file
      if (is.null(inFile)) return(NULL)
      isolate(vals$files <- c(vals$files, inFile$datapath))
    })

    # Print out files after session has ended
    session$onSessionEnded(function() {
      files <- isolate(vals$files)
      print(row.names = FALSE,
        data.frame(
          filename = files,
          exists = vapply(files, file.exists, logical(1)),
          stringsAsFactors = FALSE
        )
      )
    })
  }
)
```

R/fileupload.R
```diff
@@ -81,33 +81,41 @@ FileUploadOperation <- R6Class(
...
     },
     onJobFinished = function(jobId) {
-      .operations$remove(jobId)
+      private$operations$remove(jobId)
+    },
+    # Remove the directories containing file uploads; this is to be called when
+    # a session ends.
+    rmUploadDirs = function() {
+      all_paths <- file.path(private$basedir, private$ids)
+      unlink(all_paths, recursive = TRUE)
```
jcheng5: Can you put some kind of sanity check here, e.g. that basedir is a child of tempdir or something? Would hate for a bug now or in the future to result in "" or, heaven forbid, "/".

wch:
I've added a check that the dir to remove is under the temp dir.

PR #875, opened 2015-06-18
Separate xtable from print.xtable args. Closes #761

### LINKED ISSUE #761: renderTable() additional arguments bug
renderTable() takes additional arguments to be passed to both the xtable and print.xtable functions. 

The bug arises when one attempts to pass an argument to one function that can be partially matched as the argument of another. One either receives an "argument matches multiple arguments" error or unintentionally passes an argument to one function or the other without meaning to. The temporary workaround I used was to specify all the other arguments that could be partially matched to the argument of interest. 

A specific example (with the argument matches multiple arguments error) can be found here: https://gist.github.com/cmishra/87bad81d4eed495272ae.

wch:
Test apps, from #761:

```R
library(shiny)

serverShouldWork <- function(input, output) {
  output$bTable <- renderTable({iris}, caption=paste("test"),
    caption.placement = getOption("xtable.caption.placement", "top"),
    caption.width = getOption("xtable.caption.width", NULL)
  )
}

serverDoesWork <- function(input, output) {
  output$bTable <- renderTable({iris}, caption="test")
}

ui <- fluidPage("RenderTable Example",
  tableOutput("bTable") #table of information on beds (beds)
)

shinyApp(ui, serverDoesWork)
shinyApp(ui, serverShouldWork)
```

Neither of these apps gives warnings or errors now.


wch:
OK, I switched it to use `formals()`, so it should be good to merge now.

PR #876, opened 2015-06-19
Add support for HTTP HEAD requests

wch:
This fixes an issue where Shiny would not respect HTTP HEAD requests. The change was requested by CRAN because they now test package URLs with `getCurlHeaders()`, which is equivalent to `curl --head --location`.

This fix can be tested by running a simple test app, and then accessing it with `curl --head`.

```R
library(shiny)
runApp(
  shinyApp(bootstrapPage(), function(input, output) {}),
  port = 9000, launch.browser = FALSE
)
```

After the fix, `curl --head` gives these responses:

```
$ curl --head http://localhost:9000/
HTTP/1.1 200 OK
X-UA-Compatible: IE=edge,chrome=1
Content-Type: text/html; charset=UTF-8
Content-Length: 0

$ curl --head http://localhost:9000/foo
HTTP/1.1 404 Not Found
X-UA-Compatible: IE=edge,chrome=1
Content-Type: text/html; charset=UTF-8
Content-Length: 0
```

### DIFF
--- a/R/middleware.R
+++ b/R/middleware.R
@@ -332,6 +332,15 @@ HandlerManager <- R6Class("HandlerManager",
             headers=list('Content-Type' = 'text/html')))
         }
 
+        # Catch HEAD requests. For the purposes of handler functions, they
+        # should be treated like GET. The difference is that  they shouldn't
+        # return a body in the http response.
+        head_request <- FALSE
+        if (identical(req$REQUEST_METHOD, "HEAD")) {
+          head_request <- TRUE
+          req$REQUEST_METHOD <- "GET"
+        }
+
         response <- handler(req)
         if (is.null(response))
           response <- httpResponse(404, content="<h1>Not Found</h1>")
@@ -341,9 +350,21 @@ HandlerManager <- R6Class("HandlerManager",
           headers$'Content-Type' <- response$content_type
 
           response <- filter(req, response)
-          return(list(status=response$status,
-            body=response$content,
-            headers=headers))
+          if (head_request) {
+            headers$`Content-Length` <- nchar(response$content, type = "bytes")
+            return(list(
+              status = response$status,
+              body = "",
+              headers = headers
+            ))
+          } else {
+            return(list(
+              status = response$status,
+              body = response$content,
+              headers = headers
+            ))
+          }
+
         } else {
           # Assume it's a Rook-compatible response
           return(response)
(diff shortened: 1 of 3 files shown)


R/middleware.R
```diff
@@ -341,9 +350,11 @@ HandlerManager <- R6Class("HandlerManager",
           headers$'Content-Type' <- response$content_type
 
           response <- filter(req, response)
-          return(list(status=response$status,
-            body=response$content,
-            headers=headers))
+          return(list(
+            status = response$status,
+            body = if (head_request) "" else response$content,
```
aronatkins: You're returning a `Content-Length` of `0`, which is misleading. Since you've done the work of generating the content, use its size to compute the proper content length.

wch: @aronatkins Thanks! I tried explicitly adding `Content-Length` to `headers` with this:

```R
  headers$`Content-Length` <- nchar(response$content, type = "bytes")
```

But then that header comes up twice in the output, once with my value, and once with the computed value.

```
$ curl -I http://localhost:9000/
HTTP/1.1 200 OK
X-UA-Compatible: IE=edge,chrome=1
Content-Type: text/html; charset=UTF-8
Content-Length: 823
Content-Length: 0
```

I'll have to investigate how to avoid this.

aronatkins: In the interim, losing `Content-Length` altogether would be ok. But if it's getting added magically, you might need to do the same digging. :)

wch: I think digging will be required -- it's getting added magically, and I'm pretty sure it's not from inside the Shiny code base. :(

jcheng5: It's getting added by httpuv unfortunately.

jcheng5: https://github.com/rstudio/httpuv/blob/master/src/http.cpp#L400-L401

jcheng5: We could add a `if (_headers.find("Content-Length") != _headers.end()) { ... }` around that

aronatkins: Can you `body=NULL`?

wch: With `body=NULL`, R gives an error when the page is accessed: `Error: not compatible with requested type`

wch: @jcheng5 That sounds good to me.

jcheng5: Do you mind taking a stab at it?

wch: Sure - won't be able to today, but hopefully on Monday.

wch:
I've updated this pull request and made a PR against httpuv for the needed fix (rstudio/httpuv#50).

PR #885, opened 2015-07-02
Slider improvements

### LINKED ISSUE #689: sliderInput with date format
Hi, based on http://ionden.com/a/plugins/ion.rangeSlider/demo_advanced.html new sliderInput should work with date format, however I don't think it is working at moment.

```r
runApp(list(
    ui = basicPage(
        sliderInput('foo1', 'How many days after today?', min = Sys.Date() -10, 
                           max = Sys.Date(), value = Sys.Date() - 1)
    ),
    }
))
```

Returned an error 
```
Error in `/.difftime`(n_steps, scale_factor) : 
  second argument of / cannot be a "difftime" object
```
Do have in mind to implement it?

wch:
This pull request:

* Adds support for date and time values (closes #689)
* Adds a `dragRange` option. When TRUE and it's a range slider, the entire range can be dragged.

Example app: https://winston.shinyapps.io/slider-date

Presently, the time formatting is ridiculously verbose, and there's apparently not a simple way to do time formats in JavaScript. It would be nice to let the user set the time format, as well as the date format. I think the two best options are:

* [moment.js](http://momentjs.com/docs/#/displaying/format/)
* This library which provides a C-like [strftime](https://github.com/samsonjs/strftime) function.


wch:
I decided to use the strftime JS library, because it provides the same syntax used in R. It also supports setting the time zone for the time display on the client. I think this PR is complete - if it looks good, it's ready to merge.

RoyalTS:
Is there any way to look at the code for the example app?

jcheng5:
@RoyalTS Maybe @wch can provide the code, but in the meantime you should be able to figure it out from ?sliderInput, see the `timeFormat` parameter in particular.

PR #892, opened 2015-07-14
Events shiny:connected and shiny:disconnected

yihui:
Example:

```r
library(shiny)
shinyApp(
  ui = fluidPage(
    tags$head(tags$script(HTML(
      "$(document).on('shiny:connected', function(event) {
        alert('socket state: ' + event.socket.readyState);
      });
      $(document).on('shiny:disconnected', function(event, data) {
        alert('socket state: ' + event.socket.readyState);
      });"
    ))),
    actionButton('foo', 'Stop')
  ),
    observeEvent(input$foo, {
      stopApp()
    })
  }
)
```

Note I didn't commit the minified js files generated by grunt.

### DIFF
--- a/srcjs/shinyapp.js
+++ b/srcjs/shinyapp.js
@@ -70,6 +70,10 @@ var ShinyApp = function() {
 
     var socket = createSocketFunc();
     socket.onopen = function() {
+      $(document).trigger({
+        type: 'shiny:connected',
+        socket: socket
+      });
       socket.send(JSON.stringify({
         method: 'init',
         data: self.$initialInput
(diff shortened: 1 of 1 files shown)


jcheng5:
I was envisioning the socket being on the event object itself (`e.socket`). See the section "Example: To pass arbitrary data through an event object" on this page: http://api.jquery.com/trigger/

yihui:
Done. I thought it could be cleaner to separate the data into a different object, otherwise we will have to check if a property exists in the event object (e.g. we don't want to accidentally override `event.target`). Or maybe we use

```js
trigger({
  type: 'shiny:connected',
  data: {socket: socket}
})
```

?

jcheng5:
Yeah we do need to check (manually, not programmatically) that we're not overwriting a default property, but I think it's worth it given how strange it is for event handlers to take extra parameters. Also I think the risk is mitigated by the fact that we're talking about custom events here--so there's only ever going to be a small core of common properties, e.g. we're not dealing with mouse events and all the baggage they bring.

jcheng5:
When you do the grunt build, please also bump the version and NEWS--thx.

PR #909, opened 2015-07-22
Fix runGist

jcheng5:
Looks like GitHub changed the format of gist downloads from .tar.gz
to .zip

### DIFF
--- a/R/run-url.R
+++ b/R/run-url.R
@@ -112,7 +112,7 @@ runGist <- function(gist, destdir = NULL, ...) {
     stop('Unrecognized gist identifier format')
   }
 
-  runUrl(gistUrl, filetype = ".tar.gz", destdir = destdir, ...)
+  runUrl(gistUrl, filetype = ".zip", destdir = destdir, ...)
 }
 
 


yihui:
LGTM.

Too bad that runGist() is broken now because of this.

R/run-url.R
```diff
@@ -112,7 +112,7 @@ runGist <- function(gist, destdir = NULL, ...) {
     stop('Unrecognized gist identifier format')
   }
 
-  runUrl(gistUrl, filetype = ".tar.gz", destdir = destdir, ...)
+  runUrl(gistUrl, filetype = ".zip", destdir = destdir, ...)
```
aronatkins: The download URL that's embedded in the gist UI has a `.zip` extension. If you replace that with `.tar.gz`, you can get a tarball. Scraping the HTML doesn't feel great and I don't see that GitHub documents either of these direct-download approaches.

An alternate approach would be to look at the `files.FILENAME.raw_url` within a gist API response. Would that be less fragile?
https://developer.github.com/v3/gists/#response

jcheng5:
The `raw_url` might be less fragile but it'd also be downloading one file at a time.

Maybe what we should do is try to unzip it, and if that doesn't work, try to un-tar.gz it... :-\

aronatkins:
Oh! I hadn't realized that gists could contain multiple files. You could always use something like file magic on the downloaded archive before attempting to unarchive.

jcheng5:
That's a great idea, thanks!

PR #918, opened 2015-08-03
Fix #676: getId(el) not being used in output binding

### LINKED ISSUE #676: getId(el) not being used in output binding
I think the `getId(el)` function used for `OutputBinding` objects is being ignored.
The documentation I'm using is here: [http://shiny.rstudio.com/articles/building-outputs.html].

The relevant description for `getId(el)` is:
"Return the Shiny output ID for the element el, or null if the element doesn't have an ID and should therefore be ignored. The default implementation in Shiny.OutputBinding reads the data-output-id attribute and falls back to the element's id if not present."

It seems like the fall-back is occurring even when this function is overridden.
Below are 3 pieces of code, server.R, ui.R, and my_binding.js (located in the www/ directory).
Based on the description, I believe console.log(data) should show "my_data", but instead it is showing "my_div".

ui.R:

    library(shiny)
    shinyUI(fluidPage(
        singleton(tags$head(tags$script(src = "my_binding.js"))),
        div(id = "my_div")
    ))

server.R:

    library(shiny)
    shinyServer(function(input, output) {
        output$my_data <- renderText("my_data") ## I want this
        output$my_div <- renderText("my_div") ## I get this
    })

www/my_binding.js:

    var my_output_binding = new Shiny.OutputBinding();
    $.extend(my_output_binding, {
        find: function(scope) {
            return $(scope).find("#my_div");
         },
        getId: function(el) {
            // should be bound to server-side output$my_data
    	return "my_data";
        },
        renderValue: function(el, data) {
    	    console.log(data);
        },
        renderError: function(el, err) {
    	    console.log(err);
        },
    });
    Shiny.outputBindings.register(my_output_binding);

Per the docs, I believe the server-generated output$my_data should be bound to el, not output$my_div (which is the fallback).

jcheng5:


srcjs/init_shiny.js
```diff
@@ -207,6 +207,15 @@ function initShiny() {
   }
   exports.initializeInputs = initializeInputs;
 
+  function getIdFromEl(el) {
+    var $el = $(el);
+    var bindingAdapter = $el.data("shiny-output-binding");
+    if (!bindingAdapter)
+      return null;
+    else
+      return bindingAdapter.getId();
```
yihui: This PR looks good to me, but I guess here you meant `bindingAdapter.getId(el)`? Other than that, I think it should be ready to go after grunt/git commit -a/git rebase/force push.

jcheng5: No, it's intentional: https://github.com/rstudio/shiny/pull/918/files#diff-1f04015072adf4e6718b3ea09773dd7dR2945

yihui: Oh okay, that makes sense.

PR #930, opened 2015-08-18
More JS events

yihui:


srcjs/shinyapp.js
```diff
@@ -473,11 +499,23 @@ var ShinyApp = function() {
...
+      $(document).trigger('shiny:idle');
+    }
+  });
+
   // Progress reporting ====================================================
 
   var progressHandlers = {
     // Progress for a particular object
     binding: function(message) {
+      $(document).trigger({
```
jcheng5: I'm not sure this belongs here anymore.

srcjs/shinyapp.js
```diff
@@ -473,11 +499,23 @@ var ShinyApp = function() {
   });
 
 
+  addCustomMessageHandler('busy', function(message) {
+    if (message === true) {
+      $(document).trigger('shiny:busy');
+    } else if (message === false) {
+      $(document).trigger('shiny:idle');
```
jcheng5: I believe the existing lines where `shiny-busy` class is added/removed from the document element, should be moved to here.

jcheng5:
If you haven't already, can you create test app(s) for every event, including changing or not changing the value, and calling or not calling preventDefault()?

Did you have any ideas about how to document these events in-package? Custom .Rd page, or vignette?

jcheng5:
I think we said the `bound`/`unbound` events need an additional event property to say whether it was an input or output that was bound?

srcjs/shinyapp.js
yihui: Deleted.

srcjs/shinyapp.js
yihui: Moved.

yihui:
The binding type was added via 6b2113652117aceaf8c2f4b155a5af5786a3ebf5

I think a shiny dev center article should be good to document these events. I'm fine with .Rd if the documentation has to stay inside the package. A vignette is also a good choice.

I was puzzled for quite a while when the idle event did not work today, and it was fixed via d54530bbca0437d1d22c7315734d3142b3a91a91. I'm not sure if it is okay to change this line https://github.com/rstudio/shiny/blob/a242ae3849622062a90c0822a1bc718833574f5e/srcjs/shinyapp.js#L384 to `msgObj.hasOwnProperty(msgType)` because it happened that `msgObj[msgType]` was exactly `false` (now I changed the message to `"idle"` instead).

jcheng5:
Yes can you make the hasOwnProperty change please?

yihui:
Sure. Done.

jcheng5:
Let's have some docs in the package too (I don't care whether .Rd or vignette) only because I think the event types, properties, and behaviors are maybe likely to change in future versions.

jcheng5:
Other than docs, do you think this is ready to merge?

srcjs/input_rate.js
```diff
@@ -251,6 +251,20 @@ var InputDeferDecorator = function(target) {
   };
 }).call(InputDeferDecorator.prototype);
 
+var InputEventDecorator = function(target) {
+  this.target = target;
+};
+(function() {
+  this.setInput = function(name, value, immediate) {
+    var evt = jQuery.Event("shiny:inputchanged");
+    evt.name = name;
```
yihui: @jcheng5 The `name` here may have a suffix, e.g. `foo:shiny.action` for an action button with id `foo`. Do we want to do anything about it? The real question is, do we want to expose this kind of internal special id's to users?

yihui:
Other than the question about the `name` of the inputs, I think this should be ready to merge now. Most of the commit messages were written in haste, and I tend to squash all commits into two: one for the changes in source, and one for the minified js. Does that make sense?

BTW, I just finished an example https://gallery.shinyapps.io/107-events (source: https://github.com/rstudio/shiny-examples/tree/shiny-events/107-events), and will add a vignette later.

yihui:
A minor issue on the event names that just came to my mind: for `inputchanged` and `updateinput`, do you have a preference over the naming convention? e.g. `shiny:inputChanged`, `shiny:input-changed`, `shiny:input_changed`...

yihui:
This should be ready to merge now.

PR #932, opened 2015-08-20
Bump fontawesome version to 4.4.0, fixes #913

### LINKED ISSUE #913: Is there way to use those new fontawesome icons
I was trying to use some of the new icons fontawesome came up in 4.4 version. seems unsuccessful.
I was wondering is there way to use those new fontawesome icons?

dselivanov:
please, consider to merge


wch:
We'll also need to bump the `htmlDependency` version.

PR #938, opened 2015-08-26
Fixes #933: make sure type is of length one

### LINKED ISSUE #933: updateSliderInput gives error when trying to update label
When I use the function to update the label  It gives the following error:
```
Error in if (type == "date" || type == "datetime") { : 
  missing value where TRUE/FALSE needed
```
This function works well in old shiny version, but does not work in newest shiny version.

When looking into the source code of the function, it looks like  that it only look at the types in value, min, max. If I only want to update label, then vals is NULL leading to the error when comparing the type of NULL to date and datetime in the above if statement.

yihui:


wch:
Looks good to me.

PR #939, opened 2015-08-27
Fixes #929: defer setValue() for selectize until the options have been loaded from server

### LINKED ISSUE #929: Undefined label for selected item when updating server side selectize input element
There seems to be an issue with displayed labels when updating a selectizeInput element (server=TRUE, selected != NULL, and custom JS display function based on a data frame). The selected item is marked as `undefined` in the drop-down menu. See the following code:  

```
library(shiny)

runApp(list(
	ui = bootstrapPage(
		selectizeInput("selectizeTest", label="Selectize Test", choices=NULL)
	),
		choices <- cbind(value = paste0(rownames(mtcars),".v"), label=paste0(rownames(mtcars),".l"), 
										name=paste0(rownames(mtcars),".n"), 
										html = paste0("<strong>", rownames(mtcars), "</strong>"), mtcars)

		updateSelectizeInput(session, "selectizeTest", choices=choices, 
												 selected="Cadillac Fleetwood.v", server = TRUE, 
												 options = list(render = I(
												 	'{
												 	option: function(item, escape) {
												 		return "<div>" + item.html + "</div>"; }
													}'
												 ))
		)
		# also for return "<div>" + escape(item.name) + "</div>" 
	}
))
```

This is most likely a timing problem. When updating the server-side selectize element some time after app initialization (e.g. using invalidateLater) with a new value, this item is displayed correctly according to the provided JS `option` function. In the first case, I guess, the data frame is not yet available and the `selected` parameter is interpreted as new value without label.

See also [http://shiny.rstudio.com/articles/selectize.html](http://shiny.rstudio.com/articles/selectize.html). It may be helpful to update this page (the JS function name needs to be `option` instead of `options`; it should be clear that the data frame has to contain a `value` and a `label` column;  the return value of the `option` function should always be wrapped in a `div` or similar tag, otherwise the items are not displayed correctly).

Thanks

yihui:
I have tested most examples in the shiny-examples repo. Hopefully this is a safe change.


inst/www/shared/shiny.js
```diff
@@ -3840,6 +3841,9 @@ $.extend(selectInputBinding, {
           },
           success: function(res) {
             callback(res);
+            if (!loaded && data.hasOwnProperty('value'))
```
jcheng5: Is `!loaded` correct here? You only want to set it if it's not loaded?

jcheng5: Never mind, I see.

Issue #843, opened 2015-05-22
Add support for ifelse logic in conditionalPanel


daattali:
Many times I want to do something along the lines of

```
conditionalPanel("input.test", "yep"),
conditionalPanel("!input.test", "nope")
```

Some syntactic sugar here would be awesome :)  Hopefully the above could be translated into

```
conditionalPanel("input.test", iftrue = "yep", iffalse = "nope")
```

(clearly I'm not great at naming arguments)

daattali:
@jcheng5 I keep running into cases where I really want this functionality.

What does the shiny team think about 

```
conditionalPanel(condition, ..., ns = NS(NULL), else = NULL)
```

I used to be absolutely terrified about this idea, but now that conditionalPanel already has another named argument, I'm starting to think it's not that bad.

alandipert:
Sorry for taking so long to get to this 😄 This is a nice enhancement, I think it's safe to say we'd consider a PR.

vipulg13:
@daattali @alandipert is there any branch pertaining to this feature enhancement?

PR #943, opened 2015-08-31
Events shiny:visualchange, shiny:conditional, and shiny:recalculating

yihui:


srcjs/shinyapp.js
```diff
@@ -310,6 +310,10 @@ var ShinyApp = function() {
       var show = condFunc(scope);
       var showing = el.css("display") !== "none";
       if (show !== showing) {
+        el.trigger({
+          type: 'shiny:conditional',
```
jcheng5: This is useful in its own right, but it wasn't what I originally meant. I meant for a $(document).trigger() call at the top or bottom of $updateConditionals; this would let users create extensions that need to react any time any input or output changes. At least, I think that's what I was picturing? @daattali did you ask for something like this?

srcjs/shinyapp.js
```diff
@@ -503,6 +507,14 @@ var ShinyApp = function() {
     }
   });
 
+  addCustomMessageHandler('recalculating', function(message) {
+    if (message.hasOwnProperty('name')) return;
+    $('#' + message.name).trigger({
```
jcheng5: I don't think we can assume that the element's `id` is the same as the output ID, can we? Do we make this assumption anywhere else? If not, I think ShinyApp has a dictionary between output ID and binding/element.

jcheng5: Is it weird that we have an event for recalculating starting but not ending?

srcjs/init_shiny.js
```diff
@@ -253,7 +253,12 @@ function initShiny() {
       }
     });
     $('.shiny-bound-output').each(function() {
-      $(this).data('shiny-output-binding').onResize();
+      var $this = $(this), binding = $this.data('shiny-output-binding');
+      $this.trigger({
+        type: 'shiny:visualchange',
+        binding: binding
+      });
+      binding.onResize();
```
jcheng5: Do we also need to perform this logic in doSendOutputHiddenState?

srcjs/shinyapp.js
yihui: I have added changes according to the last three comments. For this one, I don't quite get it. I can certainly add the `$(document).trigger()` call, but I don't see how it can be useful.

jcheng5: https://github.com/rstudio/shiny/issues/843

daattali:
I like the recalculating+recalculated event, but I can't tell when exactly the other events trigger? Looks like there's an event that'll tell us whenever a uiOutput gets rendered?

It'd be awesome to have a list of these javascript events now that there are a few (including the `shiny:connected/disconnected` from before), it might be worth adding documentation for it in the main package docs. Thanks for the constant improvements!

jcheng5:
@daattali A vignette is on the way. Thanks for the feedback!

daattali:
I just came to github to look at the vignette again and it looks like it was deleted today. I found it very useful, I hope it comes back

daattali:
Nevermind, I just saw the commit message that said the vignette was moved into an article in the Shiny dev center. Sorry for spamming you guys

PR #949, opened 2015-09-09
Allow shared brush IDs

jcheng5:
Lets multiple plotOutput/imageOutput calls share the same brush ID. Brushing on a plot will dismiss any other brushes that share the same ID. This makes linked brushing implementations much easier. Brush input values now include `brushId` and `outputId` properties.


jcheng5:
@wch Can you review this? Nothing's changed since our conversation yesterday except I added comments.

srcjs/output_binding_image.js
```diff
@@ -800,17 +831,26 @@ imageutils.createBrushHandler = function(inputId, $el, opts, coordmap) {
 
   }
 
+  // Brush maintenance: When an image is re-rendered, the brush must either
+  // be removed (if brushResetOnNew) or imported (if !brushResetOnNew). The
+  // "mostRecentBrush" bit is to ensure that when multiple outputs share the
+  // same brush ID, inactive brushes don't send null values up to the server.
```
wch: Something about the logic and this comment is a little confusing. Here's my understanding of what happens each time an image is re-rendered:

* The old image is removed, and `onRemoveImg` is called. If `opts.brushResetOnNew`, then the brush will be reset.
* Various rendering things happen (the particulars aren't relevant - it's just important to note that the previous and next steps don't come right after one another).
* `createBrushHandler` (the current function) is called. If `!opts.brushResetOnNew`, the old brush will be imported.

I think the logic can be simplified in a way that makes things more understandable... I believe that we should be to simplify it to something like this, if we're properly checking for reset state in the right places.

```js
  function onRemoveImg() {
    if (opts.brushResetOnNew) {
      brush.reset();
      brushInfoSender.immediateCall();
    }
  }

  # This shouldn't do anything if the brush has been reset, I think?
  brush.importOldBrush();
  brushInfoSender.immediateCall();
```



So there are 2x2 conditions:

| `opts.brushResetOnNew` | `mostRecentBrush` |
| ---- | ---- |
| `true` | `true` |
| `true` | `false` |
| `false` | `true` |
| `false` | `false` |


| `opts.brushResetOnNew` | `mostRecentBrush` |
| ---- | ---- |
| `true` | `true` |
| `true` | `false` |
| `false` | `true` |
| `false` | `false` |

wch: After a bit of discussion with @jcheng5, I think that the current code makes sense.

PR #966, opened 2015-09-24
Fix highlighting of reactives in showcase mode

jcheng5:
This has been broken since we switched to jsonlite. Reactives
don't highlight because their srcref is in an S3(?) class that
jsonlite doesn't recognize, whereas RJSONIO would treat it as
a numeric vector.


yihui:
Weird. I don't see it is broken. Do you have a reproducible example?

jcheng5:
Try `runExample('03_reactivity')`. The outputs light up but not the reactive.

yihui:
Oh okay. Thanks! I can confirm this issue has been fixed in that example.

Issue #961, opened 2015-09-22
unicode inequalities (u2264 and u2265) not working on windows


garthtarr:
**The issue is that the options in selectInput lists with unicode characters appear to be ignored.**

This bug is specifically related to Windows (I've tested with Windows 7 and Windows 10).  The MWE below works perfectly on unix-type systems (tested on OS X 10.10.5 and Ubuntu 14.04).

The issue is related to certain unicode characters, specifically u2264 and u2265, though there probably are others.

It's not obvious if this is an R issue or a shiny issue, [though in discussion](https://groups.google.com/forum/#!topic/shiny-discuss/3KDGxupzRRQ) with @yihui, he suggested I open an issue here.  I have tested on the latest R-devel build for Windows and the issue is still present:

```
> sessionInfo()
R Under development (unstable) (2015-09-20 r69402)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 7 x64 (build 7601) Service Pack 1

locale:
[1] LC_COLLATE=English_Australia.1252  LC_CTYPE=English_Australia.1252   
[3] LC_MONETARY=English_Australia.1252 LC_NUMERIC=C                      
[5] LC_TIME=English_Australia.1252    

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] shiny_0.12.2.9003

loaded via a namespace (and not attached):
 [1] R6_2.1.1         htmltools_0.2.9  tools_3.3.0      rstudioapi_0.3.1
 [5] Rcpp_0.12.1      jsonlite_0.9.17  digest_0.6.8     xtable_1.7-4    
 [9] httpuv_1.3.3     mime_0.4
```

A work around (as suggested by @yihui) is to change the locale from English (the default on my machine) to Chinese:

```
if (.Platform$OS.type != "unix") Sys.setlocale("LC_CTYPE", "Chinese")
```

Another work around is to not use windows machines - unfortunately that's not always possible.

Here's the MWE:

```
require(shiny)
app <- shinyApp(
  ui = fluidPage(
    "\u2264 and \u2265",
    uiOutput("bin_prob_k"),
    uiOutput("bin_prob_inequality"),
    uiOutput("bin_prob_out"),
    plotOutput("pmf")
  ),
  server = function(input, output) {
    
    output$bin_prob_k = renderUI({
      numericInput("binom4_k", label = "Number of successes (k)", value = 3,
                   step = 1, min = 0, max = 10)
[...]

PR #968, opened 2015-09-25
A more fundamental fix to the Unicode issue

yihui:
After Duncan fixed the `grep()` bug https://bugs.r-project.org/bugzilla3/show_bug.cgi?id=16264 and a user reported #961, I took another look at the Unicode issue on Windows. This time I realized my previous work was actually buggy: I was testing if a file is encoded in UTF8 by reading it as UTF8 and checking if `iconv(from = 'UTF-8', to = '')` is `NA`. This is actually checking if the file content can be converted to native encoding, and does not really tell us if the file is encoded in UTF8.

The other mistake I made before was to convert everything to native encoding using `enc2native()`, which can be lossy, e.g. the UTF-8 encoded character `≤` will be converted to `=`. #961 revealed this mistake.

```r
> Sys.setlocale(,'English')
[1] "LC_COLLATE=English_United States.1252;LC_CTYPE=English_United States.1252;LC_MONETARY=English_United States.1252;LC_NUMERIC=C;LC_TIME=English_United States.1252"
> "\u2264"
[1] "≤"
> iconv("\u2264", "UTF-8")
[1] "="
```

Now I use native encoding only if the conversion is not lossy; see `tryNativeEncoding ()`. Native encoding works better than UTF8 on Windows in general (e.g. assign a value to a Unicode variable name in an environment), but if native encoding is lossy, we just use UTF8 anyway. Everything will be converted to UTF-8 in the final step, e.g. when sending the JSON message, or writing out HTML.

The major challenge with UTF8 before Duncan's fix was actually source references, e.g. via `source(keep.source = TRUE)`. The reason that it fails with UTF8 was a single line `grep("\n", fixed = TRUE)` in the source code of `srcfilecopy()`:

```r
> srcfilecopy
function (filename, lines, timestamp = Sys.time(), isFile = FALSE) 
{
    stopifnot(is.character(filename), length(filename) == 1L)
    e <- new.env(parent = emptyenv())
    if (any(grepl("\n", lines, fixed = TRUE))) 
        lines <- unlist(strsplit(sub("$", "\n", as.character(lines)), 
            "\n"))
    ....
}
```

I have provided a hack for it in a custom version of `srcfilecopy()`, and I believe the hack is pretty safe. So now source references will work well with UTF8 source code for all versions of R (you don't have to use R 3.3.0).

A few known issues left (they only occur when Unicode characters cannot be converted to native encoding, e.g. when using Chinese in an English locale):

1. We cannot use multibyte characters in input/output IDs, e.g. `output[['summary这里也可以用中文']] <- renderPrint(...)`. This can be fixed by replacing the environment with a list in the `Map` object, but I found a new issue after the replacement, so I didn't include the commit in this PR (https://github.com/yihui/shiny/commit/00a7a86f85368d4142ce73299606ee50460dba99).
2. Showcase mode does not work for some reason (I have not investigated the reason yet).
3. Reading the DESCRIPTION using `read.dcf()` does not work, since `read.dcf()` does not have an `encoding` argument. Not sure if `read.dcf(file('DESCRIPTION', encoding = 'UTF-8'))` works.

Anyway, after this PR, we will no longer suffer from corruption of certain Unicode characters, and we don't need to assume all characters can be represented in the system native encoding.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -1035,6 +1035,11 @@ sourceUTF8 <- function(file, ...) {
   source(file, ..., keep.source = TRUE, encoding = checkEncoding(file))
 }
 
+# write text as UTF-8
+writeUTF8 <- function(text, ...) {
+  text <- enc2utf8(text)
+  writeLines(text, ..., useBytes = TRUE)
+}
 
 URLdecode <- decodeURIComponent
 URLencode <- function(value, reserved = FALSE) {
--- a/R/utils.R
+++ b/R/utils.R
@@ -1022,8 +1022,27 @@ tryNativeEncoding <- function(string) {
 }
 
 # similarly, try to source() a file with UTF-8
-sourceUTF8 <- function(file, ...) {
-  source(file, ..., keep.source = TRUE, encoding = checkEncoding(file))
+sourceUTF8 <- function(file, envir = globalenv()) {
+  lines <- readUTF8(file)
+  enc <- if (any(Encoding(lines) == 'UTF-8')) 'UTF-8' else 'unknown'
+  src <- srcfilecopy(file, lines, isFile = TRUE)  # source reference info
+  # oddly, parse(file) does not work when file contains multibyte chars that
+  # **can** be encoded natively on Windows (might be a bug in base R); we fall
+  # back to parse(text) in this case
+  exprs <- tryCatch(
+    parse(file, keep.source = FALSE, srcfile = src, encoding = enc),
+    error = function(e) {
+      parse(text = lines, keep.source = FALSE, srcfile = src, encoding = enc)
+    }
+  )
+  eval(exprs, envir)
+}
+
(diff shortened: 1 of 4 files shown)


R/utils.R
```diff
@@ -1005,41 +1000,56 @@ checkEncoding <- function(file) {
...
+    parse(file, keep.source = FALSE, srcfile = src, encoding = enc),
+    error = function(e) {
+      parse(text = lines, keep.source = FALSE, srcfile = src, encoding = enc)
+    }
+  )
+  eval(exprs, envir)
 }
 
+# a workaround for https://bugs.r-project.org/bugzilla3/show_bug.cgi?id=16264
+if (getRversion() <= '3.2.2') srcfilecopy <- function(filename, lines, ...) {
```
jcheng5: Isn't this `getRversion()` check going to apply at package *build* time, not at runtime?

yihui: Addressed in #1027. Thanks!

PR #979, opened 2015-10-15
Enhancements to help profiling

wch:
There are three changes in this PR:

* There's a function `wrapFunctionLabel` which essentially renames a function so that it can have any arbitrary label to show up in the profiler. This makes it possible to make output observers show up as `output$foo` instead of as `func`, at the small cost of adding one more function on the call stack.
* A `try(tryCatch())` statement was simplified to just `tryCatch()`. This makes the call stack a little smaller, and also now prints to the console which output the error occurred in.
* Various internal functions named `func` and `.func` were renamed to more descriptive names, like `observerFunc` and `flushCallback`.

### DIFF
--- a/R/react.R
+++ b/R/react.R
@@ -72,8 +72,8 @@ Context <- R6Class(
         }
       }, add = TRUE)
 
-      lapply(.flushCallbacks, function(func) {
-        func()
+      lapply(.flushCallbacks, function(flushCallback) {
+        flushCallback()
       })
     }
   )
@@ -112,7 +112,9 @@ ReactiveEnvironment <- R6Class(
       old.ctx <- .currentContext
       .currentContext <<- ctx
       on.exit(.currentContext <<- old.ctx)
-      shinyCallingHandlers(func())
+      # Rename to assist debugging
+      contextFunc <- func
+      shinyCallingHandlers(contextFunc())
     },
     addPendingFlush = function(ctx, priority) {
       .pendingFlush$enqueue(ctx, priority)
--- a/R/react.R
+++ b/R/react.R
@@ -108,12 +108,10 @@ ReactiveEnvironment <- R6Class(
       }
       return(.currentContext)
     },
-    runWith = function(ctx, func) {
+    runWith = function(ctx, contextFunc) {
       old.ctx <- .currentContext
       .currentContext <<- ctx
       on.exit(.currentContext <<- old.ctx)
-      # Rename to assist debugging
-      contextFunc <- func
       shinyCallingHandlers(contextFunc())
     },
     addPendingFlush = function(ctx, priority) {
--- a/R/reactives.R
+++ b/R/reactives.R
@@ -487,15 +487,12 @@ Observer <- R6Class(
     .destroyed = logical(0),
     .prevId = character(0),
 
-    initialize = function(func, label, suspended = FALSE, priority = 0,
+    initialize = function(observerFunc, label, suspended = FALSE, priority = 0,
                           domain = getDefaultReactiveDomain(),
                           autoDestroy = TRUE) {
-      if (length(formals(func)) > 0)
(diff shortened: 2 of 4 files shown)


R/utils.R
```diff
@@ -1046,3 +1046,23 @@ URLencode <- function(value, reserved = FALSE) {
...
+# trace.
+wrapFunctionLabel <- function(func, name) {
+  if (name == "name" || name == "func" || name == "relabelWrapper") {
+    stop("Invalid name for wrapFunctionLabel: ", name)
+  }
+  assign(name, func, environment())
+
+  relabelWrapper <- function(...) {
+    # This `f` gets renamed to the value of `name`
+    f(...)
```
yihui: This will trigger a NOTE in R CMD check: `no visible global function definition for ‘f’`

Other than that, this PR looks good to me.

R/react.R
```diff
@@ -112,7 +112,9 @@ ReactiveEnvironment <- R6Class(
       old.ctx <- .currentContext
       .currentContext <<- ctx
       on.exit(.currentContext <<- old.ctx)
-      shinyCallingHandlers(func())
+      # Rename to assist debugging
+      contextFunc <- func
```
jcheng5: Could we rename the `runWith` argument to `contextFunc` directly?

R/reactives.R
```diff
@@ -494,9 +494,12 @@ Observer <- R6Class(
         stop("Can't make an observer from a function that takes parameters; ",
              "only functions without parameters can be reactive.")
 
+      # New name to assist debugging
+      observerFunc <- func
```
jcheng5: Same here

R/utils.R
```diff
@@ -1046,3 +1046,23 @@ URLencode <- function(value, reserved = FALSE) {
...
+  }
+
+  body(relabelWrapper)[[2]][[1]] <- as.name(name)
```
jcheng5: Rather than mutating the function body, how about this?

```r
relabelWrapper <- eval(
  substitute(function(...) {f(...)}, list(f = as.name(name)))
)
```

jcheng5:
Nothing for reactives yet, right?

R/react.R
wch: Yeah, that's a nicer solution.

R/utils.R
wch: Thanks for catching that!

R/utils.R
wch: I just tested that out, and it has a somewhat weird result. When printed, the function show something different from the actual body of the function, because of source refs stored in the function.

```R
x <- eval(substitute(
  function(...) {
    f(...)
  },
  list(f = as.name("FOO"))
))

# When printed, it says it's calling f
x
# function(...) {
#     f(...)
#   }

# But it's really calling FOO
body(x)
# {
#     FOO(...)
# }

str(x)
# function (...)  
#  - attr(*, "srcref")=Class 'srcref'  atomic [1:8] 2 3 4 3 3 3 2 4
#   .. ..- attr(*, "srcfile")=Classes 'srcfilecopy', 'srcfile' <environment: 0x93d0018> 
```

But with my method, it prints the actual  content of the function:

```R
z <- function(...) {
  f(...)
}
body(z)[[2]][[1]] <- as.name("FOO")

z
# function (...) 
# {
#     FOO(...)
# }

body(z)
# {
#     FOO(...)
# }

# No source refs
str(z)
# function (...)  
```

But other than that, I think using `substitute` is a cleaner approach, so I'll go with that.

R/utils.R
wch: It turns out this is fixed by using the `substitute()` method of creating the function.

wch:
As we discussed, there's no need to relabel reactives, since they're usually called with the name assigned by the user.

wch:
This PR should be good to merge now.

PR #983, opened 2015-10-16
updateCheckboxGroup: allow clearing all choices. Fixes #981

### LINKED ISSUE #981: You cannot clear the choices in updateCheckboxGroupInput
You can clear the choices in updateSelectInput with choices = character(0). This does not work the same way with updateCheckboxGroupInput and I would expect it to.

choices <- letters[1:5]
choices2 <- letters[5:1]
runApp(list(
  ui = basicPage(
    checkboxGroupInput('chkGrp', 'Options', choices = NULL),
    actionButton("toggle","Toggle")
  ),
    observe({
      
      is.even <- (input$toggle %% 2 == 0)
      
      if (is.null(input$toggle) || is.even)
      {
        updateCheckboxGroupInput(session,"chkGrp", choices = choices)
      }
      else
      {
        updateCheckboxGroupInput(session,"chkGrp", choices = character(0))
        #updateCheckboxGroupInput(session,"chkGrp", choices = choices2)
      }
    })
  })
)

wch:
This fixes #981. The new logic is essentially the same as for `updateSelectInput`.

As tested:
* `choices=c("a","b","c")` updates the input with those choices
* `choices=character(0)` removes all choices
* `choices=NULL` has no effect on the current choices

Test app:

```R
runApp(list(
  ui = basicPage(
    actionButton("vector", 'choices=c("a", "b", "c")'),
    actionButton("empty", "choices=character(0)"),
    actionButton("null", "choices=null (no change)"),
    checkboxGroupInput('chkGrp', 'Options', choices = NULL)
  ),
    observeEvent(input$vector, {
      updateCheckboxGroupInput(session,"chkGrp", choices = c("a", "b", "c"))
    })
    observeEvent(input$empty, {
      updateCheckboxGroupInput(session,"chkGrp", choices = character(0))
    })
    observeEvent(input$null, {
      updateCheckboxGroupInput(session,"chkGrp", choices = NULL)
    })
  })
)
```

### DIFF
--- a/R/update-input.R
+++ b/R/update-input.R
@@ -306,15 +306,16 @@ updateSliderInput <- function(session, inputId, label = NULL, value = NULL,
 updateInputOptions <- function(session, inputId, label = NULL, choices = NULL,
                                selected = NULL, inline = FALSE,
                                type = 'checkbox') {
-
-  choices <- choicesWithNames(choices)
+  if (!is.null(choices))
+    choices <- choicesWithNames(choices)
   if (!is.null(selected))
     selected <- validateSelected(selected, choices, inputId)
 
-  options <- if (length(choices))
+  options <- if (!is.null(choices)) {
     format(tagList(
       generateOptions(inputId, choices, selected, inline, type = type)
     ))
+  }
 
   message <- dropNulls(list(label = label, options = options, value = selected))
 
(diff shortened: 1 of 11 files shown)


jcheng5:
LGTM but can you also update the docs and add a test please?

wch:
I've updated the docs. Roxygen is very limited in allowing you to add information to separate doc files when a template is used, so I just added the info to the template.

Since it's an updater function, I don't think it's possible to write a unit test -- it would have to be tested on a running app.

PR #988, opened 2015-10-23
Shiny modules

jcheng5:
Article draft is here: http://shiny.rstudio-staging.com/articles/modules.html


cholcomb:
Very interesting and perfect timing, since I'm refactoring our large Shiny app. It still seems valuable from a debugging perspective to have unique names for each input/output combination, and when you don't have such a combination it seems like you just need functions. I'll try it out more. You probably already know this but the article has some names mismatching - csvUpload for example. Since this is fairly abstract, I think it would be nice to re-arrange the article where folks can copy-paste and follow along through fully executable examples. Joe, I can't remember, have I shown you our app?

jcheng5:
Typos in the article have been fixed, thanks.

Each input/output combination does have unique names, but they are formed using the namespaces. Hopefully in real-world usage the namespace qualifiers will make things less confusing, not more confusing. Once you've had a try with it on your actual app let me know if you still feel there is any downside here.

No, I don't think I've seen your app...?

jcheng5:
@cholcomb No sorry, I think I did see your app, didn't we video chat about Leaflet?

R/modules.R
```diff
@@ -0,0 +1,57 @@
+# Creates an object whose $ and $<- pass through to the parent
+# session, unless the name is matched in ..., in which case
+# that value is returned instead. (See Decorator pattern.)
+createSessionProxy <- function(parentSession, ...) {
+  e <- new.env(parent = emptyenv())
+  e$parent = parentSession
```
wch: Don't tell me we're going to start using `=` now. ;)

cholcomb:
Yea awhile back, I forgot if we showed the app, thought you might like to see an example of what your work is helping make possible in case you hadn't.

R/modules.R
yihui: Oh yeah! Follow your heart, Joe! :wink: All assignments shall be created **equal**! (Coincidently I just started this war on Twitter a couple of hours ago today)

jcheng5: Haha. Too much time in the rstudio/leaflet repository I guess.

shrektan:
ha, this is an excellent feature for me. Thanks.

jcheng5:
Once I have tested with [shiny-examples](https://github.com/rstudio/shiny-examples) I'll merge this.

maxmoro:
I like modules! Right on time for a new big application we are building.
Thanks!
Max

PR #993, opened 2015-10-28
Cleaner reactlog labels for reactives and outputs

jcheng5:
Instead of showing the code, try to just show the name
of the reactive/output. Uses a fairly flaky algorithm
for determining the name of the reactive; will only
work in cases where the definition of the reactive
begins with "foo <- reactive({".

### DIFF
--- a/R/reactives.R
+++ b/R/reactives.R
@@ -441,12 +441,45 @@ reactive <- function(x, env = parent.frame(), quoted = FALSE, label = NULL,
   # Attach a label and a reference to the original user source for debugging
   if (is.null(label))
     label <- sprintf('reactive(%s)', paste(deparse(body(fun)), collapse='\n'))
-  srcref <- attr(substitute(x), "srcref")
+  srcref <- attr(substitute(x), "srcref", exact = TRUE)
+  label <- srcrefToLabel(srcref[[1]], label)
   if (length(srcref) >= 2) attr(label, "srcref") <- srcref[[2]]
   attr(label, "srcfile") <- srcFileOfRef(srcref[[1]])
   o <- Observable$new(fun, label, domain)
   registerDebugHook(".func", o, "Reactive")
   structure(o$getValue, observable = o, class = "reactive")
+}
+
+# Given the srcref to a reactive expression, attempts to figure out what the
+# name of the reactive expression is. This isn't foolproof, as it literally
+# scans the line of code that started the reactive block and looks for something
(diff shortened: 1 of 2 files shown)


R/reactives.R
```diff
@@ -449,6 +450,38 @@ reactive <- function(x, env = parent.frame(), quoted = FALSE, label = NULL,
...
+  m <- regexec("(.*)(<-|=)\\s*reactive\\s*\\($", firstLine)
+  if (m[[1]][1] == -1) {
+    return(defaultLabel)
+  }
+  sym <- regmatches(firstLine, m)[[1]][2]
+  res <- try(parse(text = sym), silent = TRUE)
+  if (inherits(res, "try-error"))
+    return(defaultLabel)
+
+  if (length(res) != 1 || !is.symbol(res[[1]]))
```
wch: Just curious - under what conditions would it get here? I could imagine someone keeping a list of reactives like:

```R
reactives$x1 <- reactive( ... )
reactives$x2 <- reactive( ... )
```
in which case it would still be good to use the left side of the assignment as the label.

R/reactives.R
```diff
@@ -441,7 +441,8 @@ reactive <- function(x, env = parent.frame(), quoted = FALSE, label = NULL,
   # Attach a label and a reference to the original user source for debugging
   if (is.null(label))
     label <- sprintf('reactive(%s)', paste(deparse(body(fun)), collapse='\n'))
-  srcref <- attr(substitute(x), "srcref")
+  srcref <- attr(substitute(x), "srcref", exact = TRUE)
+  label <- srcrefToLabel(srcref[[1]], label)
```
wch: This will override the label if one is supplied with `reactive(label = "foo")`. Is that the intent?

R/reactives.R
jcheng5: `a = list(foo = reactive(...))`, we wouldn't want the label to be `a = list(foo`.

You're right, I hadn't thought about those cases. I could either blindly take the left side, or pattern match on `$`/`[[`/`[`, or...?

jcheng5: I'm wrong, `a = list(foo` won't parse. The (even more unrealistic) case I was originally thinking of was `a <- 1; b <- eactive(...)`. The length(res) check would take care of that, I could drop the `is.symbol` test, I think that would satisfy your case.

jcheng5:
OK, that should do it.

wch:
Looks good to me.

PR #994, opened 2015-10-29
Correct a regression within modules

RCura:
Commit 07f2792cf9f65e964d32daf6ae87567128bb9742 introduced an error, replacing `e$parent = parentSession` with `e$parent <= parentSession`, while it should have been `e$parent <- parentSession`


jcheng5:
Let's never speak of this again.

RCura:
I did love you commit msg and was forced to look at the inside ;-)

PR #1000, opened 2015-11-02
Don't require session to be explicitly passed to invalidateLater

jcheng5:
These functions were created before getDefaultReactiveDomain()
existed, so the only way to get ahold of the current session was
if the caller explicitly passed it.

This is slightly backwards incompatible, in that existing calls
to invalidateLater() that don't pass a session argument will
behave slightly differently (bound to the current session instead
of to no session), but those calls would have triggered a warning
for all but the very earliest versions of Shiny.


jcheng5:
@wch @yihui Any concerns about the "slight" backwards incompatibility here?

yihui:
I think that is fine.

PR #1005, opened 2015-11-05
Upgrade bootstrap to 3.3.5

jcheng5:
Supposedly just bug fixes.

I didn't have a pressing need for this, just noticed we were a few versions back.


wch:
I think it needs a version bump here: https://github.com/wch/shiny/blob/712c4cb/R/bootstrap.R#L39

jcheng5:
Thanks, fixed.

yihui:
Sounds good to me!

PR #1011, opened 2015-11-13
Better errors

wch:
Don't merge - I'm just creating this PR so that I can comment on the combined changes. (I think Github only allows commenting on individual commits or PRs.)


R/conditions.R
```diff
@@ -0,0 +1,120 @@
...
+# the ..stacktraceon.. and ..stacktraceoff.. calls themselves are hidden
+# too.
+#
+# Example:
+# ..stacktraceoff..({
+#   tryCatch(
+#     withVisible(..stacktraceon..(userFunc()))
+#   )
+# })
+..stacktraceon.. <- identity
```
wch: I think these would be better as `..stacktraceon.. <- function(x) x`, so that R will keep source refs to them. With `identity`, there's no source ref.

R/imageutils.R
```diff
@@ -50,9 +50,12 @@ plotPNG <- function(func, filename=tempfile(fileext='.png'),
...
   # to set par(mar = smaller_value) before they draw base graphics.
   op <- graphics::par(mar = rep(0, 4))
-  tryCatch(graphics::plot.new(), finally = graphics::par(op))
+  on.exit(graphics::par(op), add = TRUE)
+  graphics::plot.new()
+
   dv <- grDevices::dev.cur()
-  tryCatch(shinyCallingHandlers(func()), finally = grDevices::dev.off(dv))
+  on.exit(grDevices::dev.off(dv), add = TRUE)
+  func()
```
wch: I like this code way better than what we had before.

R/reactives.R
```diff
@@ -357,7 +357,9 @@ Observable <- R6Class(
         stop("Can't make a reactive expression from a function that takes one ",
              "or more parameters; only functions without parameters can be ",
              "reactive.")
-      .func <<- func
+
+      .func <<- wrapFunctionLabel(func, paste("reactive", label),
```
wch: Is it necessary to wrap a label here? The name of the reactive is already present in the stack. "count" appears at two levels in this flame graph:

image

Is it possible to turn on stack tracing between `.func` and the user code, but without the relabeling?

R/reactives.R
```diff
@@ -400,7 +404,12 @@ Observable <- R6Class(
       on.exit(.running <<- wasRunning)
 
       ctx$run(function() {
-        result <- withVisible(try(shinyCallingHandlers(.func()), silent=TRUE))
+        result <- withCallingHandlers(withVisible(.func()),
+          error = function(cond) {
+            .visible <<- FALSE
+            .value <<- try(stop(cond), silent = TRUE)
```
wch: What's the purpose of the `try(stop(cond))`? Playing around with this, it looks like it'll convert `cond` from something like this:

```
> str(cond)
List of 2
 $ message: chr "foo"
 $ call   : language func()
 - attr(*, "class")= chr [1:3] "simpleError" "error" "condition"
```

into something like this:

```
> str(.value)
Class 'try-error'  atomic [1:1] Error in func() : foo

  ..- attr(*, "condition")=List of 2
  .. ..$ message: chr "foo"
  .. ..$ call   : language func()
  .. ..- attr(*, "class")= chr [1:3] "simpleError" "error" "condition"
```

If that's the goal, it would be nice to have a comment explaining it, because it's weird.

R/reactives.R
```diff
@@ -544,14 +553,17 @@ Observer <- R6Class(
 
     initialize = function(observerFunc, label, suspended = FALSE, priority = 0,
                           domain = getDefaultReactiveDomain(),
-                          autoDestroy = TRUE) {
+                          autoDestroy = TRUE, ..stacktraceon = TRUE) {
```
wch: I'm not sure I like the name `..stacktraceon` for the argument name. I see the parallel with the function name, but I'm not sure it actually helps, since "stacktraceon" is already pretty distinctive.

R/render-plot.R
```diff
@@ -37,6 +37,7 @@ renderPlot <- function(expr, width='auto', height='auto', res=72, ...,
   } else {
     installExprFunction(expr, "func", env, quoted)
   }
+  func <- wrapFunctionLabel(func, "renderPlot", ..stacktraceon = TRUE)
```
wch: It might be useful to wrap other `render` functions.

R/reactives.R
```diff
@@ -779,14 +792,16 @@ Observer <- R6Class(
 #' @export
 observe <- function(x, env=parent.frame(), quoted=FALSE, label=NULL,
                     suspended=FALSE, priority=0,
-                    domain=getDefaultReactiveDomain(), autoDestroy = TRUE) {
+                    domain=getDefaultReactiveDomain(), autoDestroy = TRUE,
+                    ..stacktraceon = TRUE) {
```
wch: In what cases would you want `..stacktraceon` to be `FALSE`?

R/shiny.R
```diff
@@ -533,15 +533,15 @@ ShinySession <- R6Class(
       }
 
       if (is.function(func)) {
-        if (length(formals(func)) != 0) {
+        funcFormals <- formals(func)
+        func <- wrapFunctionLabel(func, paste0("output$", name), ..stacktraceon = TRUE)
```
wch: Just checking - is the intent here to wrap the original function, and not the one created in the `if` statement below?

R/reactives.R
wch: Oh, I see that it's used in below in `defineOutput`.

R/reactives.R
jcheng5: That's what I had before, but it looks quite unintuitive due to the file/line number not matching up. The srcref of the reactive name (`count`) is pointing to somewhere deep in Shiny. We can't drop that one from the stack trace either though; if a reactive expr caches an error and re-raises it when called later, then you need that `count` on the stack as a clue for which reactive raised the error.

R/reactives.R
jcheng5: Yep, I'll add a comment. It'd be even better to have a separate flag (like `.visible`) tracking whether it's an error and just storing the plain condition. But this is fine for now.

R/reactives.R
jcheng5: I want it to really stick out at the call site, because you'll be scanning the code for matching `..stacktraceoff..`/`..stacktraceon..` pairs. When you use these arguments then that's the substitute for `..stacktraceon`.

R/render-plot.R
jcheng5: Yeah it sure might. Do we want third party devs to be doing this too? It's getting complicated...

Maybe we should add this behavior as an additional argument to installExprFunction?

R/reactives.R
jcheng5: Yeah, just cases where the observer represents internal, not user, code.

R/shiny.R
jcheng5: Yeah. If we don't wrap the original, then `orig` shows up on the stack trace.

R/render-plot.R
jcheng5: I added it to installExprFunction. As long as functions are using that with default args they should be OK. I think I broke `[Shiny: Reactive]` labels in RStudio debugger though (do we even need those anymore?). cc @jmcphers

R/imageutils.R
yihui: @wch I used `tryCatch(..., finally)` intentionally on line 53. It will be too late to `par(op)` on exit. I think that caused #1020.

wch: @yihui, do you know why doing `par(op)` on exit is too late? I just want to understand the root of the problem better. Does it close the device before running the `on.exit` stuff?

@jcheng5 reverting to the `tryCatch` code should be ok, right? That is, it shouldn't interfere with the debugging changes.

yihui: The point of `plot.new()` is to guarantee at least a blank PNG when `renderPlot()` does not really contain code that generates a plot, but we need to make sure `plot.new()` always works. One situation in which it does not work is when the plot size is too small, and the error is "figure margin too large". The solution is set zero margins `par(mar = rep(0, 4))`. But we also have to make sure we do not let this `mar` setting affect user's plot, so we have to reset `par()` _immediately_ after `plot.new()`.

We can certainly use `on.exit()` if it looks cleaner, and we just need to introduce a helper function like

```r
blank_plot = function() {
  op = par(mar = rep(0, 4))
  on.exit(par(op), add = TRUE)
  plot.new()
}
```

Then we call `blank_plot()`.

wch: Got it, thanks. I think it makes sense to use the `tryCatch` as we did before (and @jcheng5 already pushed a fix with that).

PR #1014, opened 2015-11-16
added binary messages

daef:
websockets provide binary message type

to transfer large datasets json/base64 is not always a favourable solution.

this lets a shiny app register a custom handler in js and receive an arraybuffer, filled with a raw vector sent via session$sendBinaryMessage

an example app using this: http://github.com/daef/shinyThree


jcheng5:
I like the concept, but implementing it this way means that binary messages can only be used for a single purpose per app (since there is no way to tag these messages with a "type"). Perhaps binary messages could have a format like:

1. 4-byte little endian integer indicating the length of the type in bytes
2. n bytes (from 1.) containing UTF-8 encoded characters representing the type
3. The rest of the message is the payload

Or if we wanted to go further, maybe it'd even make sense to implement protobuf.js support :)

daef:
if you've read my implementation carefully you'll have noticed that one can stop propagating the message via 'return true' in the handler. (the caller will break it's loop early)

i think it's a common thing that one has one *big* dataset on a shiny app, i wanted to be able to handle the big blob myself - for obvious performance reasons. shiny doesn't provide anything into this direction so far, even thou i've found a handfull of ppl asking for similar things.

tagging with type and parsing those every time, simply as developer convenience feature, should be easily possible, but defies the purpose in my case.

daef:
updated so one can tag binary messages now.
sample repo got updated too.

daef:
what else are you missing?

PR #1025, opened 2015-12-07
Fixes #1003: correctly parse source code that can be represented with native encoding

### LINKED ISSUE #1003: file encoding issue in Windows related to ui.R and server.R
Hi, accoring to [shiny-article-unicode](http://shiny.rstudio.com/articles/unicode.html),

>For the sake of consistency and portability, Shiny requires the character encoding of all its components to be UTF-8, which include ui.R, server.R, global.R, DESCRIPTION, and/or README.md. 

However, in the latest dev version, those files seem to be encoded in native to function well (or the non ASCII character would be shown as a mess). I think it might related to one of your recent commit, since I regularly update my shiny version with the dev version.

#### ui.R
```r
library(shiny)
tagList(
  p("Chinese: 中文")
)
```

#### server.R
```r
function(input, output) {
  
}
```

### session info
```
> sessionInfo()
R version 3.2.2 (2015-08-14)
Platform: i386-w64-mingw32/i386 (32-bit)
Running under: Windows 7 x64 (build 7601) Service Pack 1

locale:
[1] LC_COLLATE=Chinese (Simplified)_People's Republic of China.936 
[2] LC_CTYPE=Chinese (Simplified)_People's Republic of China.936   
[3] LC_MONETARY=Chinese (Simplified)_People's Republic of China.936
[4] LC_NUMERIC=C                                                   
[5] LC_TIME=Chinese (Simplified)_People's Republic of China.936    

attached base packages:
[1] stats     graphics  grDevices utils     datasets 
[6] methods   base     

other attached packages:
[1] shiny_0.12.2.9005

 [1] R6_2.1.1         htmltools_0.2.9  tools_3.2.2     
 [4] rstudioapi_0.3.1 Rcpp_0.12.1      jsonlite_0.9.17 
 [7] digest_0.6.8     xtable_1.8-0     httpuv_1.3.3    
[10] mime_0.4 
```

yihui:
This is an improvement after the PR #968. When I was working on #968, I bumped into a weird issue: `parse(file, encoding = 'UTF-8')` would not work in certain case when the `file` was really encoded in UTF-8. This is probably a bug of `parse()`, and I provided a workaround in this PR by writing the source code to a temporary file with the native encoding, and parsing the temp file instead. I have tested it on my Windows machine with both English and Chinese locales for several apps with UTF8 characters.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -1073,14 +1073,14 @@ sourceUTF8 <- function(file, envir = globalenv()) {
   enc <- if (any(Encoding(lines) == 'UTF-8')) 'UTF-8' else 'unknown'
   src <- srcfilecopy(file, lines, isFile = TRUE)  # source reference info
   # oddly, parse(file) does not work when file contains multibyte chars that
-  # **can** be encoded natively on Windows (might be a bug in base R); we fall
-  # back to parse(text) in this case
-  exprs <- tryCatch(
-    parse(file, keep.source = FALSE, srcfile = src, encoding = enc),
-    error = function(e) {
-      parse(text = lines, keep.source = FALSE, srcfile = src, encoding = enc)
-    }
-  )
+  # **can** be encoded natively on Windows (might be a bug in base R); we
+  # rewrite the source code in a natively encoded temp file and parse it in this
+  # case (the source reference is still pointed to the original file, though)
+  if (isWindows() && enc == 'unknown') {
+    file <- tempfile(); on.exit(unlink(file), add = TRUE)
+    writeLines(lines, file)
+  }
+  exprs <- parse(file, keep.source = FALSE, srcfile = src, encoding = enc)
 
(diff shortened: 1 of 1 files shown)


jcheng5:
With this change, what happens to characters in the UTF-8 encoded file that are not representable in the system encoding?

yihui:
This PR does not affect that case. For a file containing UTF8 characters that are not representable in the system encoding, `parse(file, encoding = 'UTF-8')` works. This PR only addresses the case in which the UTF8 characters are representable in the system encoding.

jcheng5:
Before you do any future locale work, it'd be great if you could put together some tests (even if they can only be run, manually, under Windows, with certain locales, etc.). At this point it's impossible for anyone but you to work in this area of the code without a very high chance of introducing a regression. Thanks.

yihui:
Sure. I do have some tests on my machine that I had to manually run every time. I'll put them together tomorrow.

PR #1026, opened 2015-12-08
Fix #1010: updateSliderInput doesn't update hidden sliders

### LINKED ISSUE #1010: updateSliderInput doesn't update the value until the input is in view, which can cause problems
When using `updateSliderInput()` on a `sliderInput` that is in a tab that isn't currently visible, the slider doesn't actually get updated until it is visible.  This is not the case with other inputs I tested and causing problems if another input/output depends on that slider.

Example: this app has two tabs, each with a slider input. There is an output specifying what the value of the slider in tab2 is. Whenever the slider in tab1 is changed, the slider in tab2 is updated to match it. Note that the output *should* change whenever tab1 slider is moved, but it only gets updated after seeing tab2 slider.

```
library(shiny)
runApp(shinyApp(
  ui = 
    navbarPage(
      "title",
      tabPanel(
        "tab1",
        sliderInput("num1", "Set slider in tab2 to:", 1, 10, 1),
        textOutput("foo")
      ),
      tabPanel(
        "tab2",
        sliderInput("num2", "num", 1, 10, 1)
      )
    )
  ,
    output$foo <- renderText(sprintf("Number in tab2 is: %s", input$num2))
    observe({
      updateSliderInput(session, "num2", value = input$num1)
    })
  }
))
```

Note that if you put both sliders in the same tab, things work as expected. Also, if you change the slider input to a numeric input, then again things work as expected.

jcheng5:


jcheng5:
This should be good to go @wch

PR #1028, opened 2015-12-09
Add HTML template support

wch:
This PR provides support for HTML templates.

### Complete HTML documents

Here's an example template for a complete HTML document. It demonstrates a number of features, which are explained below.

```html
<html>
  <head>
    <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery/1.11.3/jquery.min.js"></script>
    {{ suppressDependency("jquery") }}
    {{ shinyHead }}

  </head>
  <body>
    <div>
      {{ button }}
      {{ sliderInput("x", "X", 1, 100, 50) }}
    </div>
  </body>
</html>
```

To use the template as the UI for an app, you could use this in your `app.R` and then call `runApp('path')`. You could also split the relevant pieces into `ui.R` and `server.R`.

```R
shinyApp(
  ui = htmlTemplate("template.html",
    button = actionButton("action", "Action")
  ),
)
```

The template consists of mostly HTML, but there are special code blocks within the `{{` and `}}`.

* Because this template is a complete HTML document, Shiny needs to know where to insert HTML HEAD stuff. This is what `{{ shinyHead  }}` is for.
* `suppressDependency("jquery")` tells Shiny to not include its version of jQuery, because it's already included in the HTML template.
* `{{ button }}` is replaced with `actionButton("action", "Action")`.
* Arbitrary R code can be inlined, as in `{{ sliderInput("x", "X", 1, 100, 50) }}`.

The way it works is that R code within `{{ }}` is evaluated in a special environment. This environment is  a *sibling* of the global environment, so that the template won't inadvertently get global vars when it's processed -- all variables must be passed in explicitly in the call to `htmlTemplate()`.

### HTML fragments

Templates can also be used for pieces of HTML that are inserted in a UI. For example, suppose you have this file `fragment.html`:

```html
<div class="foo">
  This is an HTML fragment named <code>{{ name }}</code>.
</div>
```

It could be used like so:

```{r eval=FALSE}
ui <- basicPage(
  h2("Inline template test"),
  htmlTemplate("fragment.html", name = "test-fragment")
)
shinyApp(ui = ui, server = function(input, output) {} )
```

###  Issues/questions:

* `{{ shinyHead }}` is a special case. In the first pass on the template, it's replaced with `<!-- SHINY_HEAD -->`, then in another pass, the actual head stuff is inserted there. We could simply get rid of the `{{ shinyHead }}` special case, and instead have users put `<!-- SHINY_HEAD -->` directly in the template.
* Because evaluation environment is created once, at the beginning of the template processing, and is a sibling of the global environment, if there are call to `library(foo)` in the template, that won't result in the foo package being visible to R code in the template. Possible solutions:
  * Create a new evaluation environment for each `{{ }}` block, but this will still require a separate block for `{{ library(foo) }}`, and the code that actually calls functions from foo. 
  * Require users to call `library(foo)` in `app.R`/`ui.R`, before calling `htmlTemplate()`.
  * Override `library()` in the evaluation environment so that it changes the evaluation environment's parent env. This might be a bad idea; the documentation for `parent.env<-` says, "The replacement function `parent.env<-` is extremely dangerous as it can be used to destructively change environments in ways that violate assumptions made by the internal C code. It may be removed in the near future."
  * Make the evaluation environment a child of the global environment.


wch:
Closing in favor of #1031.

PR #1029, opened 2015-12-10
Add knit_print.reactive

jcheng5:
Allows us to drop reactive expressions right into Rmd docs
with runtime:shiny and have them stay up-to-date.

    ---
    title: "reactive print"
    author: "Joe Cheng <joe@rstudio.com>"
    date: "December 9, 2015"
    output: html_document
    runtime: shiny
    ---

    ```{r}
    a <- reactive({
      invalidateLater(1000)
      Sys.time()
    })
    a
    ```


yihui:
Sounds like a good idea!

wch:
I'm not sure I'm on board with this -- it makes the behavior inside a knitr document very different from the console, where simply typing `a` prints out `a`.  Also, it invokes `a` without needing parens, like `a()`, which is also a break from normal behavior.

What about requiring an explicit function call that does the same thing? Something like `show_rx` (not a good name, but you get the idea).

jcheng5:
It's explicitly *not* an invocation of a; it's knit-printing the reactive expression object (i.e. all of its values over time), not its value.

It's true that it's different than the console, but I don't see why that's a problem--you can't write `renderText("foo")` at the console either. (And if it were possible to support either of these at the console, I think we would...)

BTW, this only works in a runtime:shiny document. If you try it in a non-Shiny document, RStudio will show you a nice dialog explaining what's about to break, and asks if you would like to convert to a Shiny document instead :)

jcheng5:
BTW, I struggle to think of legitimate reasons you should be composing/invoking/printing reactive expressions at the R console anyway, where there is no flushReact loop.

PR #1031, opened 2015-12-10
Use templates from htmltools

wch:
This adds support for HTML templates from htmltools 0.2.10. It is meant to go with rstudio/htmltools#33.


inst/template/default.html
```diff
@@ -0,0 +1,7 @@
+<!DOCTYPE html>
+<html>
+{{ headContent() }}
```
kevinushey: Is it intended that `headContent()` lies outside of the actual `<head>` tags?

wch: That's embarrassing. I've fixed it in ac95dcb.

PR #1034, opened 2015-12-14
Simpler fix for updating hidden sliders (#1010)

wch:
All that's needed is to add a listener for `update`. This also reverts the previous fix, #1026.


jcheng5:
Cool. Does this change the rate at which values are sent? i.e. does the slider get chattier when it's visible and moved by the user, than before this PR?

wch:
Update speed seems to be unchanged - I believe it's about 0.25 seconds after moving the slider.

PR #1035, opened 2015-12-15
Fix warnings when renderXXX is called with explicit namespace

jcheng5:
Calling shiny::renderText(...) instead of renderText(...) would
cause warnings.

```r
> invisible(shiny::renderText("foo"))
Warning message:
In assign(name, func, environment()) :
  only the first element is used as variable name
```

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -437,10 +437,15 @@ exprToFunction <- function(expr, env=parent.frame(2), quoted=FALSE,
 installExprFunction <- function(expr, name, eval.env = parent.frame(2),
                                 quoted = FALSE,
                                 assign.env = parent.frame(1),
-                                label = as.character(sys.call(-1)[[1]]),
+                                label = deparse(sys.call(-1)[[1]]),
                                 wrappedWithLabel = TRUE,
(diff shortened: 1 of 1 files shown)


jcheng5:
@wch I needed to merge this so I can reply to a customer on an overdue support ticket. Can you do a post-merge review please?

R/utils.R
```diff
@@ -437,10 +437,15 @@ exprToFunction <- function(expr, env=parent.frame(2), quoted=FALSE,
 installExprFunction <- function(expr, name, eval.env = parent.frame(2),
                                 quoted = FALSE,
                                 assign.env = parent.frame(1),
-                                label = as.character(sys.call(-1)[[1]]),
+                                label = deparse(sys.call(-1)[[1]]),
                                 wrappedWithLabel = TRUE,
                                 ..stacktraceon = FALSE) {
   func <- exprToFunction(expr, eval.env, quoted, 2)
+  if (missing(label) && length(label) > 1) {
```
wch: Is the `missing(label)` part necessary?

jcheng5: Oh--I suppose not. I'll fix that.

PR #1036, opened 2015-12-15
Fix validation error handling

jcheng5:
Validation errors were behaving too much like real errors: they were
being printed with stack traces, and passed to the options(shiny.error)
function. Also, if a reactive() cached a validation error, on future
calls the error would be re-raised (which is correct) without the
custom class names attached (which is not).


jcheng5:
@wch I needed to merge this so I can reply to a customer on an overdue support ticket. Can you do a post-merge review please? (number 2 of 2) Thanks!

PR #1037, opened 2015-12-16
Add `req` function for validating required inputs/values

jcheng5:


jcheng5:
I'm having second thoughts about this implementation. I didn't bother with non-standard evaluation for the `...` which means there's no short-circuiting. But it might be cool to be able to do:

```r
req(input$a, !is.na(input$a))
```

which will currently generate a warning.

jcheng5:
OK, should be good to go now.

R/utils.R
```diff
@@ -980,6 +980,105 @@ need <- function(expr, message = paste(label, "must be provided"), label) {
...
+}
+
+# Execute a function against each element of ..., but only evaluate each element
+# after the previous element has been passed to fun_. The return value of fun_
+# is discarded, and only invisible() is returned from dotloop.
+#
+# Can be used to facilitate short-circuit eval on dots.
+dotloop <- function(fun_, ...) {
+  for (i in 1:(nargs()-1)) {
+    fun_(eval(as.symbol(paste0("..", i)), environment()))
```
wch: I think you can leave out the `environment()` here.

wch:
Looks good!

Issue #1047, opened 2015-12-22
defer loading of iframes


aronatkins:
When a Shiny application contains a number of iframes, defer loading those iframe until after the parent application starts.

If the iframe content is allowed to load, those requests may be serviced before the parent application's server connection. In some cases, this will cause timeouts.

PR #1049, opened 2015-12-22
Defer subapp iframe loading

jcheng5:
This greatly improves robustness when apps with many subapps try to connect, especially when the app is busy serving other sessions.

Relevant issues:
https://github.com/rstudio/shiny/issues/1047
https://github.com/rstudio/connect/issues/1137


jcheng5:
Reviewed by @aronatkins

inst/www/shared/shiny.js
```diff
@@ -3477,7 +3466,6 @@ $(document).on('click', '.slider-animate-button', function(evt) {
           val.to = Math.min(slider.result.max, slider.result.to + slider.options.step);
 
         slider.update(val);
-        forceIonSliderUpdate(slider);
```
yihui: @jcheng5 Are the above changes about sliders intentional?

jcheng5: They were from an earlier commit to master that didn't get grunt-built properly. (Ideally we would have travis detect that condition)

yihui: I see. I think we can just run grunt on Travis and automatically push to the repo. That way, we no longer have to worry about these minified js files and map files. It is pretty easy to set up, but I guess you'd prefer running grunt by ourselves locally.

jcheng5: Well, right now Shiny won't run without the concatenated and/or minified JS, so we can't test changes locally without also running grunt locally. So we'd have to address that at the same time.

yihui: We do run grunt locally. I mean we don't include the output files in the PR's.

jcheng5: Oh. I thought you were hoping to get us out of needing to run grunt locally. If we're not aspiring for that, then the main advantage of having Travis do this is, what, that we avoid merge conflicts when two different PRs "simultaneously" change JS code?

yihui: Yes.

PR #1057, opened 2015-12-28
runGadget automatically handles cancel button

jcheng5:


R/server.R
```diff
@@ -802,21 +830,41 @@ runExample <- function(example=NA,
...
 #' @export
 runGadget <- function(app, server = NULL, port = getOption("shiny.port"),
-  viewer = paneViewer()) {
+  viewer = paneViewer(), stopOnCancel = TRUE) {
 
   if (!is.shiny.appobj(app)) {
     app <- shinyApp(app, server)
   }
 
+  if (isTRUE(handleCancel)) {
```
wch: Should this be `stopOnCancel`?

jcheng5: That's embarrassing.

R/server.R
```diff
@@ -802,21 +830,41 @@ runExample <- function(example=NA,
...
+}
+
+# Add custom functionality to a Shiny app object's server func
+decorateServerFunc <- function(appobj, serverFunc) {
+  origServerFuncSource <- appobj$serverFuncSource
+  appobj$serverFuncSource <- function() {
+    origServerFunc <- origServerFuncSource()
+    function(input, output, session) {
+      serverFunc(input, output, session)
+      if (length(formals(plot)) == 2)
```
wch: `plot`?

wch: Also, when we get to this line, has the user's server function already been checked for the number and name of arguments? I have a vague memory of the argument check happening when the app's server function is invoked, but I could be wrong. Even if it has already been checked for arguments, it would be helpful to have a comment about it here.

wch:
This is a corner case, but seeing that `.globals` is being modified got me thinking - what happens if there's more than one connection to the app, or if the user reloads the page?

R/server.R
jcheng5: Ugh. I must be a little rusty after the holidays ;)

I'm not sure what you mean by "already been checked". We're replacing `appobj$serverFuncSource` which is the function that returns the user's server function, so Shiny will check against the function that we return (it'll always have input, output, session).

It's origServerFunc that needs to be checked, so I'm doing that here (well, I would be if I hadn't copied-and-pasted `plot` instead of `origServerFunc`). The check that is normally performed is more complicated because it checks for both clientData and session, I haven't bothered with that here but I certainly could.

jcheng5:
More than one app or page reload are both irrelevant, because stopApp kills the server. I.e. it shouldn't be possible for two stopApp() calls to collide, if that's what you're worried about.

jcheng5:
What would be a problem is if you could have multiple *apps* running at the same time. We may need to support that if we want non-blocking gadgets, for example. If we go that direction then every use of `.globals` needs to revisited and refactored into something else.

R/server.R
wch: I meant that `origServerFunc` needs to be checked not only that it has two arguments, but that they're named `input` and `output`.

This is the code I was referring to:
  https://github.com/rstudio/shiny/blob/af0463e/R/server.R#L305-L315

Although, I suppose that simply invoking with named arguments would be sufficient, as in:

```R
origServerFunc(input = input, output = output)

origServerFunc(input = input, output = output, server = server)
```

I think `clientData` probably isn't necessary since we decided that `session$clientData` is preferred.

jjallaire:
This is currently what I see in the console when hitting the Cancel button:

screen shot 2015-12-28 at 4 09 02 pm

Whereas when I hit the Stop button the app just exits silently. Is there a way to get this to provide a more straightforward experience in the console? (note that I had done a devtools/install_github so some of this (i.e. the RStudio debugging overlay) may be an artifact of installing with source references.

jjallaire:
Note that the dialog viewer listens for the "disconnected" message and when this occurs closes the dialog and interrupts R. I wonder if we should have Cancel do this when it's running inside RStudio? (to eliminate the potentially confusing dialog chatter).

jcheng5:
Is there a programmatic way to do an interrupt from R?

jjallaire:
Not that I know of (stop is the only thing that behaves somewhat similarly)

wch:
I believe you could do it with `signalCondition` and an object like this:

```R
structure(list(), class = c("interrupt", "condition"))
```

The way I found this object is by doing this:

```R
# Run this and then press Esc or Ctrl-C to interrupt.
# z will be the interrupt condition object
z <- tryCatch(
  Sys.sleep(5),
  interrupt = function(e) {
    e
  }
)

str(z)
# list()
# - attr(*, "class")= chr [1:2] "interrupt" "condition"


# Test it to make sure it works
tryCatch(
  signalCondition(z),
  interrupt = function(e) {
    cat("Interrupt received")
  }
)
# Interrupt received
```

I think you could catch the error raised by `stop`, and then signal this interrupt.

jcheng5:
@wch I tried this earlier today, it doesn't seem to work:

```r
while (TRUE) {signalCondition(structure(list(), class = c("interrupt", "condition")))}
```

wch:
I may not be fully understanding the situation, but I had in mind something like this. You catch the `stop` and then throw an `interrupt`:

```R
tryCatch(
  
  tryCatch(
    { 
      Sys.sleep(1)
      stop("foo")
    },
    error = function(e) {
      # Throw an interrupt object, passing along the message
      e2 <- structure(
        list(message = e$message),
        class = c("interrupt", "condition")
      )
      signalCondition(e2)
    }
  ),

  interrupt = function(e) {
    cat(paste0("Interrupt received: ", e$message))
  }
)
# Interrupt received: foo
```

jcheng5:
@wch The goal is to break through the entire callstack (including whoever is calling runGadget, i.e. as part of a workflow) but not show the error at the console. I just talked to JJ by vchat though and he's fine with it as-is (I removed the call so at least it doesn't say "Error in force(returnValue): User cancel", it just says "Error: User cancel").

wch:
OK, sounds good, I'm OK with merging then.

PR #1059, opened 2015-12-30
Fix stopApp bug, and fix flex box in RStudio

jcheng5:
@wch can I get a quick code review for this? The "Return stopApp value without visibility info" in particular is a nasty one.


jcheng5:
Or if you're going to be AFK for a while I can merge it and you can review at your leisure.

PR #1067, opened 2016-01-05
Fix progress for Shiny modules

jcheng5:
Too-specific class check was being used, interfered with the
"duck typed" session-like objects we use for modules.


wch:
Looks good to me.

PR #1073, opened 2016-01-11
Tweaks for improved staticdocs rendering

jcheng5:
- \donttest -> \dontrun, otherwise staticdocs hangs
- Put NS before ns.sep so usage entries are in a better order


wch:
CRAN gave us grief before when we used `\dontrun` where we could have used `\donttest`. We ended up using `\donttest` and then wrapping the app examples with `if (interactive())`.

jcheng5:
Oh, OK. I'm going to leave this until after we go to CRAN if that's alright with you.

PR #1074, opened 2016-01-11
Reconnect

wch:
This PR adds the ability for Shiny apps to reconnect -- the existing input values are sent to a new session on the server side. It's designed to work with rstudio/shiny-server#440.

This reconnection behavior is opt-in. To enable, the app should call `session$allowReconnect(TRUE)`. Here's an example app:

```R
shinyApp(
  ui = fluidPage(sidebarLayout(
    sidebarPanel(
      sliderInput("bins", "Number of bins:", min = 1, max = 50, value = 30)
    ),
    mainPanel(
      plotOutput("distPlot")
    )
  )),
      hist(rnorm(100), breaks = input$bins)
    })

    # Set this to "force" instead of TRUE for testing locally (without Shiny Server)
    session$allowReconnect(TRUE)
  }
)
```


wch:
I've updated the code to use the notification API.  It now looks like this:

image

@jcheng5 we should discuss the UI and make it consistent with the UI from Shiny Server.

wch:
The only remaining thing is to implement `shiny.local.reconnect` option, to enable/disable reconnections when running locally. It's tricky because this option should be respected only when _not_ running under Shiny Server or Connect

wch:
@jcheng5 This should be good to go now.

PR #1080, opened 2016-01-15
Don't use sendCustomMessage for messages internal to Shiny

### LINKED ISSUE #1077: Small suggestion: make the custom JavaScript messages use more unique names to avoid collisions with users calling their own JS
I just looked at how `session$reload` [is implemented](https://github.com/rstudio/shiny/blob/877d7451ddc3a051450ae7965f0ed6885ecbd1d1/R/shiny.R#L723) (which is a great addition, thank you), and saw that it sends a message with the name `reload`.  I think it'd be a bit safer to use a name that's a bit more unique like `shiny:reload`, just to make sure that if a user ever tries writing a JavaScript function called `reload`, they wouldn't run into weird problems and be able to call their version.

Same goes for the other messages

wch:
This fixes #1077. `session$sendCustomMessage` was originally intended for app authors to use, not for internal Shiny code, and this PR removes internal uses of it. It also adds the convenience function `private$sendMessage`, which is a wrapper for `private$write`.


jcheng5:
LGTM, once it passes checks. Thanks for taking care of that.

PR #1099, opened 2016-02-06
Partial fix of debugger breakage

jcheng5:
There are two problems I'm trying to solve here.

1) Somewhere along the way, exprToFunction gained a hardcoded
   assumption that two stack frames up is a variable "expr",
   meaning anything that called installExprFunction had to have
   the first argument be exactly "expr". I think I got this
   fixed, now the only assumption made by both installExprFunc
   and exprToFunc is if they are called with quoted = FALSE,
   then the caller is merely passing through code that originated
   exactly one more level up the stack frame. If the code is
   less than one level up, i.e. an end user is directly passing
   code into installExprFunction or exprToFunction, then it won't
   work; and if the code is more than one level up (someone is
   passing code into function A which passes through to function
   B which calls installExprFunction, with quoted = FALSE) then
   it also won't work.

2) registerDebugHook calls were broken in various places by the
   name/envir registered with the hook being different than the
   name/envir through which the function was actually called.
   This generally seems fixable by moving the registerDebugHook
   call closer to the name/envir that will ultimately be called
   (e.g. call registerDebugHook directly from wrapFunctionLabel).

There still seems to be a problem here in that breakpoints in
RStudio are hit but then the IDE automatically runs "n" multiple
times. Also the unit tests don't currently pass, I haven't
investigated that yet.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -6,6 +6,11 @@ shiny 0.13.0.9000
 * Fixed #561: DataTables might pop up a warning when the data is updated
   extremely frequently.
 
+* Fixed RStudio debugger integration.
+
+* BREAKING CHANGE: The long-deprecated ability to pass functions (rather than
+  expressions) to reactive() and observe() has finally been removed.
+
 shiny 0.13.0
 --------------------------------------------------------------------------------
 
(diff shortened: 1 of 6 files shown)


jcheng5:
@jmcphers Total brain buster! Let's talk about this in person sometime next week.

PR #1100, opened 2016-02-06
Fix plot flickering in Safari and Firefox. Closes #776

### LINKED ISSUE #776: Images occasionally flicker in Firefox when updated
If you hit the button many times in the example below, in some cases you'll see the image flicker, in Firefox:

```R
library(shiny)

ui <- fluidPage(
  plotOutput("plot"),
  actionButton("newplot", "New Plot")
)
 
  output$plot <- renderPlot({
    input$newplot
    hist(jitter(rnorm(200)))
  })
}
 
shinyApp(ui, server)
```

I tried many, many variations of inserting the image, using absolute positioning, placing one image above another, and so on, but I could never get the flicker to go away. I believe it's actually a bug in Firefox:
https://bugzilla.mozilla.org/show_bug.cgi?id=705826

So we might have to use html canvas to work around this issue.

wch:
By switching from img to HTML canvas, plots no longer flicker. I've tested on Firefox, Safari, and Chrome.

Simple test app:

```R
library(shiny)

ui <- fluidPage(
  plotOutput("plot"),
  actionButton("newplot", "New Plot")
)

  output$plot <- renderPlot({
    input$newplot
    hist(rnorm(200))
  })
}

shinyApp(ui, server)
```


jcheng5:
Well, this was a pleasant surprise! :)

I think you need to clear the canvas before drawing though (try passing `bg = "transparent"` to renderPlot).

wch:
Nice catch! I've updated the code to clear the canvas on a redraw. I also streamlined some of the error handling/clearing -- previously that code was in `renderValue`, but I put it where it should be, in `renderError` and `clearError`.

inst/www/shared/shiny.js
```diff
@@ -1413,13 +1413,26 @@ $.extend(imageOutputBinding, {
...
+      $el.append(canvas);
+      $canvas = $(canvas);
+    } else {
+      // Trigger custom 'reset' event for any existing images in the div
+      canvas = $canvas[0];
+      $canvas.trigger('reset');
+    }
 
     if (!data) {
       $el.empty();
```
jcheng5: This will drop the canvas from the DOM, won't it? Is that what we want (vs. just `clearRect`)?

jcheng5:
The only downside I see from the user's perspective is 1) no compatibility with IE8, which is fine cause we've dropped support for IE8 anyway, and 2) the resizing behavior has gotten a bit worse, in terms of what it looks like before the plot is re-rendered. At the very least we should set `overflow:hidden` on `.shiny-plot-output`, which would stop scrollbars from temporarily appearing as you resizing the window to be narrower. Ideally we'd also scale the last-known-good image as we resize, too.

inst/www/shared/shiny.js
wch: It looks like I added this block in aed31c0eba9481b05ae5412a4271a452970fc541, but I actually don't know/remember the circumstances in which it will reach this line of code. I may have thought at the time that it would get here if there was an error, but that's not correct -- it doesn't get there if there's an error, since `renderError` gets called instead of `renderValue`.

Am I (now) missing some situation where it would get here?

wch:
For 2), how is the behavior different? For 0.13.0 and this PR, I see a scrollbar when making the window narrower.

Another possible drawback: in Safari, when you right click a canvas element, there's no "save image" option. In Chrome and Firefox, it does give the option to save to a png.

jcheng5:
Interesting, it looks like the resizing behavior changed between 0.10 and 0.11. No matter then, if nobody's noticed for this long.

The Safari save/copy image thing is a bit of a bummer. (I'm assuming it affects copy too?) I can imagine a solution where we float an `opacity:0` image over the top of the canvas...?

wch:
So I just tested out something that I can't believe we didn't try earlier. On master, we create a new img element and then replace the old one, and this flickers on Safari and Firefox. But if we use the same img element and simply change the value of `src`, it doesn't flicker, at least in my initial testing. It's not ready yet, because some details need to be taken care of, but this would continue IE8 support, and allow saving images from Safari.

wch:
Closing this PR in favor of #1104.

PR #1104, opened 2016-02-09
Fix plot flickering in Safari and Firefox, take 2. Closes #776

### LINKED ISSUE #776: Images occasionally flicker in Firefox when updated

wch:
Previously, a new `img` was added each time a new plot was received by the browser, but now it uses the same `img` tag and changes the value of `src` each time. This seems to fix the flickering in Safari and Firefox.


wch:
I've added a check so that it unsets attributes that were in the old img, but not present in the new data. This should be good to merge now.

PR #1105, opened 2016-02-09
Take advantage of new travis features

hadley:
This will cache package install between checks, which should make it run quite a bit faster - this is what @jimhester's has been working on

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,5 @@
 language: r
-warnings_are_errors: true
-
-r_binary_packages:
-  - Rcpp
-  - Cairo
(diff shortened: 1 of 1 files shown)


wch:
Looks good. Why is `warnings_are_errors` set to false now?

hadley:
It's not - the default has changed

jcheng5:
Need to merge this; the old `cairo` binary dependency doesn't work anymore.

wch:
FWIW, I had already fixed the `cairo` issue in master -- it changed to `Cairo`.

PR #1112, opened 2016-02-15
Enable plot replaying

wch:
When a plot is made in `renderPlot`, it records the display list with `recordPlot`. Later, if only the height or width has changed (and no other inputs), the plot can be replayed with `replayPlot`, which can give a speed boost, especially with ggplot2.

### DIFF
--- a/R/render-plot.R
+++ b/R/render-plot.R
@@ -57,14 +57,8 @@ renderPlot <- function(expr, width='auto', height='auto', res=72, ...,
   else
     heightWrapper <- NULL
 
-  # If renderPlot isn't going to adapt to the height of the div, then the
-  # div needs to adapt to the height of renderPlot. By default, plotOutput
-  # sets the height to 400px, so to make it adapt we need to override it
-  # with NULL.
-  outputFunc <- plotOutput
-  if (!identical(height, 'auto')) formals(outputFunc)['height'] <- list(NULL)
 
-  return(markRenderFunction(outputFunc, function(shinysession, name, ...) {
+  renderFunc <- function(shinysession, name, ...) {
     if (!is.null(widthWrapper))
       width <- widthWrapper()
     if (!is.null(heightWrapper))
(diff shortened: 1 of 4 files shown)


wch:
This should be good to go now.

R/render-plot.R
```diff
@@ -50,66 +55,169 @@ renderPlot <- function(expr, width='auto', height='auto', res=72, ...,
...
+        # Get coordmap error message if present
+        error = attr(coordmap, "error", exact = TRUE)
+      ))
+    }
+
+    img
+  }
+
+
+  render <- reactive({
```
jcheng5: Nit: I don't like verbs as names of reactive expressions. `render()` looks like it's a directive to render the plot, when the whole point of it is to *not* be that. A noun would be better.

R/render-plot.R
```diff
@@ -37,10 +37,15 @@
...
+#' @param replay If \code{TRUE} (the default), then when a plot is resized,
+#'   Shiny will \emph{replay} the plot drawing commands with
+#'   \code{\link[grDevices]{replayPlot}()} instead of re-executing \code{expr}.
+#'   This can result in faster plot redrawing, but there may be rare cases where
+#'   it is undesirable. If you encounter problems when resizing a plot, set this
+#'   to \code{FALSE}.
 #' @export
 renderPlot <- function(expr, width='auto', height='auto', res=72, ...,
-                       env=parent.frame(), quoted=FALSE, func=NULL) {
+                       env=parent.frame(), quoted=FALSE, func=NULL, replay=TRUE) {
```
jcheng5: Can we change the name of `replay` to something more to the effect of "do a full render on resize", and default it to `FALSE`?

wch:
I've made the changes to address the feedback: `render()` is now `plotObj()`, and the `replay` option is now `execOnResize`. I also tested that if there's an error in a replay, it will re-render when resized again.

PR #1116, opened 2016-02-19
Closes #810: check if the input file is encoded in UTF-8 and warn if not

### LINKED ISSUE #810: validUTF8() in R > 3.2.1
`shiny:::checkEncoding()` is not rigorous. We should use `validUTF8()` in the future.

https://github.com/wch/r-source/commit/1c5b9d61abc0344e5580dbed940bb71f582fde2a

yihui:
The validUTF8() function is still in R-devel, and they probably will never export it, so let's use iconv(x, from = 'UTF-8', to = 'UTF-8') to test if x is encoded in UTF-8.


wch:
Thanks!

PR #1117, opened 2016-02-19
Add ability to abort the processing of outputs

jcheng5:
`abortOutput()` leaves the state of the output unchanged, unlike `req()` (clears it), `validate()` (grey message), or `stop()` (red error message).

And help me find a better name than `abortOutput()` please!

This is a dumb example, it just halts the output if the value is less than 10. I made an error throw if n > 80 to demonstrate that abortOutput will also preserve errors, not just valid values.

```r
library(shiny)

ui <- fluidPage(
  sliderInput("n", "bins", 1, 100, 5),
  plotOutput("plot")
)

  output$plot <- renderPlot({
    if (input$n < 10) {
      abortOutput()
    } else if (input$n > 80) {
      stop("boom")
    }
    
    hist(runif(100), input$n)
  })
}

shinyApp(ui, server)
```


jcheng5:
cc @vnijs @hadley

hadley:
`useLastOutput()`? `leaveAsIs()`?, `dontChangeMe()`?

vnijs:
Nice!

`freeze()` or `freezeOutput()` ?

shrektan:
nice feature! Come up with a non-elegant name : 'valish', meaning the command vanishes here…

wch:
`cancel`, `cancelOutput`, `discard`, `discardOutput`?

Also, did you consider making it an option for req?

jcheng5:
I like cancel/discard-oriented ones--it makes it clear that we're talking about the current operation, i.e. there won't be lasting effects for future render calls.

I did consider making it an option for req, but as an alternative to having a separate function, and decided the separate function was better. But I don't know why I didn't consider _also_ making it available as a req option, that makes a lot of sense to me.

hadley:
Yeah, I like making it a `req()` option too.

vnijs:
@jcheng5 Just tried this out and it maintains the output perfectly! I was wondering, however, if it is possible to use/extend this functionality to maintain the value of a reactive as well? i.e., something like the snippet below. Now when i press a button (i.e., "glm_store_res") to store predictions from a glm the  model is re-estimated even though only the list of available variables changed.

```r
.glm_reg <- reactive({
  isolate({
    if (!is.null(input$glm_store_pred) && input$glm_store_pred > 0)
      abortOutput()
  })

  withProgress(message = 'Estimating model', value = 0,
    do.call(glm_reg, glm_inputs())
  )
})
```
I believe this would address my question on the shiny google group:

https://groups.google.com/forum/#!topic/shiny-discuss/NuUxPPG0fdo

vnijs:
If it is possible to maintain the value of reactives after a `req` or `abortOutput` this could also be useful to, for example, download plots and tables using functions that need access to the value of the 'old' reactive (e.g., pause estimation after running a model and still be able to download a plot or a set of predictions).

jcheng5:
@vnijs I don't think this mechanism will do what you want for reactives, and the correct fix would depend on what you're actually looking for. In your example, is the goal to avoid doing the work of `glm_reg`, or to avoid that PLUS whatever downstream work the observer or output was going to do?

jcheng5:
@wch This should be all set

wch:
@jcheng5 Can you add a NEWS item for this (on master)?

vnijs:
@jcheng5 `req` with the 'cancelOutput' parameter doesn't seem to work as I had expected. The following works to pause plotting so the user can change inputs without changing the plot, but the previous version of the plot is still shown:

```r
  if (is.null(input$viz_pause) || input$viz_pause == TRUE)
    abortOutput()
```

I expected the same would be possible with `req` now. `req` requires a truthy value to run the reactive and with `cancelOutput` I would expect the plot to still show when `pause == TRUE`. In other words, the reactive should update the plot when `pause == FALSE` and pause when `pause == TRUE`. Using the code below that doesn't happen, i.e., the plot disappears on 'pause'.
  
```r
req(input$viz_pause == FALSE, cancelOutput = TRUE)
```
Interestingly, when `cancelOutput` is set to the default value (i.e., FALSE) the plot is not shown at all, regardless of whether pause is set to TRUE or FALSE by the user.

req(input$viz_pause == FALSE, cancelOutput = FALSE)

I made a demo to show the results from using the 3 code snippets above:

http://www.youtube.com/watch?v=FLYJEtv9SLc

Code:
https://github.com/vnijs/radiant/blob/master/inst/base/tools/data/visualize_ui.R#L300-L304

jcheng5:
@vnijs I don't think you have the latest. The commit that added the `cancelOutput` parameter to `req` also renamed the `abortOutput` function to `cancelOutput`, so if `abortOutput` still works for you it means you need to reinstall from master. If not, let me know.

vnijs:
@jcheng5 

"@vnijs I don't think this mechanism will do what you want for reactives, and the correct fix would depend on what you're actually looking for. In your example, is the goal to avoid doing the work of glm_reg, or to avoid that PLUS whatever downstream work the observer or output was going to do?"

That is too bad. I'm not sure if everything downstream would need to stop in all cases but that would probably be fine. 

In the glm_reg example I provide `.glm_reg` is a reactive that calls non-reactive function `glm_reg` to run a logistic regression. `glm_reg` returns an output list to `.glm_reg` and I would like to access the 'old' version of that list even if a `req`-type condition is activated (e.g., because estimation was paused or a specific button is pressed). 

Demos in video linked below:
https://www.youtube.com/watch?v=NwzAyibiCls

jcheng5:
@vnijs I'm pretty sure we can give you that without any changes to shiny itself, just by writing a higher order reactive. I'll send you a gist later today.

vnijs:
That sounds great @jcheng5!

vnijs:
@jcheng5 I'm starting to wondering if it might not be easier to use an eventReactive. That said, it would still be nice to be able to switch to a normal reactive based on the value of a check-box or something.

jcheng5:
Here's my attempt. Use `reactiveWithPrev` like `reactive`, except instead of passing it a code block, pass it a function that takes a single argument `prevVal`, which is the value that was returned the last time the reactive executed *without* throwing an error (you could make it work in that case too if you wanted to, I wasn't clear on whether it was helpful for you and that impl would be harder to understand than this).

```r
reactiveWithPrev <- function(func, initialValue <- NULL) {
  val <- initialValue
  reactive({
    val <<- func(val)
    val
  })
}
```

Use it like this:

```r
.glm_reg <- reactiveWithPrev(function(prevValue) {
  if (!is.null(input$glm_store_pred) && input$glm_store_pred > 0)
    return(prevValue)

  withProgress(message = 'Estimating model', value = 0,
    do.call(glm_reg, glm_inputs())
  )
})
```

It'd be possible to write something similar for `eventReactive`, take a stab at it and let me know if you need help (hint: I'd fork the `eventReactive` code rather than wrapping it, probably).

PR #1126, opened 2016-03-04
Add code diagnostics for missing/extra commas, and for unmatched }, ), and ]

wch:
For example, with this ui.R:

```R
bootstrapPage(
  div("foo"),
  ,
  div(
    "foo"
    "bar"
  )
  123, NA, x[1],
  { 45 },
  mtcars[1,,FALSE]
))
```

It outputs:

```
> runApp('temp/commas')

Listening on http://127.0.0.1:6775
Error in parse(file, keep.source = FALSE, srcfile = src, encoding = enc) : 
  /home/winston/shiny/temp/commas/ui.R:8:5: unexpected string constant
7:     "foo"
8:     "bar"
       ^
Possible extra comma at:
5:  ,
    ^
Possible missing comma at:
8:    "bar"
      ^
Possible missing comma at:
10:  123, NA, x[1],
     ^
Possible unmatched ')' at:
13:))
    ^
Warning: Error in sourceUTF8: Error sourcing /home/winston/shiny/temp/commas/ui.R
Stack trace (innermost first):
    1: runApp [/home/winston/shiny/R/server.R#696]
```

If an error occurs sourcing ui.R, server.R, app.R, or global.R (anything that's sourced via `sourceUTF8`), shiny will run `diagnoseCode` on that file and print out this information.

Some issues that are worth considering:

* The first error message (with "unexpected string constant") is from R failing to parse the code; the remaining messages are from the new `diagnoseCode` function. It would be nice to not show R's error message and only show the messages from `diagnoseCode`, but I don't think that's doable in the general case because the code could have errored in any number of ways, not just the ones that the `diagnoseCode` function will identify.
* The diagnostics will be used for ui.R, server.R, and app.R (if an error occurs), but if any of those files `source` another file, that other file won't have diagnostics.
* The same applies to modules that are `source`d.

### DIFF
--- a/R/diagnose.R
+++ b/R/diagnose.R
@@ -1,0 +1,164 @@
+# Analyze an R file for possible extra or missing commas. Returns FALSE if any
+# problems detected, TRUE otherwise.
+diagnoseCode <- function(path = NULL, text = NULL) {
+  if (!xor(is.null(path), is.null(text))) {
+    stop("Must specify `path` or `text`, but not both.")
+  }
+
+  if (!is.null(path)) {
+    tokens <- sourcetools::tokenize_file(path)
+  } else {
+    tokens <- sourcetools::tokenize_string(text)
+  }
+
+  find_scopes <- function(tokens) {
+    # Strip whitespace and comments
+    tokens <- tokens[!(tokens$type %in% c("whitespace", "comment")),]
+
+    # Replace various types of things with "value"
+    tokens$type[tokens$type %in% c("string", "number", "symbol", "keyword")] <- "value"
+
+    # Record types for close and open brace/bracket/parens, and commas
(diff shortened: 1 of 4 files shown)


R/diagnose.R
```diff
@@ -0,0 +1,164 @@
...
+      tokens$scope,
+      tokens$err,
+      SIMPLIFY = FALSE,
+      FUN = function(type, prevType, nextType, scope, err) {
+        # If an error was already found, just return it. This could have
+        # happened in the brace/paren/bracket matching phase.
+        if (!is.na(err)) {
+          return(err)
+        }
+        if (is.na(scope)) {
```
wch: This could be pulled out into a separate function.

jcheng5:
(For the record, @wch and I code-reviewed this in person this morning)

So lovely to get this in!

kevinushey:
Awesome :smile:

PR #1129, opened 2016-03-11
Improved renderTable()

bborgesr:
renderTable() now looks prettier, taking advantage of the Bootstrap table styles. A bunch of new arguments were also added (or at least made explicit, since a few of these were already possible if passed to `xtable()` or `print.xtable()` through the `...` argument). Among other things, the improved function allows the user to select between different table styles, customize the table width and set the column alignment. An interactive demo of this function can be found [here](https://beta.rstudioconnect.com/barbara/shiny_table_demo/).


R/render-table.R
```diff
@@ -3,33 +3,103 @@
...
+#' class name \code{shiny-html-output}.
 #'
 #' @param expr An expression that returns an R object that can be used with
 #'   \code{\link[xtable]{xtable}}.
-#' @param ... Arguments to be passed through to \code{\link[xtable]{xtable}} and
-#'   \code{\link[xtable]{print.xtable}}.
+#' @param striped,condensed,hover,bordered Logicals: if \code{TRUE}, apply
+#'   the corresponding Bootstrap table format to the output table.
+#' @param width A string with the width of the table, expressed in any valid
+#'   CSS unit of length.
```
jcheng5: Numbers are valid here too, right? (Because it's being fed into `validateCssUnit`?)

I'd also add a couple of examples of valid units--see `plotOutput` for an example.

R/render-table.R
```diff
@@ -3,33 +3,103 @@
...
-#' name \code{shiny-html-output}.
+#' The corresponding HTML output tag should be \code{div} and have the CSS
+#' class name \code{shiny-html-output}.
 #'
 #' @param expr An expression that returns an R object that can be used with
 #'   \code{\link[xtable]{xtable}}.
-#' @param ... Arguments to be passed through to \code{\link[xtable]{xtable}} and
-#'   \code{\link[xtable]{print.xtable}}.
+#' @param striped,condensed,hover,bordered Logicals: if \code{TRUE}, apply
+#'   the corresponding Bootstrap table format to the output table.
```
jcheng5: Just to be clear, these are independently specified? (Like hover + striped is OK?)

R/render-table.R
```diff
@@ -3,33 +3,103 @@
...
+#' @param align A string that specifies the column alignment. If equal to
+#'   \code{'l'}, \code{'c'} or \code{'r'}, then all columns will be,
+#'   respectively, left-, center- or right-aligned. Otherwise, \code{align}
+#'   must have the same number of characters as the resulting table (if
+#'   \code{rownames = TRUE}, this will be equal to \code{ncol()+1}), with
+#'   the \emph{i}-th character specifying the alignment for the
+#'   \emph{i}-th column (again, only \code{'l'}, \code{'c'} or
+#'   \code{'r'} are permitted). If \code{NULL}, then all numeric/integer
+#'   columns (including the row names, if they are numbers) will be
+#'   right-aligned and everything else will be left-aligned.
```
jcheng5: Did you consider taking a vector of characters instead of a single string with col-per-letter? This feels less idiomatic R to me than that, though I like that it's concise. (I think @hadley does something similar in readr for col types IIRC) Just wondering what your thought process was.

jcheng5: Maybe you could also have `a` for auto or `d` for default?

R/render-table.R
bborgesr: Yes. Initially, I just add a format=c("striped", "condensed", etc), but when we discovered that Bootstrap allowed us to combine multiple table classes, we decided to separate them out, so that the user can choose the styles independently.

R/render-table.R
```diff
@@ -3,33 +3,104 @@
...
+#' @param rownames,colnames Logicals: include rownames? include colnames
+#'   (column headers)?
+#' @param digits An integer specifying the number the decimal places for
```
jcheng5: Second "the" should change to "of"

R/render-table.R
jcheng5: Ok. They could've still chosen independently with a format arg by providing multiple values though, right?

R/render-table.R
bborgesr: This is the way that `align` in xtable() works, so I thought it may be good to keep it consistent. Especially given that some users may already have been passing that argument for the old renderTable() function through the ellipsis argument. It won't be difficult to do it the way you are suggesting though, so we can definitely think if that would be a better way...

R/render-table.R
jcheng5: (I'm not just being pedantic, imho it's a little bit of an art deciding whether related but not mutually exclusive options like this are multiple logical args or a single multi-value one)

R/render-table.R
jcheng5: Oh that's totally fine then. Thanks for the clarification.

R/render-table.R
bborgesr: Updated: with some very minor changes, I just copied the documentation from `plotOutput`. It's better now and has the added perk of keeping arguments as consistent as possible across shiny functions.

R/render-table.R
```diff
@@ -3,33 +3,104 @@
...
   if (!is.null(func)) {
     shinyDeprecated(msg="renderTable: argument 'func' is deprecated. Please use 'expr' instead.")
   } else {
     installExprFunction(expr, "func", env, quoted)
   }
 
+  # A small helper function to create a wrapper for an argument that was
+  # passed to renderTable()
+  createWrapper <- function(arg) {
+    if (is.function(arg)) wrapper <- reactive({ arg() })
```
jcheng5: What was the thinking behind making this a reactive instead of just returning arg as-is?

R/render-table.R
bborgesr: Sure, that would be possible too. I guess this way felt more natural to me, but I agree that 4 args instead of 1 may be a little overwhelming to the user. Now, I'm quite torn between the two approaches... Winston, thoughts?

R/render-table.R
```diff
@@ -3,33 +3,104 @@
...
+    digits <- digitsWrapper()
+    na <- naWrapper()
+
+    # For css styling
+    classNames <- paste0("table shiny-table")
+    for (i in seq_len(length(format))) {
+      if (format[i]) {
+        classNames <- paste0( classNames, " table-", names(format)[i])
+      }
+    }
```
jcheng5: I think there's a vectorized one-liner that can replace this for loop, can you try?

R/render-table.R
```diff
@@ -3,33 +3,104 @@
...
+
     data <- func()
 
-    if (is.null(data) || identical(data, data.frame()))
-      return("")
+    # Return empty string if no data is provided
+    if (is.null(data) || identical(data, data.frame())) return("")
```
jcheng5: I see that you're returning empty string because that's what the old code did. That was extremely early code I believe, these days it's more idiomatic shiny to return NULL.

R/render-table.R
```diff
@@ -37,23 +108,96 @@ renderTable <- function(expr, ..., env=parent.frame(), quoted=FALSE, func=NULL)
     xtable_args <- dots[intersect(names(dots), xtable_argnames)]
     non_xtable_args <- dots[setdiff(names(dots), xtable_argnames)]
 
-    # Call xtable with its args
+    # A small helper function to determine if the row.names can be coerced
+    # to numeric or if they are legitimate strings
+    isNumber <- function(n) {
+      !(suppressWarnings(is.na(all(n == as.character(as.numeric(n))))))
```
jcheng5: Shouldn't this really be using as.integer? Or `grepl("^\\d$+", n)`?

R/render-table.R
bborgesr: It's the same construction used in the `renderPlot()` function (see below). Winston and I were actually thinking if it would be worth it to make `createWrapper()` a utils function, since it's being used in these two places (and we could also add it to other, I guess). Basically it allows the argument to be passed as a (reactive) function, which is what I need to make by [table demo app work](https://beta.rstudioconnect.com/barbara/shiny_table_demo/).

===============
```
if (is.function(width))
    widthWrapper <- reactive({ width() })
  else
    widthWrapper <- function() { width }

  if (is.function(height))
    heightWrapper <- reactive({ height() })
  else
    heightWrapper <- function() { height }
```

R/render-table.R
```diff
@@ -37,23 +108,96 @@ renderTable <- function(expr, ..., env=parent.frame(), quoted=FALSE, func=NULL)
...
+    # Figure out column alignment
+    ## Case 1: if align=NULL, check if row.names are numbers. If not, make
+    ## sure to left-align them (xtable right-aligns them by default, which
+    ## looks weird when the row.names are strings).
+    if (is.null(align)) {
+      cols <- if (isNumber(row.names(data))) "r" else "l"
+      for (i in seq_len(ncol(data))) {
+        cls <- class(data[,i])
+        if (cls == "numeric" || cls == "integer") cols <- paste0(cols, "r")
+        else cols <- paste0(cols, "l")
```
jcheng5: Incrementally pasting together a string like this in a for loop is a bad code smell, lots of unnecessary copies being made in memory. Numbers not large enough to matter here but still a bad habit. Can you do something vectorized and/or functional instead?

R/render-table.R
bborgesr: Done

R/render-table.R
jcheng5: I'm not sure why we did that in renderPlot. On the pro side it will cache the value (actually not totally sure if that's good or bad). On the bad side it will add noise to the reactive graph.

R/render-table.R
bborgesr: I'm not sure this is actually possible in here because I have to check the truth value of each element in `format` (we're only adding the class if the corresponding arg is set to `TRUE`). The closest I got is `paste0(" table-", names(format), collapse='')`, but this adds all classes and doesn't check each arg's truth value. If I try to do something like `paste0(" table-", if (format) names(format), collapse='')`, then it only checks the truth value of the first element and throws the familiar warning: `the condition has length > 1 and only the first element will be used`. Did I miss something?

R/render-table.R
bborgesr: Yes, that is definitely a con worth considering... But don't you think it's useful to be able to do something like `renderTable({ head(datasetInput(), n = input$obs)}, width = reactive(input$width))`? Since `input$width` is outside the reactive context here (notice that although it is an argument to `renderTable()`, it only comes after the closing of the curly brace), I don't see another way to make this possible...

R/render-table.R
jcheng5: Maybe `names(format[format])` or `names(format)[format]` or `na.omit(ifelse(format, names(format), NA))`? I feel like at least one of those should work

R/render-table.R
bborgesr: Yes, I don't know what I was thinking... regex is 1000x more elegant

R/render-table.R
bborgesr: Done

R/render-table.R
jcheng5: With that example and this code won't the effect be to double-wrap, I.e. reactive(reactive(input$width))? Anyway, this is a very subtle thing, especially if you don't yet have a deep understanding of our reactivity stuff--lets talk about it in person.

R/render-table.R
bborgesr: `names(format)[format]` did the trick. Thanks!

R/render-table.R
hadley: I think it would be nice to still add a character that means default behaviour. Maybe `_`?

R/render-table.R
hadley: I think using format with `match.arg(format, several.ok = TRUE)` would be reasonable, and should be reasonably familiar to people.

R/render-table.R
```diff
@@ -37,23 +104,96 @@ renderTable <- function(expr, ..., env=parent.frame(), quoted=FALSE, func=NULL)
...
     xtable_res <- do.call(xtable, c(list(data), xtable_args))
 
     # Set up print args
     print_args <- list(
       xtable_res,
-      type = 'html',
-      html.table.attributes = paste('class="', htmlEscape(classNames, TRUE),
-                                    '"', sep='')
-    )
+      type='html',
```
hadley: Missing spaces around `=`?

R/render-table.R
hadley: This is a dangerous test for class because the class isn't always a length 1 vector. It's safer to use `inherits()`, or in this case I think you could just use `is.numeric()`.  I'd do something like:

```R
is_num <- vapply(data, is.numeric, logical(1))
cols <- ifelse(is_num, "r", "l")
```

R/render-table.R
jcheng5: In readr doesn't `_` mean ignore? That's the only reason I didn't suggest it.

R/render-table.R
```diff
@@ -37,23 +104,96 @@ renderTable <- function(expr, ..., env=parent.frame(), quoted=FALSE, func=NULL)
...
+    )
+
+    # All further processing concerns the table headers, so we don't need to run
+    # any of this if colnames=FALSE
+    if (colnames) {
+      # Make sure that the final html table has a proper header (not included
+      # in the print.xtable() default)
+      tab <- sub("<tr>", "<thead> <tr>", tab)
+      tab <- sub("</tr>", "</tr> </thead> <tbody>", tab)
+      tab <- sub("</table>", "</tbody> </table>", tab)
```
jcheng5: I'm slightly worried about this being too naive, especially the `</table>` because it comes after an entire body full of user-generated content. I'd like to make sure it's the last `</table>`. Could you do a single sub statement, using capturing groups and backreferences? `sub("^(.*)<tr>(.*)</tr>(.*)</table>$", "\\1<thead><tr>\\2</tr></thead><tbody>\\3</tbody></table>", tab)` (untested)

jcheng5: Oh or just keep it the way it is and change `"</table>"` to `"</table>$"`. heh.

jcheng5:
Oh, thought of a couple more things. Can you also verify that this looks good in 1) an R Markdown `runtime:shiny` document, 2) an R Markdown `runtime:shiny` ioslides presentation, and 3) shinydashboard? Just to make sure there no clashing styles.

jcheng5:
Other than that--LGTM! Nice work!

R/render-table.R
hadley: Oh yeah, I guess the `readr` equivalent would be `?`

R/render-table.R
wch: I'm a bit ambivalent about which way to go, but I'm more on the side of having separate arguments. 

The one drawback is that having four separate arguments is a bit unwieldy, but the benefits are that:
* The arguments are easily discoverable with tab-completion.
* Conceptually the options are independent from each other.
* Non-binary values are possible for these options (which might matter for `condensed`).
* We're currently making `condensed` default to `TRUE`, and with a char vector argument, if someone wanted to keep that default and also use another option, they'd have to do something like `format = c("condensed", "bordered")`.

A real-world example of a function that takes a char vector is `install.packages`: the `dep` argument can be a subset of `c("Depends", "Imports", "LinkingTo", "Suggests", "Enhances")`. I can never remember what the choices are and I always have to look it up in the documentation.

R/render-table.R
wch: I feel like `NULL` is a relatively common way of saying to use the default -- @hadley, is `_` something you've been doing recently?

R/render-table.R
```diff
@@ -3,33 +3,100 @@
...
+    # For css styling
+    classNames <- paste0("table shiny-table",
+                         paste0(" table-", names(format)[format], collapse='' ))
+
     data <- func()
 
-    if (is.null(data) || identical(data, data.frame()))
-      return("")
+    # Return NULL if no data is provided
+    if (is.null(data) || identical(data, data.frame())) return(NULL)
```
wch: This isn't an issue introduced by this PR, but it's probably worth fixing now. The `identical(data, data.frame())` check is a bit too strict. It's possible for an empty data frame to have an additional class, like a `tbl_df` from dplyr, and in that case, `identical()` will return false. For example:

```R
x <- dplyr::tbl_df(data.frame())
identical(x, data.frame())
# FALSE
```

So the `identical(....)` part should be replaced with something like:

```R
(is.data.frame(data) && nrow(x) == 0 && ncol(x) == 0)
```

R/render-table.R
```diff
@@ -37,23 +104,96 @@ renderTable <- function(expr, ..., env=parent.frame(), quoted=FALSE, func=NULL)
     xtable_args <- dots[intersect(names(dots), xtable_argnames)]
     non_xtable_args <- dots[setdiff(names(dots), xtable_argnames)]
 
-    # Call xtable with its args
+    # A small helper function to determine if the row.names can be coerced
+    # to numeric or if they are legitimate strings
+    isNumber <- function(n) {
+      grepl("^\\d$+", n)
```
wch: Two things:

* I think it should be `grepl("^\\d+$", n)`. Right now:

```R
isNumber("12")
# FALSE
```
 
* For the use of `isNumber` below, you need `all()` somewhere. For example:

```R
if (isNumber(c("1", "A"))) print("number")
# [1] "number"
# Warning message:
# In if (isNumber(c("1", "A"))) print("number") :
#   the condition has length > 1 and only the first element will be used
```

So I'd suggest renaming this to `allNumbers` and use `all(grepl(....))` inside.

R/render-table.R
wch: I think this could be done a bit more conservatively, by not `paste`ing the result from `capture.output` above. Then the first row with a `<tr>` could be wrapped in `<thead>`; something similar could be done for the `<tbody>`. This would also probably be more efficient.

R/render-table.R
```diff
@@ -37,23 +104,96 @@ renderTable <- function(expr, ..., env=parent.frame(), quoted=FALSE, func=NULL)
...
+      # Make sure that the final html table has a proper header (not included
+      # in the print.xtable() default)
+      tab <- sub("<tr>", "<thead> <tr>", tab)
+      tab <- sub("</tr>", "</tr> </thead> <tbody>", tab)
+      tab <- sub("</table>", "</tbody> </table>", tab)
+
+      # Update the `cols` string (which stores the alignment of each column) so
+      # that it only includes the alignment for the table variables (and not
+      # for the row.names)
+      cols <- if (rownames) cols else substr(cols,2,nchar(cols))
```
wch: Spaces after commas here.

R/render-table.R
```diff
@@ -37,23 +104,96 @@ renderTable <- function(expr, ..., env=parent.frame(), quoted=FALSE, func=NULL)
...
+
+      # Create a vector whose i-th entry corresponds to the i-th table variable
+      # alignment (substituting "l" by "left", "c" by "center" and "r" by "right")
+      header_alignments <- c()
```
wch: This is equivalent to `header_alignments <- NULL`. You probably would want to use `character(0)` instead. But this line isn't actually be necessary in this case.

R/render-table.R
```diff
@@ -37,23 +104,96 @@ renderTable <- function(expr, ..., env=parent.frame(), quoted=FALSE, func=NULL)
...
+      for (i in seq_len(nchar(cols))) {
```
wch: I think it would be better to do something like:

```R
cols <- strsplit(cols, "")[[1]]
cols[cols == "l"] <- "left"
cols[cols == "r"] <- "right"
cols[cols == "c"] <- "center"
```

R/render-table.R
```diff
@@ -37,23 +104,96 @@ renderTable <- function(expr, ..., env=parent.frame(), quoted=FALSE, func=NULL)
...
+          if (substr(cols, i, i) == "l") "left"
+          else if (substr(cols, i, i) == "c") "center"
+          else "right"
+        }
+      }
+
+      # Align each header accordingly (this guarantees that each header and its
+      # corresponding column have the same alignment)
+      for (i in seq_len(nchar(cols))) {
+        tab <- sub(" <th",
```
wch: This use of `sub` in a loop feels a little fragile to me. If xtable were to generate th elements with two spaces in front of them, each one would get matched twice as the for-loop proceeds. Can you search for `"<th>"` instead? That way you're guaranteed not to match after adding a `style` attribute.

inst/www/shared/shiny.css
```diff
@@ -3,12 +3,9 @@ body.disconnected {
...
 }
 
-table.data {
-  width: auto;
-}
-table.data td[align=right] {
-  font-family: monospace;
-  text-align: right;
+.shiny-table th {
+  padding-left: 12px !important;
```
wch: Why is the `!important` necessary"

R/render-table.R
hadley: We're talking about individual elements of a string, like "???rrll"

R/render-table.R
hadley: It would be better to check the underlying `row.names()` attribute:

```R
df1 <- data.frame(x = 1:2)
is.numeric(attr(df1, "row.names"))
df2 <- data.frame(x = c(a = 1, b = 2))
is.numeric(attr(df2, "row.names"))
```

wch:
It just occurred to me -- what if the input is a matrix? I can think of some parts of the code that won't work for matrices. I think it would be OK to simply coerce to a data frame with `as.data.frame()`.

R/render-table.R
kevinushey: Or even use `is.numeric(.row_names_info(df, 0L))`.

R/render-table.R
bborgesr: Ok, I inserted spaces for this chunk of code now, but I feel like I don't quite grasp our coding style yet. For example, the arguments to the function itself (lines 46-49: `env=parent.frame()` etc.) don't have spaces. The very first lines of code (51: `msg="renderTable: argument...`) and some other places (ex. 161: `collapse=`\n`) don't either. These examples are all from the master branch, and I was trying to make it consistent with that (I inferred that the rule was: usually operators have spaces around them, but not if they're function arguments). Could someone please clarify this? (Personally, I also prefer having spaces, especially in this code chunk because it gets quite dense otherwise.)

R/render-table.R
bborgesr: @wch I'm not sure I understand what you're suggesting specifically... @jcheng5 Following your suggestion, I changed `"</table>"` to `"</table>$"`, but I think the second command (line 170) still falls under the naive territory (I'm not worried about the previous line because it's impossible for the user to enter anything before the first `"<tr>"`). However I suppose it is theoretically possible for one of the headers to be the string `"</tr>"`. How do we fix this?

R/render-table.R
bborgesr: Done

R/render-table.R
bborgesr: I agree that it may be better to have `header_alignments <- character(0)`, although `header_alignments <- c()` (or, as you pointed out, its equivalent `header_alignments <- NULL`) also does the trick. I think it's nice to have `header_alignments <- c()` instead of the equivalent `header_alignments <- NULL` because it primes the reader for the fact that `header_alignments` will be a vector, but I agree that it is obscuring reality a little bit. Anyway, I changed it to `character(0)` for now.

I don't understand your second point though. How is this line not necessary? If the variable is not defined at all, I can't use `header_alignments[i]` in line 182, right? For example, here's a small console example:

```
> vec                 ## making sure `vec` is not defined
Error: object 'vec' not found
> vec[1] <- "hello"   ## trying to assign a value to the first slot of `vec`
Error in vec[1] <- "hello" : object 'vec' not found
```

bborgesr: Never mind... I just read your next comment :/

R/render-table.R
bborgesr: Agreed. Done

R/render-table.R
bborgesr: Agreed. I changed it now. It also has the added advantage of feeling slightly less hack-ish :)

bborgesr:
@wch So you're suggesting changing `data <- func()` to `data <- as.data.frame(func())` (line 96), right? I also think this should be ok. I'll do that for now. If anyone thinks this introduces an issue, let me know.

R/render-table.R
bborgesr: Agreed and done. It's interesting: I thought that changing the previous line from `data <- func()` to `data <- as.data.frame(func())` would take care of this issue, but it didn't...

```{r}
x <- dplyr::tbl_df(data.frame())
str(x)
# Classes ‘tbl_df’, ‘tbl’ and 'data.frame':	0 obs. of  0 variables
x <- as.data.frame(x)
str(x)
# 'data.frame':	0 obs. of  0 variables
str(data.frame())
# 'data.frame':	0 obs. of  0 variables

identical(x, data.frame())
# FALSE
```

bborgesr:
Ok, so apparently this did introduce an issue: the integration tests at "shiny/tests/testthat/test-stacks.R" (lines 42-77) are failing. This is why. The function is expecting the following output:

```
   num                                     call loc
1   31                                        A    
2   30                                        B    
3   29                               reactive C    
4   18                                        C    
5   17                              renderTable    
6   16                                     func    
7   15 renderTable({     C() }, server = FALSE)    
8    8                                  isolate    
9    7                      withCallingHandlers    
10   6                       captureStackTraces    
11   5                               doTryCatch    
12   4                              tryCatchOne    
13   3                             tryCatchList    
14   2                                 tryCatch    
15   1                                      try 
```

But it is getting something slightly different after I changed `data <- func()` to `data <- as.data.frame(func())` (the difference is line 7):

```
   num                                     call loc
1   32                                        A    
2   31                                        B    
3   30                               reactive C    
4   19                                        C    
5   18                              renderTable    
6   17                                     func    
7   16                            as.data.frame    
8   15 renderTable({     C() }, server = FALSE)    
9    8                                  isolate    
10   7                      withCallingHandlers    
11   6                       captureStackTraces    
12   5                               doTryCatch    
13   4                              tryCatchOne    
14   3                             tryCatchList    
15   2                                 tryCatch    
16   1                                      try    
```

While I'm not sure exactly how this table is being generated, it seems to me that the issue is probably in the test, not in the `renderTable()` function. If I'm correct, @jcheng5 , @wch, how do we fix this?

R/render-table.R
bborgesr: Done (using `z`)

inst/www/shared/shiny.css
bborgesr: Because there is a bunch of Bootstrap padding defaults. If the `!important` is not included, these lines are ignored in the final rendered page...

jcheng5:
@bborgesr (I can't find the comment about `!important` to reply to it.) Using `!important` is the biggest hammer we have, best to leave it to users if they want to override our stuff. The usual way I deal with this is by providing a more-specific CSS selector than Bootstrap's rule, then ours will apply. Usually this means adding more tag and/or class names. https://www.smashingmagazine.com/2007/07/css-specificity-things-you-should-know/

R/render-table.R
bborgesr: Done

jcheng5:
LGTM.

PR #1133, opened 2016-03-15
Add tooling for ES6

wch:
I believe that tools/README.md includes all the information needed to make this work. For developoers with an existing `tools/node_modules` directory, they may need to delete that dir first, and then run `npm install`. Also, this may require a newish version of Node and NPM to work properly, though I'm not completely sure.

@jcheng I copied your eslint rules from the shiny-server-client repo. I didn't copy options from the `.eslintrc.es5.js` file -- should they be used here?


jcheng5:
@wch The ES5 rules were only needed because one particular folder of that module has to be written directly in ES5, not ES6 (it doesn't get transpiled and needs to work on Node v0.10). So you don't have to worry about it.

PR #1141, opened 2016-03-22
Notification interface

wch:
(Not ready for merge.) This adds a notification API for R and Javascript.

Basic R usage:

```R
shinyApp(
  ui = fluidPage(
    actionButton("show", "Show"),
    actionButton("remove", "Remove")
  ),
    id <- NULL
    observeEvent(input$show, {
      id <<- showNotification("Notification text")
    })
    observeEvent(input$remove, {
      removeNotification(id)
    })
  }
)
```

Messages display by default for 5 seconds. This time is customizable, and if set to NULL, the message will not automatically disappear; the user would have to click on the X to make it go away.

In Javascript, there are the following functions:

```js
Shiny.notifications.show()
Shiny.notifications.remove()
Shiny.notifications.removeAll()
Shiny.notifications.get() // Get the DOM object for a notification
Shiny.notifications.ids() // Return IDs of all notifications
```

There are four styles: "default", "message", "warning", and "error"; these correspond to R message types. The colors for all but "default" are the same as for Bootstrap's "info", "warning", and "danger" classes. Screenshot:

image


wch:
If the API looks OK, we could merge this now so that we can move forward with other things that depend on the notifications API, like reconnections. Styling and polish work could happen independently of that.

bborgesr:
I think it looks great :) Really like the colors!

srcjs/notifications.js
```diff
@@ -0,0 +1,160 @@
...
+    }
+  }
+
+  return {
+    show,
+    remove,
+    removeAll,
+    get,
+    ids
+  };
```
jcheng5: I found this style a bit confusing, as all these methods just look like regular functions. (Did you consider ES6 classes?)

srcjs/notifications.js
```diff
@@ -0,0 +1,160 @@
+exports.notifications = (function() {
+
+  // Milliseconds to fade in or out
+  const fadeDuration = 250;
+
+  function show({ html=null, duration=5000, id=null, closeButton=true,
+                  style = null } = {})
```
jcheng5: Have we called this kind of option (partial CSS class name) `style` in the past? To me it feels like an option called `style` should let you stuff CSS properties in there. `show({style: "font-weight: bold;"})`

srcjs/notifications.js
```diff
@@ -0,0 +1,160 @@
...
+    });
+  }
+
+  function removeAll() {
+    ids().forEach(id => remove(id));
+  }
+
+  // Returns an individual notification DOM object (wrapped in jQuery).
+  function get(id) {
+    return _getPanel().find('#shiny-notification-' + id);
```
jcheng5: I think `id` should be `$escape(id)`

srcjs/notifications.js
wch: I did look at ES6 classes, but I didn't think it was a good fit here, since this is a singleton object. The approach I used was for "public" functions to have normal names, and for "private" functions to start with an underscore. But I'm not necessarily tied to that particular style.

jcheng5:
I think the notification ui object should be treated as full UI objects, just like `htmlOutput`/`uiOutput`. That means we need a code chunk like this:

```javascript
exports.renderHtml(html, el, dependencies);
exports.initializeInputs(el);
exports.bindAll(el);
```

And also `unbindAll(el)` when the notification is dismissed. And bootstrap show/shown/hide/hidden events ought to be triggered as well.

I don't know whether this housekeeping stuff should be done as part of the notification object, or by the code in Shiny that calls the notification object. I'd love for notification object implementers not to have to know about this, but I don't think we can completely get around it because it's pretty abstract to the rest of Shiny how those notifications are represented, right? (i.e. we wouldn't even know how to find the right DOM nodes?)

srcjs/notifications.js
bborgesr: I think @jcheng5 has a good point. Currently, some of the Shiny inputs can take a `style` argument (actually passed in through `...`) that lets you set any CSS thing... For example, this is totally valid (taken from a [stackoverflow answer](http://stackoverflow.com/a/35871042)):

```{r}
actionButton("run", "Run Analysis", icon("paper-plane"), 
             style="color: #fff; background-color: #337ab7; border-color: #2e6da4")
```

So, `style` might be confusing... I think using `class` would be okay... App authors are probably familiar with the bootstrap classes already...

srcjs/notifications.js
jcheng5: I know how ridiculous this is going to sound coming from me, but in general even if an object is going to be a singleton, it's a good idea to make it a class or at least to have some kind of factory for creating them, for unit testing purposes. I know that we're really far from having good unit tests on the JS side but after the experience I had with shiny-server-client I think I've finally got the JS automated testing religion :)

jcheng5: After 17 minutes of reflection, I realize my comments here are a bit bikesheddy (especially in the absence of us actually doing unit tests or having coding standards) and we can always change it later if we want to.

srcjs/notifications.js
jcheng5: `class` is a bit misleading too, since the value you pass it won't literally be the CSS class, as we prepend it with `"shiny-notification-"` or whatever... what about "type" or "level" (like a log level)?

bborgesr: I like `type` because I think it's how you'd naturally think about this -- as in "what's the type of this notification?" and the answer (info, warning, ...) makes sense.

srcjs/notifications.js
wch: OK, sounds good. :) One shortcoming of ES6 classes is that they still don't have private members. In order to have private methods, you'd have to wrap the class in a closure, which takes some of the elegance away.

srcjs/notifications.js
wch: `type` sounds good to me.

wch:
@jcheng I added a function `exports.renderContent` in 3e0d8da, which is now used by `htmlOutputBinding` as well as `notifications.show`. I did something similar to the code on the R side, in 2fcb4db.

I haven't added triggering of show/shown/hide/hidden, but this code manages to show a plot in a notification, as well as take inputs from a `textInput`:

```R
shinyApp(
  ui = fluidPage(
    actionButton("show", "Show"),
    verbatimTextOutput("footxt")
  ),
    observeEvent(input$show, {
      showNotification(
        tagList(textInput("foo", "Foo"), plotOutput("plot", height = 200)),
        duration = 10
      )
    })

    output$footxt <- renderText(input$foo)

    output$plot <- renderPlot({
      par(mar = c(1,1,1,1))
      plot(rnorm(30))
    })
  }
)
```

Since that works, do you think the events are getting triggered properly from somewhere else?

PR #1143, opened 2016-03-24
Joe/feature/output arg passthrough

bborgesr:


bborgesr:
@jcheng5 I'm a little confused as to how this is supposed to work... (I realized that even before my commit, the changes you made before weren't producing anything different.) It seems to me that `useRenderFunction` (where the core of the changes are taking place) is not being called at all... I commented it all out and rebuilt the package and the result is exactly the same! What am I missing here? I can't figure out what the issue is here (except if this has always been an issue and `useRenderFunction` was just never called, but that didn't matter until now)...

bborgesr:
Also, I think that checks are failing because your commit seems to have broken some ggplot tests. The travis build seems pretty unreadable to me, but here's the relevant part of the `devtools::check` run:

```{r}
Running the tests in ‘tests/test-all.R’ failed.
Last 13 lines of output:
  
  testthat results ================================================================
  OK: 356 SKIPPED: 0 FAILED: 7
  1. Error: ggplot coordmap 
  2. Failure (at test-plot-coordmap.R#80): ggplot coordmap with facet_wrap 
  3. Error: ggplot coordmap with facet_wrap 
  4. Failure (at test-plot-coordmap.R#129): ggplot coordmap with facet_grid 
  5. Error: ggplot coordmap with facet_grid 
  6. Failure (at test-plot-coordmap.R#214): ggplot coordmap with 2D facet_grid 
  7. Error: ggplot coordmap with 2D facet_grid 
  
  Error: testthat unit tests failed
  Execution halted
* DONE
Status: 1 ERROR, 2 WARNINGs, 1 NOTE
```

I can look into this if you want, but we should probably solve the first issue first...

bborgesr:
@jcheng5, @wch Here are a couple of observation (issues?):

1) As we've noted before, it's a little redundant for some of the render functions to take in the `outputArgs` argument because the corresponding output functions do not take any arguments (except for the object's id). In particular, these functions are parameter-less:

- `verbatimTextOutput` (used in `renderPrint`)
- `tableOutput` (used in `renderTable`)
- `dataTableOutput` (used in `renderDataTable`)

For consistency, I'm currently including an `outputArgs` in these functions anyway. This means shiny won't throw any error if someone decided to do something like `renderPrint({"hello"}, outputArgs = list())` --  but it's not particularly useful either... This weirdness is not currently documented anywhere because I'm not sure if that would actually help or just confuse people further... I'm also not sure where would be the place to insert a comment about this (it feels weird to include it in the `@param` description because it would run something like -- "use `outputArgs` to pass `verbatimTextOutput` arguments through `renderPrint`... except `verbatimTextOutput` doesn't take any user parameters, so never mind"). Of course, we could just do something different for the `@param` description of these three functions in particular and just say something like "included for consistency, but redundant for this function". But that has its own weirdness. So it seems to me that just not documenting this is probably the best course of action...

2) `renderPrint` and `renderText` were the only render functions that did not have a `shinysession` arg being passed to their internal `renderFunc`, so I modified them to include it (although it doesn't seem to be used in the function's body itself). For some reason, this turned out to be important when using the new `markRenderFunction` -- for some reason, it wants to be passed a `shinysession` arg (I was getting an errors like `shiny error unused arguments (name = "outfcf15ef246224894", shinysession = <environment>)`). Just adding the `shinysession` arg fixed this. 

3) I had to modify the integration tests at `test-stacks.R` because when we included `origRenderFunc` in `markRenderFunction`, that changed the stack. Before:
```{r}
   num                                     call loc
1   32                                        A    
2   31                                        B    
3   30                               reactive C    
4   19                                        C    
5   18                              renderTable    
6   17                                     func    
7   16 renderTable({     C() }, server = FALSE)    
8    8                                  isolate    
9    7                      withCallingHandlers    
10   6                       captureStackTraces    
11   5                               doTryCatch    
12   4                              tryCatchOne    
13   3                             tryCatchList    
14   2                                 tryCatch    
15   1                                      try 
```

After:
```{r}
   num                                     call loc
1   32                                        A    
2   31                                        B    
3   30                               reactive C    
4   19                                        C    
5   18                              renderTable    
6   17                                     func    
7   16                           origRenderFunc    
8   15 renderTable({     C() }, server = FALSE)    
9    8                                  isolate    
10   7                      withCallingHandlers    
11   6                       captureStackTraces    
12   5                               doTryCatch    
13   4                              tryCatchOne    
14   3                             tryCatchList    
15   2                                 tryCatch    
16   1                                      try 
```

So I altered the tests accordingly. Is this okay?

4) @jcheng5 got rid of the (deprecated) `func` argument in his first commit to `renderPlot`. I've followed this pattern for the rest of the render functions... Let me know if I should undo this.

If you don't think any of these actually pose a problem, then this is ready to merge.

R/shinywrappers.R
```diff
@@ -12,24 +12,68 @@ globalVariables('func')
...
+#'   list, and pass through the value to \code{markRenderFunction}, to allow
+#'   app authors to customize outputs. (Currently, this is only supported for
+#'   dynamically generated UIs, such as those created by Shiny code snippets
+#'   embedded in R Markdown documents).
 #' @return The \code{renderFunc} function, with annotations.
 #'
 #' @export
-markRenderFunction <- function(uiFunc, renderFunc) {
+markRenderFunction <- function(uiFunc, renderFunc, outputArgs = list()) {
+  tracker <- new.env()
```
wch: Now that I'm looking at this again, I think it makes sense to make the tracker an R6 class. That would simplify this section of code and the code in `useRenderFunction`, and help keep things easier to understand.

bborgesr:
Update:

- Followed @wch's suggestion of making `tracker` an `R6Class` object. @jcheng5 mentioned it might be nice to have this in `utils.R` because having a simple mutable object can come in handy in other contexts too. I tried to make the code itself as obvious as possible and also added a couple of comments. What do you think?

- Backtracked on point 2) above (about adding `name` and `shinysession` to the previously argument-less `renderPrint` and `renderText`). @jcheng5 pointed out that what I did, while it worked fine for the all shiny `renderFunc`s, would not be very helpful if people have been making their own argument-less `renderFunc`s. So, I undid what I had done and replaced it with a simple check on `markRenderFunction` (line 45: `is.null(formals(origRenderFunc))`) and, accordingly, either pass or not pass `...` to the final `renderFunc`.

I think this takes care of all the issues we've talked about.

bborgesr:
@jcheng5 I made a couple of minor changes following a conversation with @wch. The only one that's relevant is adding `name` and `shinysession` back into `renderPrint` and `renderText`. Winston felt this would be better for consistency. However, the code still works if the inner `renderFunc` is parameter-less (and a warning is printed informing the user that they should have at least those two arguments). So, let me know if there's anything you still feel unsure about. Otherwise, feel free to merge this.

PR #1147, opened 2016-03-25
navbarMenu horizontal dividers

### LINKED ISSUE #888: Feature request: divider and dropdown-header in navbarMenu
As @yihui [suggested](https://groups.google.com/forum/#!topic/shiny-discuss/JFCFSmAK-YQ) I'm filing a feature request for dividers and dropdown-headers in a navbarMenu.

Below the text from my [question on SO](http://stackoverflow.com/questions/31237120/add-a-divider-and-dropdown-header-to-navbarmenu):

As my shiny app grows I'm running out of space in the navbar. I would like to add a divider and dropdown-header into a navbarMenu to combine some menu items. Adding the html directly doesn't do anything. Wrapping it inside `tabPanel` does insert something but makes the dropdown-header into a link and also adds whitespace that links to ... nothing (see example code below). So not great. Is there another way to do this?

    navbarMenu(
      tabPanel(HTML("<li class='divider'></li><li class='dropdown-header'>example header</li>")),
      tabPanel(HTML("<li class='divider'></li>"))
    )

There seems to have been an SO question by @george-dontas about this at some point or the link inside [this page](http://www.inside-r.org/questions/adding-divider-navbarmenu-shiny) is broken.

wch:
This closes #888.

Example app:

```
shinyApp(
  ui = navbarPage("App Title",
    tabPanel("Plot"),
    navbarMenu("More",
      tabPanel("Summary"),
      "----",
      "Header text",
      tabPanel("Table")
    )
  ),
  server = function(input, output) { }
)
```
image

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -27,6 +27,8 @@ shiny 0.13.1.9001
   and unmatched braces, parens, and brackets, and will print out messages
   pointing out those problems.
 
+* Added support for horizontal dividers in `navbarMenu`. (#888)
+
 shiny 0.13.1
 --------------------------------------------------------------------------------
 
--- a/R/bootstrap.R
+++ b/R/bootstrap.R
@@ -246,18 +246,20 @@ pageWithSidebar <- function(headerPanel,
 #' toggle a set of \code{\link{tabPanel}} elements.
 #'
 #' @param title The title to display in the navbar
-#' @param ... \code{\link{tabPanel}} elements to include in the page
+#' @param ... \code{\link{tabPanel}} elements to include in the page. If a
+#'   string of dashes like \code{"----"} is used in a \code{navbarMenu}, it
+#'   will display a horizontal separator in the menu.
 #' @param id If provided, you can use \code{input$}\emph{\code{id}} in your
 #'   server logic to determine which of the current tabs is active. The value
 #'   will correspond to the \code{value} argument that is passed to
 #'   \code{\link{tabPanel}}.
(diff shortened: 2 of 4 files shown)


vnijs:
Just tried out the PR. Very nice! I did notice that it will also turn an id into a dropdown-header.

```r
shinyApp(
  ui = navbarPage("App Title",
                  tabPanel("Plot"),
                  navbarMenu("More",
                             id = "Not a header",
                             tabPanel("Summary"),
                             "----",
                             "Header text",
                             tabPanel("Table")
                  )
  ),
  server = function(input, output) { }
)
```

image

wch:
Thanks for finding that case! I think that with the previous code, the `id` argument would basically get thrown away -- it wouldn't actually be used.

I think probably the right thing to do is to make sure that all the tab items are unnamed arguments, though it's possible I'm overlooking something. If we don't add that check, it's probably also makes sense to just use the behavior that you found in this pull request, and have the `id` argument (and any others) be treated like other unnamed arguments to `navbarMenu`. The first option is probably better for the sake of correctness; the second option is better for backward compatibility.

vnijs:
Does this mean that the `id` parameter can be removed from the function definition of `navbarMenu`?

wch:
`id` isn't in there: https://github.com/rstudio/shiny/blob/8a721fb/R/bootstrap.R#L398

vnijs:
Ah. I misread the diff on github. Thanks again for adding this feature!

PR #1149, opened 2016-03-26
upgrade ion-rangeSlider to 2.1.2

dmpe:
hi @wch again, 
this upgrades ion-rangeSlider to 2.1.2 as seen in https://github.com/IonDen/ion.rangeSlider/releases
also updates news file and dependency

From 2.0.12 it was mostly bug fixes and code refactoring. A few new features have been also added.


wch:
We actually use a modified version of ion.RangeSlider (https://github.com/rstudio/ion.rangeSlider/tree/shiny-mods-2.1.2), so I updated it in master using that version.

Issue #1123, opened 2016-03-02
Suppress error messages


jcheng5:
For security reasons, have an option to prevent any raw error messages from getting to the client--make them sanitized. Raw error messages can contain file paths, internal hostnames, argument values, and sometimes even source code snippets.

(Let the user throw "trusted" exceptions and turn all other exceptions into generic exceptions?)

Currently you can work around this by throwing a generic exception in `options(shiny.error = function() { ... })` but that means even logging doesn't occur (you can't get the error from within shiny.error--seems perverse, but it's modeled on `options(error)`).

We should at least have shiny.error functions with at least one argument, be passed the error. This would allow a complete workaround.

We should also consider building our own version of this and sticking it directly into our server products, so you can opt into this behavior across an entire server (or override on a per-app basis) with a config file.

Ref: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/shiny-discuss/FyMGa2R_Mgs/FNyusgdxBgAJ

cc @benb111

jcheng5:
Places to test: downloadHandler, ui.R itself

bborgesr:
Closed via #1156

PR #1156, opened 2016-04-05
Barbara/error hiding

bborgesr:
A PR to address [this issue](https://github.com/rstudio/shiny/issues/1123). I think the coding for this is more or less finished. Planning on testing it a bit more, but aside from that, this should be good to merge soon.

**An overview:**
By default, all errors are now sanitized automatically. This means that *most* error messages will be replaced with a generic error message: "An error has occurred. Check your logs or contact the app author for clarification." We can disable this feature and go back to the way things were up until now by adding `options(shiny.sanitize.errors = FALSE)`. However, if we want to sanitize most errors but there is one or two particular errors that we think would be really useful for the user to see (and which don't leak any sensitive information), then we can use the new `customStop()` function instead of the regular `stop()` function (even if we keep the default of `options(shiny.sanitize.errors = TRUE)`).

**Possible issues**

- This applies to all errors produced in a shiny app, including the trickier cases of `downloadHandler()` and `ui.R`. The aesthetics are still very different though -- these two open a new window to display the error, rather than producing an in-place, red error message. This can be something for us to prettify at some point in the future...

- This also applies to errors inside `runtime: shiny` Rmd documents. However, for that case, we're still not completely safe because sensitive information may still leak through to the end user via *warnings* and/or *messages*. This is something we'll need to think about. [The silver lining here is that if a warning or message occurs inside of a `reactive` or a `renderXXX` function, they do not show up in the document, only in the R Markdown console.]


jcheng5:
The default behavior is to sanitize, even when running Shiny apps locally? That feels like going overboard. I can understand defaulting to sanitizing in Shiny Server.

jcheng5:
How attached are you to the function name `customStop`? :) Any other candidates?

R/middleware.R
```diff
@@ -299,18 +299,37 @@ HandlerManager <- R6Class("HandlerManager",
...
-            withLogErrors(handlers$invoke(req))
+          function(req,
+                   full = getOption("shiny.fullstacktrace", FALSE),
+                   offset = getOption("shiny.stacktraceoffset", TRUE)) {
+
+             withCallingHandlers(captureStackTraces(handlers$invoke(req)),
+               error = function(cond) {
+                 # Don't print shiny.silent.error (i.e. validation errors)
+                 if (inherits(cond, "shiny.silent.error")) return()
+                 if (isTRUE(getOption("show.error.messages"))) {
```
jcheng5: What happens if `show.error.messages` is `FALSE`?

R/middleware.R
```diff
@@ -299,18 +299,37 @@ HandlerManager <- R6Class("HandlerManager",
...
+                   printError(cond, full = full, offset = offset)
+                   sanitizeErrors = getOption('shiny.sanitize.errors', TRUE)
+
+                   if (inherits(cond, 'shiny.custom.error') || !sanitizeErrors) {
+                     stop(cond$message, call. = FALSE)
```
jcheng5: Do we need this line? Isn't the default behavior of a calling handler, to let the error continue?

bborgesr: Well, I do want to throw the error at some point (because I want an error message to show up) - right? If so, I think this is the place to do it...

R/middleware.R
bborgesr: I inherited this check from the `shiny.R` code (before I added to it): `if (isTRUE(getOption("show.error.messages"))) printError(cond)`. `show.error.messages` is a general R error handling option: from my understanding, if it is `FALSE`, no errors (strictly specking, nothing produced by `stop` is printed). I don't think it's really important whether we keep this or not because I don't think this is a widely used option. It doesn't hurt though, and keeps things consistent.

bborgesr:
@jcheng5:

1. Yes, the default behavior is to sanitize, even when running Shiny apps locally. 
So, this I think it a big point of contention between you and Winston and Tareef (I'm Switzerland on this!) Their argument is: once we agree that the default is to sanitize errors in Shiny Server (and possibly Connect), then it would be good to be consistent. I.e. it'd be good for app authors working on their apps locally to have the same set of default options there as they'll have once they deploy them. And it isn't all that much more work because in local development, you have the console logs right there (don't have to go fishing for them, like with Shiny Server). That said, I understand your point of view as well. Maybe something to go over on Friday with everyone and reach a consensus?

2. Not particularly attached to `customStop` :) But I couldn't really think of a better name... Is there something in particular that you dislike about `customStop`? I feel that, in the context of the documentation, the name makes sense. Other possibilities (none of which I like better than `customStop` though): `priorityStop`, `shinyStop`, `showStop`, `unsanitizedStop` (this one is a mouthful!)

R/middleware.R
bborgesr: We've all talked about this already, so I'm just noting it here for future reference: it's necessary to have the `stop()` there (i.e. rethrow the error) because we want to set `.call = FALSE` and we can't do that if we just let the error continue being thrown.

bborgesr:
Update:

* changed `customStop(error)` to `stop(safeError(error))` as discussed with Joe -- this (hopefully) makes the concept easier for users to understand.

* `safeError()` now also works for `downloadHandler` errors. Bliss!

* refactored the code in `middleware.R` to avoid unnecessary duplication of the code used in `withLogErrors()`

bborgesr:
@jcheng5 @wch Let's merge this?

R/utils.R
```diff
@@ -947,15 +949,21 @@ columnToRowData <- function(data) {
...
+    cond <- structure(
+      list(message = error$message),
+      class = c("shiny.custom.error", errorClass, class(error))
+    )
   } else if (inherits(error, "character")) {
-    stopWithCondition(c("shiny.custom.error", errorClass), error)
+    cond <- structure(
+      list(message = error),
+      class = c("shiny.custom.error", errorClass, "error", "condition")
+    )
```
jcheng5: I was actually picturing more like this (not tested):

```r
if (inherits(error, "character")) {
  error <- simpleError(error)
}
if (!inherits(error, "error")) {
  stop("The class of the `error` parameter must be either 'error' or 'character'")
}
if (!("shiny.custom.error" %in% class(error))
  class(error) <- c("shiny.custom.error", class(error))
error
```

I'd like to use `simpleError` to convert the error string to an error obj, as it relies less on the internal implementation details of errors. And for the case where error is an error obj already, we can leave it alone as much as possible and just add the one class name.

jcheng5:
Looks great other than my one last comment!

R/utils.R
bborgesr: Ok, that makes sense. I've changed to that, with a small difference: I don't think this check is necessary: `if (!("shiny.custom.error" %in% class(error))`. `error` will never have the class `shiny.custom.error` a priori -- we'll always have to add it (right?) Also, I added the user-provided `errorClass` to `class(error)`.

R/utils.R
```diff
@@ -950,20 +950,14 @@ columnToRowData <- function(data) {
...
-    )
-  } else {
+  if (inherits(error, "character")) {
+    error <- simpleError(error)
+  }
+  if (!inherits(error, "error")) {
     stop("The class of the `error` parameter must be either 'error' or 'character'")
   }
-  return(cond)
+  class(error) <- c("shiny.custom.error", errorClass, class(error))
```
jcheng5: I didn't notice errorClass before--what's the intended usage of this?

jcheng5: Cause I think it might make more sense to just say, if you want to provide a custom error class, just give us a custom error object.

bborgesr: Hmmm... Initially, I just took errorClass from `stopWithCondition` that has an argument just like that. I then thought it was worth keeping around to allow app authors to style their apps however they want (yes, we do give them the "shiny.custom.error" class, but maybe one class is not enough). I doubt this will be used a lot, but it seemed unobtrusive enough that it was worth keeping it around to give app authors more control over their errors.

bborgesr:
Final updates:

- removed `errorClass` argument from the `safeError` function

- changed sanitization default to `FALSE` (on local development)

PR #1157, opened 2016-04-05
Add modal dialogs

wch:
This adds modal dialogs to Shiny.

Below are three examples (these are in the documentation). We might want to have a simplified API for handling cases like the password/username, where some event has to happen once (successfully) to make the modal go away, and sets a reactive value.

Another thing to consider: the `dismissable` argument isn't a great name for what it does, so I'm open to other ideas. EDIT: I've changed it to `easyClose` and updated the examples below.

```R
# Display an important message that can be dismissed only by clicking the
# dismiss button.
shinyApp(
  ui = basicPage(
    actionButton("show", "Show modal dialog")
  ),
    observeEvent(input$show, {
      showModal(modalDialog(
        title = "Important message",
        "This is an important message!"
      ))
    })
  }
)
```

image

```R
# Display a message that can be dismissed by clicking outside the modal dialog,
# or by pressing Esc.
shinyApp(
  ui = basicPage(
    actionButton("show", "Show modal dialog")
  ),
    observeEvent(input$show, {
      showModal(modalDialog(
        title = "Somewhat important message",
        "This is a somewhat important message.",
        easyClose = TRUE,
        footer = NULL
      ))
    })
  }
)
```
image

```R
# Display a modal that requires valid username and password input.
shinyApp(
  ui = basicPage(
    actionButton("show", "Show modal dialog"),
    verbatimTextOutput("loginInfo")
  ),
    # A string with the current login status. This is in a reactiveValues
    # object so that it can trigger reactivity.
    vals <- reactiveValues(loginStatus = "Not logged in.")

    # Attempt logging in with a username and password, returning TRUE if
    # successful and FALSE if not.
    login <- function(username, password) {
      # In a real-world use case, this would check against some sort of user
      # database instead of just checking that the values are identical to
      # hard-coded values.
      if (identical(username, "user1") && identical(password, "pass1")) {
        vals$loginStatus <- paste0('Logged in as "', username, '"')
        return(TRUE)

      } else {
        vals$loginStatus <- "Not logged in."
        return(FALSE)
      }
    }

    # Return the UI for a modal dialog with username/password inputs.
    # If 'failed' is TRUE, then display a message that the previous username
    # and password were invalid.
    loginModal <- function(failed = FALSE) {
      modalDialog(
        textInput("username", "Username"),
        passwordInput("password", "Password"),
        span('(Try logging in with "user1" and "pass1")'),
        if (failed)
          div(tags$b("Invalid username/password", style = "color: red;")),

        footer = tagList(
          modalButton("Cancel"),
          actionButton("login", "Log in")
        )
      )
    }

    observeEvent(input$show, {
      showModal(loginModal())
    })

    # When login button is pressed, attempt to log in. If successful, remove the
    # modal. If not show another modal, but this time with a failure message.
    observeEvent(input$login, {
      if (login(input$username, input$password)) {
        removeModal()
      } else {
        showModal(loginModal(failed = TRUE))
      }
    })

    # Display current login status
    output$loginInfo <- renderText({
      vals$loginStatus
    })
  }
)
```

image


nuno-agostinho:
You could add the option of having a small or a large modal dialog through an optional argument `size`. If `size` is `large` then `class = "modal-dialog modal-lg"`, else if `small` then `class = "modal-dialog modal-lg"`, else just let it be `class = "modal-dialog"`.

Small modal windows are nice for alerts! :)

Regarding alerts, it'd also be nice if we could change the ID of `modal-header` to allow to style modals like this using CSS: http://bootsnipp.com/snippets/featured/colored-modal-headings

wch:
@nuno-agostinho You can select the `modal-header` with a selector like `#shiny-modal .modal-header`. Also, although sizes could be useful, I think in this case the costs (complexity, maintenance) outweigh the benefits. There's also the new notification system if you want to show smaller pieces of information.

R/modal.R
```diff
@@ -0,0 +1,172 @@
...
+#' )
+#' }
+#' @export
+modalDialog <- function(..., title = NULL, footer = modalButton("Dismiss"),
+                        easyClose = FALSE) {
+
+  div(id = "shiny-modal", class = "modal fade", tabindex = "-1",
+    `data-backdrop` = if (!easyClose) "static",
+    `data-keyboard` = if (!easyClose) "false",
+
```
nuno-agostinho: @wch I don't think it adds that much complexity since it's just allowing a new argument `size` in the `modalDialog` function (as `NULL` by default) and adding:
```r
size <- switch(size, large= "modal-lg", small=" modal-sm", NULL)
div (class = paste0("modal-dialog", size))
```

Anyway, I can't wait to try the new notification system. :)

PR #1166, opened 2016-04-20
Allow to pass arguments to the icon tag function in the icon function

nuno-agostinho:
This allows to add classes/id to an icon making it easy to use CSS selectors to apply styles.


wch:
We discussed this PR and decided to not merge it, but of course you can always use `tags$i()` directly.

PR #1174, opened 2016-04-25
insertUI and removeUI

bborgesr:


jcheng5:
Did you mention somewhere that this wasn't working well in the RStudio Viewer? What exactly was the problem there?

bborgesr:
@jcheng5 It might be a simple case of me having to run `npm install` since this is running on ES6 (I think Chrome supports that by default). I'm finishing up the app now and will look into that soon. I'll post here when I'm sure it's bug-less :) But I think the meat of the code is finished. Let me know if you don't like something in there (oh, and I'll probably move the js part to "shinyapp.js" since it doesn't really seem to belong to "reactlog.js").

R/insert-ui.R
```diff
@@ -0,0 +1,194 @@
...
+  where = c("beforeBegin", "afterBegin", "beforeEnd", "afterEnd"),
+  ui,
+  immediate = FALSE,
+  container = if (inline) "span" else "div",
+  inline = FALSE,
+  session = getDefaultReactiveDomain()) {
+
+  force(selector)
+  force(ui)
+  force(session)
```
jcheng5: Let's also `force(multiple)` and `force(container)` here.

R/insert-ui.R
```diff
@@ -0,0 +1,194 @@
...
+  if (missing(where)) where <- "afterEnd"
+  where <- match.arg(where)
+
+  sendInsertUI <- function() {
+    session$sendCustomMessage("shiny-insert-ui",
```
jcheng5: Sorry, I think we're supposed to use `session$sendMessage` instead of `sendCustomMessage` for Shiny-internal features. I've been in the bad habit of using `sendCustomMessage` in the past but @wch fixed all of them I think. When you use `sendMessage` the way you receive messages on the client side is a little different, you use `addMessageHandler` instead of `exports.addCustomMessageHandler`.

R/insert-ui.R
```diff
@@ -0,0 +1,194 @@
...
+  }
+}
+
+
+#' Remove UI objects
+#'
+#' Remove a UI object from the app.
+#'
+#' This function allows you to remove any part of your UI. Once you call
+#' \code{removeUI} on some element, it is gone forever.
```
jcheng5: "Once you call" is fairly accurate if you pass immediate=TRUE, but the default behavior is to remove the UI at the end of the current... uh... flush cycle? I guess we don't have a name for it... anyway, the UI will be removed the next time the outputs/errors are flushed to the client.

inst/www/shared/shiny.js
```diff
@@ -5006,6 +5006,41 @@ var _typeof = typeof Symbol === "function" && typeof Symbol.iterator === "symbol
...
+        });
+      }
+    }
+  );
+
+  exports.addCustomMessageHandler("shiny-remove-ui",
+    function(message) {
+      let els = $(message.selector);
+      els.each((i, el) => {
+        $(el).remove();
```
jcheng5: We need to do something here to unbind any inputs/outputs before they are removed. I'm... actually not exactly sure what the right way to do this is. Definitely `exports.unbindAll(el)` is part of it, but there's an issue if `el` itself is an input or output--I don't think it'll be unbound in that case. We might need to take each element we're about to remove, and move it into a temporary div, which we then call unbindAll on... pretty gross I guess...

jcheng5: Oh and also need to sendImageSize/sendOutputHiddenState. I have to run but ping me or @wch if you have trouble figuring out what that means.

man/insertUI.Rd
```diff
@@ -0,0 +1,95 @@
...
+}
+Adapted from
+\href{https://developer.mozilla.org/en-US/docs/Web/API/Element/insertAdjacentHTML}{here}.}
+
+\item{ui}{The UI object you want to insert. This can be anything that
+you usually put inside your apps's  \code{ui} function.}
+
+\item{immediate}{Whether the UI object should be immediately inserted into
+the app when you call \code{insertUI}, or whether Shiny should wait for the
+next time your app is updated for some other reason (default).}
```
jcheng5: "wait for the next time your app is updated for some other reason" seems overly vague--how about "wait until all outputs have been updated and all observers have been run"?

R/insert-ui.R
bborgesr: Done

R/insert-ui.R
bborgesr: Done

man/insertUI.Rd
bborgesr: That does sound better. I wasn't sure how to say "flush cycle" in layman's terms, but I think this is pretty good. Thanks! Changed it.

inst/www/shared/shiny.js
bborgesr: Alright, I did my best educated guess for this. I think the `unbindAll` is probably working, but am pretty unsure about the `sendImageSize`, `sendInputHiddenState` parts. Should we talk about this later today during our 1:1?

R/insert-ui.R
bborgesr: I changed it to "Once `removeUI` is executed on some element, it is gone forever." -- does that sound better? My only intention with this is to let the user know that `removeUI` doesn't just hide the UI, or even remove it in a way that is later retrievable. So, I didn't want to go into details of exactly the timing of this and how it works in relation to flush cycles -- only that, once it happens (whenever that may be), your UI is gone.

bborgesr:
@jcheng5, @wch:
As I was changing things from `sendCustomMessage` to `session$sendMessage`, I had to change this chunk of code in the `insertUI` function (and also the corresponding bit in `removeUI`):

**Before:**
```R
if (!immediate) {
  session$onFlushed(sendUI, once = TRUE)
} else {
  sendUI()
```

**After:**
```R
if (!immediate) {
  session$onFlushed(session$sendInsertUI, once = TRUE,
                    selector = selector,
                    multiple = multiple,
                    where = where,
                    content = processDeps(ui, session),
                    container = container)
} else {
  session$sendInsertUI(selector = selector,
                       multiple = multiple,
                       where = where,
                       content = processDeps(ui, session),
                       container = container)
}
```

This is necessary because the `sendInsertUI` function is no longer defined in this file, and we need to have a way to pass it the values it needs to do its thing.

However, the `onFlushed` method only took two arguments: `func` (the function's name) and `once`. So there was no way to pass in the `func`'s arguments. To allow for this, I changed `onFlushed` to the following (I did a similar thing on `onFlush` for consistency):

*Documentation*
```R
#' \item{onFlushed(func, once=TRUE, ...)}{
#'   Registers a function to be called after the next time (if \code{once=TRUE})
#'   or every time (if \code{once=FALSE}) Shiny flushes the reactive system.
#'   The \code{...} argument allows you to pass arguments that should be applied
#'   to \code{func} when it is called. Returns a function that can be called
#'   with no arguments to cancel the registration.
#' }
```

*Code*
```R
onFlushed = function(flushedCallback, once = TRUE, ...) {
  if (!isTRUE(once)) {
    return(private$flushedCallbacks$register(flushedCallback))
  } else {
    dereg <- private$flushedCallbacks$register(function() {
      dereg()
      if (is.null(formals(flushedCallback))) flushedCallback()
      else flushedCallback(...)
    })
    return(dereg)
  }
},
```

This seems to work fine in my demo app and shouldn't affect any previous code. Are you both okay with this? Is there a better way to do it?

srcjs/shinyapp.js
```diff
@@ -631,6 +631,91 @@ var ShinyApp = function() {
...
+        $([]),
+        message.content.html,
+        message.content.deps
+      );
+    } else {
+      targets.each((i, target) => {
+        let container = document.createElement(message.container);
+        target.insertAdjacentElement(message.where, container);
+        exports.renderContent(container, message.content);
+        $(container).trigger('shown');
```
wch: You might need to check if the `container` is visible before triggering `shown`.

wch: Actually, I don't think `shown` needs to be triggered. If you look at `htmlOutputBinding`, it doesn't explicitly trigger any events, even though it adds or removes DOM content.

inst/www/shared/shiny.js
wch: I think you can just call `exports.renderContent` and pass it empty content. This is what `htmlOutputBinding` does if it receives empty content from the server, and we know that it handles the bindings correctly.

wch: Actually, that might not work because `renderContent` expects there to be a wrapper DOM element, but that might not be true here.

wch: Could we modify `unbindAll` to take an optional argument that tells it to unbind the object passed in (and not just the object's children)?

wch:
I think you can do something like:

```R
insertUI <- function() {
   session$sendInsertUI(selector = selector,
                        multiple = multiple,
                        where = where,
                        content = processDeps(ui, session),
                        container = container)
}

if (!immediate) {
  session$onFlushed(insertUI, once = TRUE)
} else {
  insertUI()
}
```

srcjs/shinyapp.js
```diff
@@ -631,6 +631,91 @@ var ShinyApp = function() {
     window.location.reload();
   });
 
+  // Helper function for addMessageHandler('shiny-insert-ui').
+  // Turns out that Firefox does not support insertAdjacentElement().
+  // So we have to implement our own version for insertUI.
+  // Code adapted from here: http://forums.mozillazine.org/viewtopic.php?t=445587
+  HTMLElement.prototype.insertAdjacentElement = function(where, parsedNode) {
```
wch: I'd suggest making this a standalone function in `utils.js`, instead of modifying `HTMLElement.prototype`.

srcjs/shinyapp.js
```diff
@@ -631,6 +631,91 @@ var ShinyApp = function() {
...
+        $(el).appendTo(temp);
+        exports.unbindAll(temp);
+        $(temp).remove();
+
+        // The size of each image may change either because the browser window was
+        // resized, or because a tab was shown/hidden (hidden elements report size
+        // of 0x0). It's OK to over-report sizes because the input pipeline will
+        // filter out values that haven't changed.
+        // Need to register callbacks for each Bootstrap 3 class.
+        var bs3classes = ['modal', 'dropdown', 'tab',
```
wch: It shouldn't be necessary to re-register these event handlers.

srcjs/shinyapp.js
```diff
@@ -631,6 +631,38 @@ var ShinyApp = function() {
...
+          return message.multiple;
+        });
+      }
+    });
+
+  addMessageHandler('shiny-remove-ui', function (message) {
+    var els = $(message.selector);
+    els.each(function (i, el) {
+      exports.unbindAll(el, true);
+      $(el).trigger('hide');
```
wch: I think this `hide` and `hidden` stuff isn't necessary, for a couple reasons. From a conversation, I understand that the reason this is here is so that `sendImageSize` and `sendOutputHiddenState` get triggered when a UI element is removed. There are two reasons I don't think that's needed:

1. Those functions update the clientData width and height, and hidden state. Those values are only needed for re-rendering a plot, but if the plot output UI is removed, there's no need to update those values.
2. In the `htmlOutputBinding` (`uiOutput` + `renderUI`), it's also possible to remove a plot output, by simply replacing it with nothing. But in that case, we don't bother triggering `hide` or `hidden` events, and everything seems to work fine.

R/shiny.R
```diff
@@ -743,12 +743,12 @@ ShinySession <- R6Class(
       } else {
         dereg <- private$flushCallbacks$register(function() {
           dereg()
-          flushCallback()
+          flushedCallback()
```
wch: Erroneous variable name change?

R/shiny.R
```diff
@@ -743,12 +743,12 @@ ShinySession <- R6Class(
...
         })
         return(dereg)
       }
     },
-    onFlushed = function(flushedCallback, once = TRUE) {
+    onFlushed = function(flushedCallback, once = TRUE, ...) {
```
wch: I think the `...` isn't necessary?

srcjs/shinyapp.js
jcheng5: Hidden state absolutely needs to be updated for a plot you're removing, otherwise the server will continue to render it even though it will never be shown. However, sendOutputHiddenState is (very sensibly) called by unbindOutputs--I had forgotten that. So we're covered there.

But we still do need to at least call sendImageSize from shiny-remove-ui. (This doesn't need to be called from htmlOutputBinding because that code path calls `renderContent`>`bindAll`>`bindOutputs`>`sendImageSize`.) This matters for cases where the removal of an element could cause a different element to change size--for example, fillRow, fillCol, flexdashboard, or even just the height of the page shrinking to where the scrollbar is no longer needed.

bborgesr: How about this, then? I added a call to `sendImageSize` from inside the `unbindOutputs` function. This way the function mimics `bindOutputs` in that it calls both `sendImageSize` and `sendOuputHiddenState`. I realize this may not always be necessary for `unbindOutputs`, but in practice, we pretty much only unbind things when we're modifying them or removing them, so, most times, this will be necessary. And for the other times, it doesn't hurt, right? It feels like this is the cleaner solution... But if any of you have objections, here are two alternatives:

1- Change `shiny-remove-ui` back to this:
```js
addMessageHandler('shiny-remove-ui',
    function(message) {
      let els = $(message.selector);
      els.each((i, el) => {
        exports.unbindAll(el, true);
        $(el).trigger('hide');
        $(el).hide();
        $(el).trigger('hidden');
        $(el).remove();
        // If `multiple` is false, returning false terminates the function
        // and no other elements are removed; if `multiple` is true,
        // returning true continues removing all remaining elements.
        return message.multiple;
      });
    }
  );
```


2- Call `sendImageSize` directly from `shiny-remove-ui` (no need to worry about `sendOutputHiddenState` because, as Joe mentions, that is taken care of by the call to `unbindAll`):
```js
addMessageHandler('shiny-remove-ui',
    function(message) {
      let els = $(message.selector);
      els.each((i, el) => {
        exports.unbindAll(el, true);
        setTimeout(sendImageSize, 0);
        $(el).remove();
        // If `multiple` is false, returning false terminates the function
        // and no other elements are removed; if `multiple` is true,
        // returning true continues removing all remaining elements.
        return message.multiple;
      });
    }
  );
```

jcheng5: sendImageSize from within unboundOutputs sounds fine to me!

bborgesr:
I think I've tested this pretty thoroughly by now, and I think it may be ready to merge. Also, here's the [demo app](https://beta.rstudioconnect.com/barbara/insertUI/).

wch:
Looks good to me now.

PR #1185, opened 2016-05-09
Barbara/showcase update

bborgesr:
Still need to figure out code highlighting for non-R files. Implement dropdown for www files. Option settable only from Description file. Defaults to True (i.e. always show .js, .css and .html files included in the www directory). Maybe this is the wrong way to go? I also feel that the option might be too "lenient" at this point. It is meant to be a boolean but if the user enters anything that is not coercible to an R logical, then the default kicks in and the www files are shown... Should we throw an error instead?


bborgesr:
I think this is ready to be merged now.

*Updates:*
- code highlighting works for everything now :)
- www files show up in a dropdown menu, so as to not take up too much space
- changed Description option name from `wwwFiles` to `IncludeWWW`
- `IncludeWWW` is optional (defaults to `TRUE`), but an error will be thrown if the user enters a value that is not coercible to an R boolean (these are all valid: `TRUE`, `True`, `T`, `true` and the false equivalents).

bborgesr:
To get the highlighting to work:

- Clone @jcheng5's highlight.js repo from [here](https://github.com/rstudio/highlight.js/tree/r/tools) (make sure to checkout to the `r` branch).
- Navigate to the tools subdirectory and run, for instance, `python build.py r xml js css`. In this case, this will create a `r.js`, a `xml.js`, a `js.js` and a `css.js` file. It will also overwrite the previous `highlight.pack.js` file.
- Replace shiny's copy of of this file (path: shiny/inst/www/shared/highlight) with the file you've just generated.
- Your highlighter now supports highlighting for R, XML/HTML, JS and CSS. Inspect a page that is highlighting some code to figure out the class names. Then add the corresponding css code to the `rstudio.css` file (path is the same:  shiny/inst/www/shared/highlight). For example, to highlight css pseudo elements in dark green, I added:

```css
.language-css .pseudo {
  color: rgb(13, 105, 18);
}
```

cc @jjallaire

R/showcase.R
```diff
@@ -77,38 +77,82 @@ appMetadata <- function(desc) {
...
+  })
+}
+
 # Returns tags containing the application's code in Bootstrap-style tabs in
 # showcase mode.
 showcaseCodeTabs <- function(codeLicense) {
   rFiles <- list.files(pattern = "\\.[rR]$")
+  wwwFiles <- list()
+  if (isTRUE(.globals$IncludeWWW)) {
+    path = paste0(getwd(), "/www")
```
wch: This should be `file.path(getwd(), "www")`. And it should use `<-`. :)

wch:
Looks good to me, other than the one comment I had.

PR #1189, opened 2016-05-13
insertUI and removeUI -- round 2 

bborgesr:
-- removed constraint that forced elements inserted with `insertUI` to be wrapped in a div/span


bborgesr:
I think this is ready for merging now.

PR #1190, opened 2016-05-13
Add shiny options

wch:
This adds functionality for Shiny to have its own set of options, which can be set globally, and also overridden on a per-app basis. If options are set within an app, they will be restored to their original values outside the app.


wch:
Closing because this version is more complicated than necessary.

Issue #1177, opened 2016-04-29
Feature request: clearBrush() function


cpsievert:
http://stackoverflow.com/a/36927826/1583084

wch:
This was implemented as `resetBrush` in #1197.

PR #1197, opened 2016-05-25
Add ability to reset brush with session$resetBrush/Shiny.resetBrush

jcheng5:
Necessary for some linked brushing scenarios involving Crosstalk


srcjs/shinyapp.js
```diff
@@ -641,19 +641,19 @@ var ShinyApp = function() {
...
+      // the side-effect of rendering the deps, singletons,
+      // and <head> still occur
+      exports.renderHtml($([]), message.content.html, message.content.deps);
+    } else {
+      targets.each(function (i, target) {
+        exports.renderContent(target, message.content, message.where);
+        return message.multiple;
+      });
+    }
+  });
```
jcheng5: Just un-indenting by one level.

wch:
Does this actually reset the brush and clear the div? It's not clear to me how that works.

jcheng5:
All the image/plot divs listen for `shiny-internal:brushed`, and clear themselves when it's triggered. I added that a while back so that multiple image/plot outputs could share the same brushId.

wch:
Oh right, now I remember.

Does this also close #1177?

Issue #1200, opened 2016-05-27
segfault on OS X when closing shiny app


jotsetung:
Hi,
I frequently stumble over the problem that I get a segfault whenever I want to close running shiny app. The apps run without problems but when it gets to closing either the browser tab or hitting Ctrl-C on the R-terminal I get the segfault.
Example:

> runExample("03_reactivity")

Listening on http://127.0.0.1:4435
^C

 *** caught segfault ***
address 0xfffffffd0ba655ac, cause 'memory not mapped'

Traceback:
 1: order(x, na.last = na.last, decreasing = decreasing, method = "radix")
 2: sort.int(x, na.last = na.last, decreasing = decreasing, ...)
 3: sort.default(as.integer(.callbacks$keys()), decreasing = TRUE)
 4: sort(as.integer(.callbacks$keys()), decreasing = TRUE)
 5: private$closedCallbacks$invoke(onError = printError, ..stacktraceon = TRUE)
 6: shinysession$wsClosed()
 7: handler()
 8: (function (handle) {    ws <- .wsconns[[as.character(handle)]]    ws$.handle <- NULL    rm(list = as.character(handle), pos = .wsconns)    for (handler in ws$.closeCallbacks) {        handler()    }})("140212578892336")
 9: eval(substitute(expr), envir, enclos)
10: evalq((function (handle) {    ws <- .wsconns[[as.character(handle)]]    ws$.handle <- NULL    rm(list = as.character(handle), pos = .wsconns)    for (handler in ws$.closeCallbacks) {        handler()    }})("140212578892336"), <environment>)
11: doTryCatch(return(expr), name, parentenv, handler)
12: tryCatchOne(expr, names, parentenv, handlers[[1L]])
13: tryCatchList(expr, names[-nh], parentenv, handlers[-nh])
14: doTryCatch(return(expr), name, parentenv, handler)
15: tryCatchOne(tryCatchList(expr, names[-nh], parentenv, handlers[-nh]),     names[nh], parentenv, handlers[[nh]])
16: tryCatchList(expr, classes, parentenv, handlers)
17: tryCatch(evalq((function (handle) {    ws <- .wsconns[[as.character(handle)]]    ws$.handle <- NULL    rm(list = as.character(handle), pos = .wsconns)    for (handler in ws$.closeCallbacks) {        handler()    }})("140212578892336"), <environment>), error = function (x) x, interrupt = function (x) x)
[...]

PR #1201, opened 2016-05-27
avoid overflow in R 3.3.0 radix sort

kevinushey:
This PR works around a bug in R 3.3.0's radix sort implementation (where overflows can occur when attempting to sort integers with INT_MAX included). Should resolve https://github.com/rstudio/shiny/issues/1200.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -597,7 +597,10 @@ Callbacks <- R6Class(
     .callbacks = 'Map',
 
     initialize = function() {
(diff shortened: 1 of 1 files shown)


wch:
Could you add a comment that mentions why the `- 10` is there?

kevinushey:
Will do!

kevinushey:
Updated (and it looks like just using `-1` is sufficient here); the Travis CI build seems to be failed for separate reasons.

PR #1209, opened 2016-06-13
Bookmarkable state

wch:


wch:
I just ran into an issue with Shiny docs. Example:

    ---
    title: "Shiny subapp"
    output: html_document
    runtime: shiny
    ---
    
    ```{r cars}
    library(shiny)
    shinyApp(
      ui = basicPage(),
    )
    ```

The embedded app has a URL like http://127.0.0.1:6861/appf973a07fb4e84c1d57d6ac505661c351/?w=&__subapp__=1. This raises two issues:

1. Shiny gives `Warning: Error in value[[3L]]: Failed to parse URL parameter "w"`. This is because `parseQueryString` isn't able to handle empty values. This is easy to fix.
2. Should we try to restore from a bookmarked state in a subapp? If no, then when there's a `__subapp__=` in the query string, we could just not try to restore state. If yes, then we need to figure out how to deal with query parameters in URL.

bborgesr:
I can't really foresee a use case scenario in which someone wants to bookmark a subapp... If you're including the app as part of a larger document, and you change the app's defaults, you probably need to change some of the context around, right? I mean, if this would be something relatively easy to support, maybe why not... On the other hand, I think it might just get confusing. Especially if you have several subapps (maybe even with the same input/output ids)... So my intuition would be, let's not support this.

jcheng5:
I think supporting subapps makes sense from a user perspective, but we should probably punt for this release unless you already have a clear picture for how it could be implemented cleanly.

wch:
OK, I think it would be better to not support subapps for now.

R/bookmark-state-local.R
```diff
@@ -0,0 +1,30 @@
...
+#
+# @param id A session ID to save.
+# @param callback A callback function that saves state to or restores state from
+#   a directory. It must take one argument, \code{stateDir}, which is a
+#   directory to which it writes/reads.
+
+saveInterfaceLocal <- function(id, callback) {
+  # Try to save in app directory, or, if that's not available, in the current
+  # directory.
+  appDir <- getShinyOption("appConfig")$appDir %OR% getwd()
```
jcheng5: The case where `appDir` is `NULL` is `shinyApp(ui, server)`, right? Perhaps capture `appDir <- getwd()` during the `shinyApp` call rather than late-binding here?

R/bookmark-state.R
```diff
@@ -0,0 +1,960 @@
...
+        values = self$values
+      )
+    }
+  ),
+
+  private = list(
+    # Given a query string with a _state_id_, load saved state with that ID.
+    loadStateQueryString = function(queryString) {
+      values <- parseQueryString(queryString, nested = TRUE)
+      id <- values[["_state_id_"]]
```
jcheng5: Need to ensure this is a valid-looking id (no special characters, especially not `"../"`, etc.).

R/bookmark-state.R
```diff
@@ -0,0 +1,960 @@
...
+
+      # This function is passed to the loadInterface function; given a
+      # directory, it will load state from that directory
+      loadFun <- function(stateDir) {
```
jcheng5: For some common error cases, can we throw explicit errors (since they will be displayed to the user)? stateDir doesn't exist, missing files that we expect, etc.

R/bookmark-state.R
```diff
@@ -0,0 +1,960 @@
...
+
+      loadInterface <- getShinyOption("load.interface", default = loadInterfaceLocal)
+      loadInterface(id, loadFun)
+
+      invisible()
+    },
+
+    # Given a query string with values encoded in it, restore saved state
+    # from those values.
+    decodeStateQueryString = function(queryString) {
```
jcheng5: I'd like to see some unit tests for this, as it is I'd be very wary of making changes myself here.

Is there always an `_input_` and always a `_values_`, or are one or both optional? Where can other (non-bookmark) parameters appear--before our parameters, after them?

Also, did you consider using regexec with capturing groups for this? Assuming neither are optional:

```
m <- regexec("(^|&)_input_(&.*)?&_values_(&.*)?$", queryString)
regmatches(queryString, m)
```

jcheng5: Or running `parseQueryString` on the whole thing and picking out where `names(x) == "_input_" & x == ""`, etc.

R/bookmark-state.R
```diff
@@ -0,0 +1,960 @@
...
+  private = list(
+    values = NULL,
+    pending = character(0),
+    used = character(0)     # Names of values which have been used
+  ),
+
+  public = list(
+    initialize = function(values) {
+      private$values <- new.env(parent = emptyenv())
+      list2env(values, private$values)
```
jcheng5: Just curious, is this different than `private$values <- list2env(values, parent = emptyenv())`?

R/bookmark-state.R
```diff
@@ -0,0 +1,960 @@
...
+
+#' Update URL in browser's location bar
+#'
+#' This function updates the URL in the client browser's location bar. It
+#' typically is called from an observer.
+#'
+#' @param queryString The new query string to show in the location bar.
+#' @param session A Shiny session object.
+#' @export
+updateLocationBar <- function(queryString, session = getDefaultReactiveDomain()) {
```
jcheng5: I'd expect a function called `updateLocationBar` to let you change the entire URL (or at least the path, hash, and query string). How would you feel about `updateQueryString`?

R/bookmark-state.R
```diff
@@ -0,0 +1,960 @@
...
+#'
+#' @seealso configureBookmarking
+#' @inheritParams actionButton
+#' @export
+bookmarkButton <- function(label = "Bookmark...",
+  icon = shiny::icon("link", lib = "glyphicon"),
+  title = "Bookmark this application's state and get a URL for sharing.",
+  ...)
+{
+  actionButton("__bookmark__", label, icon, title = title, ...)
```
jcheng5: I don't suppose you've tested this with `submitButton`? I have a feeling the id needs to start with `.`.

R/bookmark-state.R
```diff
@@ -0,0 +1,960 @@
...
+#' @param subtitle Text to display underneath URL.
+#' @export
+urlModal <- function(url, title = "Bookmarked application link", subtitle = NULL) {
+
+  subtitleTag <- NULL
+  if (!is.null(subtitle)) {
+    subtitleTag <- tagList(
+      br(),
+      span(class = "text-muted", subtitle),
+      span(id = "shiny-bookmark-copy-text", class = "text-muted")
```
jcheng5: We don't include the "Press Ctrl-C to copy" if you don't provide a subtitle?

R/bookmark-state.R
```diff
@@ -0,0 +1,960 @@
...
+  )
+}
+
+#' Enable bookmarking for a Shiny application
+#'
+#' There are two types of bookmarking: saving an application's state to disk on
+#' the server, and encoding the application's state in a URL. For state that has
+#' been saved to disk, the state can be restored with the corresponding state
+#' ID. For URL-encoded state, the state of the application is encoded in the
+#' URL, and no server-side storage is needed.
```
jcheng5: Maybe a sentence or two about why you might pick one type vs the other

R/bookmark-state.R
```diff
@@ -0,0 +1,960 @@
...
+#'
+#' For restoring state to work properly, the UI must be a function that takes
+#' one argument, \code{request}. In most Shiny applications, the UI is not a
+#' function; it might have the form \code{fluidPage(....)}. Converting it to a
+#' function is as simple as wrapping it in a function, as in
+#' \code{function(request) \{ fluidPage(....) \}}.
+#'
+#' By default, all input values will be bookmarked, except for the values of
+#' actionButtons and passwordInputs. FileInputs will be saved if the state is
+#' saved on a server, but not if if the state is encoded in a URL.
```
jcheng5: Double "if". Also I wouldn't capitalize "fileInput". And these three functions should all be \code{\link{..}}

R/bookmark-state.R
wch: There are a bunch of unit tests here:
https://github.com/wch/shiny/blob/bookmarkable-state/tests/testthat/test-bookmarking.R

`_inputs_` and `_values_` are both optional. Non-bookmark params can occur only before bookmark ones. That could be changed, though.

R/bookmark-state.R
wch: Nope, the result is exactly the same. I'll use your version because it's slightly simpler.

R/bookmark-state.R
wch: `updateQueryString` sounds good. One thing to note: previously it was possible to trigger an update every time an input changed, with this:

```R
observe({
  reactiveValuesToList(input)
  updateLocationBar( .... )
})
```

However, the `onBookmark` observer is currently hard-coded to look for `input[['__bookmark__']]`, so the location bar can only be updated when the bookmark button is pressed. If we want to make it possible to trigger `onBookmark` for any input change, the `observeEvent` needs to have a configurable `eventExpr`.

R/bookmark-state.R
wch: Argh, `submitButton`!

wch:
I've addressed all the comments, so if there's nothing else, this should be good to merge. I don't know for sure, but I think appveyor errors are happening because this branch doesn't yet have the recent `appveyor.yml` file.

bborgesr:
Yes, Winston, I think you're right re the AppVeyor error. I think that, if that's the only thing in question, we should go ahead and merge. I'm 99% sure the build will be successful then.

bborgesr:
On a second thought, we may have to look a bit better into that... The error occurs when trying to install `digest`, which seems odd, since other packages were already successfully installed before:

```
* installing *source* package 'digest' ...
** package 'digest' successfully unpacked and MD5 sums checked
** libs
c:/Rtools/mingw_32/bin/gcc  -I"c:/R/include" -DNDEBUG     -I"d:/Compiler/gcc-4.9.3/local330/include"     -O3 -Wall  -std=gnu99 -mtune=core2 -c aes.c -o aes.o
/bin/sh: c:/Rtools/mingw_32/bin/gcc: No such file or directory
make: *** [aes.o] Error 127
Warning: running command 'make -f "c:/R/etc/i386/Makeconf" -f "c:/R/share/make/winshlib.mk" SHLIB="digest.dll" OBJECTS="aes.o crc32.o digest.o init.o md5.o pmurhash.o raes.o sha1.o sha2.o sha256.o xxhash.o"' had status 2
ERROR: compilation failed for package 'digest'
* removing 'c:/RLibrary/digest'
The downloaded source packages are in
	'C:\Users\appveyor\AppData\Local\Temp\1\Rtmp80cAZP\downloaded_packages'
Warning messages:
1: running command '"c:/R/bin/i386/R" CMD INSTALL --no-multiarch -l "c:\RLibrary" C:/Users/appveyor/AppData/Local/Temp/1\Rtmp80cAZP/downloaded_packages/digest_0.6.10.tar.gz' had status 1 
2: In install.packages(commandArgs(TRUE), repos = "https://cran.rstudio.com",  :
  installation of package 'digest' had non-zero exit status
```

In any case, I still think we should merge and deal with this on master *if* it persists...

wch:
I think it's saying it can't find gcc? But at any rate, I agree, it would be best to merge and then fix that.

bborgesr:
Ah, ok, I looked closer into it and you're right. It can't find gcc. My guess is that this is related to Rtools, in which case it should disappear when we merge...

R/reactives.R
```diff
@@ -118,18 +127,54 @@ ReactiveValues <- R6Class(
...
+      .metadata[[key]][[metaKey]] <<- value
+    },
+
+    # Mark a value as invalid. If accessed while invalid, a shiny.silent.error
+    # will be thrown.
+    invalidate = function(key) {
+      setMeta(key, "invalid", TRUE)
+    },
+
+    unInvalidate = function(key) {
```
jcheng5: Should setting a value cause it to automatically `unInvalidate`? Honest question, not rhetorical.

jcheng5:
Everything looks good except that I still have to review the docs, and my eyes crossed from looking at shiny.R too long. Maybe we can review that one file together.

R/reactives.R
wch: Hm, that's a tough one. One thing that's a little different about invalidating values this way is that the behavior becomes dependent on the order that the reactives execute (imagine that you access `input$x` in one reactive and invalidate it in another, but those two reactives don't depend on each other). So we're already in a bit of a murky area.

Right now I don't see any advantage to doing it either way. For simplicity then, I'd say we leave it as is. But if you can see reasons to go one way or the other, I'm definitely open to that.

Two things to note:
* I think the names `invalidate` and `invalidateReactiveValue` are too similar to the *other* kind of invalidation that we do, so we should change the name.
* @bborgesr this code throws `shiny.silent.errors` but isn't yet in master, so you may need to make changes to this code later related to `reactiveStop()`.

R/reactives.R
```diff
@@ -60,32 +61,40 @@ ReactiveValues <- R6Class(
...
+        .dependents[[dep.key]] <- ctx
         ctx$onInvalidate(function() {
-          rm(list=dep.key, pos=.dependents, inherits=FALSE)
+          rm(list=dep.key, envir=.dependents, inherits=FALSE)
         })
       }
 
-      if (!exists(key, where=.values, inherits=FALSE))
+      if (isInvalid(key))
+        stopWithCondition(c("validation", "shiny.silent.error"), "")
```
bborgesr: @wch, is this the only place where a `shiny.silent.error` is thrown? Once #1272 is merged, we can simply change this line to: `reactiveStop()`

R/shiny.R
```diff
@@ -420,9 +431,136 @@ ShinySession <- R6Class(
...
+                  rc <- getCurrentRestoreContext()
+                  if (rc$active) {
+                    restoreState <- getCurrentRestoreContext()$asList()
+                    private$restoreCallbacks$invoke(restoreState)
+                  }
+                })
+              ),
+              error = function(e) {
+                showNotification(
+                  paste0("Error calling onRestore callback: ", e$message),
```
jcheng5: Should probably print the error and stack to the console here

R/shiny.R
```diff
@@ -502,12 +653,157 @@ ShinySession <- R6Class(
...
           self$makeScope(ns(namespace))
+        },
+
+        setBookmarkExclude = function(names) {
+          bookmarkExclude <<- names
+        },
+        getBookmarkExclude = function() {
+          bookmarkExclude
+        },
+        onBookmark = function(fun) {
```
jcheng5: Do you think the order that `onBookmark(ed)`/`onRestore(d)` is invoked matters? Just curious.

wch: They'll get invoked in the order that they were registered, so if the module has an `onBookmark` callback and`callModule()` is called before the app's own `onBookmark`, then the module's callback will be invoked first. But if the app calls `onBookmark()` and then `callModule()`, the app's callback will be invoked first. But the `state$values` code is written so that order won't matter.

R/shiny.R
bborgesr: Isn't this done in line 517 with the call to `withLogErrors`?

PR #1222, opened 2016-06-26
Don't use randomly-chosen port numbers that Chrome deems unsafe

jcheng5:
Still OK to use these ports if the user asks for them explicitly

Chrome actually blocks these URLs from loading.

image


jcheng5:
@wch @bborgesr Any concerns, or can we merge?

wch:
LGTM

jcheng5:
One of the few places a do-while loop would've been nice I guess

PR #1256, opened 2016-07-22
Fix #931: Observer memory leak

### LINKED ISSUE #931: Memory leak in nested render/output functions
There seems to be a memory leak when accessing the output object within a render function (e.g. for rendering another output element). Maybe related to the use of the same environment, see attached example. 

I guess, using nested render functions is bad practice and the alternative based on reactiveValues works fine (commented in example). However, I was not aware of this problem, which may arise  accidentally.

```
library(shiny)
library(pryr)

runApp(list(
	ui = bootstrapPage(
		actionButton("testAction", "Test"), br(), br(), 
		uiOutput("test1"), br(), 
		uiOutput("test2")
	),
		
		revals <- reactiveValues(message="")
		
		output$test1 <- renderText({
			if (is.null(input$testAction))
				return(NULL)
			if (input$testAction == 0)
				return(NULL)
			
			# nested render function (memory leak)
			output$test2 <- renderText({ paste("Test output message", sample(1:100,1)) })
			# alternative (no memory leak)
			# revals$message <- paste("Test output message", sample(1:100,1))
			
			# large object
			bigVec <- 1:(10^7)
			memUsed <- round(as.numeric(pryr::mem_used()) / 1024 / 1024)
			
			return(paste(memUsed, "MB"))
		})
		
		# alternative
		output$test2 <- renderText({ 
			revals$message
		})
	}
))
```

Thanks

jcheng5:
Observers were being prevented from being garbage collected by
their own onReactiveDomainEnded() event handlers. This commit
fixes that by making sure that those event handlers are only
registered when autoDestroy=TRUE, and that they are unregistered
both on destruction and when autoDestroy is changed.

### DIFF
--- a/R/reactives.R
+++ b/R/reactives.R
@@ -591,6 +591,10 @@ Observer <- R6Class(
     .domain = 'ANY',
     .priority = numeric(0),
     .autoDestroy = logical(0),
+    # A function that, when invoked, unsubscribes the autoDestroy
+    # listener (or NULL if autodestroy is disabled for this observer).
+    # We must unsubscribe when this observer is destroyed, or else
+    # the observer cannot be garbage collected until the session ends.
     .autoDestroyHandle = 'ANY',
     .invalidateCallbacks = list(),
     .execCount = integer(0),
--- a/R/reactives.R
+++ b/R/reactives.R
@@ -721,12 +721,12 @@ registerDebugHook("observerFunc", environment(), label)
           if (.domain$isEnded()) {
             destroy()
           } else {
-            .autoDestroyHandle <<- onReactiveDomainEnded(.domain, self$.onDomainEnded)
+            .autoDestroyHandle <<- onReactiveDomainEnded(.domain, .onDomainEnded)
           }
         }
       } else {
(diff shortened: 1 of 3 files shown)


R/reactives.R
```diff
@@ -706,11 +708,28 @@ registerDebugHook("observerFunc", environment(), label)
...
       .autoDestroy <<- autoDestroy
-      if (!is.null(.domain) && .domain$isEnded()) {
-        destroy()
+
+      if (autoDestroy) {
+        if (!.destroyed && !is.null(.domain)) { # Make sure to not try to destroy twice.
+          if (.domain$isEnded()) {
+            destroy()
+          } else {
+            .autoDestroyHandle <<- onReactiveDomainEnded(.domain, .onDomainEnded)
```
wch: It would be useful to have a comment saying that the handle is a function that clears the callback.

wch:
Looks good. Do you mind also adding tests for the memory leak fixed by #1254? That could be a separate PR if you want.

jcheng5:
@wch The leak in 1254 is covered by one of the tests already in here (case "e"). Good call though.

Issue #971, opened 2015-09-26
runApp / shiny.options:  Cast port to numeric.


adilapapaya:
This line does not throw an error. Instead, it outputs a message that makes it look like everything is working fine.

```r
runApp(".", port = "8003", launch.browser = TRUE)
#> Listening on domain socket 8003
```

I think the code should either
1. `as.numeric(port)` then throw an error if the result is `NA`
2. check if `typeof(port) == "double"` and throw an error if it isn't.

PR #1263, opened 2016-07-27
Barbara/error handling/port as socket

bborgesr:
In issue #971, the user requests that ports be cast to numeric when they're passed in as a string. However, there is a legitimate use case when the `port` argument may be passed as a string, which is when it is being used *not* to specify the actual port, but to provide a domain socket instead. Fortunately, if this is the case, `port` must include a `mask` attribute. So if that is not present, we know that `port` is being used incorrectly. In that situation, throw an error with a (hopefully) helpful message.

### DIFF
--- a/R/server.R
+++ b/R/server.R
@@ -447,6 +447,9 @@ startApp <- function(appObj, port, host, quiet) {
       message('\n', 'Listening on domain socket ', port)
     }
     mask <- attr(port, 'mask')
+    if (is.null(mask)) {
+      stop("`port` is not a valid domain socket (missing `mask` attribute)")
+    }
     return(startPipeServer(port, mask, handlerManager$createHttpuvApp()))
   }
 }
--- a/R/server.R
+++ b/R/server.R
@@ -448,7 +448,10 @@ startApp <- function(appObj, port, host, quiet) {
     }
     mask <- attr(port, 'mask')
     if (is.null(mask)) {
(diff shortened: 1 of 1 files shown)


bborgesr:
Hi @jcheng5, I actually wanted to talk to you about this before merging/dropping it... The whole premise of this is that `mask` exists and is non-null. But from the `httpuv` docs, it seems like there are legitimate use cases of domain sockets with a NULL `mask`: https://github.com/rstudio/httpuv/blob/master/R/httpuv.R#L360 If that's the case, this fix doesn't work.

If that's true, I see two ways out of this: provide an extra argument that indicates whether to use a TCP socket (default) or a Unix domain socket / Windows pipe -- which will be a breaking change if anyone was using the latter. Or, just drop this whole thing and maybe only add a more explicit message here: https://github.com/rstudio/shiny/blob/master/R/server.R#L445 E.g.:

```r
message('\n', 'Using `port` as path for a domain socket (on Unix-like systems) ',
              'or the name of the named pipe (on Windows). ',
              'Listening on domain socket/pipe ', port)
```

jcheng5:
The domain socket stuff should be be deprecated and removed. It turned out not to be stable under load. I can almost guarantee that nobody is using this functionality except for very very old versions of Shiny Server.

bborgesr:
Should we do a soft deprecate for this Shiny release?

Issue #1270, opened 2016-08-02
cancelOutput() does not behave the same as req(FALSE) when called inside an observer.


bborgesr:
Which is what the documentation says it does... So we should either fix it or change the documentation (which I think is what makes the most sense) to specify that `cancelOutput()` should only be used in an output context (after all, it *is* right there in the name)...

We should also add some examples to the documentation of both `cancelOutput` and `req`.  I also feel like they should be documented together. And finally, it may be useful to write a small article/app that uses these types of simple examples for people to understand the difference between them. I don’t think it is completely intuitive and these may help. In any case, I didn’t want to move ahead with any of these before we agree on how we want to proceed.

Here is an example of the bug:

```r
ui <- fluidPage(
  textInput('txt', 'Enter text'),
  textOutput('check')
)
server <- function(input, output, session) {
  obs <- observe({
    req(FALSE)  ## this silently throws, does not affect the app
    # req(FALSE, cancelOutput = TRUE) ## but these two end the session
    # cancelOutput()
    input$txt
  })
  
  output$check <- renderText({
    input$txt
  })
}
shinyApp(ui, server)
```

Here are a bunch of examples that work as expected, that we can include in the documentation...

```r
ui <- fluidPage(
  textInput('txt', 'Enter text'),
  textOutput('check')
)
server <- function(input, output, session) {
  output$check <- renderText({
    if (input$txt == 'hi') return('hi')
    else if (input$txt == 'bye') return('bye')
    else cancelOutput()
  })
}
shinyApp(ui, server)

#*******************************#
#*******************************#

ui <- fluidPage(
  textInput('txt', 'Enter text'),
  textOutput('check')
)
server <- function(input, output, session) {
  output$check <- renderText({
    req(input$txt)
    if (input$txt == 'hi') return('hi')
    else if (input$txt == 'bye') return('bye')
  })
}
shinyApp(ui, server)

#*******************************#
#*******************************#

ui <- fluidPage(
  textInput('txt', 'Enter text'),
  textOutput('check')
)
server <- function(input, output, session) {
  output$check <- renderText({
    if (req(input$txt == 'hi')) return('hi')
    else if (req(input$txt == 'bye')) return('bye')
  })
}
shinyApp(ui, server)

#*******************************#
#*******************************#

ui <- fluidPage(
  textInput('txt', 'Enter text'),
  textOutput('check')
)
server <- function(input, output, session) {
  output$check <- renderText({
[...]

PR #1272, opened 2016-08-04
Introduce reactiveStop()

bborgesr:
@jcheng5, @wch: I think everything behaves as it is supposed to, per what we discussed in #1270. Are we good to merge?


R/reactives.R
```diff
@@ -614,10 +614,10 @@ registerDebugHook("observerFunc", environment(), label)
...
           else
             observerFunc(),
-          validation = function(e) {
-            # It's OK for a validation error to cause an observer to stop
-            # running
-          }
+          # It's OK for validation errors or shiny.output.cancel error to
+          # cause an observer to stop running
+          validation = function(e) NULL,
+          shiny.output.cancel = function(e) NULL
```
wch: I was thinking that the observer would catch something like `reactive.error` instead of these two separate things. (Probably need a better name though)

R/utils.R
```diff
@@ -973,6 +973,65 @@ safeError <- function(error) {
...
+#'   observe({
+#'     val()
+#'   })
+#' }
+#'
+#' shinyApp(ui, server)
+#' }
+#' @export
+reactiveStop <- function(message = "", class = NULL,
+                         type = c("validation", "shiny.output.cancel")) {
```
wch: I think `type` should default to `NULL`.

R/utils.R
```diff
@@ -973,6 +973,65 @@ safeError <- function(error) {
...
+  type <- match.arg(type)
+  stopWithCondition(c(type, "shiny.silent.error", class), message)
```
wch: This `"shiny.silent.error"` would be `"reactive.error"`, to match the condition that's caught in the observer.

R/utils.R
bborgesr: I think if we create a new class `reactive.error`, we can get rid of the `type` argument completely...

bborgesr: Although, doesn't `shiny.silent.error` then lose its raison d'etre? Are we just subbing one error class by another?

bborgesr:
@jcheng5, @wch: if there are no issues here, I can (squash and) merge this and then also update [this line](https://github.com/rstudio/shiny/pull/1209#discussion_r73568224) of the bookmarkable state commit to use the internal `reactiveStop` function.

The only open question I still have is whether or not we export `isTruthy` -- see [here](https://github.com/rstudio/shiny/issues/1270#issuecomment-238261772).

PR #1276, opened 2016-08-08
updates bootstrap too

dmpe:


wch:
Thanks; we've just made this update ourselves.

PR #1277, opened 2016-08-08
Solve updateSelectInput bug in IE 11.0

### LINKED ISSUE #1257: Server-side updateSelectizeInput Fails for Internet Exporer and Edge Browsers
When `updateSelectizeInput` is used to change the list of choices, it doesn't work if I use Microsoft-developed browsers. The same Shiny application works in Opera 38, for example. It also works in the Microsoft=developed browsers if `server` is set to `FALSE`.

    server = shinyServer(function(input, output, session) {
      
      observeEvent(input[["dataChoice"]],
                   {
                     options <- switch(input[["dataChoice"]], Letters = LETTERS[1:5], Numbers = 1:5)
                     updateSelectizeInput(session, "itemChoice", choices = options, server = TRUE)
                   })
    })
    
    interface = fluidPage(
      selectizeInput("dataChoice", "Choose Data Type", c("Letters", "Numbers")),
      selectizeInput("itemChoice", "Choose An Item", "")
    )
    
    shinyApp(server = server, ui = interface)

Will the cross-browser compatibility of Shiny be improved?

bborgesr:
Fixes #1257


wch:
Are there any drawbacks to using `cache: false`?

jcheng5:
Doesn't seem like there's much downside. It adds some crud to the URL to bust the cache. I guess if the data hasn't changed and you hit the URL several times, it might be making unnecessary requests?

Our other option is to do this on the server side--add the cache-busting parameter to the `registerDataObj` method that forms the URL that is sent to the client. The benefits of this are 1) this problem will be solved for all clients--nobody else need remember set `cache:false`, and 2) you'd actually get some desirable caching behavior (i.e. you wouldn't have the problem in the previous paragraph).

bborgesr:
Ok, changed it to your suggestion, Joe. It seems to work equally well. I basically added a pseudoparam that sticks in the timestamp in the URL, tricking the browser into thinking that it's a new page so caching doesn't occur. This is how I think it's done by ajax anyway: " It works by appending "_={timestamp}" to the GET parameters." (http://api.jquery.com/jquery.ajax/). What do you think?

srcjs/input_binding_select.js
```diff
@@ -59,7 +59,7 @@ $.extend(selectInputBinding, {
       selectize.settings.load = function(query, callback) {
         var settings = selectize.settings;
         $.ajax({
-          cache: false,
+          //cache: false,
```
jcheng5: Why not just delete this?

R/shiny.R
```diff
@@ -1348,10 +1348,11 @@ ShinySession <- R6Class(
     registerDataObj = function(name, data, filterFunc) {
       # abusing downloads at the moment
       self$downloads$set(name, list(data = data, filter = filterFunc))
-      return(sprintf('session/%s/dataobj/%s?w=%s',
+      return(sprintf('session/%s/dataobj/%s?w=%s?timestamp=%s',
```
jcheng5: You've added a second `?` here--you want `&`.

Also, instead of timestamp, I'd personally use `createUniqueId(8)` or something like that, but I don't have a rational justification for it. :) So either way, just giving you the option.

srcjs/input_binding_select.js
bborgesr: Oh, no reason, just because I wasn't sure this was really better than the alternative, so I was just experimenting with it, pending feedback. But it seems like there's no objections, so I'll get rid of it.

R/shiny.R
bborgesr: Oops, sorry, changed it to `&`. I actually prefer the `createUniqueId` too, it looks much neater than `as.character(as.numeric(Sys.time()))`. Good to merge?

PR #1281, opened 2016-08-10
Bookmarking fixes

### LINKED ISSUE #1273: Remaining bookmarkable state issues
- [x] The `onBookmark`, `onBookmarked`, `onRestore`, and `onRestored` observers should print errors and stack trace to the console. UPDATE: It turns out that they already do this.
- [x]  In `ReactiveValues`, the `invalidate` and `uninvalidate` should be renamed to something less confusing.
- [x]  Calling `updateXxxInput` functions inside of the `onRestored` callback doesn't work because it puts things the `inputMessageQueue` after `session$flushOutput()` is called, but there's no `msg` to trigger another flush.

wch:
This closes #1273.


R/shiny.R
```diff
@@ -962,20 +962,37 @@ ShinySession <- R6Class(
...
+          length(private$inputMessageQueue) != 0
+        )
+      }
+
       # ..stacktraceon matches with the top-level ..stacktraceoff..
       private$flushCallbacks$invoke(..stacktraceon = TRUE)
-      # ..stacktraceon matches with the top-level ..stacktraceoff..
-      on.exit(private$flushedCallbacks$invoke(..stacktraceon = TRUE))
 
-      if (length(private$progressKeys) == 0
```
wch: Previously, the check for `private$progressKeys` didn't quite have the right logic, I believe. If there was anything in `private$progressKeys`, *all* of the subsequent code in this function would run, including sending a message that didn't contain anything from `private$progressKeys`:

```R
 private$sendMessage(
         errors = as.list(errors),
         values = as.list(values),
         inputMessages = inputMessages
       )
```

I think that instead, `private$progressKeys` should just get reset every time this function is called. (That's how I've implemented it, below.)

wch:
This should be good to merge now.

R/shiny.R
```diff
@@ -1552,17 +1568,37 @@ outputOptions <- function(x, name, ...) {
 #'
 #' @export
 onFlush <- function(fun, once = TRUE, session = getDefaultReactiveDomain()) {
-  session$onFlush(fun)
+  session$onFlush(fun, once = once)
```
jcheng5: Heh, good catch

R/shiny.R
```diff
@@ -961,21 +958,40 @@ ShinySession <- R6Class(
...
+        if (hasPendingUpdates() || .getReactiveEnvironment()$hasPendingFlush()) {
+          scheduleFlush()
+        }
+      })
 
+      # Clear any progress keys
       private$progressKeys <- character(0)
 
+      if (!hasPendingUpdates()) {
+        return(invisible())
```
jcheng5: It actually may not be safe to return early here. I think shiny.js treats the reception of an errors/values message (can't remember which) as an indication that the flushReact loop has completed, and uses that knowledge to stop showing progress.

jcheng5: https://github.com/rstudio/shiny/blob/master/inst/www/shared/shiny.js#L1023-L1025

R/shiny.R
jcheng5: The original code was correct. The existence of progress keys means we need to send a `values` message (even if empty) as this has a side effect of clearing out progress indication on the client. See the other note I added.

jcheng5:
Other than those progress-related fixes that need to be made, this looks good. Let me review one more time after you fix those.

wch:
OK, I've restored the original logic for `progressKeys`.

PR #1287, opened 2016-08-11
update use to `stopWithCondition` to `reactiveSTop`

bborgesr:
Specifically, do so in the bookmarkable code that was introduced when the idea of `reactiveStop` was still in flux. As per Winston's suggestion [here](https://github.com/rstudio/shiny/pull/1209#discussion_r73452037) (my follow-up comment [here](https://github.com/rstudio/shiny/pull/1209#discussion_r73568224)). I'm pretty sure this is a completely safe merge (I used this line in other parts of the code to sub exactly the same thing). But better safe than sorry.

cc @jcheng5, @wch

### DIFF
--- a/R/reactives.R
+++ b/R/reactives.R
@@ -82,7 +82,7 @@ ReactiveValues <- R6Class(
       }
 
       if (isInvalid(key))
-        stopWithCondition(c("validation", "shiny.silent.error"), "")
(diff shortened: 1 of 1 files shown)


wch:
I think it doesn't need the `"validation"` class added.

jcheng5:
@wch will make this change independently

PR #1291, opened 2016-08-11
fixes #1093

### LINKED ISSUE #1093: updateRadioButtons errors if numeric vector used as choices
There seems be a problem with using multiple updateRadioButtons, see my question here:

[http://stackoverflow.com/questions/35056668/update-two-sets-of-radiobuttons-shiny/35061269?noredirect=1#comment57845550_35061269]

In short, if there are more than one set of radioButtons to update reactively you must explicitly state selected=" " in the updateRadioButtons argument. This isn't a big deal but it took me a day to figure out why I couldn't seem to render to sets of radioButtons reactively.

bborgesr:
A very simple casting of `choices` and `selected` to "character" inside the `updateRadioButtons` function.

I spent some time trying to figure out exactly why this bug was occurring (i.e. why it only affects the second set of radio buttons) and didn't get very far.

I also wonder how many other `updateXXX` functions could benefit from a similar type casting... Is there a reason we didn't cast things before?

### DIFF
--- a/R/update-input.R
+++ b/R/update-input.R
@@ -511,8 +511,8 @@ updateCheckboxGroupInput <- function(session, inputId, label = NULL,
 #' @export
 updateRadioButtons <- function(session, inputId, label = NULL, choices = NULL,
                                selected = NULL, inline = FALSE) {
-  choices <- as.character(choices)
-  selected <- as.character(selected)
+  if (!is.null(choices)) choices <- as.character(choices)
+  if (!is.null(selected)) selected <- as.character(selected)
 
   # you must select at least one radio button
   if (is.null(selected) && !is.null(choices)) selected <- choices[[1]]
--- a/R/update-input.R
+++ b/R/update-input.R
@@ -511,6 +511,9 @@ updateCheckboxGroupInput <- function(session, inputId, label = NULL,
 #' @export
 updateRadioButtons <- function(session, inputId, label = NULL, choices = NULL,
                                selected = NULL, inline = FALSE) {
+  choices <- as.character(choices)
+  selected <- as.character(selected)
+
   # you must select at least one radio button
   if (is.null(selected) && !is.null(choices)) selected <- choices[[1]]
   updateInputOptions(session, inputId, label, choices, selected, inline, type = 'radio')


bborgesr:
Here's a simple example:

```r
library(shiny)

ui <- fluidPage(
  radioButtons("rd1", "RD1", 1:3, 3),
  radioButtons("rd2", "RD2", 1:3, 3),
  actionButton("btn", "Go")
)

  observeEvent(input$btn,
    {
      updateRadioButtons(session, "rd1", 
         choices = (as.numeric(input$rd1) + 1):(as.numeric(input$rd1) + 3),
         selected = as.numeric(input$rd1) + 1)
      updateRadioButtons(session, "rd2",
         choices = (as.numeric(input$rd2) + 1):(as.numeric(input$rd2) + 3),
         selected = as.numeric(input$rd2) + 1)         
    }
  )
}

shinyApp(ui, server)
```

wch:
When I run your test app, I do see that only the first group gets updated. With `options(shiny.trace=T)`, you can see that the content being sent to both is the same:

```JSON
{  
   "errors":[  

   ],
   "values":[  

   ],
   "inputMessages":[  
      {  
         "id":"rd1",
         "message":{  
            "options":"<div class=\"shiny-options-group\">\n  <div class=\"radio\">\n    <label>\n      <input type=\"radio\" name=\"rd1\" value=\"3\" checked=\"checked\"/>\n      <span>3</span>\n    </label>\n  </div>\n  <div class=\"radio\">\n    <label>\n      <input type=\"radio\" name=\"rd1\" value=\"4\"/>\n      <span>4</span>\n    </label>\n  </div>\n</div>",
            "value":3
         }
      },
      {  
         "id":"rd2",
         "message":{  
            "options":"<div class=\"shiny-options-group\">\n  <div class=\"radio\">\n    <label>\n      <input type=\"radio\" name=\"rd2\" value=\"3\" checked=\"checked\"/>\n      <span>3</span>\n    </label>\n  </div>\n  <div class=\"radio\">\n    <label>\n      <input type=\"radio\" name=\"rd2\" value=\"4\"/>\n      <span>4</span>\n    </label>\n  </div>\n</div>",
            "value":3
         }
      }
   ]
}
```

In the JS console, I see:

```
Uncaught TypeError: val.replace is not a function
```

=====

With your fix, this is what gets sent:

```json
{  
   "errors":[  

   ],
   "values":[  

   ],
   "inputMessages":[  
      {  
         "id":"rd2",
         "message":{  
            "options":"<div class=\"shiny-options-group\">\n  <div class=\"radio\">\n    <label>\n      <input type=\"radio\" name=\"rd2\" value=\"3\" checked=\"checked\"/>\n      <span>3</span>\n    </label>\n  </div>\n  <div class=\"radio\">\n    <label>\n      <input type=\"radio\" name=\"rd2\" value=\"4\"/>\n      <span>4</span>\n    </label>\n  </div>\n</div>",
            "value":"3"
         }
      },
      {  
         "id":"rd1",
         "message":{  
            "options":"<div class=\"shiny-options-group\">\n  <div class=\"radio\">\n    <label>\n      <input type=\"radio\" name=\"rd1\" value=\"3\" checked=\"checked\"/>\n      <span>3</span>\n    </label>\n  </div>\n  <div class=\"radio\">\n    <label>\n      <input type=\"radio\" name=\"rd1\" value=\"4\"/>\n      <span>4</span>\n    </label>\n  </div>\n</div>",
            "value":"3"
         }
      }
   ]
}
```

The only difference I see is that instead of `"value":3`, it's `"value":"3"`. Your PR seems to fix it.

======

I think that the reason that only one group gets replaced with the original code is because it successfully replaces one, and then an error is thrown before it gets to the second group.

The error happens here, in the `$escape` function:
  https://github.com/rstudio/shiny/blob/20f0566/srcjs/utils.js#L202
It gets called from:
  https://github.com/rstudio/shiny/blob/20f0566/srcjs/input_binding_radio.js#L11

I think it would make sense to coerce `val` to a string before doing the `.replace` on it.

This bug also affects `checkboxGroupInput`, and fixing the `$escape` function should fix that as well.

PR #1292, opened 2016-08-11
fixes #1122

### LINKED ISSUE #1122: downloadHandler error not seen by client
The following test case produces a Shiny application which never responds to download requests.

This case intentionally errs within the `downloadHandler`; the bug is that the browser is not sent an error response. Instead, the browser sits endlessly waiting.

```r
library(shiny)

  plots <- reactiveValues(scatter=NULL, histres=NULL, resx=NULL, reshyat=NULL);
  output$save <- downloadHandler(
      filename = paste("Scatterplot.png"),
      content = function(file) {
        png(file, height=400, width=600)
        print(isolate(plots$scatter))
        dev.off()
      },
      contentType = "image/png"
  );
};

ui <- fluidPage(
    titlePanel("bogus download"),
    downloadButton("save", "Download")
);

shiny::shinyApp(server = server, ui = ui);
```

When running locally, the following output is produced:

```
Listening on http://127.0.0.1:5801
NULL
Error opening file: 2
Error reading: 9
```

bborgesr:
sends a "No content" HTTP error 

@jcheng5: for some reason, this need to be a 400 code -- I tried the 204 code and the message "No content" shows up in the RStudio viewer, but not in any of the browsers (Chrome, Firefox, Safari). It doesn't seem all that wrong to have a 400 in here either because the content function is never meant to be empty or malformed (i.e. falls under the general "Bad Request" category)... What do you think?

### DIFF
--- a/R/shiny.R
+++ b/R/shiny.R
@@ -1269,7 +1269,9 @@ ShinySession <- R6Class(
         ext <- tools::file_ext(filename)
         if (nzchar(ext))
           ext <- paste(".", ext, sep = "")
-        tmpdata <- tempfile(fileext = ext)
+        tmpdata <- tempfile( tmpdir = tempdir(), fileext = ext)
+        snapshot <- utils::fileSnapshot(tempdir())
+
         # ..stacktraceon matches with the top-level ..stacktraceoff..
         result <- try(shinyCallingHandlers(Context$new(getDefaultReactiveDomain(), '[download]')$run(
           function() { ..stacktraceon..(download$func(tmpdata)) }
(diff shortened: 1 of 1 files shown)


R/shiny.R
```diff
@@ -1269,7 +1269,9 @@ ShinySession <- R6Class(
         ext <- tools::file_ext(filename)
         if (nzchar(ext))
           ext <- paste(".", ext, sep = "")
-        tmpdata <- tempfile(fileext = ext)
+        tmpdata <- tempfile( tmpdir = tempdir(), fileext = ext)
+        snapshot <- utils::fileSnapshot(tempdir())
```
wch: It looks like his function was added in R 3.1.0 (see the history of https://github.com/SurajGupta/r-source/blob/db0ce24/src/library/utils/R/changedFiles.R#L19). I think that's probably OK, but we should also rev the required R version in the DESCRIPTION file.

bborgesr: Any objections, @jcheng5?

jcheng5: (as we discussed over vchat) the test should simply change to `file.exists(tmpdata)`, no need to snapshot and compare.

jcheng5:
Replaced by #1314

PR #1297, opened 2016-08-12
updateDateInput: better handling of malformed dates. Closes #1179

### LINKED ISSUE #1179: updateDateInput
I 'm facing a problem with "updateDateInput" :

for example, while trying to update my dateInput with the value "2016-04-02" :

updateDateInput(session, "date.ddn" , label = NULL, min=NULL, max=NULL, value="2016-04-02")

The display within the shiny application correspond to the "2016-04-01"

This bug doesn't exist in shiny 0.13.1 (which I'm using for developpement, on another device that run macosx Yosemite with R.3.2.2), but there, I'm trying to deploy my shiny app on a shiny 0.13.2, on macOSX leopard (R 3.2.1)

Thanks

wch:
Test app. Move the slider and check that the min and max are one day before and after the selected value.

```R
ui <- fluidPage(
  sliderInput("controller", "Controller", 1, 30, 5),
  dateInput("inDate", "Input date")
)
  observe({
    x <- input$controller

    updateDateInput(session, "inDate",
      label = paste("Date label", x),
      value = paste("2013-04-", x, sep=""),
      min = paste("2013-04-", x-1, sep=""),
      max = paste("2013-04-", x+1, sep="")
    )
  })
}
shinyApp(ui, server)
```

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -90,6 +90,8 @@ shiny 0.13.2.9005
 * Fixed #1144: updateRadioButton and updateCheckboxGroupInput break controls
   when used in modules (thanks, @sipemu!).
 
+* Fixed #1179: updateDateInput sometimes didn't work with malformed dates.
+
 shiny 0.13.2
 --------------------------------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


jcheng5:
Would you expect this to break any apps today that users might reasonably expect to work?

If not, LGTM.

wch:
I can't think of any reason that this would break an app. Well, I suppose there could be a case where someone passes a string that the client would treat as a date, but wouldn't be correctly converted by `as.Date` in R, like in the other date issues where a year of 99999 is valid in Javascript but not in R. That's really esoteric though, and I don't think we need to worry about it.

PR #1298, opened 2016-08-12
Don't crash on malformed date input values. Closes #803

### LINKED ISSUE #803: dateInput(): entering numbers into field crashes app
What happens
--------------------
If you enter digits directly into dateInput() field, you can get the whole app to crash with
```r
Error in charToDate(x) : 
  character string is not in a standard unambiguous format
```

How to replicate
----------------------

* Have a dateInput() field in a shiny app
* put your cursor in it and just start pressing '1'. 
* after a few presses, app will crash because of unparseable date.

Expected behaviour
---------------------------
A user entering an incorrect/unparseable date into a form field shouldn't crash the app. 

Running R3.1.3, shiny-0.11.1 on rstudio-0.99.335

wch:
Test app. Entering very large numbers for the year (like 99999) previously would crash the app.

```R
app <- shinyApp(
  ui = bootstrapPage(
    dateRangeInput('dateRange', 'Date Range Input'),
    dateInput('date', 'Date Input')
    
  ),
    observe({
      print(input$dateRange)
      print(input$date)
    })
  }
)
runApp(app)
```

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -94,7 +94,10 @@ shiny 0.13.2.9005
   means a more-expensive no-op than necessary).
 
 * Fixed #1179: updateDateInput sometimes didn't work with malformed dates.
+
 * Added inputTextArea. (Thanks to @nuno-agostinho)
+
+* Fixed #803: Malformed date input values crashed app.
 
 shiny 0.13.2
 --------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


jcheng5:
LGTM. (Just to be clear, you mean ends the session, right? Or does it literally cause the entire app to exit, even other sessions?)

wch:
Right, it would just result in the one session ending.

Issue #896, opened 2015-07-16
Allow `dateInput` to have an unspecified value


aronatkins:
It would be nice to allow a `dateInput` control to have an unspecified value. Right now, if you provide `value=NULL`, you get "today" by default.

The use case is one where I want to know if the user has made a explicit choice of a date value. Not choosing a date value could mean that a search is unbounded, for example.

hadley:
`as.Date(NA)` would be a natural value to use here

gmeurice:
I've tried to use the as.Date(NA), but still, when the user erase the date from dateInput widget, the returned value is still Sys.Date(), and not "", as I expected.

How could a user can choose to leave the fields empty (or to erase a wrong date input to leave it empty) whithout systematically returning "Sys.Date()" ?

PR #1299, opened 2016-08-12
Allow updateDateInput to set input fields to empty

gaborcsardi:
You need to set it to the empty string, and then it'll be empty. I am not sure if this is the best option, maybe some other special value, like `NA_character_` is better.

If you like this, then I can do the same for date ranges as well, and also add a sentence to the manual.

I have re-generated the minified js files as well, let me know if this is ok, or you want me to remove them.

Here is a simple app to test it:

```r
shinyApp(
  ui = fluidPage(
    div(
      dateInput("date", "Date", min = "2014-06-10")
    ),
    actionButton("btn", "reset date")
  ),
    observeEvent(input$btn, updateDateInput(session, "date", value = ""))
  }
)
```

UPDATE: the app to test with `NA` values:

```r
shinyApp(
  ui = fluidPage(
    div(
      dateInput("date", "Date", min = "2014-06-10"),
      dateRangeInput("daterange", "Date range:",
               start = "2001-01-01",
               end   = "2010-12-31")
    ),
    actionButton("btn1", "reset date"),
    actionButton("btn2", "reset date range start"),
    actionButton("btn3", "reset date range end"),
    actionButton("btn4", "reset date range start and end")
  ),
    observeEvent(input$btn1, updateDateInput(session, "date", value = NA))
    observeEvent(input$btn2, updateDateRangeInput(session, "daterange", start = NA, end = "2005-01-01"))
    observeEvent(input$btn3, updateDateRangeInput(session, "daterange", start = "2003-01-01", end = NA))
    observeEvent(input$btn4, updateDateRangeInput(session, "daterange", start = NA, end = NA))
  }
)
```


wch:
I think this is a good idea. However, we recently made a change to `updateDateInput` where it coerces the value with `as.Date()` and then calls `format(x, "%Y-%m-%d")` on it, which won't work with empty strings. I think it makes more sense to use `NA`.

Could you implement that for 

Also for future reference, this fixes #896.

gaborcsardi:
> I think this is a good idea. However, we recently made a change to updateDateInput where it coerces the value with as.Date() and then calls format(x, "%Y-%m-%d") on it, which won't work with empty strings. I think it makes more sense to use NA.

OK, I like NA in general, but it is somewhat harder to deal with in JS as it has no good representation in JS. Do you just use the default representation coming out of jsonlite?

> Could you implement it that way for updateDateInput and updateDateRangeInput? Also, please add a bullet point to NEWS.

Sure.

wch:
> OK, I like NA in general, but it is somewhat harder to deal with in JS as it has no good representation in JS. Do you just use the default representation coming out of jsonlite?

Yes, I think it gets turned into NULL, so you should be able to check for `date === null` in JS.

wch:
@gaborcsardi Do you mind updating this PR? We're trying to wrap things up soon for a release.

gaborcsardi:
Sure, I'll try today....

gaborcsardi:
@wch updateDateRangeInput is buggy I believe. Here:
https://github.com/rstudio/shiny/blob/42dbb128be53d41a8df7c1a4d6794dab1dbf7f50/R/update-input.R#L277

If `start` is not updated, i.e. it is `NULL`, then `c(start, end)` will drop it, and `end` will take its place. Do you want to fix this here, or separately? Maybe separately makes more sense.

wch:
Sure, a separate PR works. Thanks!

gaborcsardi:
PTAL.

srcjs/input_binding_date.js
```diff
@@ -16,10 +16,17 @@ $.extend(dateInputBinding, {
   setValue: function(el, value) {
     var date = this._newDate(value);
     // If date is invalid, do nothing
+    // Note: R's NA is null here, and isNaN(null) is false
```
wch: I just want to check: `this._newDate(null)` is `null`?

gaborcsardi: It is now. :) See my change below at https://github.com/rstudio/shiny/pull/1299/files/3707cd813f1262410681a8f1f0b2f1186081ad73#diff-664d9420c5db5c8d072c778dcb0be113R155

Is that OK? I am a bit worried that I cannot unit test this stuff....

wch: I think the change in behavior to `this._newDate()` is OK, but I'd prefer logic like this:

```js
if (value === null) {
  $(el).find('input').val('').datepicker('update');
  return;
}

// original code here
```

It's not a big deal - it's just a little easier to mentally process.

gaborcsardi: OK, updated.

wch: Sorry for being unclear -- I meant checking `value === null` right at the top of this function.

gaborcsardi: Sure, NP. I guess you want the same with the date range then. 

And it is not needed to modify `_newDate`. Right?

wch: Right, I think `_newDate` can stay as it was.

gaborcsardi:
OK, please take a look.

wch:
Great, thanks!

PR #1309, opened 2016-08-18
Allow app authors to set bookmark trigger event

### LINKED ISSUE #1305: Make bookmark input ID customizable
This would be useful for when there are tabs.

wch:
Closes #1305. This adds a function call `setBookmarkEvent`, which allows the app author to set the event that triggers bookmarking. The default is to use `input[["._bookmark_"]]`, which corresponds to the `bookmarkButton`'s default ID.

This still needs an example in the documentation for `setBookmarkEvent`, but it also needs a better syntax than what I have so far. For example, this app has two buttons on different tabs that each can be used to trigger bookmarking, but the expression in `setBookmarkEvent()`  is really ugly:

```R
ui <- function(request) {
  fluidPage(
    tabsetPanel(id = "tabs",
      tabPanel("One",
        checkboxInput("chk", "Checkbox"),
        bookmarkButton(id = "bookmark1")
      ),
      tabPanel("Two", bookmarkButton(id = "bookmark2"))
    )
  )
}

  setBookmarkExclude(c("bookmark1", "bookmark2"))

  setBookmarkEvent({
    # This is hideous - need a better way of expressing this
    if (!isTruthy(input$bookmark1) & !isTruthy(input$bookmark2))
      NULL
    else 
      TRUE
  })
}

enableBookmarking(store = "url")
shinyApp(ui, server)
```


wch:
Closing in favor of #1310, which is a better way of doing it.

Issue #1063, opened 2016-01-02
using shinyAppDir() does not seem to recognize global.R


daattali:
I'm using shiny server pro (if you need the version number, then please remind me how to find out?) and I've noticed that when I try to run apps in a different directory with `shinyAppDir()` , objects defined in `global.R` are not recognized.

Example:

test1/global.R:

```
myvar <- 10
```

test1/ui.R:

```
fluidPage("HELLO", myvar)
```

test1/server.R:

```
server <- function(input, output, session) {
  cat(myvar)
}
```

If I go to the URL for this app, it works.

But if I have in a different directory

test2/app.R:

```
setwd("../test1")
shiny::shinyAppDir(".")
```

And go to that app directory, I get error `ERROR: object 'myvar' not found`.  If I add a `source("global.R")`, then it works.

Repeating this on my local computer does not result in an error.

daattali:
Opened this issue with shiny-server instead
https://github.com/rstudio/shiny-server/issues/180

jcheng5:
Discussion [here](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/shiny-discuss/ibAXs3zLWAs/YOVRVRU2AQAJ)

PR #1313, opened 2016-08-22
Add enableBookmarking argument to shinyApp()

### LINKED ISSUE #1312: Add enableBookmarking as an option to shinyApp()
Some of our early testers have requested the ability to do something like `shinyApp(enableBookmarking="url")`.

wch:
This closes #1312. Opinions on the interface? If this looks good, I'll need to make a few more changes to the documentation before we merge.

cc @daattali

### DIFF
--- a/R/app.R
+++ b/R/app.R
@@ -27,6 +27,11 @@
 #'   request to determine whether the \code{ui} should be used to handle the
 #'   request. Note that the entire request path must match the regular
 #'   expression in order for the match to be considered successful.
+#' @param enableBookmarking Can be one of \code{"url"}, \code{"server"}, or
+#'   \code{"disable"}. This is equivalent to calling the
+#'   \code{\link{enableBookmarking}()} function just before calling
+#'   \code{shinyApp()}. See \code{\link{enableBookmarking}} for more
+#'   information.
 #' @return An object that represents the app. Printing the object or passing it
 #'   to \code{\link{runApp}} will run the app.
 #'
@@ -61,7 +66,7 @@
 #' }
 #' @export
 shinyApp <- function(ui=NULL, server=NULL, onStart=NULL, options=list(),
-                     uiPattern="/") {
+                     uiPattern="/", enableBookmarking = NULL) {
   if (is.null(server)) {
     stop("`server` missing from shinyApp")
   }
@@ -73,6 +78,11 @@ shinyApp <- function(ui=NULL, server=NULL, onStart=NULL, options=list(),
 
   serverFuncSource <- function() {
     server
+  }
+
+  if (!is.null(enableBookmarking)) {
+    bookmarkStore <- match.arg(enableBookmarking, c("url", "server", "disable"))
+    enableBookmarking(bookmarkStore)
   }
 
   # Store the appDir and bookmarking-related options, so that we can read them
--- a/R/app.R
+++ b/R/app.R
@@ -30,8 +30,9 @@
 #' @param enableBookmarking Can be one of \code{"url"}, \code{"server"}, or
 #'   \code{"disable"}. This is equivalent to calling the
 #'   \code{\link{enableBookmarking}()} function just before calling
-#'   \code{shinyApp()}. See \code{\link{enableBookmarking}} for more
-#'   information.
+#'   \code{shinyApp()}. With the default value (\code{NULL}), the app will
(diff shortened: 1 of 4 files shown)


daattali:
Thank you for listening to our feedback, this is much better IMO.  A few small comments that I personally think could be improved:

- I'd think it makes more sense to have the param called `bookmark` or `bookmarking` rather than `enableBookmarking` (shorter, and since it's not a function call then it makes less sense to have it be a verb)
- Is there any reason to include the `disable` option?  If I want it disabled I just won't include the param or set it to NULL, but why would I ever explicitly set it to `disabled`?
- Is it easy to add this param to all the other ways of running a shiny app? For example, on my shiny server I often use `shinyAppDir()` so ideally that function will also take this param (and any other similar functions, I don't know what all the ways are)

wch:
Point taken for the argument name, but after some discussion between me and @jcheng5, we decided to stick with `enableBookmarking`.

For `disable`, that's present for the `enableBookmarking()` function because we want to give people a way to turn it off, if they've turned it on. I think it makes sense to keep the argument the same here. If `NULL` is used here, then it leaves bookmarking unchanged (so a previous `enableBookmarking()` could have an effect).

These are the scenarios that we need to make sure work:
- shinyApp()
  - enableBookmarking()
  - enableBookmarking = "url"
- shinyAppDir()
  - enableBookmarking()
  - enableBookmarking = "url"
- shinyAppFile()
  - enableBookmarking()
  - enableBookmarking = "url"
- runApp() with a directory containing ui.R and server.R
  - enableBookmarking()
- runApp() with a directory containing app.R
  - enableBookmarking()
- runApp() with a list
  - enableBookmarking()

daattali:
Have you considered completely removing the `enableBookmarking()` function that needs to be called independently of a shiny app, and **only** having it as a parameter that is attached to a specific app?

wch:
If you remove `enableBookmarking()`, how would you handle cases where people call `runApp()` in a directory with server.R/ui.R or app.R? We don't want to make bookmarking an argument to `runApp()`; we want to make it something that is a property of the app, not the way that it's invoked.

daattali:
I'm not sure I have a better solution to offer...

> We don't want to make bookmarking an argument to runApp(); we want to make it something that is a property of the app

I agree, but the way it's currently done is *not*  a property of the app, and I do personally think it's safer and more consistent to ask the user to provide it as a param to `runApp()` rather than relying on them to call a function before. 

The best option would be to redesign the implementation so that it doesn't have to be called before the app gets run, and it'll be truly part of the app, that would make it a lot simpler.  But I assume that is out of the question :)

Another random thought I had was "it can be in the DESCRIPTION file" but I haven't seen that file being used much, I feel like it's not something you're really pushing for people to use.

I don't have any more thoughts on the topic, if it remains as it is right now it's good enough for my purposes since I never use runApp, so I won't argue anymore! But it still feel like a very weird interaction to have to call `enableBookmarking()` every time before an app is run

wch:
> I agree, but the way it's currently done is not a property of the app, and I do personally think it's safer and more consistent to ask the user to provide it as a param to runApp() rather than relying on them to call a function before.

But that's not how they are to use it. If they're using `runApp` with an app.R or server.R/ui.R/global.R, then the app author puts `enableBookmarking()` in the app.R or global.R. (Or, with this PR, for app.R, they can use `shinyApp(enableBookmarking = "url")`.)

wch:
This should be good to go now.

After some discussion, @jcheng5  and I decided that `shinyAppDir` and `shinyAppFile` don't need the `enableBookmarking` argument. In both of those cases, the app author has the ability to enable bookmarking in the global.R or app.R, by calling `enableBookmarking()` or `shinyApp(enableBookmarking = X)`.

`shinyApp` has arguments that `shinyAppDir` and `shinyAppFile` don't have, such as `onStart`, so this would be following in that pattern.

daattali:
This will be a bit annoying to use on shiny servers because of https://github.com/rstudio/shiny/issues/1063 since it forces us to put a function in global.R even though global.R isn't recognized when using `shinyAppDir()`.  I guess you can't make everyone happy :)

jcheng5:
Do you mean on shiny server when the app is in a package?

daattali:
Yes that's my main use case, but also any other time when you want to use `shinyAppDir()`.  But yes you're correct, every time somebody asks about how to run an app within a package my answer is to use shinyAppDir() and people generally really like that solution

PR #1329, opened 2016-08-30
Add old-style rendering option to progress bars

### LINKED ISSUE #1327: Think about progress indicator back compat
If we decide not to make any changes, at least move the NEWS item to Breaking Changes.

wch:
Closes #1327. This adds an argument, `style`, to `withProgress` and `Progress$new`, which lets the user choose new notification-based progress bars with `"notification"` (the default), or the older style progress bars with `"old"`.

This is so that users who have custom-styled their progress bars with CSS can keep that CSS. It will require them to add the `style="old"` argument, however.

We had also discussed adding a "raw" option where the progress bars would get CSS classes but we wouldn't actually have any CSS that styles the bars. I think it would be better to wait on that until the next version of Shiny.


NEWS.md
```diff
@@ -17,7 +17,9 @@ shiny 0.13.2.9005
   `showNotification` and `hideNotification`. From JavaScript, there is a new
   `Shiny.notification` object that controls notifications. (#1141)
 
-* Progress indicators now use the notification API. (#1160)
+* Progress indicators can now either use the new notification API, using
+  `style="notification"`, or be displayed with the previous styling, using'
+  `style="old"`. (#1160, #1327, #1329)
```
jcheng5: Should be moved to "breaking changes" section and reworded to make clear that the default behavior is the new style.

bborgesr: I can take care of that, since I'm working on NEWS. You can just undo the change to NEWS, @wch if you want (or we can work it out later because there will be a merge conflict otherwise).

wch: I'll take care of it here, since it makes sense to put it in this PR. I'll probably need to add something about it to the release notes after it's merged.

PR #1330, opened 2016-08-30
fix a bug in the modal example

shrektan:
`exists(input$dataset)` will return error if `input$dataset` is `""` or `NULL`


wch:
Thanks for letting us know about this! I fixed it a different way, in ad171d6.

shrektan:
It's me that should thank you for making such a fantastic pkg  :bow: 

Ha, your way is better than mine ~ Didn't know `nzchar` before :smiley:

PR #1344, opened 2016-08-31
Let modules get at the root scope (undocumented for now)

jcheng5:
This is needed for plotly. I really hope people are not tempted to abuse this to do inter-module communication.


jcheng5:
cc @cpsievert

PR #1360, opened 2016-09-02
Check whether hosting environment supports bookmarking

wch:
This adds a check for whether the hosting environment supports saved-to-server bookmarking. If not, then, when the user tries to bookmark state, it shows an error in a notification.

I've tested this with:
* the latest version of Shiny Server (1.4.4.804)
* an older version of Shiny Server (1.4.4.801)
* [shinyapps.io](https://winston.shinyapps.io/bookmark-server/)
* [connect](https://beta.rstudioconnect.com/content/1701/)
* Shiny Server Pro

Bookmarking works with the new version of Shiny Server. With all the others, the error message comes up as expected.

### DIFF
--- a/R/bookmark-state.R
+++ b/R/bookmark-state.R
@@ -76,7 +76,25 @@ saveShinySaveState <- function(state) {
 
   # Pass the saveState function to the save interface function, which will
   # invoke saveState after preparing the directory.
-  saveInterface <- getShinyOption("save.interface", default = saveInterfaceLocal)
+
+  # Look for a save.interface function. This will be defined by the hosting
+  # environment if it supports bookmarking.
+  saveInterface <- getShinyOption("save.interface")
+
+  if (is.null(saveInterface)) {
+    if (inShinyServer()) {
+      # We're in a version of Shiny Server/Connect that doesn't have
+      # bookmarking support.
+      saveInterface <- function(id, callback) {
+        stop("The hosting environment does not support saved-to-server bookmarking.")
+      }
+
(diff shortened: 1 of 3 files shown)


R/bookmark-state.R
```diff
@@ -76,7 +76,25 @@ saveShinySaveState <- function(state) {
...
+  # Look for a save.interface function. This will be defined by the hosting
+  # environment if it supports bookmarking.
+  saveInterface <- getShinyOption("save.interface")
+
+  if (is.null(saveInterface)) {
+    if (inShinyServer()) {
+      # We're in a version of Shiny Server/Connect that doesn't have
+      # bookmarking support.
+      saveInterface <- function(id, callback) {
+        stop("The hosting environment does not support saved-to-server bookmarking.")
```
jcheng5: This waits until someone actually attempts to save a bookmark to error. Can we also give an error when the app starts, or something?

jcheng5: Or maybe it's too harsh to error on startup--a warning notification?

bborgesr: Yes, I vote for the warning notification -- it makes sense to let people know that this functionality is not available (so they don't waste time getting their state ready to bookmark), but an error does feel too harsh (especially when this is something that will, in all likelihood, work fine in the future).

wch: A warning notification seems reasonable.

wch:
Should be good to go now. I tested the change on all the platforms above.

PR #1370, opened 2016-09-13
Fix bookmark fileinput 

### LINKED ISSUE #1368: Problem with fileInput and bookmarkable state
Using the example app below:

1. Upload a file
2. Create a bookmark URL
3. Open the new URL
4. Create a bookmark URL
5. Open the new URL

You get an error like this (note that the path is two absolute URLs pasted together):

```
  cannot open file '/Users/jcheng/Desktop/20160913-EARLLondon-bookmark/shiny_bookmarks/e980d9618cab972f//Users/jcheng/Desktop/20160913-EARLLondon-bookmark/shiny_bookmarks/98d0f7a4b2090fce/0': No such file or directory
```

```r
library(shiny)
library(ggplot2)

ui <- function(request) {
  fluidPage(
    sidebarLayout(
      sidebarPanel(
        fileInput("file", "CSV file"),
        uiOutput("column_ui"),
        bookmarkButton()
      ),
      mainPanel(
        tabsetPanel(id = "tabs",
          tabPanel("Plot",
            plotOutput("plot")
          ),
          tabPanel("Model",
            verbatimTextOutput("model")
          )
        )
      )
    )
  )
}

  dataset <- reactive({
    req(input$file)
    read.csv(input$file$datapath)
  })
  
  output$column_ui <- renderUI({
    tagList(
      selectInput("x", "X variable", c("", names(dataset()))),
      selectInput("y", "Y variable", c("", names(dataset())))
    )
  })
  
  output$plot <- renderPlot({
    req(input$x, input$y)
    ggplot(dataset(), aes_string(x = input$x, y = input$y)) +
      geom_point()
  })
  
  output$model <- renderPrint({
    req(input$x, input$y)
    lm(as.formula(paste(input$x, "~", input$y)), data = dataset())
  })
}

enableBookmarking("server")

shinyApp(ui, server)
```

wch:
Fixes #1368. The problem was that a restored fileInput wasn't marked with the correct serializer because it uses a different code path from a file that was uploaded the normal way.

This PR also:
* Copies the restored file to a temp dir. Previosuly, it let the restored session access the bookmarked file directly, which could be a problem if the user's code modifies the file.
* Has a better check that the restored file doesn't contain a path, so that a malicious user can't access any other parts of the filesystem.
* Uses `createUniqueId` in the `FileUploadContext`, for consistency.

### DIFF
--- a/R/fileupload.R
+++ b/R/fileupload.R
@@ -94,7 +94,7 @@ FileUploadContext <- R6Class(
     },
     createUploadOperation = function(fileInfos) {
       while (TRUE) {
-        id <- paste(as.raw(p_runif(12, min=0, max=0xFF)), collapse='')
+        id <- createUniqueId(12)
         private$ids <- c(private$ids, id)
         dir <- file.path(private$basedir, id)
         if (!dir.create(dir))
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,13 @@
 shiny 0.14.0.9000
 =================
 
+## Minor new features and improvements
+
+* Restored file inputs are now copied on restore, so that the restored application can't modify the bookmarked file. ([#1370](https://github.com/rstudio/shiny/issues/1370))
+
+## Bug fixes
+
+* Fixed [#1368](https://github.com/rstudio/shiny/issues/1368): If an app with a file input was bookmarked and restored, and then the restored app was bookmarked and restored (without uploading a new file), then it would fail to restore the file the second time. ([#1370](https://github.com/rstudio/shiny/issues/1370))
 
 shiny 0.14
 ==========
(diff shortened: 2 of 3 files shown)


R/fileupload.R
```diff
@@ -94,7 +94,7 @@ FileUploadContext <- R6Class(
     },
     createUploadOperation = function(fileInfos) {
       while (TRUE) {
-        id <- paste(as.raw(p_runif(12, min=0, max=0xFF)), collapse='')
+        id <- createUniqueId(12)
         private$ids <- c(private$ids, id)
```
wch: This is a separate issue, but it seems to me that the `id` shouldn't be recorded here. Instead, it should be recorded after the `dir.create()` succeeds. Although realistically, there's no need to attempt the `dir.create` and then loop if it fails, because the probability of a name collision is so low. It seems more likely that the `dir.create` would fail due to a permission problem, and then this code would get stuck in an infinite loop.

PR #1374, opened 2016-09-17
Fix datepicker conflicts

### LINKED ISSUE #1218: Update datepicker library to a newer version
I can't figure out what version of datepicker shiny is using, but based on the fact that [the javascript file](https://github.com/rstudio/shiny/blob/master/inst/www/shared/datepicker/js/bootstrap-datepicker.js) hasn't changed since Dec 2014 I think it's at least as old as that.  There are some new functionality that have been implemented since then that can be useful, for example in Jan 2015 they added a way to [clear the date](https://github.com/eternicode/bootstrap-datepicker/commit/1d38515ef82dda5933e9bf7d394725a0d1fccecc) which I can't figure out how to do with the current datepicker.  

As a sidenote, it'd be nice to have a list somewhere that says what version each library within shiny is using

### LINKED ISSUE #1346: jquery-ui 1.11.4 datepicker has bad interaction with bootstrap datepicker
We have an app where we are trying to include both a dateRangeInput selector and a ggvisOutput for a graph in the same page.  This seems to not work at this time.  The underlying cause for this is that there seems to be a bad interaction between jquery-ui's datepicker and the bootstrap datepicker.  Below is a minimal example for how to reproduce the problem. As I've not really spent a lot of time in front-end dev I'm just not sure I'm equipped to fix this (yet).

WHAT IS EXPECTED:
Both the date range picker and the ggvis plot should be able to be displayed on the same page with data filled in.

HOW TO REPRODUCE:
server.R file:
```
shinyServer( function(input, output) 
{
  mtcars_reactive <- reactive({ mtcars })
  mtcars_reactive %>% ggvis(~wt, ~mpg) %>% layer_points() %>% bind_shiny("plot")
}
```

ui.R file:
```
shinyUI(fluidPage(
dateRangeInput('dateRange', label = 'Date range: ', start = '2016-08-15', end = '2016-08-25', max = '2016-08-26'),
ggvisOutput('plot')
)
)
```
The effect should be that no data will load in the data range picker and the plot will not show up.  If you remove either component from the ui.R file the remaining component will show up as expected.

Inspecting the DevTools in Chrome I can confirm both jquery-ui's datepicker and bootstrap datepicker are being loaded in this scenario.  Another note is that this app previously worked before 0.13.2

wch:
This closes #1218 and closes #1346. The upgraded datepicker has some significant API changes, which I think I've addressed, although `updateDateInput` and `updateDateRangeInput` could still use some testing.

This also updates jQuery UI to a full build 1.12.1, instead of the 1.11.4 build that was missing the datepicker component due to the conflict with bootstrap-datepicker.


jcheng5:
LGTM

PR #1387, opened 2016-09-21
options render for updateSelectizeInput did not worked in modules  

sipemu:
Custom rendering for updateSelectizeInput do not work in modules. This fix should solve this issue.


jcheng5:
Thanks!

PR #1402, opened 2016-09-30
Fix #1399: Duplicate binding error with insertUI and nested uiOutput

### LINKED ISSUE #1399: Module namespace issue?
I'm very confused in the following SO question I encountered, can someone help here?
My code structure is pretty close to the [example](https://gallery.shinyapps.io/insertUI-modules/) , however my code always trigger an error '_Uncaught Duplicate binding for ID_'

[SO Question](http://stackoverflow.com/questions/39737031/improve-the-robustness-of-shiny-module)

jcheng5:
This is a pretty low-level change. Basically, we find a list of uiOutput elements and bind them one by one. But element 0 is the parent of element 1. Binding element 0 causes element 1 to be removed and replaced with element 1a (which shares an `id` with element 1). But we continue down the list we captured earlier, so element 1 is the next thing we bind. Not only are we binding an element that's been removed from the document, but, we're also re-binding an `id` that has just been bound.


wch:
This *looks* OK to me but I'm reluctant to merge it for the upcoming release because there could be unforeseen consequences.

wch:
@jcheng5 Can you add a NEW entry? You'll probably want to rebase this branch to get your NEWS.md up to date first.

jcheng5:
@wch Should be all set now, assuming CI tests pass. Thanks.

PR #1416, opened 2016-10-11
Use yarn instead of npm

wch:
Facebook just [announced yarn](http://blog.npmjs.org/post/151660845210/hello-yarn), which is a replacement for npm. After playing around with it a little bit, I think it's a big improvement over using npm for managing packages, because it's much, much faster to install everything (I'd say around 10x for Shiny), and the result is deterministic, so we can be sure that everyone has the exact same set of packages installed.


jcheng5:
They changed their package name from `yarnpkg` to `yarn`, let's update that before merging.

Ordinarily I'd say that moving to a dependency management tool literally the day it's announced, is too aggressive. But in this case, it's hard to imagine how we'll get into worse trouble than with npm (we don't even use shrinkwrap today)... so I'm fine with merging this.

wch:
Wow, the docs were out of date the same day I wrote them. :)

The commentary on the HN thread was uniformly positive, so that helps. And I agree, it's not likely to be worse than npm.

PR #1421, opened 2016-10-14
Export input handlers

wch:
This makes it possible to apply input handlers to input values from outside of Shiny, which can be used for testing.


jcheng5:
lgtm

PR #1428, opened 2016-10-15
allow shiny.trace option to specify which messages to relay; fixes #1422

### LINKED ISSUE #1422: When using the `shiny.trace` option, allow specifying to only log SEND or RECV messages (or both)
It's great to have the ability to debug apps using `options(shiny.trace=TRUE)`.  But often I find the information to be overwhelming and I'm often only interested in seeing either the RECV or the SEND messages, so it'd be nice if we had a way to specify that.

I  can submit a PR if a shiny team member thinks this is acceptable.

Relevant code: I believe the only changes would happen [here](https://github.com/rstudio/shiny/blob/9613c58bf8120bcfdab35801b17167418b5464ac/R/server.R#L221-L226) and [here](https://github.com/rstudio/shiny/blob/9613c58bf8120bcfdab35801b17167418b5464ac/R/shiny.R#L417)

daattali:


wch:
Can you add an entry to NEWS.md?

daattali:
done

wch:
Can you also add a link to this PR at the end of the NEWS item? That's the format we've been using recently.

daattali:
updated (with slightly different format; I added the word "PR" to make it clear that it precedes a PR rather than an issue , since you can't distinguish between the two without looking at the URL)

bborgesr:
LGTM

bborgesr:
Oh, also, re @daattali's comment about having a tab (instead of a single space) after the `SEND`or `RECV` keywords, I think that would be sweet... (see: https://github.com/rstudio/shiny/issues/1422#issuecomment-253967164). I feel a little claustrophobic as it is now. @wch, any preference?

wch:
I don't have strong feelings either way for the spacing. I wouldn't want actual tabs, but I would be OK with a few more spaces.

wch:
I did a manual squash-merge of this PR in bc4dcee.

Issue #1437, opened 2016-10-25
HTML elements as labels for radio buttons


stla:
Hello, 

 It would be nice to have the possibility to put HTML in the labels of the radio buttons. 
Here is the function I wrote myself to do so (and it works fine):

```
radioButtonsHTML <- function(inputId, label, choices, choicesLabels=NULL, selected = NULL, inline = FALSE, width = NULL){
  if(is.null(choicesLabels)){
    return(radioButtons(inputId, label, choices, selected, inline, width))
  }else{
    if(length(choicesLabels) != length(choices)){
      stop("")
    }else{
      rb <- radioButtons(inputId, label, choices, selected, inline, width)
      for(i in seq_along(choices)){
        rb$children[[2]]$children[[1]][[i]]$children[[1]]$children[[2]]$children[[1]] <- HTML(choicesLabels[i])
      }
      return(rb)
    }
  }
}
```

bborgesr:
I'm having trouble understanding what this function does. Could you post a usage example?

stla:
This is the problem as [here](http://stackoverflow.com/questions/32169748/how-can-i-have-html-elements-as-labels-for-radio-buttons). 
Below is a usage example of my function:
```
        radioButtonsHTML("parameter", "Calculate:",
                     c("delta", "sigma", "p"), 
                     choicesLabels = c("&delta;", "&sigma;", "p"))
```
This gives:
![capture du 2016-10-25 14 28 00](https://cloud.githubusercontent.com/assets/4466543/19685980/69af19cc-9abf-11e6-98e7-a69f41c33a72.png)

daattali:
Similarly, I've seen many people ask for a way to have an image as the option (so this would also be solved by allowing any HTML)

bborgesr:
Closed via d62a2fc (#1521)

PR #1439, opened 2016-10-25
allow for Radio Buttons to have choice names using HTML

bborgesr:
This is a native (and pretty simple) solution to #1437. I feel bad that this was such an easy fix and it has been suggested/requested over an year ago :( ([here](http://stackoverflow.com/questions/32169748/how-can-i-have-html-elements-as-labels-for-radio-buttons)). Still, better late than never...

Sample code:
```r
library(shiny)
ui <- radioButtons("dist", "d",
                   c("Normal &mu;<sub>1</sub>" = "norm1",
                     "Normal &amp;mu;&lt;sub&gt;1&lt;&#47;sub&gt;" = "norm2"))
server <- function(input, output, session) {}
```

Before this PR:
screen shot 2016-10-25 at 2 04 03 pm

With with PR:
screen shot 2016-10-25 at 2 04 18 pm

This example also illustrates the only downside with this approach (at least the only one I could find): it may break previous apps that were purposefully using unescaped HTML (i.e. you can see that if you actually want the option name to be `Normal &mu;<sub>1</sub>"` previously you just typed that verbatim, but now you have to type `Normal &amp;mu;&lt;sub&gt;1&lt;&#47;sub&gt;`).

I feel like this would be a non-existent to negligible problem, but if you think otherwise, we could make this 100% back-compatible by having an additional arg to `radioButtons` that would be something like `choicesAsHTML = FALSE`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,8 @@ shiny 0.14.1.9001
 
 ### Minor new features and improvements
 
+* HTML markup can now be used in the choices' names for radio buttons. ([#1439](https://github.com/rstudio/shiny/pull/1439))
+
 * Added a `fade` argument to `modalDialog()` -- setting it to `FALSE` will remove the usual fade-in animation for that modal window. ([#1414](https://github.com/rstudio/shiny/pull/1414))
 
 * Exported function to apply input handlers to input values. This can be used for testing Shiny applications. ([#1421](https://github.com/rstudio/shiny/pull/1421))
--- a/R/input-utils.R
+++ b/R/input-utils.R
@@ -51,10 +51,10 @@ generateOptions <- function(inputId, choices, selected, inline, type = 'checkbox
       # If inline, there's no wrapper div, and the label needs a class like
       # checkbox-inline.
       if (inline) {
-        tags$label(class = paste0(type, "-inline"), inputTag, tags$span(name))
+        tags$label(class = paste0(type, "-inline"), inputTag, tags$span(HTML(name)))
       } else {
         tags$div(class = type,
-          tags$label(inputTag, tags$span(name))
+          tags$label(inputTag, tags$span(HTML(name)))
         )
       }
     },
--- a/tests/testthat/test-bootstrap.r
+++ b/tests/testthat/test-bootstrap.r
@@ -33,15 +33,15 @@ test_that("Repeated names for selectInput and radioButtons choices", {
   x <- radioButtons('id','label', choices = c(a='x1', a='x2', b='x3'))
   choices <- x$children
 
-  expect_equal(choices[[2]]$children[[1]][[1]]$children[[1]]$children[[2]]$children[[1]], 'a')
+  expect_equal(choices[[2]]$children[[1]][[1]]$children[[1]]$children[[2]]$children[[1]], HTML('a'))
   expect_equal(choices[[2]]$children[[1]][[1]]$children[[1]]$children[[1]]$attribs$value, 'x1')
   expect_equal(choices[[2]]$children[[1]][[1]]$children[[1]]$children[[1]]$attribs$checked, 'checked')
 
-  expect_equal(choices[[2]]$children[[1]][[2]]$children[[1]]$children[[2]]$children[[1]], 'a')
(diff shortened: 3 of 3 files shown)


shrektan:
I feel that using raw strings without escape is not consistent with other shiny functions. 

maybe add a new param like `escape = FALSE`?

wch:
I think we need to support the old behavior and have it be the default, because we don't want to break code that has `<`, `>`. or `&` in it. I'd lean toward a name like `choicesAsHTML=FALSE` instead of `escape`, because the meaning is much clearer.

The behavior of jQuery's `.html()` would be confusing for Shiny users who are far removed from that code. If you pass something like `"a < b"` to jQuery's `.html()` function, it will automatically convert it to "a &lt; b"; but then if you pass `"a <br> b"`, it won't escape the `<br>`. This inconsistency may be useful in some cases, but I think it would be confusing here.

It would be nice if we could use tag functions so that people could use `icon()` (and other tag functions), but that won't work because the text is the name part of the vector.

A couple other thoughts:

* Does `updateRadioButtons` also need a change?
* `checkboxGroupInput` should have the same change, right?

NEWS.md
```diff
@@ -5,6 +5,8 @@ shiny 0.14.1.9001
 
 ### Minor new features and improvements
 
+* HTML markup can now be used in the choices' names for radio buttons. ([#1439](https://github.com/rstudio/shiny/pull/1439))
```
wch: Could you also refer to the issues closed by this PR?

bborgesr:
Following a chat with @wch, we've arrived at the conclusion that this is not necessarily as easy as it seems. How things stand so far, while we have a very small and simple change that would already give our users a lot more flexibility, it's not perfect.

In particular, as Winston pointed out, "it would be nice if we could use tag functions so that people could use icon() (and other tag functions)". There is no easy or quick solution for this and we were wondering which set of tradeoffs are the least bad. Here's the three possibilities we outlined:

1. What we have in the PR has the advantage of simplicity, but doesn't allow for people to use familiar tag functions, and doesn't automatically import html dependencies (important for the font-awesome icon library -- pretty unfortunate, since this would probably be a common use case). It would also be the only time in Shiny so far we require users to write pure HTML (and don't even require -- since we can't -- that it be wrapped in the `HTML()` function). For back-compat, we'd also have to introduce a new argument (like `choicesAsHTML = FALSE`).

2. Winston proposed doing something like having  `choices = htmlChoices(x = icon())`, where the `htmlChoices()` function uses the reverse for name/value pairings. (In his opinion, that's the ordering that should have been used from the beginning). This has the advantages of not needing another argument, and could potentially import dependencies, but adds the possibility of confusion with the reversed name/value order. Also requires a lot more testing than option 1.

3. I proposed that, instead of having a boolean arg like `choicesAsHTML`, we add two new args `names` and `values` — that can be used as usual (so `choices = c(name1 = "a", name2 = "b")` would be equivalent to `names = c("name1", "name2"), values = c("a", "b")`) — but in which cases the `names` would always be interpreted as R code before being cast to characters. This would be one of those functions that if you specify `choices`, you can’t specify the other two and vice versa (`choices` would be the default for back-compat). In this case, the`names` and `values` args could also import dependencies, and there is no potential confusion arising from reversing the name/value pairings like option 2 implies. But it does add two extra arguments and moves the names and values away from each other (this would make it harder to see which name is paired with which value, especially for large lists). Like option 2, it also requires a lot more testing.


So... Any preferences, @jcheng5, @jjallaire and @yihui? 
(Also, this is very far from urgent!)

yihui:
If I were to design this function again, I'd definitely use two arguments, like `function(choices, names = base::names(choices))` (or maybe `choices` and `values`).

At this stage, I don't have a strong preference (slightly leaning towards two arguments).

shrektan:
@yihui Yes, I agree. With two arguments, we can then use `HTML(names)` directly if we intend to use the name without being escaped.

jcheng5:
Since this isn't the implementation we're going to use, I'm going to close the PR.

jcheng5:
@wch and I discussed it and we both think some variation of option 2 is the ideal solution. Whether we do two arguments (`choices, names`) or one, I feel less strongly about.

PR #1443, opened 2016-10-25
Fix #1365: dateInput minViewMode option

### LINKED ISSUE #1365: dateInput minViewMode option
I work with monthly data a lot and it would be great to have the dateInput function and dateRangeInput to take a minViewMode option.  I really like how startview = "year" looks and it would be awesome if I could just have my user pick the month they want from that and never have to pick a specific date.

brascoball:
Based on this posted answer (http://stackoverflow.com/questions/31152960/display-only-months-in-daterangeinput-or-dateinput-for-a-shiny-app-r-programmin), added the ability to set the bootstrap-datepicker minViewMode parameter for both dateInput and dateRangeInput, keeping "days" as the default. Second and third commit were to fix documentation/roxygen2 errors.

### DIFF
--- a/R/input-date.R
+++ b/R/input-date.R
@@ -70,7 +70,7 @@
 #'
 #'   # Start with decade view instead of default month view
 #'   dateInput("date6", "Date:",
-#'             startview = "decade")
+#'             startview = "decade"),
 #'                         
 #'   # Allow the user to select a month instead of a day
 #'   dateInput("date7", "Date:",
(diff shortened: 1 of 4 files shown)


brascoball:
@wch jj mentioned I should let you know I submitted this fix, and would love to see it make it into master. Thanks!

wch:
@brascoball We have some other changes that we want to do to date inputs, and we'll probably try to make all of the changes at the same time, in the next couple of months. When we make changes like this that alter the public API, it has the potential to break existing code (for example, if a user called the function with many unnamed arguments), and in cases like that we'd like to make all the changes in a single release.

brascoball:
@wch that makes sense. I'll eagerly await the update. thanks for the response!

PR #1446, opened 2016-10-28
Add list2env wrapper, for R <3.2.0

wch:
This fixes a problem where Shiny applications would crash in R < 3.2.0. See rstudio/shiny-server-pro#560.


wch:
After some hard work to get packages installed on R 3.1, @fereshtehRS has managed to test this, and it works.

PR #1452, opened 2016-11-01
Updated font-awesome to 4.7.0

benfasoli:
Nothing fancy, just updated font-awesome to enable use of icons released in 4.7.0


wch:
I've implemented this in 174a1fe. We generally prefer to do library updates ourselves because it's difficult to verify if binary or minified files are correct.

PR #1457, opened 2016-11-03
Trigger shiny:inputchanged only if input actually changes.

### LINKED ISSUE #1455: shiny:inputchanged event fires too often for textInputs
If you run this example app with the browser's JS console open, you'll see that the events fire more often than they should. According to the docs at http://shiny.rstudio.com/articles/js-events.html, the event should only fire if an input value changes.

```R
shinyApp(
  fluidPage(
    textInput("txt", "Text: Enter text, then move cursor, or click outside"),
    textInput("pass", "Password: Like text"),
    numericInput("num", "Number: delete number and re-type quickly", 0),
    dateInput("date", "Date: Set value (event happens twice), then set same value"),
    dateRangeInput("daterange", "Date Range: Same as date"),
    sliderInput("slider", "Slider: Slide up and then back to 0 quickly",
      min = 0, max = 10, value = 0),

    tags$script(HTML('
      $(document).on("shiny:inputchanged", function(event) {
        console.log(event.name + ":" + event.value);
      });'
    ))
  ),
  function(input, output) { }
)
```

All of the inputs above have this problem. Others don't, probably because they update immediately; these include `checkboxInput`, `checkboxGroupInput`, `radioButtons`, and `selectInput`.

wch:
This fixes #1455.


wch:
After talking with @jcheng about this, we decided not to change the current behavior. The `shiny:inputchanged` name is a bit of a misnomer (a more accurate name would be `shiny:inputpossiblychanged`), and the documentation needs to be updated.

It would also be useful to have another event that's triggered in `InputNoResendDecorator.setInput()`. The events would be passed post-transformation value, and would be deduplicated.

PR #1458, opened 2016-11-04
allow other packages to contain examples

0tertra:
adding an additional package parameter would allow other packages to contain examples with the ease of using runExample


daattali:
this would be a nice small addition, so that I don't have to implement my own function that does exactly the same in every package of mine :+1:

krlmlr:
@schloerke: This might have been closed accidentally. Any plans to incorporate the ideas here?

schloerke:
@krlmlr Correct. This was accidentally closed.

I've added it to my TODO list for the next release

PR #1464, opened 2016-11-15
Add support for injecting JS code when in test mode

wch:
This PR adds support for injecting Javascript code when the app is launched with the `shiny.testmode` option set to TRUE. This makes it possible to record test events interactively.


jcheng5:
Just a thought, how would you feel about making this more like the launch.browser param/app-option rather than like a global R option?

Also, can we emit a message saying that we're running in test mode when that's the case?

wch:
> Just a thought, how would you feel about making this more like the launch.browser param/app-option rather than like a global R option?

I kind of like that. A few questions/thoughts:

* Are you thinking that the param would get its default value from a global option, like `launch.browser` does?
* We would need to make sure this would work when an app is served with SS/Connect.
* Currently the `shiny.testmode` option is also used to enable the test endpoint URL (for getting input, output, and exported values). I recall discussing whether this could be dynamically set, or had to be set when the app started, and we thought it was a good idea to allow it to be dynamically set. Do you remember if that's right, and if so, why?

> Also, can we emit a message saying that we're running in test mode when that's the case?

Yes, that's a good idea.

PR #1475, opened 2016-11-21
Improve documentation for submitButton and change 07_widgets example to use an action button

### LINKED ISSUE #303: submitButton should have a way to specify which inputs it puts on hold
Presently, a submit button "suspends" all the inputs on a page until the button is clicked. It would be really useful to have more granular control over which inputs are paused.

Here is one possibility that we discussed:

```
# Named submit group
submitGroup(submitId = "submit1"
  textInput("foo", "bar")
)
submitButton("submit1", "Click me")


# Automatic ID if submitButton is included
submitGroup(
  textInput("foo", "bar")
  submitButton(label = "Click me")
)
```

### LINKED ISSUE #1355: 07_widgets example shouldn't use submitButton

### LINKED ISSUE #1426: Cannot have more than one submit button in an app working correctly
I've never really used submit buttons but I was trying to use them tonight and noticed that if you have more than one submit button in an app, they ignore their enclosing `<form>` and they act as if the entire app is one form. Submit buttons should be isolated within each form, but it looks like activating any submit in an app will cause all inputs in the app to fire.

I also noticed that even if I create a submit button with pure HTML I still run into the issue, so the problem is probably something weird that shiny is doing.

Example:

```
library(shiny)

ui <- fluidPage(
  tags$form(
    textInput("text1", "text1"),
    tags$button("submit1", type="submit", class="btn btn-default")
  ),
  tags$form(
    textInput("text2", "text2"),
    tags$button("submit2", type="submit", class="btn btn-default")
  )
)

  observe({
    input$text1
    cat("submit 1")
  })
  observe({
    input$text2
    cat("submit 2")
  })
}

```

A related question is: can we deprecate submit buttons since they're so awkward (they don't even take an id!), that would be the best :)

bborgesr:
Closes #303 (won't fix -- "submitButton should have a way to specify which inputs it puts on hold"), closes #1355 ("07_widgets example shouldn't use submitButton"), closes #1426 ("Cannot have more than one submit button in an app working correctly").

In addition to the changes here, there will also be related PR to shiny-examples (to update the 07_widgets example to reflect the changes here) and to shiny-dev-center (to update the action button article along these same lines -- mostly highlight the very limited cases when you can use a submit button).

### DIFF
--- a/R/input-submit.R
+++ b/R/input-submit.R
@@ -1,56 +1,27 @@
 #' Create a submit button
 #'
-#' Create a submit button for an input form. Forms that include a submit
+#' Create a submit button for an app. Apps that include a submit
 #' button do not automatically update their outputs when inputs change,
 #' rather they wait until the user explicitly clicks the submit button.
+#' The use of \code{submitButton} is generally discouraged in favor of
+#' the more versatile \code{\link{actionButton}} (see details below).
 #'
-#' @section Warning:
-#' Submit buttons are very particular shiny objects. One the weirdest
-#' things about them is they have no IDs, and there is no way to regulate
-#' which inputs they should (and which ones they should not) be applied
-#' to. This means they are very unwieldy for anything but the simplest
-#' apps (see the example below). For example, having \emph{two} submit
-#' buttons in the same app will probably not do what you'd expect because
-#' submit buttons are not isolated in any way, so activating any of the
-#' two buttons will cause all inputs in the app to fire. Another problem
-#' with submit buttons is that dynamically created submit buttons (for
-#' example, with \code{\link{renderUI}} or \code{\link{insertUI}}) will
-#' not work.
+#' Submit buttons are unusual Shiny inputs, and we recommend using
+#' \code{\link{actionButton}} instead of \code{submitButton} when you
+#' want to delay a reaction.
+#' See \href{http://shiny.rstudio.com/articles/action-buttons.html}{this
+#' article} for more information (including a demo of how to "translate"
+#' code using a \code{submitButton} to code using an \code{actionButton}).
 #'
-#' For all of these reasons, \strong{our general reccomendation is to
-#' avoid submit buttons} and
-#' \href{http://shiny.rstudio.com/articles/action-buttons.html#pattern-2---delay-reactions}{
-#' use action buttons instead when you want to delay a reaction}. Any
-#' code that uses a submit button can be converted to code that uses an
-#' action button instead (while the reverse is almost never true). So,
-#' if you find your app getting increasingly complicated, opt for using
-#' an action button (or more since there is no problem with multiple
-#' action buttons in the same app). The link above shows an example of
-#' a delayed reaction using an action button. For another example, the
-#' code below converts the example in this page (at the bottom) to an
-#' equivalent app using an action button (using the same pattern
-#' described in the link above):
-#'
-#' \preformatted{
-#' shinyApp(
-#'   ui = basicPage(
-#'     numericInput("num", label = "Make changes", value = 1),
-#'     actionButton("update" ,"Update View", icon("refresh"),
-#'                  class = "btn btn-primary"),
-#'     helpText("When you click the button above, you should see",
-#'              "the output below update to reflect the value you",
-#'              "entered at the top:"),
-#'     verbatimTextOutput("value")
-#'   ),
-#'   server = function(input, output) {
-#'     currentValue <- eventReactive(input$update, {
-#'       input$num
-#'     }, ignoreNULL = FALSE)
-#'
-#'     output$value <- renderPrint({ currentValue() })
-#'   }
-#' )
-#' }
+#' In essence, the presence of a submit button stops all inputs from
+#' sending their values automatically to the server. This means, for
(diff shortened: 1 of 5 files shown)


bborgesr:
shiny-examples PR: rstudio/shiny-examples#54

bborgesr:
shiny-dev-center PR: rstudio/shiny-dev-center#34

R/input-submit.R
```diff
@@ -4,6 +4,54 @@
...
+#' if you find your app getting increasingly complicated, opt for using
+#' an action button (or more since there is no problem with multiple
+#' action buttons in the same app). The link above shows an example of
+#' a delayed reaction using an action button. For another example, the
+#' code below converts the example in this page (at the bottom) to an
+#' equivalent app using an action button (using the same pattern
+#' described in the link above):
+#'
+#' \preformatted{
+#' shinyApp(
```
wch: I think it's best to leave out this example, because I don't think people will understand how it's a translation of the original `submitButton` version -- it's notably more complex, and uses a function (`eventReactive`) they're likely not familiar with. I think the people will be better served by reading the article. For the same reason, I also changed the article link (above) to point to the top level of the actionButton article instead of jumping to the `eventReactive` example.

R/input-submit.R
```diff
@@ -4,6 +4,54 @@
 #' button do not automatically update their outputs when inputs change,
 #' rather they wait until the user explicitly clicks the submit button.
 #'
+#' @section Warning:
+#' Submit buttons are very particular shiny objects. One the weirdest
```
wch: Here's a trimmed-down version that tells people right off the bat to use actionButtons instead:

> Submit buttons are unusual Shiny inputs, and we recommend using \code{\link{actionButton}} instead of \code{submitButton} when you want to delay a reaction. See \href{http://shiny.rstudio.com/articles/action-buttons.html}{this article} for more information.
>
> The presence of a submit button stops all inputs from sending their values automatically to the server, making them unwieldy for all but the simplest apps. If there are \emph{two} submit buttons in the same app, clicking either one will cause all inputs in the app to send their values to the server. Another issue with submit buttons is that dynamically created submit buttons (for example, with \code{\link{renderUI}} or \code{\link{insertUI}}) will not work.

R/input-submit.R
bborgesr: Yeah, I was also not feeling great about doing the "translation" right in the documentation. But since we're also moving the aside about submit buttons to the end of the Action Button article, do you think we could include both the example and the translation there (i.e. right at the end of the article)?

wch: Yeah, I think that makes sense. 

For simplicity, instead of using `eventReactive()`, you could just use `isolate()` in the `renderPrint()` call, like in the example below. However, looking at the [article](http://shiny.rstudio.com/articles/action-buttons.html), I don't see this as one of the 5 patterns there. It is shown in the [isolation article](http://shiny.rstudio.com/articles/isolation.html) though.

@jcheng5, does it seem right to you that the action button article barely talks about using `isolate()`? That's the way I almost always use action buttons.

```R
shinyApp(
  ui = basicPage(
    numericInput("num", label = "Make changes", value = 1),
    actionButton("update" ,"Update View", icon("refresh"),
      class = "btn btn-primary"),
    helpText("When you click the button above, you should see",
      "the output below update to reflect the value you",
      "entered at the top:"),
    verbatimTextOutput("value")
  ),
    output$value <- renderPrint({ 
      input$update
      isolate(input$num)
    })
  }
)
```

bborgesr: I agree that your version of the example looks simpler. I was using `eventReactive()` exactly because it matches up with Pattern 2 in the Action Button article. And I agree, Winston: this article should talk about `isolate()` more explicitly. Maybe we could do something like Pattern 2a and Pattern 2b: both delay reactions, but one uses `eventReactive()` and the other uses `isolate()`... Seems clunky, but for that pattern, I really do think these are equally valid (and the best one depends on the complexity of your app) -- for small things like this, `isolate()` is neater than introducing a special reactive for compute the value...

bborgesr:
@wch: I followed most of your suggestions. It especially makes sense to tell people right off the bat to use actionButtons. It was my concern with that admonition that made me call the section "Warning" (so people wouldn't skip it!). Also, again, appreciate the conciseness :) I still modified the Details section text a bit from your version to add the info about the "code translation" part of the article (up soon!) and make the second paragraph flow better.

I also changed the initial description a bit, to make it clear from the start that we generally recommend action buttons (otherwise, I felt like we were misleading the reader -- I feel like this is too important a piece of info to only introduce it in the Details section). I also changed the wording from "creating a submit button for a form" to "creating a submit button for an app". I feel like we don't actually provide any context for what a form is and most apps that use submit buttons probably don't make their use of the form tag explicit. So it felt like an unnecessary and potentially confusing use of HTML jargon. Let me know if you don't agree with something in here...

Finally, do these changes qualify for a NEWS entry? Since we're just essentially updating examples and documentation to follow a pattern we already advocate (and are not really changing anything technical or in the API), I thought it wasn't worth it... But maybe that's the wrong way to think about it, and every PR that changes content (besides fixing typos, etc) should have a NEWS entry..?

PR #1478, opened 2016-11-22
Fixes #1321: mention support for the optgroup tag in the documention for selectInput

### LINKED ISSUE #1321: Add documentation to say that selectInput() has support for optgroup
After seeing https://github.com/rstudio/shiny/pull/520 I finally learned that the dropdown input does indeed support groups. I doubt many people know about this, it would be great to advertise a bit more

bborgesr:
Fixes #1321: Mention support for the optgroup tag in the documentation for selectInput (specifically in the `choices` arg). Added example app too.

### DIFF
--- a/R/input-select.R
+++ b/R/input-select.R
@@ -15,7 +15,12 @@
 #'
 #' @inheritParams textInput
 #' @param choices List of values to select from. If elements of the list are
-#'   named then that name rather than the value is displayed to the user.
+#'   named, then that name rather than the value is displayed to the user.
+#'   This can also be a named list whose elements are (either named or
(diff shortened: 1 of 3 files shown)


bborgesr:
I feel like it's overkill to also mention this in the [selectize article](http://shiny.rstudio.com/articles/selectize.html)... That's a concise piece devoted mostly to the difference between select vs selectize, and within that, the difference between client-side selectize and server-side selectize. So I don't feel like there's any good place to make a note of the optgroup support... And I also feel this really belongs in the documentation. Thoughts?

wch:
I think this is the right place and includes the right amount of info.

PR #1480, opened 2016-11-22
Closes #1357: verbatimTextOutput should optionally be hidden if no content

### LINKED ISSUE #1357: verbatimTextOutput should be hidden if no content
Right now, verbatimTextOutput renders as an empty, squat rounded rectangle if it contains no content (even `req(FALSE)`). Better if it doesn't appear at all if it doesn't contain content. (Though we'd have to make sure it doesn't get accidentally marked as hidden from the shiny perspective, and stop rendering)

bborgesr:
Can be done with only:

* a new parameter to `verbatimTextOutput` (called `placeholder`, and defaulting to TRUE -- which means no old apps will look different from before);

* a new parameter to `textOutput` (`...` so that we can send through the new class we add in `verbatimTextOutput`);

* a little CSS.

No JS needed!

Example app (one with the default, another with `placeholder = FALSE`):

```r
library(shiny)

ui <- basicPage(
  textInput("txt", "Enter the text to display below:"),
  verbatimTextOutput("default"),
  verbatimTextOutput("new", placeholder = FALSE),
  p("Dummy text to show there is no extra spacing when", 
    "the second output is empty")
)

  output$default <- renderText({ input$txt })
  output$new <- renderText({ input$txt })
}

shinyApp(ui, server)

```

### DIFF
--- a/R/bootstrap.R
+++ b/R/bootstrap.R
@@ -935,21 +935,34 @@ textOutput <- function(outputId, container = if (inline) span else div, inline =
 #' Render a reactive output variable as verbatim text within an
 #' application page. The text will be included within an HTML \code{pre} tag.
 #' @param outputId output variable to read the value from
+#' @param placeholder if the output is empty or \code{NULL}, should an empty
+#'   rectangle be displayed to serve as a placeholder? (does not affect
+#'   behavior when the the output in nonempty)
 #' @return A verbatim text output element that can be included in a panel
 #' @details Text is HTML-escaped prior to rendering. This element is often used
-#' with the \link{renderPrint} function to preserve fixed-width formatting
-#' of printed objects.
+#'   with the \link{renderPrint} function to preserve fixed-width formatting
+#'   of printed objects.
 #' @examples
-#' mainPanel(
-#'   h4("Summary"),
-#'   verbatimTextOutput("summary"),
-#'
-#'   h4("Observations"),
-#'   tableOutput("view")
-#' )
+#' ## Only run this example in interactive R sessions
+#' if (interactive()) {
+#'   shinyApp(
+#'     ui = basicPage(
+#'       textInput("txt", "Enter the text to display below:"),
+#'       verbatimTextOutput("default"),
+#'       verbatimTextOutput("placeholder", placeholder = TRUE)
+#'     ),
+#'     server = function(input, output) {
+#'       output$default <- renderText({ input$txt })
+#'       output$placeholder <- renderText({ input$txt })
+#'     }
+#'   )
+#' }
 #' @export
-verbatimTextOutput <- function(outputId) {
-  textOutput(outputId, container = pre)
+verbatimTextOutput <- function(outputId, placeholder = FALSE) {
+  pre(id = outputId,
+      class = paste(c("shiny-text-output", if (!placeholder) "noplaceholder"),
+                    collapse = " ")
+      )
 }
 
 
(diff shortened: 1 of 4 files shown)


wch:
Hm, it looks like there were changes to the generated JS code. I think this means you probably have a different version of Babel (or something like that) installed. Have you reinstalled the JS toolchain since we made the switch to yarn?

R/bootstrap.R
```diff
@@ -948,8 +953,8 @@ textOutput <- function(outputId, container = if (inline) span else div, inline =
 #'   tableOutput("view")
 #' )
 #' @export
-verbatimTextOutput <- function(outputId) {
-  textOutput(outputId, container = pre)
+verbatimTextOutput <- function(outputId, placeholder = TRUE) {
+  textOutput(outputId, container = pre, class = if (!placeholder) "noplaceholder")
```
wch: I think it would make more sense to directly call `pre()` here instead of calling `textOutput()`.  This is so that you won't have to modify `textOuput` to accept `...`.

wch:
It's nice that this can be done with just CSS!

R/bootstrap.R
bborgesr: Sure, that makes sense. Updated it.

bborgesr:
@wch, you were right! I hadn't updated after we switched to yarn. Everything matches up now. However, I've polluted the commit history a bit, so we should squash-merge. I've also added a NEWS entry for this feature.

On an unrelated note, the Travis check is failing in a bizarre way (and that is not reproducible locally):

```
  testthat results ================================================================
  OK: 526 SKIPPED: 0 FAILED: 1
  1. Failure: unreferenced observers are garbage collected (@test-gc.r#32) 
```

Is this Travis being weird or did I mess something up?

R/bootstrap.R
```diff
@@ -948,8 +951,9 @@ textOutput <- function(outputId, container = if (inline) span else div, inline =
 #'   tableOutput("view")
 #' )
 #' @export
-verbatimTextOutput <- function(outputId) {
-  textOutput(outputId, container = pre)
+verbatimTextOutput <- function(outputId, placeholder = TRUE) {
+  pre(id = outputId,
+      class = paste("shiny-text-output", if (!placeholder) "noplaceholder"))
```
wch: A nit: if `!placeholder`, this will result in `"shiny-text-output "`, with a trailing space. One way to work around this is to use `paste()` on a vector and use the `collapse` argument, as in:

```R
paste(c("shiny-text-output", if (!placeholder) "noplaceholder"), collapse = " ")
```

wch:
The `placeholder` arg seems like a decent name to me. @jcheng5, what do you think?

bborgesr:
Used `collapse = " "` per Winston's suggestion, rebased and ready to merge!

bborgesr:
An better and updated example:

```r
library(shiny)

ui <- basicPage(
  textInput("txt", "Enter the text to display below:"),
  p("OLD:"),
  verbatimTextOutput("old", placeholder = TRUE, 
                     style = "background-color: #cdb38b;"),
  p("NEW (and default from now on)"),
  verbatimTextOutput("default", 
                     style = "background-color: #fa8072;"),
  p("Dummy text to show there is no extra spacing when", 
    "the second output is empty")
)

  output$old <- renderText({ input$txt })
  output$default <- renderText({ input$txt })
}

shinyApp(ui, server)
```

PR #1481, opened 2016-11-23
Fixes #969: allow navbarPage's `fluid` param to control both containers

### LINKED ISSUE #969: "container"/"container-fluid" setting is not controllable for the navigation tab part (header) on navbarPage
In the current setting of `navbarPage`, the bootstrap container for the nav-tab part seems to be hard coded as "container" while the default container for the tab-content part is set to be "container-fluid". I mean, it looks okay in regular cases but it looks super weird when you zoom out or when the default resolution of the screen is very high, as the contents are splashed all over the screen while the header part is squeezed and center in the middle. 

You can change the container setting of the tab-content part to "container" by the "fluid" option. However, for the header, it seems like there is no easy way to change this setting (correct me if I'm wrong). 

Maybe we should just let `fluid` control both?

``` html
<nav class="navbar navbar-default navbar-static-top" role="navigation">
  <div class="container"> 
    <div class="navbar-header">
      <span class="navbar-brand">App Title</span>
    </div>
    <ul class="nav navbar-nav">
      <li class="active">
        <a href="#tab-3367-1" data-toggle="tab" data-value="Plot">Plot</a>
      </li>
      <li>
        <a href="#tab-3367-2" data-toggle="tab" data-value="Summary">Summary</a>
      </li>
      <li>
        <a href="#tab-3367-3" data-toggle="tab" data-value="Table">Table</a>
      </li>
    </ul>
  </div>
</nav>
<div class="container-fluid">
  <div class="tab-content">
    <div class="tab-pane active" data-value="Plot" id="tab-3367-1"></div>
    <div class="tab-pane" data-value="Summary" id="tab-3367-2"></div>
    <div class="tab-pane" data-value="Table" id="tab-3367-3"></div>
  </div>
</div> 
```

bborgesr:
Fixes #969: a super simple change (too bad it took us this long to get to it -- over a year!) I also don't think this warrants a NEWS entry, right @wch?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,6 +9,8 @@ shiny 0.14.2.9000
 
 ### Bug fixes
 
+* Fix [#969](ttps://github.com/rstudio/shiny/issue/969): allow navbarPage's `fluid` param to control both containers. ([#1481](https://github.com/rstudio/shiny/pull/1481))
+
 * Fix [#1438](https://github.com/rstudio/shiny/issue/1438): `unbindAll()` should not be called when inserting content with `insertUI()` ([#1449](https://github.com/rstudio/shiny/pull/1449))
 
 * Fixed bug causing `<meta>` tags associated with HTML dependencies of Shiny R Markdown files to be rendered incorrectly. ([#1463](https://github.com/rstudio/shiny/pull/1463))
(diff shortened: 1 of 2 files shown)


wch:
I think this does deserve a NEWS entry. Otherwise, LGTM.

PR #1482, opened 2016-11-23
Fixes #1358: more informative error message when calling runApp inside of an app's app.R

### LINKED ISSUE #1358: Give warning when runApp is called within runApp
EDIT: sorry, figured out the real problem, the 'Run App' button calls runApp with as argument the directory, however if you have an `app.R` with some custom `shinyApp` calls it fails. So in the example below if you simple select all and execute, it works fine. But you should not use the 'Run App' button. 

In the documentation for `shinyApp` one of the examples  shows how to make a shinyApp from an 'an explicit UI/server pair', however this example doesn't work. The example in question is 

```
app <- shinyApp(
  ui = fluidPage(
    numericInput("n", "n", 1),
    plotOutput("plot")
  ),
    output$plot <- renderPlot( plot(head(cars, input$n)) )
  }
)

runApp(app)
```

which fails with 

```
Listening on http://127.0.0.1:3280
Warning: Error in handlers$add: Key / already in use
Stack trace (innermost first):
    43: handlers$add
    42: handlerManager$addHandler
    41: startApp
    40: runApp
     1: shiny::runApp
Error in handlers$add(handler, key, tail) : Key / already in use
```

bborgesr:
Fixes #1358: more informative error message when calling runApp inside of an app's app.R (or inside ui.R or server.R).

This used to be: `stop("Key ", key, " already in use")`

Now it is: `stop("Please remove any calls to runApp from inside your source code. (Details: Key ", key, " already in use)")`.

The only way I see this may be problematic is if this error is raised in another context (that isn't calling `runApp` from the app's source code). But I don't think that ever happens (?)

Again, I don't think this merits a NEWS entry... Let me know if you disagree

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,6 +9,8 @@ shiny 0.14.2.9000
 
 ### Minor new features and improvements
 
+* Addressed [#1358](https://github.com/rstudio/shiny/issues/1358): more informative error message when calling `runApp()` inside of an app's app.R (or inside ui.R or server.R). ([#1482](https://github.com/rstudio/shiny/pull/1482))
+
 * Added a more descriptive JS warning for `insertUI()` when the selector argument does not match anything in DOM. ([#1488](https://github.com/rstudio/shiny/pull/1488))
 
 * Added support for injecting JavaScript code when the `shiny.testmode` option is set to `TRUE`. This makes it possible to record test events interactively. ([#1464]https://github.com/rstudio/shiny/pull/1464))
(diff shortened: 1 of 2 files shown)


wch:
A couple things:

* This message would work better: "Can't call `runApp` from within `runApp`. If your application code contains `runApp()`, please remove it."
* `runApp` could check this at the top:

```
  if (.globals$running) {
    stop("HEY!")
  }
  .globals$running <- TRUE
  on.exit({
    .globals$running <- FALSE
  },add = TRUE)
```

* And `.globals$running <- FALSE` would have to be already set.

PR #1483, opened 2016-11-23
Fix #1359: shinyApp options argument ignored when passed to runApp

### LINKED ISSUE #1359: shinyApp options argument ignored when passed to runApp
In the below example the app is not opened in the external browser, maybe this can be made clear in the documentation for shinyApp, or (preferable IMO, but more work) let runApp read the options from the app if an app object is passed.

```
app <- shinyApp(
  ui = fluidPage(
    numericInput("n", "n", 1),
    plotOutput("plot")
  ),
    output$plot <- renderPlot( plot(head(cars, input$n)) )
  },
  options = list(display.mode="showcase", launch.browser=T)
)
runApp(app)
```

bborgesr:
Fix #1359: shinyApp options argument ignored when passed to runApp.

The `options` arg to `shinyApp()` now works as expected, accepting the following shiny options:
* `port`
* `launch.browser`
* `host`

There are other params that `runApp()` accepts that `shinyApp()`'s `options` argument does not yet accept (these are `workerId`, `quiet`,`display.mode`). It may be worth having a discussion about whether any (or all, for consistency) of these should be promoted to a shiny option...

Here's the matrix of possibilities and the correspoding demo app:

1. no options are explicitly specified in `shinyApp()` or in `runApp()` -- defaults kick in (same behavior as before this PR):

```r
library(shiny)
app <- shinyApp(
  ui = fluidPage(
    numericInput("n", "n", 1),
    plotOutput("plot")
  ),
    output$plot <- renderPlot( plot(head(cars, input$n)) )
  },
)
runApp(app)
```

2. options are only explicitly specified in `shinyApp()` -- these should be followed (this did not work before this PR):

```r
library(shiny)
app <- shinyApp(
  ui = fluidPage(
    numericInput("n", "n", 1),
    plotOutput("plot")
  ),
    output$plot <- renderPlot( plot(head(cars, input$n)) )
  },
  options = list(launch.browser=rstudioapi::viewer, port=4000)
)
runApp(app)
```

3. options are only explicitly specified in `runApp()` -- these should be followed (same behavior as before this PR):

```r
library(shiny)
app <- shinyApp(
  ui = fluidPage(
    numericInput("n", "n", 1),
    plotOutput("plot")
  ),
    output$plot <- renderPlot( plot(head(cars, input$n)) )
  },
)
runApp(app, launch.browser=T, port=3000)
```

4. no options are explicitly specified both in `shinyApp()` and in `runApp()` -- options from `runApp` win (same behavior as before this PR):

```r
library(shiny)
app <- shinyApp(
  ui = fluidPage(
    numericInput("n", "n", 1),
    plotOutput("plot")
  ),
    output$plot <- renderPlot( plot(head(cars, input$n)) )
  },
  options = list(launch.browser=rstudioapi::viewer, port=4000)
)
runApp(app, launch.browser=T, port=3000)
```

### DIFF
--- a/R/server.R
+++ b/R/server.R
@@ -588,30 +588,22 @@ runApp <- function(appDir=getwd(),
   # I tried to make this as compact and intuitive as possible,
   # given that there are four distinct possibilities to check
   runOpts <- appParts$options
-  passedArgs <- as.list(match.call())[-1]
-  thisEnv <- environment()
-
-  # Figure out which value to use for an option
-  priorityAssign <- function(arg) {
-    # if it was passed directly to `runApp`, use that value
-    if (arg %in% names(passedArgs)) val <- passedArgs[[arg]]
-
-    # else if it was passed directly to `shinyApp`, use that value
-    else if (arg %in% names(runOpts)) val <- runOpts[[arg]]
-
-    # else, just use the default value specified in `runApp`
-    else val <- thisEnv[[arg]]
-
-    # make sure that Ts and Fs are kept as logicals
-    if (is.name(val) && val == "T") val <- TRUE
-    if (is.name(val) && val == "F") val <- FALSE
-
-    assign(arg, val, envir = thisEnv)
+  assignVal <- function(arg, default) {
+    if (arg %in% names(runOpts)) runOpts[[arg]] else default
   }
 
-  lapply(list("port", "launch.browser", "host",
-              "quiet", "display.mode", "test.mode"),
-         priorityAssign)
+  if (missing(port))
+    port <- assignVal("port", port)
+  if (missing(launch.browser))
+    launch.browser <- assignVal("launch.browser", launch.browser)
+  if (missing(host))
+    host <- assignVal("host", host)
+  if (missing(quiet))
+    quiet <- assignVal("quiet", quiet)
+  if (missing(display.mode))
+    display.mode <- assignVal("display.mode", display.mode)
+  if (missing(test.mode))
+    test.mode <- assignVal("test.mode", test.mode)
 
   if (is.null(host) || is.na(host)) host <- '0.0.0.0'
 
--- a/R/server.R
+++ b/R/server.R
@@ -561,13 +561,59 @@ runApp <- function(appDir=getwd(),
     .globals$options <- oldOptionSet
   },add = TRUE)
 
-  if (is.null(host) || is.na(host))
-    host <- '0.0.0.0'
-
   # Make warnings print immediately
   # Set pool.scheduler to support pool package
   ops <- options(warn = 1, pool.scheduler = scheduleTask)
   on.exit(options(ops), add = TRUE)
+
+  appParts <- as.shiny.appobj(appDir)
+
+  # The lines below set some of the app's running options, which
+  # can be:
+  #   - left unspeficied (in which case the arguments' default
+  #     values from `runApp` kick in);
+  #   - passed through `shinyApp`
+  #   - passed through `runApp` (this function)
+  #   - passed through both `shinyApp` and `runApp` (the latter
+  #     takes precedence)
+  #
+  # Matrix of possibilities:
+  # | IN shinyApp | IN runApp | result       | check                                                                                                                                  |
+  # |-------------|-----------|--------------|----------------------------------------------------------------------------------------------------------------------------------------|
+  # | no          | no        | use defaults | exhaust all possibilities: if it's missing (runApp does not specify); THEN if it's not in shinyApp appParts$options; THEN use defaults |
+  # | yes         | no        | use shinyApp | if it's missing (runApp does not specify); THEN if it's in shinyApp appParts$options; THEN use shinyApp                                |
+  # | no          | yes       | use runApp   | if it's not missing (runApp specifies), use those                                                                                      |
+  # | yes         | yes       | use runApp   | if it's not missing (runApp specifies), use those                                                                                      |
+  #
+  # I tried to make this as compact and intuitive as possible,
+  # given that there are four distinct possibilities to check
+  runOpts <- appParts$options
+  passedArgs <- as.list(match.call())[-1]
+  thisEnv <- environment()
+
+  # Figure out which value to use for an option
+  priorityAssign <- function(arg) {
+    # if it was passed directly to `runApp`, use that value
+    if (arg %in% names(passedArgs)) val <- passedArgs[[arg]]
+
+    # else if it was passed directly to `shinyApp`, use that value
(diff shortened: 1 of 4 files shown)


bborgesr:
Another note: as for `workerId`, `quiet` and `display.mode`: we don't *need* to make these shiny options in order for `shinyApp()`'s `options` argument to accept them and respect them. But it makes more sense to me to make them options and then do this than just doing this. This is mostly because the argument in question here is called `options` -- which is obviously reminiscent of shiny-options (at least, that is my impression). 

I think this is a bit unfortunate because we don't want to allow *all* shiny-options to be able to be passed through `shinyApp()`'s `options` argument (right?? For this PR I worked under the assumption that we only wanted to allow `port`, `launch.browser` and `host` to be passed through here).

If the goal instead is that `shinyApp()`'s `options` argument  accepts all argument that `runApp()` also accepts, then it should be called something different -- like `runningOptions`, or even just the familiar `...` with the description "args to be passed on to `runApp`" (even if that is not exactly what we do under the hood).

bborgesr:
So, in summary:

* The `options` arg to `shinyApp()` now works as expected, accepting the following `runApp()` params:
    * port
    * launch.browser
    * host
    * quiet
    * display.mode

* `shinyApp`'s documentation improved (for the `options` arg -- the 5 params above are now explicitly mentioned)

jcheng5:
LGTM.

R/server.R
```diff
@@ -551,6 +551,28 @@ runApp <- function(appDir=getwd(),
     handlerManager$clear()
   }, add = TRUE)
 
+  appParts <- as.shiny.appobj(appDir)
+
+  if (missing(port)) {
```
wch: Are these `missing()` checks necessary? In general, it's best to avoid using `missing()`, because it can make it hard to call the function programmatically.

R/server.R
```diff
@@ -551,6 +551,28 @@ runApp <- function(appDir=getwd(),
     handlerManager$clear()
   }, add = TRUE)
 
+  appParts <- as.shiny.appobj(appDir)
```
wch: I'm not 100% sure, but I don't think it's safe to move this line up here. This probably merits a discussion with @jcheng5.

R/server.R
```diff
@@ -561,14 +561,60 @@ runApp <- function(appDir=getwd(),
...
+    if (arg %in% names(passedArgs)) val <- passedArgs[[arg]]
+
+    # else if it was passed directly to `shinyApp`, use that value
+    else if (arg %in% names(runOpts)) val <- runOpts[[arg]]
+
+    # else, just use the default value specified in `runApp`
+    else val <- thisEnv[[arg]]
+
+    # make sure that Ts and Fs are kept as logicals
+    if (is.name(val) && val == "T") val <- TRUE
```
wch: Do the strings `"T"` and `"F"` ever get passed in?

R/server.R
```diff
@@ -561,14 +561,60 @@ runApp <- function(appDir=getwd(),
...
+  # |-------------|-----------|--------------|----------------------------------------------------------------------------------------------------------------------------------------|
+  # | no          | no        | use defaults | exhaust all possibilities: if it's missing (runApp does not specify); THEN if it's not in shinyApp appParts$options; THEN use defaults |
+  # | yes         | no        | use shinyApp | if it's missing (runApp does not specify); THEN if it's in shinyApp appParts$options; THEN use shinyApp                                |
+  # | no          | yes       | use runApp   | if it's not missing (runApp specifies), use those                                                                                      |
+  # | yes         | yes       | use runApp   | if it's not missing (runApp specifies), use those                                                                                      |
+  #
+  # I tried to make this as compact and intuitive as possible,
+  # given that there are four distinct possibilities to check
+  runOpts <- appParts$options
+  passedArgs <- as.list(match.call())[-1]
```
wch: I think the previous way, with the `missing` calls, had clearer intent. The `match.call()` and `environment()` assignment makes things kind of confusing. You could do something like this, where `f` is like `runApp`; `x` is like the app object; and, `arg1` is an argument like `port`:

```R
f <- function(x, arg1 = 1) {
  objValueOrDefault <- function(obj, name, default) {
    if (name %in% names(obj))
      obj[[name]]
    else
      default
  }
  
  if (missing(arg1)) arg1 <- objValueOrDefault(x, "arg1",  arg1)
  
  arg1
}
```

The drawback is that `arg1` appears 4 times on one line, and that the `if(missing())` part needs to be done for each argument. But it's easier to understand what it's doing.

It passes the following:

```R
library(testthat)
expect_identical(f(list()),              1)
expect_identical(f(list(), 2),           2)
expect_identical(f(list(arg1 = 3)),       3)
expect_identical(f(list(arg1 = 3), 2),    2)
expect_identical(f(list(arg1 = NULL)),    NULL)
expect_identical(f(list(arg1 = NULL), 2), 2)
```

The one tricky case is in the fifth test: the object has `arg1=NULL`, and that NULL gets used. I think that behavior makes the most sense.

If we didn't do things that way (if it returned `f`'s default value for `arg1`, which is 1), then `f` could be simplified:

```R

f <- function(x, arg1 = 1) {
  `%OR2%` <- function(x, y) {
    if (!is.null(x)) x
    else             y
  }
  
  if (missing(arg1)) arg1 <- x$arg1 %OR2% arg1
  
  arg1
}
```

R/server.R
```diff
@@ -561,14 +561,52 @@ runApp <- function(appDir=getwd(),
...
+  # |-------------|-----------|--------------|----------------------------------------------------------------------------------------------------------------------------------------|
+  # | no          | no        | use defaults | exhaust all possibilities: if it's missing (runApp does not specify); THEN if it's not in shinyApp appParts$options; THEN use defaults |
+  # | yes         | no        | use shinyApp | if it's missing (runApp does not specify); THEN if it's in shinyApp appParts$options; THEN use shinyApp                                |
+  # | no          | yes       | use runApp   | if it's not missing (runApp specifies), use those                                                                                      |
+  # | yes         | yes       | use runApp   | if it's not missing (runApp specifies), use those                                                                                      |
+  #
+  # I tried to make this as compact and intuitive as possible,
+  # given that there are four distinct possibilities to check
+  runOpts <- appParts$options
```
wch: Would the name `appOpts` make more sense here? Because these are options from the app object (as opposed to those given to `runApp`).

R/server.R
```diff
@@ -561,14 +561,52 @@ runApp <- function(appDir=getwd(),
...
+  assignVal <- function(arg, default) {
```
wch: This function doesn't assign values, so maybe a name like `defaultVal` would be better. Or maybe `valueDefault`?

PR #1488, opened 2016-11-28
fix #1487: better error handling for insertUI when selector does not match anything in DOM

### LINKED ISSUE #1487: Provide Better Mistake Handling for insertUI
I just spent about an hour tracking down a problem that was caused by a missing `div` ID that I intended to but forgot to put into the `ui.R` file. I was providing a value to `selector` that didn't exist but Shiny didn't emit any error. The only clue is from the internet browser's console:

    SCRIPT438: Object doesn't support property or method 'replace'
    output_binding_html.js (191,7)

It would be better if Shiny output an error such as `Error: The element selector is not in the application's DOM.`

bborgesr:
This makes the JS error more descriptive. (Still no error is printed to the R console or any indication appears on app that something went wrong).

Example app (notice that I use the wrong selector `#add1`, instead of `#add`):

```r
library(shiny)
options(shiny.launch.browser=F, shiny.minified=F, shiny.port = 9000)

## Only run this example in interactive R sessions
if (interactive()) {
  # Define UI
  ui <- fluidPage(
    actionButton("add", "Add UI")
  )
  
  # Server logic
    observeEvent(input$add, {
      insertUI(
        selector = "#add1",
        where = "afterEnd",
        ui = textInput(paste0("txt", input$add),
                       "Insert some text")
      )
    })
  }
  
  # Complete app with UI and server components
  shinyApp(ui, server)
}
```

cc @wch


srcjs/shinyapp.js
```diff
@@ -663,7 +663,11 @@ var ShinyApp = function() {
       // render the HTML and deps to a null target, so
       // the side-effect of rendering the deps, singletons,
       // and <head> still occur
-      exports.renderHtml($([]), message.content.html, message.content.deps);
+      try {
+        exports.renderHtml($([]), message.content.html, message.content.deps);
+      } catch(err) {
+        throw "The selector you chose could not be found in the DOM.";
```
wch: If I understand correctly, if `targets.length` is 0, it will always reach this `throw` statement. This is because in line 192 of output_binding_html.js, it calls `val.replace(self._reSingleton, findNewPayload)`, but `val` is `[]`, so there's no `replace` method.

If I'm right about that, then this code path (where `targets.length===0`) never worked, and that should be fixed. That should probably be done first. When would it be desirable to render this stuff when there are no targets?

jcheng5:
I don't think we should throw, rather just log a warning. The `dispatchMessage` mechanism isn't designed for message handlers to throw exceptions (though maybe/probably it should be) and anyway this isn't that crazy of a thing to do, especially with `multiple = TRUE`.

srcjs/shinyapp.js
jcheng5: Yeah it looks like this call has transposed arguments, should be:

`exports.renderHtml(message.content.html, $([]), message.content.deps)`

bborgesr:
@jcheng5, just `console.log`, then?

PR #1510, opened 2016-12-14
Add reactive debounce and throttle functions

jcheng5:
These are higher-order reactive functions to help slow down the effects of reactivity.

```r
r <- reactive(input$chattyInput)

# r_throttled is a reactive expression with the same values as r, except it
# will update/invalidate no more than once every 500 millis
r_throttled <- r %>% throttle(500)

# Same, but will wait for r to be idle for at least 500 millis before
# updating/invalidating.
r_debounced <- r %>% debounce(500)
```

I also added `options(device.ask.default = FALSE)` to all the examples that use renderPlot. This prevents the `Hit <Return> to see next plot` from occurring while you're trying to run those apps.

- [x] Add unit tests


KZARCA:
Hi!
It would be great if the r argument could be an observer as well.

PR #1513, opened 2016-12-15
Add session$userData feature

### LINKED ISSUE #1512: Any simple clean way to attach info to a session?
[x-post from shiny-discuss](https://groups.google.com/forum/#!topic/shiny-discuss/RlFGtOJMKm0)

Suppose I have a package with an exported function that can be called from other people's shiny apps' server code.  I want to know whether it's the first time they call this function or not. What's the best way to do this?

I can add an attribute to the `session` object but that sounds like a bad idea. I feel like there should be a simple solution I'm not seeing

Example pseudocode:

```
# This function would be in a package
myfunc <- function() {
  if (firsttime) {
    ...
  } else {
    ..
  }
}
```

And any shiny app can call it in server code, potentially multiple times

```
ui <- fluidPage()
  mypkg::myfunc()
}
```

If I was working on a regular R package without a shiny app, I would use a global environment to store this global value. But in this case I want this to be per-session.

This isn't an issue, it's a question that's gotten no response on shiny-discuss. I might be abusing the Issues system, but I thought this would be a bit more considerate than my usual method of personally emailing @jcheng5

jcheng5:
Fixes #1512

This implements `session$userData` as an environment, not `reactiveValues`. @wch @bborgesr feel free to weigh in. I could go either way.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -15,6 +15,8 @@ shiny 0.14.2.9000
 
 * Added ability through arguments to the `a` tag function called inside `downloadButton()` and `downloadLink()`. Closes [#986](https://github.com/rstudio/shiny/issues/986). ([#1492](https://github.com/rstudio/shiny/pulls/1492))
 
+* Added a `userData` environment to `session`, for storing arbitrary session-related variables. Generally, session-scoped variables are created just by declaring normal variables that are local to the Shiny server function, but `session$userData` may be more convenient for some advanced scenarios. ([#1512](https://github.com/rstudio/shiny/issues/1512))
+
 ### Bug fixes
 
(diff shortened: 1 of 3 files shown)


bborgesr:
I vote for leaving it as an environment. Joe, in #1512, you said:

>  Only question I have is whether `session$userData` should be a regular environment, or a `reactiveValues`? (If the former, if you needed reactivity you could create a `reactiveValues` object and put it into a slot in the regular environment; but if the common case is needing reactivity it might be nicer to do the latter.)

Your parenthetical is good enough for me. I think if people need it, they should go through the trouble of doing what you suggest. The reason I don't think this should be the default is because I think that, in most cases, this will be used just for counters and regular objects, in which case, it would be a pain to have to do `isolate(session$userData$counter)` each time you want to access the value without taking a dependency on it (or even when you're outside of a reactive context -- if you're just accessing the value from a regular function, for example).

wch:
I think using an environment is a good idea.

Issue #1348, opened 2016-09-01
Feature request: Images associated with radioButton inputs


philchalmers:
Hi all,

Could something like the following be added naively to shiny? http://stackoverflow.com/questions/35783446/can-you-have-an-image-as-a-radiobutton-choice-in-shiny

I have interests in using the `radioButtons()` input along with .png files, where each response option is associated with a specific image. It also would have to be included naively (trying to avoid the `:::` operator for CRAN packages and stability across versions). Cheers.

wch:
This is a duplicate of #1202. I think I misunderstood that issue the first time around -- I actually think it's a good idea to allow HTML to be used for radio button choices. For this particular case, a choice would need a `<img>` tag and the image would need to be stored in a www/ subdirectory.

philchalmers:
Thanks for the quick reply, and sorry for not noticing that earlier related issue. Support for `HTML()` constructed input objects for each element in `choices` would be perfect for my situation (for my purpose the www/ subdirectory isn't necessarily needed, but an explicit path to the image file should be sufficient).

bborgesr:
Closed via d62a2fc (#1521)

PR #1521, opened 2016-12-30
Allow arbitrary UI code in the choices for radio buttons and checkbox group input

bborgesr:
Example app:

```r
ui <- fluidPage(
  sidebarLayout(
    sidebarPanel( 
      radioButtons("normal", "Normal radio buttons:",
        c("Normal" = "norm", "Uniform" = "unif",
         "Log-normal" = "lnorm", "Exponential" = "exp")),
      hr(),
      radioButtons("fancy", "Fancy radio buttons:",
      choiceNames = list(tags$span(HTML("Normal &mu;<sub>1</sub>")),
        icon("calendar", "fa-3x"),
        icon("bed"),
        icon("cog", lib = "glyphicon"),
        img(src = "https://www.rstudio.com/wp-content/uploads/2016/09/RStudio-Logo-Blue-Gray-250.png")),
      choiceValues = c("html", "icon1", "icon2", "icon3", "img"))),
    mainPanel(textOutput("rb1"), textOutput("rb2"))
  )
)

  output$rb1 <- renderText({
    paste("You chose", input$normal)
  })
  
  output$rb2 <- renderText({
    paste("You chose", input$fancy)
  })
}

shinyApp(ui, server)
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,8 @@ shiny 1.0.0.9001
 
 ### Breaking changes
 
+* The functions `radioButtons()`, `checkboxGroupInput()` and `selectInput()` (and the corresponding `updateXXX()` functions) no longer accept a `selected` argument whose value is the name of a choice, instead of the value of the choice. This feature had been soft-deprecated since Shiny 0.10 (printing a warning message, but still trying to match the name to the right choice) and it's now been completely deprecated.
+
 ### New features
 
 ### Minor new features and improvements
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,6 +9,8 @@ shiny 1.0.0.9001
 
 ### Minor new features and improvements
 
+* Addressed [#1348](https://github.com/rstudio/shiny/issues/1348) and [#1437](https://github.com/rstudio/shiny/issues/1437) by adding two new arguments to `radioButtons()` and `checkboxGroupInput()`: `choiceNames` (list or vector) and `choiceValues` (list or vector). These can be passed in as an alternative to `choices`, with the added benefit that the elements in `choiceNames` can be arbitrary UI (i.e. anything created by `HTML()` and the `tags()` functions, like icons and images). While the underlying values for each choice (passed in through `choiceValues`) must still be simple text, their visual representation on the app (what the user actually clicks to select a different option) can be any valid HTML element. See `?radioButtons` for a small example. ([#1521](https://github.com/rstudio/shiny/pull/1521))
+
 * Updated `tools/README.md` with more detailed instructions. ([##1616](https://github.com/rstudio/shiny/pull/1616))
 
 * Fixed [#1565](https://github.com/rstudio/shiny/issues/1565), which meant that resources with spaces in their names return HTTP 404. ([#1566](https://github.com/rstudio/shiny/pull/1566))
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,7 +5,7 @@ shiny 1.0.0.9001
 
 ### Breaking changes
 
-* The functions `radioButtons()`, `checkboxGroupInput()` and `selectInput()` (and the corresponding `updateXXX()` functions) no longer accept a `selected` argument whose value is the name of a choice, instead of the value of the choice. This feature had been soft-deprecated since Shiny 0.10 (printing a warning message, but still trying to match the name to the right choice) and it's now been completely deprecated.
+* The functions `radioButtons()`, `checkboxGroupInput()` and `selectInput()` (and the corresponding `updateXXX()` functions) no longer accept a `selected` argument whose value is the name of a choice, instead of the value of the choice. This feature had been deprecated since Shiny 0.10 (it printed a warning message, but still tried to match the name to the right choice) and it's now completely unsupported.
 
 ### New features
 
--- a/R/input-utils.R
+++ b/R/input-utils.R
@@ -66,10 +66,10 @@ generateOptions <- function(inputId, choices = NULL, selected, inline,
       # checkbox-inline.
       if (inline) {
         tags$label(class = paste0(type, "-inline"), inputTag,
-                   tagList(tags$span(pd$html), pd$dep))
+                   tags$span(pd$html, pd$dep))
       } else {
         tags$div(class = type, tags$label(inputTag,
-                 tagList(tags$span(pd$html), pd$dep)))
+                 tags$span(pd$html, pd$dep)))
       }
     },
     SIMPLIFY = FALSE, USE.NAMES = FALSE
--- a/R/input-utils.R
+++ b/R/input-utils.R
@@ -35,12 +35,49 @@ validateSelected <- function(selected, choices, inputId) {
 }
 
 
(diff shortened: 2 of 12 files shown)


bborgesr:
Fixes #1348 and fixes #1437

jcheng5:
I'd prefer `choiceNames` and `choiceValues` instead of `choicesNames` and `choicesValues`.

I don't like that `choices`, `choicesNames`, and `choicesValues` propagate beyond `checkChoicesArgs`; I think `choices` should go away, since it can be thought of as a shorthand for `choicesNames <- names(choices)` and `choicesValues <- unname(choices)`. Then the rest of the code (like `validateSelected`) can just deal with one way of specifying these values, instead of two.

I tend to think of functions that are named `checkFoo` and `validateFoo` as simply returning `TRUE` or `FALSE`, or throwing warnings or errors. In these cases we're using them to do normalization. Not a big deal and you don't have to change the names in this PR but maybe something to keep in mind moving forward.

(Don't forget the inspecting/merging of the API branch needs to happen before the conference, this PR can be merged after the conference...)

R/input-utils.R
```diff
@@ -2,30 +2,65 @@ controlLabel <- function(controlName, label) {
...
+      stop("`choicesNames` and `choicesValues` must have the same length.")
+    }
+    if (!is.null(names(choicesNames)) || !is.null(names(choicesValues))) {
+      stop("`choicesNames` and `choicesValues` must not be named.")
+    }
+  } else {
+    if (lenNames != 0 || lenVals != 0) {
+      warning("Using `choices` argument; ignoring `choicesNames` and
+                `choicesValues`.")
+      choicesNames = NULL
```
wch: Should be `<-`.

R/input-utils.R
```diff
@@ -2,30 +2,65 @@ controlLabel <- function(controlName, label) {
...
+           alternatively, for both `choicesNames` and `choicesValues`).")
+    }
+    if (lenNames != lenVals) {
+      stop("`choicesNames` and `choicesValues` must have the same length.")
+    }
+    if (!is.null(names(choicesNames)) || !is.null(names(choicesValues))) {
+      stop("`choicesNames` and `choicesValues` must not be named.")
+    }
+  } else {
+    if (lenNames != 0 || lenVals != 0) {
```
bborgesr: check for NULL instead

R/input-utils.R
```diff
@@ -2,30 +2,65 @@ controlLabel <- function(controlName, label) {
...
+
+  if (is.null(choices)) {
+    if (lenNames == 0 || lenVals == 0) {
+      stop("Please specify a non-empty vector for `choices` (or,
+           alternatively, for both `choicesNames` and `choicesValues`).")
+    }
+    if (lenNames != lenVals) {
+      stop("`choicesNames` and `choicesValues` must have the same length.")
+    }
+    if (!is.null(names(choicesNames)) || !is.null(names(choicesValues))) {
```
bborgesr: use `anyNamed` instead

bborgesr:
I think I've addressed all points and this should be good to squash-merge. 

The one possible hiccup is what we're breaking backward compatibility (but only for unexported functions!) as we change the name of some of the internal utility functions (switch from "validate"/"check" prefixes to "normalize", per Joe's comment above) and their expected arguments (pass in `choiceNames` and `choiceValues` instead of `choices`).

jcheng5:
Maybe we can review this together tomorrow. It's been so long I don't remember :)

R/input-utils.R
```diff
@@ -2,30 +2,62 @@ controlLabel <- function(controlName, label) {
...
+  }
+
+  return(list(choiceNames = choiceNames, choiceValues = choiceValues))
+}
 
 # Before shiny 0.9, `selected` refers to names/labels of `choices`; now it
 # refers to values. Below is a function for backward compatibility. It also
 # coerces the value to `character`.
-validateSelected <- function(selected, choices, inputId) {
+normalizeSelected <- function(selected, inputId, choiceNames, choiceValues) {
```
wch: Since 0.10.0 (which was released in June 2014) users have gotten a warning if `selected` was a name instead of a value. So I think it's safe to assume that no one is doing that anymore. And that means that we can remove this function entirely. We will need to do `selected <- as.character(selected)` where we previously called `validateSelected`, for the reasons in the comment directly below.

R/update-input.R
```diff
@@ -452,16 +452,18 @@ updateSliderInput <- function(session, inputId, label = NULL, value = NULL,
 
 
 updateInputOptions <- function(session, inputId, label = NULL, choices = NULL,
-                               selected = NULL, inline = FALSE,
-                               type = 'checkbox') {
-  if (!is.null(choices))
-    choices <- choicesWithNames(choices)
+                               selected = NULL, inline = FALSE, type = NULL,
```
wch: It would be good to error if `type` is NULL.

R/update-input.R
```diff
@@ -452,16 +452,18 @@ updateSliderInput <- function(session, inputId, label = NULL, value = NULL,
...
+                               choiceNames = NULL, choiceValues = NULL) {
+  args <- normalizeChoicesArgs(choices, choiceNames, choiceValues)
+
   if (!is.null(selected))
-    selected <- validateSelected(selected, choices, session$ns(inputId))
+    selected <- normalizeSelected(selected, session$ns(inputId),
+      args$choiceNames, args$choiceValues)
 
-  options <- if (!is.null(choices)) {
+  options <- if (!is.null(args$choiceValues)) {
```
wch: This should be `if (!is.null(choices) || !is.null(choiceNames) || !is.null(choiceValues))`.

R/input-utils.R
```diff
@@ -2,30 +2,62 @@ controlLabel <- function(controlName, label) {
...
+  } else {
+    if (!is.null(choiceNames) || !is.null(choiceValues)) {
+      warning("Using `choices` argument; ignoring `choiceNames` and `choiceValues`.")
+    }
+    choices <- choicesWithNames(choices) # resolve names if not specified
+    choiceNames <- names(choices)
+    choiceValues <- unname(choices)
+  }
+
+  return(list(choiceNames = choiceNames, choiceValues = choiceValues))
```
wch: If `choices`, `choicesNames`, and `choicesValues` are all NULL, this should return a list like:

```
list(choiceNames = NULL, choiceValues = NULL)
```

And also have a test for it.

wch: Also, `choiceNames` and `choiceValues` should always be lists (or NULL), and not character vectors.

NEWS.md
```diff
@@ -5,6 +5,8 @@ shiny 1.0.0.9001
 
 ### Breaking changes
 
+* The functions `radioButtons()`, `checkboxGroupInput()` and `selectInput()` (and the corresponding `updateXXX()` functions) no longer accept a `selected` argument whose value is the name of a choice, instead of the value of the choice. This feature had been soft-deprecated since Shiny 0.10 (printing a warning message, but still trying to match the name to the right choice) and it's now been completely deprecated.
```
wch: It would be more accurate to say that it had been deprecated (not soft-deprecated) since 0.10.0.

PR #1529, opened 2017-01-09
Relax naming requirements for addResourcePath

jjallaire:
First character no longer needs to be a letter. See https://github.com/rstudio/tutor/issues/4 for discussion.

### DIFF
--- a/R/server.R
+++ b/R/server.R
@@ -52,7 +52,7 @@ registerClient <- function(client) {
 #' @export
 addResourcePath <- function(prefix, directoryPath) {
   prefix <- prefix[1]
-  if (!grepl('^[^\\.][a-z0-9\\-_.]*$', prefix, ignore.case=TRUE, perl=TRUE)) {
+  if (!grepl('^[a-z0-9\\-_][a-z0-9\\-_.]*$', prefix, ignore.case=TRUE, perl=TRUE)) {
     stop("addResourcePath called with invalid prefix; please see documentation")
   }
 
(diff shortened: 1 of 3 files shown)


R/server.R
```diff
@@ -52,7 +52,7 @@ registerClient <- function(client) {
 #' @export
 addResourcePath <- function(prefix, directoryPath) {
   prefix <- prefix[1]
-  if (!grepl('^[a-z][a-z0-9\\-_.]*$', prefix, ignore.case=TRUE, perl=TRUE)) {
+  if (!grepl('^[^\\.][a-z0-9\\-_.]*$', prefix, ignore.case=TRUE, perl=TRUE)) {
```
wch: I think this just checks that the first character is not `.`, so other weird characters end up getting through. For example:

```
> grepl('^[^\\.][a-z0-9\\-_.]*$', '#abc', ignore.case=TRUE, perl =TRUE)
[1] TRUE
```

If the goal is to prevent a leading `.`, then I think `^[a-z0-9\\-_][a-z0-9\\-_.]*$` is better:

```
> grepl('^[a-z0-9\\-_][a-z0-9\\-_.]*$', '#abc', perl =TRUE)
[1] FALSE
> grepl('^[a-z0-9\\-_][a-z0-9\\-_.]*$', '.abc', perl =TRUE)
[1] FALSE
> grepl('^[a-z0-9\\-_][a-z0-9\\-_.]*$', 'abc', perl =TRUE)
[1] TRUE
> grepl('^[a-z0-9\\-_][a-z0-9\\-_.]*$', '_abc', perl =TRUE)
[1] TRUE
> grepl('^[a-z0-9\\-_][a-z0-9\\-_.]*$', '-abc', perl =TRUE)
[1] TRUE
> grepl('^[a-z0-9\\-_][a-z0-9\\-_.]*$', '1abc', perl =TRUE)
[1] TRUE
```

jjallaire:
Good point. Just made that change.

PR #1559, opened 2017-01-26
Add shiny:filedownload Javascript event

wch:
This PR does a few things that are needed for testing file downloads:

* It adds a `shiny:filedownload` event in Javascript. The event object contains the ID of the downloadButton/Link. It does not contain the href, because it changes each time the application is run (if the ID is `downloadData`, the link looks something like `session/4b3259d8737d682d272e02ecc3faba5e/download/downloadData?w=`).
* It adds the function `markOutputAttrs`, which can be used to mark the object returned from a render function (which is a reactive) with an attribute. When the reactive is assigned to an output value, the attribute will be copied to the corresponding observer.
* It marks `downloadHandlers` so that their values are not snapshotted. As mentioned above, the value -- the href -- changes with each run of the application, so it doesn't make sense to snapshot it for testing.


srcjs/output_binding_downloadlink.js
```diff
@@ -8,3 +8,11 @@ $.extend(downloadLinkOutputBinding, {
   }
 });
 outputBindings.register(downloadLinkOutputBinding, 'shiny.downloadLink');
+
+
+// Trigger shiny:filedownload event whenever a downloadButton/Link is clicked
+$(document).on('click.shinyDownloadLink', 'a.shiny-download-link', function(e) {
+  var evt = jQuery.Event('shiny:filedownload');
+  evt.name = this.id;
```
wch: Should set `evt.href` here.

R/shinywrappers.R
```diff
@@ -88,6 +88,26 @@ as.tags.shiny.render.function <- function(x, ..., inline = FALSE) {
   useRenderFunction(x, inline = inline)
 }
 
+
+#' Mark a render function with attributes that will be used by the output
+#'
+#' @inheritParams markRenderFunction
+#' @param snapshotExclude If TRUE, exclude the output from test snapshots.
+#'
+#' @export
```
wch: Don't export.

PR #1563, opened 2017-01-27
Barbara/user info

bborgesr:
This exports `session$user` (if it exists) to the client-side. It's accessible in the Shiny object: `Shiny.user`


srcjs/shinyapp.js
```diff
@@ -631,7 +631,8 @@ var ShinyApp = function() {
   });
 
   addMessageHandler('config', function(message) {
-    this.config = message;
+    this.config = {workerId: message.workerId, sessionId: message.sessionId};
+    if (message.user != null) exports.user = message.user;
```
bborgesr: `!=` is on purpose (instead of `!==`) because it checks for both `null` and `undefined` (see [here](http://stackoverflow.com/questions/2647867/how-to-determine-if-variable-is-undefined-or-null))

bborgesr:
@wch @jcheng5 ready to merge (?) I checked that it works fine with SSP: if an app requires authentication `Shiny.user` shows up with the username; if it doesn't require authentication, there is no `Shiny.user`.

srcjs/shinyapp.js
jcheng5: Next time please make a comment directly in the code to that effect.

PR #1568, opened 2017-01-31
Add shiny:sessionInit event

bborgesr:
This PR triggers a new JS event called `shiny:sessioninitialized `, which is fired at the end of the initialize method of the Session object (piggybacking on the `config` message, which is the last command of the initialize method). This allows us to listen for the `shiny:sessioninitialized ` event when we want to get the value of things like `Shiny.user`.


srcjs/shinyapp.js
```diff
@@ -632,7 +632,12 @@ var ShinyApp = function() {
 
   addMessageHandler('config', function(message) {
     this.config = {workerId: message.workerId, sessionId: message.sessionId};
+    // != is on purpose (instead of !==) because it checks for both null and undefined
+    // (see here: http://stackoverflow.com/a/21273362/6174455)
+    /*eslint-disable */
     if (message.user != null) exports.user = message.user;
```
wch: It's more idiomatic to use `if (message.user)` (to look for truthy values in general) or, if you want to be more precise, you could be explicit with `(message.user !== undefined && message.user !== null)`. Either way is cleaner than adding the (necessary) comments explaining why it's written that way.

srcjs/shinyapp.js
```diff
@@ -632,7 +632,12 @@ var ShinyApp = function() {
...
+    /*eslint-enable */
+    $(document).trigger('shiny:sessionInit');
```
wch: For consistency with other events, this should be all lower case. I think that `shiny:sessioninitialized` would be better, since it indicates that it happens after the initialization.

bborgesr:
After a chat with @jcheng5, we're going to mull this over a little bit. For the particular use case that we wanted (fetching `Shiny.user`), we can use the following:

```js
$(document).one('shiny:idle', function() {
    console.log(Shiny.user);
});
```

instead of the new way this PR provides, which is:

```js
$(document).on('shiny:sessioninitialized', function() {
    console.log(Shiny.user);
});
```

The question is if there is still value in having a `shiny:sessioninitialized` event, or alternatively, if we could just move `shiny:connected` to fire only when `shiny:sessioninitialized` is currently firing (which would make it redundant and therefore unnecessary). For this last point, there are both questions of whether or not there is legitimate use of Shiny between the current `shiny:connected` and `shiny:sessioninitialized`, and back compatibility.

As of now, the order in which each event fires is:

connected -> sessioninitialized -> idle

wch:
It would be a good idea to look at a [search for `shiny:connected`](https://github.com/search?q=%22shiny%3Aconnected%22&type=Code&utf8=%E2%9C%93) to see if any of them need to `shiny:connected` to happen when it currently does.

wch:
After we merge this, we should also update the documentation to tell people that this event is probably what they want to use, instead of `shiny:connected`.

bborgesr:
Internal note: the plan is for us to merge this right after CRAN accepts 1.0.1

PR #1573, opened 2017-02-08
Assign individual html ids to radiobutton and checkbox choices

steffilazerte:
This assigns an html id value to the individual options created for both radiobuttons and checkboxes. By referencing this id individual choices can be 'disabled' using shinyjs (etc.).

Here's an example:

```
library(shiny)
library(shinyjs)

b <- seq(10, 50, by = 10)

ui <- fluidPage(
  shinyjs::useShinyjs(),
  checkboxGroupInput("disable", "Disable Buttons", choices = b, inline = TRUE),
  radioButtons("bins", "Number of bins:", choices = b, inline = TRUE),
  plotOutput("distPlot")
)


  observeEvent(input$disable, {
    lapply(b, FUN = function(x) shinyjs::toggleState(id = paste0("bins_", x), condition = !(x %in% input$disable)))
    if(input$bins %in% input$disable) updateRadioButtons(session, "bins", selected = b[!(b %in% input$disable)][1])
  })

    req(input$bins)
    x <- faithful[, 2]
    bins <- seq(min(x), max(x), length.out = as.numeric(input$bins) + 1)
  })
}

```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,6 +9,8 @@ shiny 1.0.0.9000
 
 ### Minor new features and improvements
 
+* Added automatic ids to radiobutton and checkbox choices. Id format: `inputId_choice`
+
 * Fixed [#1565](https://github.com/rstudio/shiny/issues/1565), which meant that resources with spaces in their names return HTTP 404. ([#1566](https://github.com/rstudio/shiny/pull/1566))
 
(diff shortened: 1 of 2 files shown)


wch:
I don't think this is needed - you should be able to use `toggleState`'s `selector` argument to control specific choices with shinyjs.

steffilazerte:
You're absolutely right, works like a charm, wish I'd realized that earlier!

PR #1579, opened 2017-02-10
Improve escapeHTML

### LINKED ISSUE #1577: About escapeHTML in util.js
https://github.com/rstudio/shiny/blob/master/srcjs/utils.js#L1

After reading about replacing one char after another is not a best practice (due to the order dependency of replacing and xss risk), I would suggest something like

```
function escape(s) {
  var escaped = {
    '&': '&amp;',
    '<': '&lt;',
    '>': '&gt;',
    "'": '&#39;',
    '"': '&quot;'
  };
  return s.replace(/[&<>'"]/g, function (m) {
    return escaped[m];
  });
}
```

Credits to https://gist.github.com/WebReflection/df05641bd04954f6d366.

If you think there is some point here, I may prepare a PR.

albertosantini:
Replacing one char after another is not a best practice, due to the order dependency of replacing, xss risk and performance.

Fix #1577

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -24,6 +24,8 @@ in shiny apps. For more info, see the documentation (`?updateQueryString` and `?
 
 * Added support for plot interactions with ggplot2 > 2.2.1. ([#1578](https://github.com/rstudio/shiny/pull/1578))
 
+* Improved escapeHTML (util.js) in terms of the order dependency of replacing, XSS risk attack and performance. ([#1579](https://github.com/rstudio/shiny/pull/1579))
+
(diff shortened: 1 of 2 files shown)


wch:
Thanks. Could you fill out the contributor agreement? See https://github.com/rstudio/shiny/blob/master/CONTRIBUTING.md

albertosantini:
RStudio Individual Contributor Agreement signed and sent via email.

wch:
Great, can you also add an entry to NEWS.md under "Minor new features and improvements
"?

PR #1582, opened 2017-02-10
Fix #1581

### LINKED ISSUE #1581: validateIcon in actionLink
In the current version of shiny, `actionLink` is defined in the following way
```
actionLink <- function(inputId, label, icon = NULL, ...) {
...codes...
  list(validateIcon(icon), label)
...codes...
}
```

That basically means that everything putting as the third item after `inputId` and `label` will be treated as an `icon` and will need to pass the `validateIcon` check. As a result, when users don't want an icon but forgot to define `icon = NULL`, the `...` gets his chance to be checked by `validateIcon` and in most cases, it will return an error message saying `Error in validateIcon: Invalid icon. Use Shiny's 'icon()' function to generate a valid icon`.

One example is the first issue in giphyr
 https://github.com/haozhu233/giphyr/issues/1
I fixed it by defining `icon = NULL` https://github.com/haozhu233/giphyr/commit/62662d137b1fad390391a5129ca02e5ab377ab83

haozhu233:
Change the order of `...` and `icon = NULL` in `actionLink()` to avoid pass non-icon contents into `validateIcon` when `icon` is not specified.

### DIFF
--- a/R/input-action.R
+++ b/R/input-action.R
@@ -54,7 +54,7 @@ actionButton <- function(inputId, label, icon = NULL, width = NULL, ...) {
 
 #' @rdname actionButton
 #' @export
-actionLink <- function(inputId, label, icon = NULL, ...) {
+actionLink <- function(inputId, label, ..., icon = NULL) {
   value <- restoreInput(id = inputId, default = NULL)
 
   tags$a(id=inputId,


wch:
Changing the argument order at this point would break existing user code, and for something that's not really a benefit, in my opinion.

PR #1589, opened 2017-02-17
Enable hiding of modals, not just removing

### LINKED ISSUE #1588: Prevent removeModal from resetting actionButton value
I'm trying to use shiny's default modals (as opposed to shinyBS) to provide users with additional filtering options. The modalDialog includes a few selectInputs that are created using renderUI. The problem I'm encountering is that whenever removeModal() is triggered, all values of the selectInputs and the actionButton are reset. Is there a way to prevent removeModal from resetting things?

Specifically, I need input$actionButton2 to not reset.

```
# Modal Code
observeEvent(input$actionButton1, { # actionButton1 is outside of the modal in the main window
    showModal(modalDialog(
      title = "MODAL TITLE",
      size = "l",
      easyClose = FALSE,
      fluidRow(
        column(width = 12,
               uiOutput("selectInput1"),
               uiOutput("selectInput2"))
       ),
      footer = tagList(
        modalButton("Cancel"),
        actionButton("actionButton2", "ACTION BUTTON 2") # I need this to not reset.
      )
    ))
  })
```

bborgesr:
Closes #1588.

Example app:

```r
library(shiny)

ui <- fluidPage(
  actionButton("go", "Show modal dialog")
)

  observeEvent(input$go, {
    showModal(modalDialog(
      radioButtons("radio", "Some choices", c(1,2,3)),
      footer = tagList(
        modalButton("Cancel", hideOnly = TRUE)
      )
    ))
  })
}

shinyApp(ui, server)
```

This is my first crack at this. It can probably benefit from some changes...

I also think there's at least one case in which this fails: if the user creates a hideOnly modal, and then closes (hides) it and after that, we wants to create a second modal. Shiny will detect that there is already a modal on the screen (hidden one) and it will display that one, instead of the actual one. There's probably an easy fix for this, if we want to pursue it.


bborgesr:
test from Github to Asana

bborgesr:
➤ Barbara Borges Ribeiro commented:
test

bborgesr:
After discussion with @wch, we decided not to pursue this because this is not the original intent behind Shiny modals and we feel like the trouble of complicating the mental model is not justified (especially having to deal with cases when the user might have several modals, of different kinds... whereas now, we only have to worry about the current modal and forget all about that if another is opened later). For now, users who need "persistent" modals should use other "tools" (bs modals, or create their own) or use a workaround through `reactiveValues`.

PR #1592, opened 2017-02-23
fix: dateRangeInput did not respect weekstart arg

akersting:


wch:
Can you provide a minimal example app that we can use to test this fix?

akersting:
```R
server <- function(input, output) {}

ui <- fluidPage(
  dateRangeInput("date_range", "Date Range Input", weekstart = 1)
)

```

With the fix the weeks in the calendars correctly start on Monday. Without the fix they start on Sunday, which is correct only for `weekstart = 0` (the default).

You can also compare with the following code from input-date.R:
```R
tags$input(type = "text",
               class = "form-control",
               `data-date-language` = language,
               `data-date-week-start` = weekstart,
               `data-date-format` = format,
               `data-date-start-view` = startview,
               `data-min-date` = min,
               `data-max-date` = max,
               `data-initial-date` = value
    )
```
Here it also says `data-date-week-start` and not `data-date-weekstart`.

wch:
Thanks!

Sbrikky:
For me weekstart is only used when the language is not "en"

so for example:

```
server <- function(input, output) {}

ui <- fluidPage(
  dateRangeInput("date_range", "Date Range Input", weekstart = 1, language = "en-GB"),
  dateRangeInput("date_range2", "Date Range Input", weekstart = 1, language = "en")
)

```

The first dateRangeinput will correctly start on Mondays, but the second one will default to Sunday.

wch:
@Sbrikky Are you using the latest development version of Shiny? This fix was merged just last week.

Sbrikky:
I totally was not, and that did fix it. Thanks!

PR #1595, opened 2017-02-24
Remove redundant calls to setInput

wch:
The `exports.bindAll` function explicitly [calls](https://github.com/rstudio/shiny/blob/0f13075/srcjs/init_shiny.js#L210-L212) `inputs.setInput` for all inputs. However, `inputs.setInput` is already called: `exports.bindAll` calls `_bindAll`, which [calls](https://github.com/rstudio/shiny/blob/0f13075/srcjs/init_shiny.js#L199) `bindInputs`, which [calls](https://github.com/rstudio/shiny/blob/0f13075/srcjs/init_shiny.js#L165) `valueChangeCallback`, and inside of that function, it [calls](https://github.com/rstudio/shiny/blob/0f13075/srcjs/init_shiny.js#L109) `inputs.setInput`.

(This is relevant to other things I'm working on, where I want to call `setInput` with an options object that includes the `binding` and `el`. In `exports.bindAll`, the binding and element aren't available, so those can't be passed to `setInput` there.)


jcheng5:
LGTM other than building

PR #1596, opened 2017-02-27
Add binding and el fields to shiny:inputchanged event

wch:
With this PR, the `shiny:inputchanged` JavaScript event gets two new fields, `binding` and `el`, which contain the input binding and DOM element, respectively. Additionally, `Shiny.onInputChange()` now accepts an optional argument, `opts`, which can contain the same fields.

### DIFF
--- a/srcjs/init_shiny.js
+++ b/srcjs/init_shiny.js
@@ -93,8 +93,14 @@ function initShiny() {
     });
   });
 
-  exports.onInputChange = function(name, value) {
-    inputs.setInput(name, value);
+  exports.onInputChange = function(name, value, opts) {
+    opts = Object.assign({
+      immediate: false,
+      binding: null,
+      el: null
+    }, opts);
+
+    inputs.setInput(name, value, opts);
   };
 
   var boundInputs = {};
@@ -106,7 +112,9 @@ function initShiny() {
       var type = binding.getType(el);
       if (type)
         id = id + ":" + type;
-      inputs.setInput(id, value, !allowDeferred);
+
+      let opts = { immediate: !allowDeferred, binding: binding, el: el };
+      inputs.setInput(id, value, opts);
     }
   }
 
--- a/srcjs/init_shiny.js
+++ b/srcjs/init_shiny.js
@@ -224,9 +224,9 @@ function initShiny() {
   exports.bindAll = function(scope) {
     // _bindAll returns input values; it doesn't send them to the server.
     // export.bindAll needs to send the values to the server.
-    var currentValues = _bindAll(scope);
-    $.each(currentValues, function(name, val) {
-      inputs.setInput(name, val.value, val.opts);
+    var currentInputItems = _bindAll(scope);
+    $.each(currentInputItems, function(name, item) {
+      inputs.setInput(name, item.value, item.opts);
     });
 
     // Not sure if the iframe stuff is an intrinsic part of bindAll, but bindAll
--- a/srcjs/init_shiny.js
+++ b/srcjs/init_shiny.js
@@ -82,16 +82,24 @@ function initShiny() {
   var inputsRate = new InputRateDecorator(inputsEvent);
   var inputsDefer = new InputDeferDecorator(inputsEvent);
 
-  // By default, use rate decorator
-  var inputs = inputsRate;
-  $('input[type="submit"], button[type="submit"]').each(function() {
+  var inputsValidate;
+  if ($('input[type="submit"], button[type="submit"]').length > 0) {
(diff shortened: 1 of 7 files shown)


jcheng5:
I'd prefer to walk through this PR together.

srcjs/init_shiny.js
```diff
@@ -93,8 +93,14 @@ function initShiny() {
     });
   });
 
-  exports.onInputChange = function(name, value) {
-    inputs.setInput(name, value);
+  exports.onInputChange = function(name, value, opts) {
+    opts = Object.assign({
```
jcheng5: Make sure Object.assign is polyfilled

srcjs/init_shiny.js
```diff
@@ -128,7 +136,13 @@ function initShiny() {
 
         var type = binding.getType(el);
         var effectiveId = type ? id + ":" + type : id;
-        currentValues[effectiveId] = binding.getValue(el);
+        inputItems[effectiveId] = {
```
jcheng5: From discussion, it seems like the behavior we have today would cause immediate=FALSE when the exported bindAll is called. Please follow up to see if indeed immediate=FALSE. In any case we want this to be immediate going forward.

srcjs/input_rate.js
```diff
@@ -211,84 +213,109 @@ var InputBatchSender = function(shinyapp) {
...
 }).call(InputDeferDecorator.prototype);
 
+
 var InputEventDecorator = function(target) {
   this.target = target;
 };
 (function() {
-  this.setInput = function(name, value, immediate) {
+  this.setInput = function(name, value, opts) {
+    opts = Object.assign({
```
wch: It might be better to have a global default options object instead of creating it each time.

srcjs/input_rate.js
```diff
@@ -211,84 +213,109 @@ var InputBatchSender = function(shinyapp) {
...
 }).call(InputEventDecorator.prototype);
 
+
 var InputRateDecorator = function(target) {
   this.target = target;
   this.inputRatePolicies = {};
 };
 (function() {
-  this.setInput = function(name, value, immediate) {
+  this.setInput = function(name, value, opts) {
```
wch: Instead of checking `opts` at each level of input decorator, it would be cleaner to add another level, like `inputValidatorDecorator` to the top of the stack, which checks that `name` isn't `null`, the `opts` are valid, and so on.

srcjs/shinyapp.js
```diff
@@ -29,7 +32,10 @@ var ShinyApp = function() {
 
     $.extend(initialInput, {
       // IE8 and IE9 have some limitations with data URIs
-      ".clientdata_allowDataUriScheme": typeof WebSocket !== 'undefined'
+      ".clientdata_allowDataUriScheme": {
```
wch: This can be moved init_shiny.

srcjs/shinyapp.js
```diff
@@ -4,6 +4,9 @@ var ShinyApp = function() {
   // Cached input values
   this.$inputValues = {};
 
+  // Input values at initialization (and reconnect)
+  this.$initialInput = {};
```
wch: `this.$initialInput` and `this$inputValues` should not keep `opts`, because it would prevent the `el` from being garbage collected. The `opts` are only used in the input decorator stack; there's no need to keep them outside of that.

srcjs/init_shiny.js
wch: PR #1595 (commit 0ef15fa) removed the following code in `bindInputs` because we thought it was redundant:

```js
        if (shinyapp.isConnected()) {
          valueChangeCallback(binding, el, false);
        }
```

It turns out that the `false` in this call is important, and wasn't being used in the other call to `inputs.setInput` from `exports.bindAll`. So  we'll need to need to make sure that the value of `immediate` is set properly here.

srcjs/init_shiny.js
```diff
@@ -82,24 +82,27 @@ function initShiny() {
...
-      immediate: false,
-      binding: null,
-      el: null
-    }, opts);
+  } else {
+    // By default, use rate decorator
+    inputsValidate = new InputValidateDecorator(inputsRate);
+  }
 
+  var inputs = inputsValidate;
```
jcheng5: Total nitpick but I would've had `var inputs` above the if/else block, and have `inputs = inputsRate` and `inputs = inputsDefer` as before, and then after the if/else block, `inputs = new InputValidateDecorator(inputs)`.

wch: That makes sense. I can make that change.

PR #1610, opened 2017-03-16
Don't use data-drag-interval for non-range sliders. Fixes #1605

### LINKED ISSUE #1605: sliderInput unintentional behavior change
Repro:

```r
ui <- fluidPage(sliderInput("slidername", NULL, min=0, max=100, value=0))
server <- function(input, output, server) {}
shinyApp(ui, server)
```

Probably an artifact of updating the ion range slider library and not noticing this change... We could just document this change and accept, or we can try to preserve the old behavior...

Origin: http://stackoverflow.com/questions/42764365/shiny-sliderinput-behaviour-change

Working fine in 0.13.2; different/broken behavior in 0.14.

wch:
This closes #1605.


bborgesr:
LGTM

PR #1613, opened 2017-03-21
Change NS() to return a vectorized function

wch:
Previously, `NS` had this behavior:

```R
ns <- NS(c("one", "two"))
ns(c("three", "four"))
# [1] "one-two-three-four"
```

But it make more sense for the `ns()` function (the one which was returned by `NS()`) to return a vector. Now it does this:

```R
ns <- NS(c("one", "two"))
ns(c("three", "four"))
# [1] "one-two-three" "one-two-four" 
```

The code ended up being more complex than I expected, but I didn't see a simpler way to do it.

### DIFF
--- a/R/shiny.R
+++ b/R/shiny.R
@@ -386,7 +386,7 @@ NS <- function(namespace, id = NULL) {
     if (length(ns_prefix) == 0)
       return(id)
 
-    paste(ns_prefix, id, sep = "-")
+    paste(ns_prefix, id, sep = ns.sep)
   }
 
   if (missing(id)) {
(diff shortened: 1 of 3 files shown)


R/shiny.R
```diff
@@ -375,12 +375,24 @@ NULL
...
+  else
+    ns_prefix <- paste(namespace, collapse = ns.sep)
+
+  f <- function(id) {
+    if (length(id) == 0)
+      return(ns_prefix)
+    if (length(ns_prefix) == 0)
+      return(id)
+
+    paste(ns_prefix, id, sep = "-")
```
jcheng5: `sep = "-"` should be `sep = ns.sep`?

wch: Oops, fixed.

jcheng5:
LGTM!

PR #1614, opened 2017-03-21
Add reactiveVal() for single reactive value

jcheng5:
If you have just a single reactive value, instead of:

```r
rv <- reactiveValue(value = 0)
rv$value # read
rv$value <- 10 # write
```

You can do:

```r
r <- reactiveVal(0)
r() # read
r(10) # write
```

The advantages of the latter are 1) you only have to think of a single meaningful name (instead of `rv` and `value` it's just `r`), and 2) you can pass `r` to anything that's expecting a reactive expression, like into a module or a higher order reactive function like debounce/throttle.

### TODO

- [x] Automatically populate label
- [x] Either support `freezeReactiveValue`, or officially don't support it


jcheng5:
@wch @bborgesr This should be good to go.

R/reactives.R
```diff
@@ -38,6 +38,224 @@ Dependents <- R6Class(
...
+      } else {
+        c("reactiveVal:", label)
+      }
+    }
+  )
+)
+
+#' Create a (single) reactive value
+#'
+#' A \code{reactiveVal} is an object used for reading and writing a value, like
```
bborgesr: change "object" to "function"..? I'm afraid the different semantics between reading/writing to `reactiveValues` values vs a `reactiveVal` value will trip people up in any case (i.e. for the former isn't callable). But if we label `reactiveVal` a function from the start, maybe that confusion can be mitigated.. However, I think that will leave people thinking about the difference between `reactive()` and `reactiveVal()`...

R/reactives.R
```diff
@@ -38,6 +38,224 @@ Dependents <- R6Class(
...
+#'   observeEvent(input$plus, {
+#'     newValue <- value() + 1
+#'     value(newValue)
+#'   })
+#'
+#'   output$value <- renderText({
+#'     value()
+#'   })
+#' }
+#'
```
bborgesr: It may be helpful to use this opportunity to compare and contrast `reactiveVal()` and `reactiveValues()`?

E.g. you could change the server function to this:
```r
# The comments below accomplish the same behavior using `reactiveValues()`
    value <- reactiveVal(0)     # rv <- reactiveValues(value = 0)
    
    observeEvent(input$minus, {
        newValue <- value() - 1 #
        value(newValue)         # rv$value <- rv$value - 1
    })
    
    observeEvent(input$plus, {
        newValue <- value() + 1 #
        value(newValue)         # rv$value <- rv$value + 1
    })
    
    output$value <- renderText({
        value()                 # rv$value
    })
}
```

R/reactives.R
```diff
@@ -38,6 +38,224 @@ Dependents <- R6Class(
...
+      private$frozen <- TRUE
+    },
+    thaw = function() {
+      private$frozen <- FALSE
+    },
+    isFrozen = function() {
+      private$frozen
+    },
+    format = function(...) {
+      label <- capture.output(print(base::format(private$value, ...)))
```
wch: Why is the `capture.output(print())` needed?

R/reactives.R
```diff
@@ -446,7 +667,7 @@ str.reactivevalues <- function(object, indent.str = " ", ...) {
 #' @export
 freezeReactiveValue <- function(x, name) {
   domain <- getDefaultReactiveDomain()
-  if (is.null(getDefaultReactiveDomain)) {
```
wch: Whoa, did this ever work correctly? It doesn't have a `()` on it.

R/reactives.R
```diff
@@ -461,6 +682,7 @@ Observable <- R6Class(
   'Observable',
   portable = FALSE,
   public = list(
+    .origFunc = 'function',
```
wch: Is the only purpose of storing `.origFunc` is so that the body can be deparsed in the `format.reactiveExpr` method?

R/reactives.R
jcheng5: I tried it with a data frame value, format doesn't convert data frames to strings. It just formats each individual cell or something. You still get a data frame as a result.

R/reactives.R
jcheng5: It didn't, but the only effect is a worse error message.

R/reactives.R
jcheng5: Yes exactly.

R/reactives.R
```diff
@@ -682,18 +905,26 @@ srcrefToLabel <- function(srcref, defaultLabel) {
...
 #' @rdname reactive
-is.reactive <- function(x) inherits(x, "reactive")
+is.reactive <- function(x) {
+  inherits(x, "reactive")
+}
 
 # Return the number of times that a reactive expression or observer has been run
 execCount <- function(x) {
-  if (is.reactive(x))
+  if (inherits(x, "reactiveExpr"))
```
wch: Maybe add `is.reactiveExpr` and `is.reactiveVal` methods? Or maybe `is.reactiveexpr` and `is.reactiveval` to match the capitalization of the existing `is.reactivevalues` method? (yuck)

R/reactives.R
```diff
@@ -682,18 +905,26 @@ srcrefToLabel <- function(srcref, defaultLabel) {
 }
 
 #' @export
+format.reactiveExpr <- function(x, ...) {
+  fun <- attr(x, "observable", exact = TRUE)$.origFunc
+  label <- sprintf('reactive(%s)', paste(deparse(body(fun)), collapse='\n'))
```
wch: Instead of reaching into the Observable object for `.origFunc` and then deparsing it, you could add a method to Observable that returns the text.

wch:
Looks good to me.

With the magic of hindsight I can see a bunch of things that I wish we had done differently, like generalizing `freezeReactiveValue` to an S3 generic like `freezeReactive` (or just `freeze`). Another thing we could have done is define a `$.reactive` method which would let you use `$` to call methods on the reactive expressions, as in `r$freeze()` (and similar for observers).

R/reactives.R
wch: Could you add a comment about that?

R/reactives.R
jcheng5: Interesting. `reactiveVal` is a function used to create a "reactive value" object; and to read from that object, you call it like a function with no arguments. Clear as mud.

R/reactives.R
jcheng5: Great idea

R/reactives.R
jcheng5: Do you think there are legitimate reasons for wanting to know the difference? I can see is.reactive being very important but not the others.

R/reactives.R
jcheng5: 👍

jcheng5:
@wch I guess it's not too late to define `$.reactive`. Also I generally like the idea of objects that you can also call as a function--in C++ they call a callable object a "functor" (although that term has a very different meaning elsewhere).

Maybe R6 could be extended to make this easy to do?

R/reactives.R
wch: One thing that someone might want to do is test if it's a `reactiveVal`, so that they can know if they can set it with something like `r(123)`.

On the other hand, the issues with capitalization and class names make me thing it might not be worth doing right now.

NEWS.md
```diff
@@ -7,6 +7,8 @@ shiny 1.0.0.9001
 
 ### New features
 
+* Added `reactiveVal` function, for storing a single value which can be (reactively) read and written. Similar to `reactiveValues`, except that `reactiveVal` just lets you store a single value instead of storing multiple values by name.
```
wch: This should have a link to the PR.

wch:
In the case of a reactive, we're talking about a function that you can also use `$` on. For R6, it would be the reverse: an object that you can also use `()` on. I don't think the latter is possible, though.

jcheng5:
For R6, I'm not saying it would _actually_ be an object you can use `()` on, but that you'd simulate that so it's all but indistinguishable to the user.

jcheng5:
But it would be really cool if `(` or whatever was a generic :)

PR #1633, opened 2017-03-29
Fix #1632: Showcase mode comes up almost blank in IE9 & 11

### LINKED ISSUE #1632: Showcase mode comes up almost blank in IE9 in some conditions
image

The problem seems to occur when the browser window is extremely wide. Resizing the window to be smaller fixes the problem; but then widening the window can bring the problem back.

jcheng5:
If the width is made very wide in showcase mode with side-by-side
arrangement, the app shrinks to almost nothing. For some reason the
zoom CSS property (which we set using jQuery.animate) is set to
"1%" instead of "1".

Numbers and percentages are equally valid here, and the issue goes
away if we use percentage.

### DIFF
--- a/inst/www/shared/shiny-showcase.js
+++ b/inst/www/shared/shiny-showcase.js
@@ -217,7 +217,7 @@
     var app = document.getElementById("showcase-app-container");
     $(app).animate({
         width: appWidth + "px",
-        zoom: zoom
+        zoom: (zoom*100) + "%"
       }, animate ? animateMs : 0);
   };
 


jcheng5:
@wch @bborgesr I don't have a strong opinion whether we should merge this before or after 1.0.1. But I think it would be good if the gallery were updated with this PR whenever we merge it (rather than waiting for the next release), I think that's the most likely place people will encounter showcase-mode apps in IE.

bborgesr:
I think we should merge it now, unless Winston has any objections... It's a super safe change, and then, Winston will have the correct version of shiny when he redeploys all the apps to the gallery

PR #1652, opened 2017-04-04
allow the `choices` argument in `checkboxGroupInput()` to be `NULL`

### LINKED ISSUE #1649: Empty list of choices checkboxGroupInput
I see that the latest release of shiny (V 1.0.1) change the way you are allowed to define a group of checkboxes. This change clashes with my use case of checkboxes and I just wanna check that my case has been considered at design.

I use checkboxes as a user controlled list where the user can add items (done with updateCheckboxGroupInput) and remove them again by checking and push a remove button. Think of the way you delete files in rstudio. This use case require me to initialize an empty checklist but the new release prevent me from doing that by throwing an error: "Please specify a non-empty vector for `choices`"

I find my example a valid case where you have to be able to initialize a checkbox group with choice = c(). What do you think? Has this use case been considered? 

Henrik

bborgesr:
So that we keep backward compatibility with Shiny < 1.0.1 

Fixes #1649.

### DIFF
--- a/R/input-checkboxgroup.R
+++ b/R/input-checkboxgroup.R
@@ -70,6 +70,9 @@
 checkboxGroupInput <- function(inputId, label, choices = NULL, selected = NULL,
   inline = FALSE, width = NULL, choiceNames = NULL, choiceValues = NULL) {
 
+  # keep backward compatibility with Shiny < 1.0.1 (see #1649)
+  if (all(is.null(c(choices, choiceNames, choiceValues)))) choices <- character(0)
+
   args <- normalizeChoicesArgs(choices, choiceNames, choiceValues)
 
   selected <- restoreInput(id = inputId, default = selected)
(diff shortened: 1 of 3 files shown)


R/input-checkboxgroup.R
```diff
@@ -70,6 +70,9 @@
 checkboxGroupInput <- function(inputId, label, choices = NULL, selected = NULL,
   inline = FALSE, width = NULL, choiceNames = NULL, choiceValues = NULL) {
 
+  # keep backward compatibility with Shiny < 1.0.1 (see #1649)
+  if (all(is.null(c(choices, choiceNames, choiceValues)))) choices <- character(0)
```
wch: I think this will work, but not in the way you're expecting. `is.null()` isn't vectorized. However, `c(NULL, NULL)` is just `NULL`, so it will evaluate to `TRUE` if all the items are NULL. The `all()` doesn't have any effect.

It would be more correct to use `vapply` and `is.null`.

PR #1685, opened 2017-04-28
Convert examples to single file apps

mine-cetinkaya-rundel:
- Convert all example apps to single file `app.R` file
- Make relevant updates to READMEs to match up with single file app structure
- Add color to plots (RStudio blue)
- Add extensive comments to app.R files and use consistent formatting of comments across examples
- Example specific edits:
    - In `04_mpg`: Show outliers by default, as opposed to hide, since this is more routine, use `pch = 19` for outlier so it's more easily visible
    - In `06_tabsets` and `08_html`: Don't name random data vector `data` (name `d` instead)
    - In `08_html`: Add headers to clarify what each output is and show only head of data to simplify app
    - In `09_upload`: Use `req()` to check for NULL entry, update comments to reflect single file upload, show only head of uploaded file by default, with an option to view all if requested
    - In `10_download`: `row.names = FALSE` in `write.csv()`
 - Added entry to NEWS.md briefly summarizing changes.


bborgesr:
@mine-cetinkaya-rundel, I skimmed through this and it looks great; I trust that you were able to run each example with no problems, right?

@wch, @jcheng5, anything holding this up or can we merge?

mine-cetinkaya-rundel:
Thanks, and yes I was able to run each one.

mine-cetinkaya-rundel:
Echoing @bborgesr's question from earlier: @wch, @jcheng5, anything holding this up?

I have a few updates to tutorial/articles that depend on the final code that goes in the package, so I wanted to make sure I have this bit reviewed first.

jcheng5:
I'm tempted to replace all the `shinyApp(ui = ui, server = server)` with `shinyApp(ui, server)`--what do you think? Other than that, LGTM (I only skimmed)

mine-cetinkaya-rundel:
@jcheng5 One could use arbitrary names for the `ui` and `server` functions, and then assign those to the `ui` and `server` arguments in the `shinyApp` function, right? I mean, this might not be best practice, but it seems doable. If so, it makes sense to leave them as is to me, to make the distinction between argument name and input.

mine-cetinkaya-rundel:
Alternatively leave as is for first couple examples, then shorten?

mine-cetinkaya-rundel:
Implemented @jcheng5's suggestion for examples 3-11 except for custom HTML example

PR #1719, opened 2017-05-26
fileInput JS: Allow uploading the same file.

### LINKED ISSUE #1508: overwrite file in fileInput
Today I started using fileInput. I noticed that I am only able to upload a file once. In other words, I am not able to overwrite the resulting data.frame by uploading a second file. I was wondering if the widget can be modified to allow for this situation.

Thank you

alandipert:
* Fixes #1508 for IE9 and other browsers.
* The same fix—clearing the file input's value after a successful upload—was applied in both the "IE8" (IE9) and non-IE8 code paths. ([StackOverflow](https://stackoverflow.com/a/22521275))


tools/README.md
```diff
@@ -84,7 +84,16 @@ One of the tasks concatenates all the .js files in `/srcjs` together into `/inst
 
 During development of Shiny's Javascript code, it's best to use `grunt watch` so that the minified file will get updated whenever you make changes the Javascript sources.
 
+#### Auto build and browser refresh
 
+An alternative to `grunt watch` is to use `entr` to trigger `grunt` when sources
```
wch: In our projects, the .md files generally don't use hard line breaks within a paragraph.

Also, this only works on Mac, right? It would be good to mention that here.

PR #1724, opened 2017-06-02
implement updateFileInput(); fixes #1721

### LINKED ISSUE #1721: [Idea] updateFileInput() - use a server-side file in place of an uploaded file, would be especially useful for testing
Somewhat to #1720 

When developing apps that require file uploads, I often find myself wishing I could programmatically set the value of a fileInput because I don't want to have to perform multiple clicks and find the right file every time I'm testing/developing. I do have a solution for myself to do that automatically, but it would be nice if shiny had a proper way to define the file. And extra points if it supported `NULL`.

daattali:
This PR attempts to add a `updateFileInput()` function. The main reason for adding this is to allow users to programmatically set the files of a file input. I tried to make it mimic other update functions as closely as possible.

Notes:

- do we need to pass the `session` param even though we can figure it out without the user passing it in? I kept it for consistency with other update functions, but wondering about this
- I'm using `do.call(rbind, rows)` which I know is not the fastest, but I don't think it's an inefficiency that will cause any problems
- I added a `for` attribute to the file input's label, so that the JS code for updating the label will work the same way as for other inputs. But a side effect of that is that now clicking the label triggers the file selection window. Is that ok?
- I didn't add update params for all the params of `fileInput()`. Only `label` and `value` are supported. 
- one "issue"  is that if the file input is set to `multiple = FALSE`, the update function will still allow you to set multiple files. But i found that this inconsistency is also the case for other inputs as well so maybe it's not too bad
- Since we cannot literally set the file on the client side, I had to tinker with the `input` object, and on the client side I tried to mimic a file selection. I found one difference between manually selecting a file and doing it with this function: the progress bar doesn't get shown if you update before ever selecting a file.

@jcheng5 @wch


daattali:
Example:

```
ui <- fluidPage(
  fileInput("file", "File", multiple = TRUE),
  actionButton("btn_1", "Select the shiny DESCRIPTION"),
  actionButton("btn_2", "Select the shiny NAMESPACE and DESCRIPTION"),
  tableOutput("table"), br(),
  "Contents of first file:",
  verbatimTextOutput("text")
)

  observeEvent(input$btn_1, {
    updateFileInput(session, "file", label = "label can change",
                    value = system.file("DESCRIPTION", package = "shiny"))
  })
  
  observeEvent(input$btn_2, {
    updateFileInput(session, "file", label = "yup it's changed",
                    value = c(system.file("NAMESPACE", package = "shiny"),
                              system.file("DESCRIPTION", package = "shiny")))
  })
  
  output$table <- renderTable({
    input$file
  })
  
  output$text <- renderPrint({
    if (is.null(input$file)) return()
    readLines(input$file$datapath[1])
  })
}

shinyApp(ui, server)
```

wch:
This is a tricky one to implement because `fileInput` does a lot more work than most inputs. 

Looking at this PR, on the server side, `updateFileInput` sets the input value directly.

* It doesn't call the usual [`@uploadEnd` method](https://github.com/rstudio/shiny/blob/379d523/R/shiny.R#L1490-L1497). I think the only functionality that would have to be duplicated is the `setMeta` part, though.
* It currently allows the user to point to files in an arbitrary place on disk. I think this shouldn't be allowed for various reasons - the files should be copied to the same place as regular file uploads, in the temp dir.

On the client side, the normal upload code path isn't being used, so `shiny:inputchanged` event isn't triggered (I think), and the progress bar isn't shown. The client-side code that normally decides what to display in the text box is also duplicated on the server side (it's the `value` that's sent).

Ideally we'd use the same code paths as much as possible for real file uploads and simulated ones with `updateFileInput`. It may require more extensive changes, but it will make things simpler and  make maintenance easier in the long run. Also please note that @alandipert is doing some work in the file input right now. Hopefully any code conflicts will be minimal.

Sorry I can't give more detailed suggestions right now, but hopefully this is helpful.

daattali:
Thanks @wch these are all great points, very useful. Will fix

daattali:
@wch there are some conflicts in the minified shiny.js file, I'll handle them once the commits are LGTMed. Can you look at my fixes since your comment and see if there are any other issues. Your comments were really useful.

This also has the side effect of fixing #1729

wch:
After some discussion with @barbara and @alandipert, I think we have another way which would reuse more existing code and reduce code duplication even more. Sorry for not realizing and mentioning this earlier. I could probably do an implementation of this based on your code, but it likely would not be until next week.

Currently, your version sets the value on the server, and it sets a value on the client, but those two things are essentially independent. It would be simpler to use the existing bookmarking restore code with some small modifications.

File inputs are different from other inputs in that they don't set a value on the client. So, for example, when you upload a file to an input named `x`, the value of `Shiny.shinyapp.$inputValues.x` on the client doesn't change, but the value of `input$x` on the server does change. However, when restoring a bookmarked file, the value of `Shiny.shinyapp.$inputValues.x` *does* get set, and then it passes through the `shiny.file` handler on the server side:
  https://github.com/rstudio/shiny/blob/379d523/R/server-input-handlers.R#L189-L220

This code path could be reused for `updateFileInput` as well. The idea is that `updateFileInput` would do the following:
* Copy the file on disk to a new temp dir (for security reasons).
* Send the file info data frame to the client.

For a first pass of the client code, you could to implement a `receiveMessage` method that:
* Sets the `data-restore` attribute to the file info data.
* Triggers a `change` event.

You'd also have to modify the `shiny.file` input handler to look in the correct temp dir, instead of the restore context's temp dir, and do it in a way that doesn't cause security problems (like letting the client look at arbitrary dirs/files on the server). Right now it's not 100% clear to me how best to do that, but it should be doable:
  https://github.com/rstudio/shiny/blob/379d523/R/server-input-handlers.R#L205

If that works on the client, then I think the next step would be to modify it so `receiveMessage` doesn't actually set the `data-restore` attribute, but instead calls a `setValue` method with the data. (The bookmark restore code would also call `setValue`.)

It would be important to still make sure the JS event gets triggered when `updateFileInput` is used (but not when a file is restored from bookmarking) and that the progress bar gets updated. It should be possible to use the code you've already written for this.

If you want to talk over video about this, feel free to get in touch directly.

daattali:
I think I'll let it go and let you guys handle it then. I'm embarrassed that it completely escaped my mind that bookmarking essentially has this feature already!

I doubt you'll actually end up using my code, but just for completeness I want to mention a few outstanding items I didn't address:
- The MIME type should default to `""` (`mime::guess_type(..., unknown = "", empty = ""))`
- The size should be integer instead of numeric, to be consistent with the regular file upload
- Need to clean up the created folders (`session[['.__enclos_env__']]$private$fileUploadContext[['.__enclos_env__']]$private$ids`)

wch:
I'm closing this now -- we'll refer back to this PR when we implement it on our side.

PR #1735, opened 2017-06-08
Support modules in conditionalPanel

alandipert:


alandipert:
Here is [Winston's reproduction of the problem](https://github.com/rstudio/shiny/issues/1586#issuecomment-305822900) modified to work using the new `ns` argument of `conditionalPanel`:

```R
library(shiny)

condpanelUI <- function(id) {
  ns <- NS(id)

  tagList(
    checkboxInput(ns("checkbox"), "Make panel visible"),
    conditionalPanel(condition = "input.checkbox",
                     ns = ns,
                     "This is the conditional panel")
  )
}

ui <- fluidPage(
  condpanelUI("foo")
)

}

shinyApp(ui, server)
```

srcjs/shinyapp.js
```diff
@@ -375,6 +375,18 @@ var ShinyApp = function() {
     }
   };
 
+
+  // Narrows a scopeComponent -- an input or output object -- to one constrained
+  // by nsPrefix. Returns a new object with keys removed and renamed as
+  // necessary.
+  function narrowScope(scopeComponent, nsPrefix) {
+    return mapKeys(pickBy(scopeComponent, function(val, key) {
```
jcheng5: 1. Feel free to use arrow functions
2. I found the indentation a little deceptive somehow, i.e. hard to tell whether the second closure belongs to mapKeys or pickBy
3. Can you make it so the lodash-alike functions are actually accessed via a `_` object? So we can swap them out in the future and also to make it obvious that the semantics of these functions must not be changed (unless it's to match lodash even more closely).

srcjs/shinyapp.js
```diff
@@ -404,7 +416,15 @@ var ShinyApp = function() {
         el.data('data-display-if-func', condFunc);
       }
 
-      var show = condFunc(scope);
+      // The data-ns-prefix attribute is always present, but might be empty. If
+      // it's empty, no keys are removed or renamed by narrowScope.
+      var nsPrefix = el.attr('data-ns-prefix');
```
jcheng5: Can you check for null/undefined here actually? It's possible for existing HTML templates to have conditional panels hard coded. Our support for such code is not clearly defined but in this case it's so trivial to not break them.

srcjs/shinyapp.js
jcheng5: The common case by far will be no prefix, exit early in that case?

srcjs/utils.js
```diff
@@ -246,3 +246,27 @@ function mapValues(obj, f) {
   }
   return newObj;
 }
+
+// Creates an object composed of the object properties predicate returns truthy
+// for. The predicate is invoked with two arguments: (value, key). Like the
+// pickBy function from lodash.
+function pickBy(obj, predicate) {
+  const newObj = {};
+  for (let key in obj) {
```
jcheng5: @wch Can we use Object.keys? I can't remember.

jcheng5:
Shaping up nicely! It will be a real relief to me when we get this merged!

srcjs/utils.js
wch: Yup, that should be safe.

alandipert:
Introducing Lodash, even with just the functions we use, increase the size of `shiny.min.js.gz` from 21kb to 33kb. @wch does not believe the cleaner code justifies the extra code size, and I'm on the fence.

jcheng5:
I agree with @wch fwiw...

R/bootstrap.R
```diff
@@ -534,8 +536,8 @@ mainPanel <- function(..., width = 8) {
 #'    )
 #' )
 #' @export
-conditionalPanel <- function(condition, ...) {
-  div('data-display-if'=condition, ...)
+conditionalPanel <- function(condition, ..., ns = NS(NULL)) {
+  div('data-display-if'=condition, 'data-ns-prefix'=ns(""), ...)
```
wch: Nit: backticks here would be more idiomatic R.

srcjs/shinyapp.js
```diff
@@ -375,6 +375,17 @@ var ShinyApp = function() {
...
 
+
+  // Narrows a scopeComponent -- an input or output object -- to one constrained
+  // by nsPrefix. Returns a new object with keys removed and renamed as
+  // necessary.
+  function narrowScope(scopeComponent, nsPrefix) {
+    return Object.keys(scopeComponent)
+      .filter(k => k.startsWith(nsPrefix))
+      .map(k => ({[k.substring(nsPrefix.length)]: scopeComponent[k]}))
+      .reduce((obj, pair) => Object.assign(obj, pair), {});
```
wch: `Object.assign` isn't supported by IE at all, and it's not polyfilled, so we should use `$.extend()` instead.

alandipert:
I *think* this is now in an IE9-compatible state, but I still need to test and confirm. Will report back soon.

wch:
Looks good to me. If it works correctly on IE9, then I think it can be merged.

alandipert:
@wch can confirm: works on IE9. Branch and small demo deployed here: https://beta.rstudioconnect.com/content/2858/

jcheng5:
🎉

PR #1449, opened 2016-10-29
Fix #1438: `unbindAll()` should not be called when inserting content with `insertUI()`

bborgesr:
It seems like this one was pretty painless!

PR #1736, opened 2017-06-08
Fix condition for calling exports.unbindAll()

wch:
This is a correction to #1449. That change had no effect on behavior.

With the example app from rstudio/DT#369, if you select rows in the first datatable, and then add another table by moving the slider, the selected rows get unselected, with the current master.

```R
library(shiny)
library(DT)

numUI <- 0

ui <- shinyUI(fluidPage(
  mainPanel(
    sliderInput("number","Number of tables",1,10,1,1),
    tags$div(id="tables")
  )))

server <- shinyServer(function(input, output, session) {

  observe({
    if (input$number > numUI) {
      for (num in (numUI+1):input$number) {
        insertUI("#tables", "beforeEnd", DT::dataTableOutput(paste0("table", num)))
        output[[paste0("table",num)]] <- DT::renderDataTable(head(mtcars), server = FALSE)
      }
    }

    if (input$number < numUI) {
      for (num in (input$number+1):numUI) {
        removeUI(paste0("#table", num))
      }
    }

    numUI <<- input$number
  })

})
# Run the application
```

After this change, the rows do not get unselected when adding a table.


wch:
This also should have some tests with shinytest, possibly in a new repo added with a git submodule or subtree.

bborgesr:
Here's a simpler test app:

```r
ui <- fluidPage(
  actionButton("add", "Add slider"),
  tags$div(id = "sliders")
)

  observeEvent(input$add, {
    containerID <- paste0("ui-", input$add)
    sliderID <- paste0("slider-", input$add)
    insertUI("#sliders", "beforeEnd", uiOutput(containerID))
    output[[containerID]] <- renderUI(sliderInput(sliderID, sliderID, 0, 10, 5))
  }, ignoreInit = TRUE)
}

shinyApp(ui, server)
```

I have started a test repo ([rstudio/shinyTestResults](https://github.com/rstudio/shinyTestResults)), but neither this app nor the first one (using DT) can be tested properly using shinytest (see rstudio/shinytest#105 and rstudio/shinytest#106). So, I think we should merge for now while Winston works on those shinytest edge cases (since the problem is with shinytest, not with this PR). It is clear that the PR works if we test manually.

PR #1741, opened 2017-06-13
Enable CORS headers

drewlanenga:
Currently set via environment variables.  Maybe a little sketchy, but also unobtrusive.


drewlanenga:
Sorry, had the wrong upstream set. :(

PR #1743, opened 2017-06-14
Expose a 'shiny:invalidated' event to resolve #1688

### LINKED ISSUE #1688: Expose a shiny:invalidated JS event (as shiny:recalculating can fire much later than invalidation)
Hi,

It appears that there is a mismatch between when the `shiny:recalculating` event is triggered vs when the `.recalculating` class is added to a shiny output. In particular the `.recalculating` class is added as soon as the output is invalidated, whereas the JS event is only triggered when the actual observer corresponding to the output starts running. If there are multiple outputs depending on a long-running reactive, this can cause a large gap between when the corresponding JS event is fired.

I am not sure if this deliberate, however the JS event behaviour seems not very intuitive and would expect `shiny:recalculating` to be triggered as soon as the output is invalidated.

The below example shows that there is 5s difference between the two outputs receiving the `shiny:recalculating` event as shown in the JS console, even though they depend on the same reactive. The `.recalculating` class is applied at the same time however.

```
library(shiny)

ui <- function() {
  fluidPage(
    shiny::singleton(
      tags$head(
        tags$script(
          shiny::HTML("
$(document).on('shiny:recalculating', function(event) {
  console.log('Recalc for %s running at %s',event.target.id,(new Date));
});
           "))
      )
    ),
    actionButton("recalc","recalc"),
    plotOutput("plot_1"),
    plotOutput("plot_2")
  )
}

  plot_r <- reactive({
    input$recalc
    Sys.sleep(5) 
  })
  output$plot_1 <- renderPlot({
    plot_r()
    plot(
      x = runif(1e4), y = runif(1e4)
    )
  })
  output$plot_2 <- renderPlot({
    plot_r()
    plot(
      x = runif(1e4), y = runif(1e4)
    )
  })
}

```

andrewsali:
This PR adds triggering of `shiny:invalidated` event when an output gets invalidated, thereby resolving #1688.

The rationale is that it's hard to capture via jQuery events when an output gets invalidated, which might be __much__ sooner than recalculation starts (and thus `shiny:recalculating` is triggered).

The modification is tiny, it's essentially one extra call added to trigger the event after binding.showProgress(true) is called.


wch:
This has been implemented in #1758.

PR #1747, opened 2017-06-16
In test mode, send message to client even when no outputs change

wch:
Normally, if an input is changed but no outputs change, the Shiny process will not send anything to the client in response. With this change, when `shiny.testmode` is TRUE, then Shiny will send a response to the client in response to an input changing, even if no outputs change.

This makes it possible for shinytest to detect when an input has been received and processed by the server, even if no outputs change. (Shinytest detects this by listening on the client for a message that contains a `values` field.)

Here is a minimal sample app:

```R
options(shiny.trace=TRUE)
shinyApp(
  fluidPage(sliderInput("test", "Test", 0, 10, 5)),
  function(input, output, session) {}
)
```

Normally, when running this app, if you move the slider multiple times, the R console prints this:

```
RECV {"method":"update","data":{"test":6}}
RECV {"method":"update","data":{"test":7}}
RECV {"method":"update","data":{"test":8}}
```

In test mode (with `options(shiny.testmode=TRUE)`), it prints this:

```
RECV {"method":"update","data":{"test":6}}
SEND {"values":[]}
RECV {"method":"update","data":{"test":7}}
SEND {"values":[]}
RECV {"method":"update","data":{"test":8}}
SEND {"values":[]}
```

This change will make (apparent) issues like rstudio/shinytest#106 much less common, and simplify the procedure of creating tests, because less manual intervention will be needed.


bborgesr:
LGTM. I think this will be a super elegant solution to those scenarios in shinytest!

PR #1758, opened 2017-06-21
Triggers a new `shiny:outputinvalidated` event

### LINKED ISSUE #1688: Expose a shiny:invalidated JS event (as shiny:recalculating can fire much later than invalidation)

bborgesr:
When an output gets invalidated (at the same time that the `recalculating` CSS class is added to the corresponding html object), trigger a new JS event called `shiny:outputinvalidated`. (fixes #1688)

This is a slight adaptation of #1743, so thank you to @andrewsali for his contribution! I also added the `binding` and the `name` (`key`) to the event.

### Repro (adapted from #1688)
Open the JS console to see the print statements and their timing.

```r
options(shiny.launch.browser = F, shiny.minified = F, shiny.port = 6347)
library(shiny)

ui <- function() {
  fluidPage(
    shiny::singleton(tags$head(tags$script(HTML("
      $(document).on('shiny:recalculating', function(event) {
        console.log('Recalc for %s running at %s',event.target.id,(new Date));
      });
      $(document).on('shiny:outputinvalidated', function(event) {
        console.log('Inval for %s running at %s',event.target.id,(new Date));
      });
    ")))),
    actionButton("recalc", "recalc"),
    plotOutput("plot_1"),
    plotOutput("plot_2")
    )
  }

  data <- reactive({
    input$recalc
    Sys.sleep(1)
    cars
  })

  output$plot_1 <- renderPlot({ plot(data()) })
  output$plot_2 <- renderPlot({ plot(data()) })
}

shinyApp(ui, server)
```


jcheng5:
LGTM

PR #1770, opened 2017-06-27
[WIP] onAppExit functionality

### LINKED ISSUE #1213: Feature request: `on.exit` function equivalent for Shiny apps
This may be very handy (for example, to disconnect a database connection when exiting the app).

bborgesr:
Fixes #1213 and rstudio/pool#38.


bborgesr:
This is ready for a code review.

Catching up @alandipert and @jcheng5 after meeting with @wch: it turns out we already had this mechanism (called `onEnd` and used internally). It's still pretty simple, but now we know it's battle-tested. So all this PR does it is expose it (and ensure it's compatible with all app structures). Considering this, I'm hoping we can get this merged soonish (next week?)

I've also tested it with pool and it works great :)

Since I was at it, I also fixed #1772 which Winston and I discovered when we talked.

_This should be squash-merged_

cc @daattali 

----

Repro:

```r
# devtools::install_github("rstudio/shiny@barbara/on-app-exit")

library(shiny)

ui <- fluidPage()
server <- function(input, output, session) {}
onStop(function() { print("bye")})
shinyApp(ui, server)

# or alternatively:
# shinyApp(ui, server, onStop = function() { print("bye") })
```

R/app.R
```diff
@@ -212,7 +217,10 @@ shinyAppDir_serverR <- function(appDir, options=list()) {
     if (file.exists(file.path.ci(appDir, "global.R")))
       sourceUTF8(file.path.ci(appDir, "global.R"))
   }
-  onEnd <- function() {
+  onStop <- function() {
+    onStopFunc <- getShinyOption("onStop", default = function() {})
+    if (!is(onStopFunc, "function")) warning("`onStop` must be a function")
+    else onStopFunc()
```
jcheng5: The logic below must run even if onStopFunc errors

NEWS.md
```diff
@@ -6,9 +6,12 @@ shiny 1.0.3.9001
 ### Breaking changes
 
 ### New features
+* Added an `onStop` argument to `shinyApp()`, which can be used to register a function callback to be run after the app has finished running (all sessions have been disconnected). This is useful if you have finalization/cleaning up code that should be run after the app is closed. If you are creating an app interactively at the R console, this would be the equivalent of code that comes immediately after the `shinyApp()` call. We've also added an `onStop()` function (whose only argument must be the callback function), which has the same effect. This is only necessary for apps that use the two-file format (ui.R and server.R), but it has the same effect regardless of your app structure. In case that you supply both an `onStop()` function _and_ an `onStop` argument to `shinyApp()`, the latter will be the one respected. ([#1770](https://github.com/rstudio/shiny/pull/1770)
```
jcheng5: "all sessions have been disconnected" is a bit misleading? Since all sessions can disconnect without the app stopping.

R/app.R
```diff
@@ -496,3 +506,17 @@ knit_print.reactive <- function(x, ..., inline = FALSE) {
...
+#' This function allows you to register a callback function that is invoked
+#' after your app has finished running (all sessions have been disconnected).
+#'
+#' @param fun A function that will be called after the app has finished running.
+#'
+#' @examples
+#' onStop(function() { print("goodbye") })
+#' @export
+onStop <- function(fun) {
+  shinyOptions(onStop = fun)
```
jcheng5: I don't think we want to use this model, where if you call onStop multiple times, the last one wins. Instead we should use a Callback(? I'm on my phone) model so that multiple calls will all be registered and executed, and they can be individually cancelled as well. Like we do for onFlushed. However, one thing I'm not sure we do for onFlushed is execute the callbacks in the reverse order of their registration--this is important for cleanup scenarios (e.g. imagine creating a pool that needs to be cleaned up, and then checking out a connection from that pool; you'd want to clean up the connection and then the pool).

bborgesr: I liked the fact that this mirrors `onStart` and it stays simple, yet powerful. I also wonder if allowing multiple `onStop` calls won't make it easy for app authors to use it for session-specific things (i.e. calling `onStop` in reactives or observers, since it wouldn't always fail -- I guess we could enforce this if we wanted to). Having just one callback possible also both forces the app author to think about order and shifts that burden to them. 

I guess that what I'm getting at is why multiple `onStop` calls would be useful if this is a global thing, rather than session-specific. And I'm pretty sure there's one answer to rule them all: modules, right?

Assuming we go your way, is the best way to implement this to use a list in `.globals`? What's the API you imagine for this + cancellation of individual callbacks? If we didn't have cancellation, it could stay the same, except that multiple calls would result in multiple callbacks, instead of the last one winning out. With cancellation in mind, the API that makes most sense to me is a la `scheduleTask()` in pool. Applied to this, the return value of an `onStop` call would a function that cancels the callback originally registered:

```r
cancel <- onStop(function(){}) # this registers the callback
cancel() # this cancels the callback
```

R/app.R
bborgesr: It's enough to just wrap `onStopFunc()` inside a `try()` call (here and in the other place), right? I did that just now and it seems to work fine (I tested by looking at the value of `getwd()`)

NEWS.md
bborgesr: That's true... By "all sessions have been disconnected," I just wanted to make it clear that the scope here is global and not session-level. I felt like "after the app has finished running" part alone could easily be misunderstood as a session ending, especially because people often don't even realize these are different at all. Do you think there's a better way of saying that or we should we just leave the parens out and keep the rest the same?

R/app.R
alandipert: @bborgesr As far as I can tell, one principle utility of multiple `onStop` calls is to support the presence of (multiple) libraries that manage resources. Being able to register multiple `onStop` calls makes it possible for users' applications to use multiple libraries, each of which might manage a resource like a database or connection, without the user needing to necessarily install an `onStop` function that calls of the appropriate shutdown/cleanup APIs for each library. The libraries can register to perform this cleanup on their own.

R/app.R
```diff
@@ -496,3 +506,17 @@ knit_print.reactive <- function(x, ..., inline = FALSE) {
...
   }), inline = inline)
 }
+
+#' Run code after app has finished
+#'
+#' This function allows you to register a callback function that is invoked
+#' after your app has finished running.
+#'
+#' @param fun A function that will be called after the app has finished running.
+#'
```
wch: It would help to have a link to `onSessionEnded`, and vice versa.

R/app.R
```diff
@@ -212,7 +217,10 @@ shinyAppDir_serverR <- function(appDir, options=list()) {
     if (file.exists(file.path.ci(appDir, "global.R")))
       sourceUTF8(file.path.ci(appDir, "global.R"))
   }
-  onEnd <- function() {
+  onStop <- function() {
+    onStopFunc <- getShinyOption("onStop", default = function() {})
+    if (!is(onStopFunc, "function")) warning("`onStop` must be a function")
```
wch: Using `is.function` would be more idiomatic. But it probably makes more sense to check the type of the object when it's registered, either in `onStop` or in `Callbacks$register`.

jcheng5:
I wouldn't use `try()`, we don't catch errors from callbacks usually. Use `on.exit` instead, or alternately, `tryCatch`/`finally` to make sure we do the `setwd` and monitor cleanup while still letting the exception propagate.

PR #1782, opened 2017-07-04
drag/drop-able fileInputs

alandipert:
## TODO
- [x] Correctly handle states and events
- [x] Allow uploading multiple files
- [x] Make glow size and colors more subtle and Bootstrap-esque
- [x] ~~IE9~~: Fix `'Symbol' is undefined`, figure out and remove whatever in the code requires polyfill
  - was `for..of`
  - now "active", "over" styles work. "drop" does not (yet)
- [x] ~~IE9~~: `drop` broken
  - It's not possible for IE9 to work because it doesn't support FileList, which is the only way we could get a dropped file.
- [x] IE10/11/Edge: `SCRIPT5045: Assignment to read-only properties is not allowed in strict mode`
  - Offending code is `el.files = e.originalEvent.dataTransfer.files;`
- [x] Confirm how to configure style (just color? argument to `fileInput`? Set of classes?)
  - Using a settable class
- [x] Validate that `multimethod` approach is worthwhile
  - I showed it to Joe and he didn't get mad
- [x] Fix dragenter/leave handlers on document in IE10+
- [x] Add workaround (mousing in to page) that resets dropzone highlighting on IE10/11
- [x] Confirm `unsubscribe` works property with dynamic UI
  - Tested with `conditionalPanel`
- [x]  Customize via class
  - ~~`:hover` pseudoclass is how "over" style is conveyed, vs. 2nd class~~
    - Can't work because `dragover` doesn't trigger `:hover`
  - defaults to `shiny-file-input-active`, `shiny-file-input-over` classes we own
  - Customization is via overriding classes in user CSS, not by specifying classnames in R.
- [x] Investigate unit test harness (for `multimethod`)
- [x] Finalize styles
  - Demo: https://beta.rstudioconnect.com/content/2921/
  - Demo has a text area in which CSS can be live-edited and applied to the page.


alandipert:
I thought a little about how this could interoperate with the "one big dropzone" idea that Tareef had and that we've discussed a bit.

I think the two ideas coexist nicely, because there is an opportunity in this code to handle a drop event that did not occur over a `fileInput`. So, it should work fine, even in the presence of `fileInput`s.

If there are no `fileInputs` on the page, we'd need to install a separate global handler. When the first `fileInput` is created and installs its global handlers, we'd need to check and account for the presence of a pre-existing global handler (if we decide "one big dropzone" should be optional).

srcjs/input_binding_fileinput.js
```diff
@@ -181,6 +181,31 @@ $.extend(FileUploader.prototype, FileProcessor.prototype);
 }).call(FileUploader.prototype);
 
 
+function uploadDroppedFilesIE10Plus(el, files) {
```
wch: This function looks like it largely duplicates code from `uploadFiles`. Could you refactor the functions so that the code is shared?

srcjs/input_binding_fileinput.js
```diff
@@ -266,11 +293,169 @@ $.extend(fileInputBinding, {
...
   unsubscribe: function(el) {
-    $(el).off('.fileInputBinding');
+    let $el   = $(el),
+        $zone = this.getZone(el);
+
+    $el.removeData("state");
+
+    // Clean up local event handlers.
+    $el.off(".fileInputBinding");
+    $zone.off(this.zoneEvents);
```
wch: Does `$zone`'s CSS also need to be cleared here?

srcjs/input_binding_fileinput.js
```diff
@@ -266,11 +293,169 @@ $.extend(fileInputBinding, {
...
+      el.files = e.originalEvent.dataTransfer.files;
+    }
+  },
   subscribe: function(el, callback) {
-    $(el).on('change.fileInputBinding', uploadFiles);
+    let $el        = $(el),
+        $zone      = this.getZone(el),
+        getState   = () => $el.data("state"),
+        setState   = (newState) => $el.data("state", newState),
+        transition = multimethod()
```
wch: I think it would be helpful to have a short description of the state machine, and how the JS events and state transitions are wired together.

srcjs/utils.js
```diff
@@ -246,3 +246,230 @@ function mapValues(obj, f) {
   }
   return newObj;
 }
+
+// Binary equality function used by the equal function.
+function _equal(x, y) {
+  if ($.type(x) === "object" && $.type(y) === "object") {
```
wch: Am I correct that you're using `$.type()` because there's not a good plain-JS way of checking if an object is a regular object (and not, for example, a Date object)?

srcjs/utils.js
```diff
@@ -246,3 +246,230 @@ function mapValues(obj, f) {
...
+//    var handleEvent = multimethod()
+//     .dispatch(e => [e.target.tagName.toLowerCase(), e.type])
+//     .when(["h1", "click"], e => "you clicked on an h1")
+//     .when(["p", "mouseover"], e => "you moused over a p"})
+//     .else(e => {
+//       let tag = e.target.tagName.toLowerCase();
+//       return `you did ${e.type} to an ${tag}`;
+//     });
+//
+//    $(document).on("click mouseover mouseup mousedown", e => console.log(handle(e)))
```
wch: Should it be `handleEvent(e)` instead of `handle(e)`?

srcjs/utils.js
```diff
@@ -246,3 +246,230 @@ function mapValues(obj, f) {
...
+    if (!trampolining) throw new Error("recur can only be called inside a method");
+    return new Sentinel(args);
+  };
+
+  invoke.dispatch = (newDispatch = (firstArg) => firstArg) => {
+    dispatch = newDispatch;
+    return invoke;
+  };
+
+  invoke.test = (newTest = _equal) => {
```
wch: Should this be `equal` instead of `_equal`, for consistency? Also, it seems to me that a default value isn't really necessary here, or for `invoke.dispatch`.

srcjs/utils.js
alandipert: @wch That's correct.

PR #1785, opened 2017-07-07
Add a function to test if the app is running

klmr:
Some time ago [I asked how to test whether Shiny is running](https://groups.google.com/forum/#!topic/shiny-discuss/cGQwYkGpDRo/discussion). I found a preliminary solution but [there are cases where it breaks](https://github.com/klmr/modules/issues/69). The solution is also generally quite hacky and brittle.

I noticed that adding this functionality to Shiny should be trivial; the actual implementation is a single line of code that just inspects `.global$running`. I therefore hope that this PR is straightforward.

---

I didn’t add a unit test since it feels that an integration test would be more appropriate, and I couldn’t find any other unit tests for runtime Shiny behaviour in the code base. I did run `devtools::check()` to verify that the change didn’t break anything obvious. I also added documentation, filled the `NEWS.md`, and signed the contributor agreement.


wch:
Thanks! I made some edits and squash-merged in 4cb428bb.

PR #1786, opened 2017-07-07
Use a more suitable content type for file uploads

jcheng5:
application/x-www-form-urlencoded;charset=UTF-8 is the default, which shinyapps.io
cares about for some reason and tries to parse the data as such. By setting the
content type to the more accurate application/octet-stream, no middleware should
be tempted to futz with the contents.


wch:
LGTM

PR #1790, opened 2017-07-12
Don't reduce warn level when running app

### LINKED ISSUE #1680: options(warn=2) is not respected
In R/server.R we blindly call options(warn=1) [here](https://github.com/rstudio/shiny/blob/master/R/server.R#L580). We shouldn't do this if warn > 1.

wch:
This fixes #1680.

Prior to this change, this app would print out a warning and keep running. After the change, it stops and gives the call stack frame selector:

```R
library(shiny)
options(shiny.error = recover, warn = 2)
shinyApp(
  fluidPage(),
  function(input, output) {
    observe({
      warning("foo")
    })
  }
)
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -45,6 +45,8 @@ shiny 1.0.3.9001
 
 * Fixed [#1763](https://github.com/rstudio/shiny/issues/1763): Shiny's private random stream leaked out into the main random stream. ([#1768](https://github.com/rstudio/shiny/pull/1768))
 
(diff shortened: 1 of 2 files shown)


jcheng5:
This permanently sets the `warn` though, right? The `ops` only contains options that were actually set.

I happened to fix this yesterday locally to debug a problem I was having, I used this:

```r
ops <- options(warn = max(1, getOption("warn", 1)), pool.scheduler = scheduleTask)
```

wch:
Oops, good catch. Your solution is better, so I'll change it to that.

PR #1792, opened 2017-07-13
Add 'serializer' function

wch:
Previously, if the author of an input wanted to add a serializer function, they would have to do something like this:

```
session <- getDefaultReactiveDomain()
.subset2(session$input, "impl")$setMeta(name, "shiny.serializer", myInputSerializer)
```

With this change, they can do:

```
serializer(name, myInputSerializer)
```

The interface is similar to the recently-added [`snapshotPreprocessInput`](https://github.com/rstudio/shiny/blob/86e9cc4896630f06f5fba3aaf19640fd843d6ce6/R/snapshot.R#L30) function.


R/serializers.R
```diff
@@ -1,3 +1,21 @@
+#' Add a function for serializing an input before bookmarking application state
+#'
+#' @param inputId Name of the input value.
+#' @param fun A function that takes the input value and returns a modified
+#'   value. The returned value will be used for the test snapshot.
+#' @param session A Shiny session object.
```
jcheng5: Could this use an example? Or are you thinking this is so advanced that nobody is really going to need it? (If so maybe `@keyword internal` or whatever that thing is?)

R/serializers.R
```diff
@@ -1,3 +1,21 @@
...
+#'
+#' @export
+serializer <- function(inputId, fun, session = getDefaultReactiveDomain()) {
```
jcheng5: How would you feel about `session$setSerializer(inputId, fun)` instead? Or at least renaming to `setSerializer`? This looked more to me like a constructor at first, something that makes a serializer out of a function; rather than the side-effecty, accessor-y operation that it actually is.

jcheng5:
Other than the nits above, LGTM.

R/serializers.R
wch: Yeah, `@keywords internal` probably makes sense for this.

R/serializers.R
wch: `setSerializer` sounds good. I also felt like the name wasn't quite right.

PR #1820, opened 2017-08-18
Escape newline characters in conditionalPanel expression

### LINKED ISSUE #1818: conditionalPanel error when there is a newline in expression (was: breaking changes in 1.0.4?)
It seems Shiny 1.0.4 introduces some breaking changes compared to 1.0.3. The issue was first reported to me here: https://github.com/radiant-rstats/radiant/issues/25

The app (link below) works fine with 1.0.3 but the UI is broken by changes in 1.0.4. i.e., it looks like conditional panel just stopped working. I don't see any messages or errors so I'm not even sure where to start looking.

https://github.com/radiant-rstats/radiant.data

In the screen shot below all UI components are shown, including inputs that should only be shown if file-type csv is selected.

Shiny 1.0.4:

image

Here is what the ui *should* look like if rds is selected. This works as intended in 1.0.3

Shiny 1.0.3:

image

wch:
This fixes #1818.

Test app:

```R
shinyApp(
  fluidPage(
    selectInput("select", "Select",
      c(First = "first", Second = "second")
    ),
    conditionalPanel(
      condition = "input.select ==
                   'second'",
      "This should be displayed only when 'Second' is selected."
    )
  ),
  function(input, output) { }
)
```


jcheng5:
@wch Can you imagine any of these other escape sequences being needed to avoid these problems? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Grammar_and_types#Using_special_characters_in_strings

wch:
@jcheng5 I think there needs to be a couple more characters added to the escaping, namely `\r` and `\b`. Here's an example:

```js
# This string contains the weird characters listed at
# https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Grammar_and_types#Using_special_characters_in_strings
# including one Unicode character.
var expr =  '\0\b\f\n\t\r\v\'"\\\u5434';

var expr_escaped = expr
    .replace(/[\\"']/g, '\\$&')
    .replace(/\u0000/g, '\\0')
    .replace(/\n/g, '\\n')
    .replace(/\r/g, '\\r')
    // \b has a special meaning; need [\b] to match backspace char.
    .replace(/[\b]/g, '\\b');

eval(`"${expr_escaped}"`);
```

Without the added substitutions for `\r` and `\b`, the `eval()` gives an error. I can implement this if it looks good to you.

wch:
I've updated it in e74592a6.

PR #1822, opened 2017-08-21
Add safe wrapper for fromJSON

wch:
Previously, if the text passed to `jsonlite::fromJSON` contained a file or URL, it would attempt to read the file or fetch the URL. This is almost never what we want in Shiny, so the `safeFromJSON` function makes sure the input is valid JSON before passing it along to `jsonlite::fromJSON`.


jcheng5:
LGTM. Feel free to merge after CI checks.

PR #1825, opened 2017-08-22
Fix #1824: HEAD request on static files causes app to stop

### LINKED ISSUE #1824: HEAD request on static files causes app to stop
[crashtest.tar.gz](https://github.com/rstudio/shiny/files/1243100/crashtest.tar.gz)

Run the attached app, then hit it with curl:

```
curl -I http://localhost:PORT/foo.doc
```

jcheng5:
The problem was that for HEAD requests specifically, we implement
an explicit Content-Length header (normally we let httpuv figure
out the Content-Length based on the content, but for HEAD we don't
return any content but still want to include the Content-Length).

The Content-Length header was only implemented correctly for string
values, not for raw vectors or file-by-path. This change implements
the value correctly for all currently valid httpuv content.

### DIFF
--- a/R/middleware.R
+++ b/R/middleware.R
@@ -360,7 +360,9 @@ HandlerManager <- R6Class("HandlerManager",
 
           response <- filter(req, response)
           if (head_request) {
-            headers$`Content-Length` <- nchar(response$content, type = "bytes")
+
+            headers$`Content-Length` <- getResponseContentLength(response, deleteOwnedContent = TRUE)
+
             return(list(
               status = response$status,
               body = "",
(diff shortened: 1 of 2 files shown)


jcheng5:
@wch, I didn't include code to only unlink if in a temp dir. It's clear from the `unlink` docs that if `recursive=FALSE` it won't try to remove a directory, not even an empty one; I thought that was good enough. Let me know if you disagree.

R/middleware.R
```diff
@@ -383,6 +385,30 @@ HandlerManager <- R6Class("HandlerManager",
...
+    if (deleteOwnedContent && isTRUE(response$content$owned)) {
+      on.exit(unlink(response$content$file, recursive = FALSE, force = FALSE), add = TRUE)
+    }
+    file.info(response$content$file)$size
+  } else {
+    warning("HEAD request for unexpected content class ", class(response$content)[[1]])
+    NULL
+  }
+
+  if (is.na(result)) {
```
wch: When would the `result` be `NA`? The cases I can think of are an `NA` string input, and when `file.info()` is called on a file that doesn't exist. A comment about this would be helpful so we remember.

R/middleware.R
```diff
@@ -383,6 +385,30 @@ HandlerManager <- R6Class("HandlerManager",
   )
 )
 
+getResponseContentLength <- function(response, deleteOwnedContent) {
+  force(deleteOwnedContent)
+
+  result <- if (is.character(response$content)) {
```
wch: It looks like you're trying to be strict about checking the input, and so it might be good to check that the length is 1.

PR #1826, opened 2017-08-23
Add Shiny.version to Javascript

wch:
This makes the Shiny version number accessible from Javascript.


jcheng5:
LGTM. I'll let you merge it because I'm not sure what you're doing w.r.t. branches and the Shiny 1.0.5 CRAN release.

Issue #1834, opened 2017-08-29
Allow ability to customize url hostname when using bookmarks


manny-goti:
I've got a few shiny apps iframed into my own website. I'd like to be able to customize the bookmark links to also have the same hostname as my website. Then from the server side of my website I could take the url path that shiny creates and feed the correct link to the iframe. That way I won't have an ugly shiny server address link.

wch:
Closing because there is a simple solution using `onBookmarked`. See comment in #1838.

PR #1838, opened 2017-08-30
Add arguments to doBookmark function for link customizability (#1834)

manny-goti:


wch:
I think these changes add a fair bit of complexity and future maintenance costs for a very specific use case. 

The same goal can be accomplished with `onBookmarked()`. For example:

```R
ui <- function(request) {
  fluidPage(
    textInput("txt", "Text to capitalize:"),
    verbatimTextOutput("out"),
    bookmarkButton()
  )
}
  output$out <- renderText({
    toupper(input$txt)
  })
  onBookmarked(function(url) {
    url <- sub("//127.0.0.1", "//something.com", url)
    showBookmarkUrlModal(url)
  })
}

shinyApp(ui, server, enableBookmarking = "url")
```

PR #1839, opened 2017-08-30
Reactlog scripting fix

jekriske-lilly:
Modern browsers will block mixed content by default. 
This corrects the blocking issue when using reactlog with a Shiny or Rstudio server using TLS.


bborgesr:
Thanks. We've opted by another solution to the problem, which is to use HTTPS instead of HTTP in those script tags. Here's is the PR: #1844.

PR #1861, opened 2017-09-29
fixes #1859

### LINKED ISSUE #1859: Significantly speed-up server-side selectizeInput
[This line of code](https://github.com/rstudio/shiny/blob/e30fac02eddbb7cd510d448e385c559c82009a24/R/update-input.R#L683-L685) slows down **A LOT** on large lists (>10000). The problem in in `apply()` function and thousands of function calls. This could be done much faster with `rowSums()`. Something like:
```r
matches = rowSums(matches)
if(query$conju == 'and')
  idx = idx | (matches == length(key))
else 
  idx = idx | matches
```

dselivanov:


wch:
A few general comments:

* The style we use in Shiny is `<-` instead of `=`.
* It would help if you could provide some minimal test apps so that we can verify that things work properly. Ideally, they would exercise the various different formats that can be used (for example, unnamed vectors, named vectors, unnamed and named lists, and nested lists).
* I'm not sure the existing code for `updateSelectizeInput` works with nested lists -- but it should support that.

R/update-input.R
```diff
@@ -667,10 +666,23 @@ selectizeJSON <- function(data, req) {
   # convert a single vector to a data frame so it returns {label: , value: }
   # later in JSON; other objects return arbitrary JSON {x: , y: , foo: , ...}
   data <- if (is.atomic(data)) {
-    data.frame(label = names(choicesWithNames(data)), value = data,
+    lab = if(is.null(names(data))) {
+      as.character(data)
+    } else {
+      lab = names(data)
```
wch: The nested assignment of `lab` is confusing. It would be better to do `lab <-` in each of the paths of the if-else.

dselivanov:
I've updated code addressing assignment style and `lab` variable comment. Minimal test app on the bottom of this comment.

My opinion regarding other issues:

1. `choices` should be converted to dataframe before passing `session$registerDataObj`. On large lists (starting from 1M choices) even primitive `tolower` function (which is called on each input update) becomes bottleneck. Ideally `selectizeJSON` should just run `grepl(k, data[[v]], fixed = TRUE)` and `data[[v]]` should be already character and lowercase.
1. `choicesWithNames` is very generic and produce `list`. So it will be always magnitudes slower than something which produce vectors. Not sure about implications if we will change "fastpath"  output of the `choicesWithNames` to character vector (possibly named).


Here are couple of test cases. Not sure how to include them to unit tests.
```r
# WORKS
ch1 = c(1, 2)
ch2 = c(a = 1, 2)
ch3 = c(a = 1, B = 2, 3)
ch4 = list(a = 1, B = 2, 3)
ch5 = list(a = 1, B = 2, c = 3)
ch6 = list(a = 1, B = list(B = 2), c = list(3))
# FAILS
ch7 = list(a = 1, list(B = 2), list(3))

ui <- fluidPage(
  selectizeInput("inp", "inp", choices = NULL),
  textOutput("txt")
)

  updateSelectInput(session, "inp",
                    choices = ch6,
                    selected = NULL)

  output$txt = renderText({
    input$inp
  })

}

shinyApp(ui, server)

```

wch:
Named and nested lists (cases 4-7 above) don't work when `server=TRUE`. This PR didn't introduce the bug, though -- they also don't work in the release version of Shiny (1.0.5). This PR doesn't necessarily need to fix that, but it should be kept in mind that the code will eventually need to support that.

Here is an app that tests these things:

```R
bigvec <- paste0("a", 1:1e5)

test_set <- list(
  "Unnamed vector" = c(1, 2),
  "Unnamed list" = c(1, 2),
  "Nested list" = list(a = 1, B = list(B = 2), c = list(3)),
)

ui <- fluidPage(
  sidebarPanel(
  ),
  mainPanel(
    verbatimTextOutput("txt")
  )
)

      return(FALSE)
      return(FALSE)

    TRUE
  }

  observe({
      choices = test_set[[input$set]], selected = NULL, server = input$server
    )
    cat("done\n")
  })

  output$txt = renderPrint({
    if (!validateSelection())
    input$select
  })
}

# not be as fast.
```

* I agree that `choices` should be converted before being passed to `session$registerDataObj`. 
* If `choices` is a flat (non-nested) list, it would be good to have it take a fast path in the code, which may involve converting to a vector. However, the act of checking whether a list is flat may itself be expensive, so this might not be a viable thing to do. If that's the case, then we should just document the performance differences.

dselivanov:
Can it be merged?

fabiangehring:
Would love to see this one merged as well. Often dealing with large (unnamed) choice vectors

wch:
@dselivanov, the two bullet points from my last comment still should be addressed:
* `choices` should be converted before being passed to `session$registerDataObj`.
* It would be great if you could at least test out whether the second item is possible. If that ends up being too difficult, I'd be OK with letting that go (though it would be good to file an issue about that).

dselivanov:
@wch your request looks really confusing to me. My PR has particular scope - issue #1859 (which can be solved in several stages). It clearly resolves biggest bottleneck without touching any other internals. I don't see any compatibility break and so on. 
Honestly speaking PR is pending for 3 months without clear feedback and know you ask to work on some other issue which is orthogonal to the scope of current PR. If you would ask it 3 months back mb I would find some time for further investigation. Now obviously I don't remember all the details about shiny internals which I had to understand before creating PR. Now it will take another several hours to refresh everything in my mind. 
Long story short - if you have concerns about this particular PR and logic it touches, I will be happy to improve it. I have no intention to work on something additional (`session$registerDataObj` can be improved future in a separate PR mb by another person). If this doesn't work - feel free to close PR.

wch:
@dselivanov  It sounds like there was a miscommunication here.

>  and now you ask to work on some other issue which is orthogonal to the scope of current PR.

I don't think the `session$registerDataObj` is orthogonal to the issue being fixed. It is related to speeding up server-side selectInputs. The request is not something new that I have added; it is something that you agreed should be done in your comment on Oct 4. In your comment, you agreed that `choices` should be converted to a data frame before being passed to `session$registerDataObj`, so I thought you were going to do make that change, and was waiting for that. (Note that the change does *not* involve modifying `session$registerDataObj` itself.)

If you have decided that you don't want to do that, that's fine, but we had no way of knowing that was your intention.

dselivanov:
@wch ok, looks like miscommunication. I've put enhancement related to `choices` (it was not that hard). Please review.

R/update-input.R
```diff
@@ -642,8 +642,34 @@ updateSelectizeInput <- function(session, inputId, label = NULL, choices = NULL,
...
+
+  # server side updateSelectizeInput
   value <- unname(selected)
   attr(choices, 'selected_value') <- value
+
+  # convert a single vector to a data frame so it returns {label: , value: }
+  # other objects return arbitrary JSON {x: , y: , foo: , ...}
+  choices <- if (is.atomic(choices)) {
+    # fast path
+    lab <- if(is.null(names(choices))) {
```
wch: The nested assignment of `lab` is somewhat confusing -- it's assigned both here on line 654, and inside the if-else, on 657. I think it would be clearer if the `lab <-` was not here, but instead on the next line, 655. Also, then line 663 could be removed.

R/update-input.R
```diff
@@ -667,22 +692,47 @@ selectizeJSON <- function(data, req) {
   # convert a single vector to a data frame so it returns {label: , value: }
   # later in JSON; other objects return arbitrary JSON {x: , y: , foo: , ...}
   data <- if (is.atomic(data)) {
```
wch: I think the whole `data <-` block can be removed, since `data` should already be a data frame at this point, right?

dselivanov: oops, somehow missed it during merge of upstream

wch:
Thanks!

PR #1877, opened 2017-10-29
Add autoInput

jiho:
Creates an input automatically based on the the class and name of some input data. See the examples for usage.


jiho:
The use-case behind the function is dynamic UIs, in which the user selects some variables and the app creates UI selectors for those variables. The ui selectors can easily be created with `autoInput`.

bborgesr:
This looks really well thought out and documented. Have you considered making it its own package that extends Shiny? We discussed the PR at our team meeting today and we feel like that approach makes more sense for this code than incorporating it into the Shiny codebase itself.

One reason for this is basically conceptual. Shiny is already a big package and we tend to encourage new and niche features to be developed in a separate package that integrates with Shiny. The idea is to keep all the building blocks in Shiny, which can be extended in other packages.

Another reason is maintainability. Especially for things that are somewhat automagical, it tends to be easier to debug and maintain code if it is relatively isolated. In a related note, given the amount of logic that exists in one function, it may be worth it to consider breaking it up and including unit tests.

Hope this makes sense!

Issue #1410, opened 2016-10-06
height="auto" does not work for plotOutput


bborgesr:
If we specify `height = "auto"` in a `plotOutput`, it will not render anything (no issue with `width = "auto"`). E.g. here nothing appears on the screen:

``` r
library(shiny)

ui <- plotOutput("plt", height = "auto")

server <- function(input, output, session) {
  output$plt <- renderPlot({
    plot(cars)
  })
}

shinyApp(ui = ui, server = server)
```

But if we use "300px" or some other valid css unit, it's fine:

``` r
library(shiny)

ui <- plotOutput("plt", height = "300px")

server <- function(input, output, session) {
  output$plt <- renderPlot({
    plot(cars)
  })
}

shinyApp(ui = ui, server = server)
```

bborgesr:
We just need to make this clear in the documentation and then make sure to fail or at least warn if this is the case.

DarioS:
It already is described in the documentation. `?plotOutput` shows

> Note that, for height, using ‘"auto"’ or ‘"100%"’ generally
          will not work as expected, because of how height is computed
          with HTML/CSS.

It's a weird limitation, though, since it's acceptable for `width`. I can't understand why it's acceptable for the CSS specification of ordinary webpages, but not Shiny applications.

shrektan:
@DarioS I remember it's due to the fluid layout.

wch:
The general problem is that CSS doesn't always make it possible to use 100% height. However, if the container's height is set to a particular value, you can use 100% height. For example:

```R
library(shiny)

ui <- div(style = "height: 300px;",
  plotOutput("plt", height = "100%")
)

server <- function(input, output, session) {
  output$plt <- renderPlot({
    plot(cars)
  })
}

shinyApp(ui = ui, server = server)
```

PR #1932, opened 2018-01-25
Async

jcheng5:


R/imageutils.R
```diff
@@ -28,35 +58,44 @@
...
-
-  dv <- grDevices::dev.cur()
+  dv <- startPNG(filename, width, height, res, ...)
   on.exit(grDevices::dev.off(dv), add = TRUE)
   func()
 
   filename
 }
+
+#' @importFrom grDevices dev.set dev.cur
```
wch: Any particular reason these were imported this way, instead of using `grDevices::dev.set`?

jcheng5: I don't think so? Just lots of calls to them maybe?

R/react.R
```diff
@@ -9,20 +27,25 @@ Context <- R6Class(
...
     },
     run = function(func) {
       "Run the provided function under this context."
-      withReactiveDomain(.domain, {
-        env <- .getReactiveEnvironment()
-        .graphEnterContext(id)
-        on.exit(.graphExitContext(id), add = TRUE)
-        env$runWith(self, func)
+
+      promises::with_promise_domain(reactivePromiseDomain(), {
```
wch: We have nested `with_xx` functions in a bunch of places. Instead of doing that, we may want to do what withr does, which is provide a `local_xx` equivalent for each `with_xx` function. These functions would set the domain until the current function exits.

R/react.R
```diff
@@ -9,20 +27,25 @@ Context <- R6Class(
...
+        withReactiveDomain(.domain, {
+          env <- .getReactiveEnvironment()
+          .graphEnterContext(id)
+          on.exit(.graphExitContext(id), add = TRUE)
+          env$runWith(self, func)
+        })
       })
     },
     invalidate = function() {
```
wch: Should there be a `identical(.pid, processId())` here as well?

R/react.R
```diff
@@ -118,16 +137,21 @@ ReactiveEnvironment <- R6Class(
     hasPendingFlush = function() {
       return(!.pendingFlush$isEmpty())
     },
+    # Returns TRUE if anything was actually called
```
wch: I'm curious: do we actually use the return values anywhere?

jcheng5: I think I was calling it in a while loop originally??

R/reactives.R
```diff
@@ -1035,16 +1038,21 @@ registerDebugHook("observerFunc", environment(), label)
...
+              shinyCallingHandlers(run())
+            }
+          },
+          catch = function(e) {
+            printError(e)
+            if (!is.null(.domain)) {
+              .domain$unhandledError(e)
+            }
+          },
+          finally = .domain$decrementBusyCount
```
wch: Needs `()`, since `finally` is an expression that will be evaluated, as opposed to a function that will be run.

wch: Oh wait, nevermind. I was looking at the help for `tryCatch`

R/render-plot.R
```diff
@@ -55,37 +55,20 @@ renderPlot <- function(expr, width='auto', height='auto', res=72, ...,
 
   args <- list(...)
 
-  if (is.function(width))
+  if (is.reactive(width))
+    widthWrapper <- width
+  else if (is.function(width))
```
wch: Why do we turn functions into reactives but turn constant values into functions (below)?

R/render-plot.R
```diff
@@ -106,164 +89,178 @@ renderPlot <- function(expr, width='auto', height='auto', res=72, ...,
   session <- NULL
   outputName <- NULL
 
+  # Calls drawPlot, invoking the user-provided `func` (which may or may not
+  # return a promise). The idea is that the (cached) return value from this
+  # reactive can be used for varying width/heights, as it includes the
+  # displaylist, which is resolution independent.
+  drawReactive <- reactive({
```
wch: The old `plotObj` reactive had `label = "plotObj"`. Should we keep it here? I think it helped profiles to look nice, but I can't remember for sure if it helped in other ways.

R/render-plot.R
```diff
@@ -106,164 +89,178 @@ renderPlot <- function(expr, width='auto', height='auto', res=72, ...,
...
-        # Get coordmap error message if present
-        error = attr(coordmap, "error", exact = TRUE)
-      ))
-    }
-
-    img
+    )
   }
 
+  # If renderPlot isn't going to adapt to the height of the div, then the
```
wch: Is this related to the async stuff, or is it something that you happened to add on? Is this because of https://github.com/rstudio/shiny/issues/1410 ?

R/render-plot.R
```diff
@@ -106,164 +89,178 @@ renderPlot <- function(expr, width='auto', height='auto', res=72, ...,
...
+}
 
-    # Resolution multiplier
-    pixelratio <- session$clientData$pixelratio %OR% 1
+resizeSavedPlot <- function(name, session, result, width, height, pixelratio, res) {
+  if (result$img$width == width && result$img$height == height &&
+      result$pixelratio == pixelratio && result$res == res) {
+    return(result$img)
+  }
+  #result$plotResult
```
wch: Extraneous comments

R/react.R
jcheng5: Added

R/render-plot.R
jcheng5: Good idea, I didn't mean to drop that.

R/render-plot.R
jcheng5: We want the functions to be cached, so they go into reactives. For the constants, using a reactive rather than a function would just add overhead with no benefit.

R/react.R
jcheng5: Do you feel strongly enough about this to want to bring in the withr package as a dependency? It doesn't feel like that big of a deal to me to let these things nest a little... we're not asking our users to write code like this.

R/render-plot.R
jcheng5: It's preexisting: https://github.com/rstudio/shiny/blob/cad20a0bfea9d02dc9f163f028a0c498462fdd90/R/render-plot.R#L263-L264

And it only has to do with Rmd's (where you output `renderPlot(...)` without a corresponding `plotOutput`).

R/react.R
wch: No, I don't mean that we'd bring in withr. I just mean that we might be able to implement our own `local_x` functions that work similarly to the ones in withr, for reactive domains, promise domains, and restore contexts. Now that I look at the actual cases some more, I don't think it's worth the trouble.

DESCRIPTION
```diff
@@ -72,7 +77,10 @@ Imports:
     htmltools (>= 0.3.5),
     R6 (>= 2.0),
     sourcetools,
-    tools
+    later (>= 0.5.0.9001),
```
wch: Should use 0.7.1 here.

R/shinywrappers.R
```diff
@@ -281,15 +323,74 @@ renderPrint <- function(expr, env = parent.frame(), quoted = FALSE,
...
+        close(domain$conn)
+      }
+    )
   }
 
   markRenderFunction(verbatimTextOutput, renderFunc, outputArgs = outputArgs)
 }
 
+createRenderPrintPromiseDomain <- function(width) {
+  f <- file()
```
wch: Where is the data written when `description` is `""`? I can't tell from reading `?file`.

R/utils.R
```diff
@@ -1581,3 +1581,98 @@ Mutable <- R6Class("Mutable",
...
+  }
+
+  if (!visible) {
+    invisible(value)
+  } else {
+    (value)
+  }
+}
+
+is.not.null <- function(x) {
```
wch: This isn't used anywhere, is it?

DESCRIPTION
jcheng5: 👍

R/shinywrappers.R
jcheng5: It's at the end of the Details section:

> If for a file or (on most platforms) a fifo connection the description is "", the file/fifo is immediately opened (in "w+" mode unless open = "w+b" is specified) and unlinked from the file system. This provides a temporary file/fifo to write to and then read from.

Pretty nice!

R/utils.R
jcheng5: Nope, deleted

jcheng5:
Ugh, I didn't realize there are failing tests.

jcheng5:
Remember to discuss async performance

jcheng5:
@wch Any idea about the appveyor build failures? Seems to not be finding g++ when installing promises and later?

wch:
@jcheng5 I think that it's not installing Rtools, because shiny doesn't have a `src/` directory ( krlmlr/r-appveyor#41). Unfortunately, that means that Rtools is unavailable for the upstream dependencies pulled in by shiny.

In the [log](https://ci.appveyor.com/project/rstudio/shiny/build/1.0.1685#L42), it says:

```
== 02/13/2018 23:24:27: Skipping download of Rtools because src/ directory is missing.
```

I've filed krlmlr/r-appveyor#104 -- hopefully there's a workaround for the issue.

wch:
@jcheng5 I just pushed a fix for the Appveyor builds.

PR #1943, opened 2018-02-14
give host param to `runGitHub`

andharris:
This will allow users that have shiny apps in github enterprise repos to be able to easily specify their enterprise host and use `runGitHub`


andharris:
@jcheng5 thanks for the feedback! I've addressed the two things mentioned, however I'm still struggling to ensure redirects are always followed. I see the flag in [`download`](https://github.com/rstudio/shiny/blob/master/R/utils.R#L356-L357) to follow redirects for `curl` but not sure how to follow redirects for other methods. I can get this to work as is when passing the method to `curl` but not sure if it's acceptable to only support `curl` with this param.

PR #1965, opened 2018-02-28
Fix #1964: renderPlot cache breaks when width/height throw

### LINKED ISSUE #1964: renderPlot cache breaks when width/height throw
Thanks to @vnijs for reporting a version of this from radiant.data.

In pre-async versions of Shiny, the following code works; click Go to make the plot appear. After the async merge, this breaks. Note that the `w`/`h` `eventReactive`s throw a `req(FALSE)` exception the first time they're called; post-async, this seems to permanently break the `drawReactive`.

```r
library(shiny)

ui <- fluidPage(
  actionButton("go", "Go"),
  plotOutput("plot")
)

  w <- eventReactive(input$go, { 400 })
  h <- eventReactive(input$go, { 400 })
  
  output$plot <- renderPlot({
    plot(cars)
  }, width = w, height = h)
}

shinyApp(ui, server)
```

jcheng5:
Fixed by moving the isolate(getDims()) call into the (effectively)
try/catch that does a non-isolated getDims() if an error occurs

### DIFF
--- a/R/render-plot.R
+++ b/R/render-plot.R
@@ -94,11 +94,13 @@ renderPlot <- function(expr, width='auto', height='auto', res=72, ...,
   # reactive can be used for varying width/heights, as it includes the
   # displaylist, which is resolution independent.
   drawReactive <- reactive(label = "plotObj", {
-    # Don't invalidate when width/height changes.
-    dims <- if (execOnResize) getDims() else isolate(getDims())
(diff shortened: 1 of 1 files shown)


jcheng5:
Test notes: See linked issue for repro case (I've also checked that example into the shiny-examples repo so we will know if it regresses).

PR #1968, opened 2018-03-07
Improve error handling when `addResourcePath()` fails (especially for `runtime: shiny_prerendered` documents)

bborgesr:
A copy of Yihui's PR for rmarkdown (https://github.com/rstudio/rmarkdown/pull/1171/) to avoid the error `"Error in normalizePath: path[1]="": No such file or directory"` when running any tutorial.

In order the repro of the issue *and* the solution, you should have the latest CRAN versions of the following:

```r
install.packages("rmarkdown")
install.packages("learnr")
```

You can repro the issue with either the latest CRAN version, or the github master version of shiny:

```r
install.packages("shiny")   # OR
remotes::install_github("rstudio/shiny")
```

Then run any `leanr` tutorial (I just used the first learnr example from the package itself). Assuming you have the package cloned to some directory `learnr`, you can do the same through this one liner:

```r
rmarkdown::run("<__your_path_to_learnr__>/learnr/examples/00-setup/00-setup.Rmd")
```

That should result in a fatal problem, that causes the tutorial not to render at all:

repro

If you restart your R session and install `shiny` from my new branch, the problem disappears:

```r
remotes::install_github("rstudio/shiny@barbara/learnr-compat")
rmarkdown::run("<__your_path_to_learnr__>/learnr/examples/00-setup/00-setup.Rmd")
```

withshinyfix

---

I don't know how this was uncaught in the great battles of Pandoc 2 last year... But here's my rough thought process from today. After using a local build of shiny, the error message is a little more helpful, pointing to file & line `/shiny/R/server.R#782`:

reprowithlocalshiny

Over in shiny land, line 782 is the beginning of a call to `..stacktraceoff..`. Instead of trying to understand that, I just did CMD + F for "normalizePath" and only found one call to it in that file. So I tried the same change as Yihui's PR https://github.com/rstudio/rmarkdown/pull/1171/, and to my complete befuddlement, it worked!

However, I worry, that the `mustWork=TRUE` arg was there for a reason and that we may need to wrap it in a conditional. I mostly still don't understand the nature of this bug. Any help is super appreciated, @jcheng5 or @wch!

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,6 +9,8 @@ shiny 1.0.5.9000
 
 ### Minor new features and improvements
 
+* Improved the error handling inside the `addResourcePath()` function, to give end users more informative error messages when the `directoryPath` argument cannot be normalized. This is especially useful for `runtime: shiny_prerendered` Rmd documents, like `learnr` tutorials. [#1968](https://github.com/rstudio/shiny/pull/1968)
+
 * Changed script tags in reactlog ([inst/www/reactive-graph.html](https://github.com/rstudio/shiny/blob/master/inst/www/reactive-graph.html)) from HTTP to HTTPS in order to avoid mixed content blocking by most browsers. (Thanks, [@jekriske-lilly](https://github.com/jekriske-lilly)! [#1844](https://github.com/rstudio/shiny/pull/1844))
 
 * The version of Shiny is now accessible from Javascript, with `Shiny.version`. There is also a new function for comparing version strings, `Shiny.compareVersion()`. ([#1826](https://github.com/rstudio/shiny/pull/1826), [#1830](https://github.com/rstudio/shiny/pull/1830))
--- a/R/server.R
+++ b/R/server.R
@@ -62,7 +62,7 @@ addResourcePath <- function(prefix, directoryPath) {
          "please use a different prefix")
   }
 
-  directoryPath <- normalizePath(directoryPath, mustWork=TRUE)
+  directoryPath <- normalizePath(directoryPath, mustWork=FALSE)
 
   existing <- .globals$resources[[prefix]]
 
--- a/R/server.R
+++ b/R/server.R
@@ -53,21 +53,22 @@ registerClient <- function(client) {
 #' @export
 addResourcePath <- function(prefix, directoryPath) {
   prefix <- prefix[1]
-  if (!grepl('^[a-z0-9\\-_][a-z0-9\\-_.]*$', prefix, ignore.case=TRUE, perl=TRUE)) {
+  if (!grepl('^[a-z0-9\\-_][a-z0-9\\-_.]*$', prefix, ignore.case = TRUE, perl = TRUE)) {
     stop("addResourcePath called with invalid prefix; please see documentation")
   }
-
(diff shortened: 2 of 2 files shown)


bborgesr:
May be tangentially related to this: https://github.com/rstudio/leaflet/issues/467..?

bborgesr:
After talking to @wch, we tracked the root of this problem to the lack of a cache invalidation mechanism in `rmarkdown` and I submitted an issue for that there: https://github.com/rstudio/rmarkdown/issues/1278. As with other problems arising from cache invalidation, the solution is as simple as deleting the .html file and then re-rendering the document from scratch. After we discovered this, my initial reaction was just to close this PR, but @wch suggested leaving the code as it was, and instead wrapping it in a `tryCatch` with a more helpful error message. So, I'm planning on adapting this PR into that.

bborgesr:
@jcheng5, this is ready for review. I tried to make the error message as informative as possible. Here's the current result when a dependency fails:

screen shot 2018-03-16 at 9 45 57 pm

bborgesr:
@jcheng5, I was wondering if it was worth it to add a line saying "You may want to delete the html files from old runs," but I settled on concise and googable, and I'll make sure to have an issue explaining this error message in the `learnr` repo when we have this on master. Let me know what you think

jcheng5:
@bborgesr Can you add a note telling @shalutiwari how to test this fix?

bborgesr:
## Context
@shalutiwari, this fix comes from trying to render a shiny prerendered RMD document (like any `learnr` tutorial) using the latest versions of the `shiny` and `rmarkdown` AND having an old .html file from a previous run. Since shiny prerendered docs try to cache all of its dependencies, if your .html file has a currently out of date dependency, you get an error message. This PR simply gives a better error message than before.

## Setup
So, in order to repro both the original error message and the one after this PR, the first thing you need is to clone the `learnr` repo and navigate to a commit when we know that the repo's "examples" directory (which has a few `learnr` tutorials) had a dependency ("highlight.js 1.1") that is out of date:

```
git clone git@github.com:rstudio/learnr.git old-learnr
cd old-learnr/
git checkout eeae534fa792dcd369075a90b59b042ad26f945f
```

## Reproing initial unhelpful error
Open and R session and install the latest CRAN version of `rmarkdown`, `shiny` and `learnr`:

```r
install.packages("rmarkdown")
install.packages("learnr")
install.packages("shiny")
```

Then run the old .Rmd:

```r
rmarkdown::run("<__your_path_to_old_learnr__>/old-learnr/examples/00-setup/00-setup.Rmd")
```

You should see the same error as my screenshot in my first comment in this thread (https://github.com/rstudio/shiny/pull/1968#issue-173362672):

image

## Reproing slight more helpful error after this PR
Restart your R session. Keep the `rmarkdown` and `learnr` packages from CRAN, but download the latest shiny development github version:

```r
remotes::install_github("rstudio/shiny")
```

Do the same step as above:

```r
rmarkdown::run("<__your_path_to_old_learnr__>/old-learnr/examples/00-setup/00-setup.Rmd")
```
You should now see the same screenshot as my more recent comment:

image

Issue #1837, opened 2017-08-30
selectizeInput option allowEmptyOption = TRUE is ignored (update selectize.js >= 0.12.2 needed)


mnazarov:
I discovered that selectizeInput's option `allowEmptyOption = TRUE` doesn't work properly.
This is a bug in the selectize.js library version 0.12.1, as described in https://github.com/selectize/selectize.js/issues/739 - and was fixed in the version 0.12.2, so I propose to update the library.

A minimal example to see the issue:
```
shinyApp(
    ui = selectizeInput("letters", "Letters:", 
        choices  = c("(none)" = "", letters[1:5]), 
        multiple = FALSE, 
        options  = list(allowEmptyOption = TRUE)),
    server = function(input, output) {}
)
```

jakelever:
+1

phlogisticfugu:
making a fix for this, as I needed the updated version too

alandipert:
Thank you, this was fixed by #2028

PR #1976, opened 2018-03-16
upgrading selectize.js v0.12.1 -> v0.12.4

phlogisticfugu:
fix for https://github.com/rstudio/shiny/issues/1837

to upgrade the selectize plugin

* Note that, if folks are open to this, there should be a build process to pull in javascript dependencies using yarn.


jcheng5:
@wch Is it too late to take this for v1.1?

wch:
I'm OK with upgrading selectize as long as we test it well, but we usually do library updates ourselves just so that we can make sure that it's done properly.

alandipert:
Thank you @phlogisticfugu for alerting us to this. We ended up doing the upgrade in a separate PR, #2028, so this is now resolved.

PR #1984, opened 2018-03-20
Lazily format stack traces

jcheng5:
With deep stack traces enabled, whenever then() is called, we need
to grab the current stack, just in case a downstream callback throws
an error and we need to form a deep stack trace.

Previously, we were calling formatStackTrace at the time that we
grab the current stack (i.e. no error has happened yet) because I
wasn't sure whether holding a reference to sys.calls() for a long
time was a good idea from a garbage collection perspective; would it
prevent the stack frame environments from being collected? But the
answer is no, sys.calls() is just calls, which can be confirmed with
.Internal(inspect(sys.calls()).

By deferring the formatStackTrace call to when we actually need to
print the stack trace, we save ourselves a ton of work--it turns out
it's quite expensive to format the stack traces, much more expensive
than sys.calls() alone.

### DIFF
--- a/R/conditions.R
+++ b/R/conditions.R
@@ -136,7 +136,7 @@ createStackTracePromiseDomain <- function() {
       force(onFulfilled)
       # Subscription time
       if (deepStacksEnabled()) {
-        currentStack <- formatStackTrace(sys.calls())
+        currentStack <- sys.calls()
         currentDeepStack <- .globals$deepStack
       }
       function(...) {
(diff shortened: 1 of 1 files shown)


jcheng5:
@shalutiwari Testing notes!

`shiny::runGitHub("shiny-examples", "rstudio", subdir = "129-async-perf")` (or just run `129-async-perf` out of the shiny-examples repo if you have it locally)

With deep stacks checkbox checked, you should see times in the hundreds of milliseconds without this PR, low tens of milliseconds with this PR.

jcheng5:
Reviewed in person by @wch

PR #1990, opened 2018-03-27
Better stack traces

jcheng5:
- [x] Fix unit tests
- [x] Testing notes

This PR adds some features to stack trace denoising:

- `..stacktraceon/off..` work across deep stacks now
- Use `lobstr::cst` style tree pruning
- Deprecate some existing functions that work only on individual stack traces (not deep)

See https://github.com/rstudio/shiny/wiki/Stack-traces-in-R for design notes.

---

### Testing notes

Run the app `shiny-examples/129-async-perf/app.R`, and watch the R console while you click the Failure button. You should get different results with

1. deep stack trace enabled (checkbox) (you should get two compact stack traces)
2. deep stack trace disabled (you should get "[No stack trace available]")

Then, stop the app, run `options(shiny.fullstacktrace=TRUE)`, relaunch the app, and again:

1. deep stack trace enabled (you should get two long stack traces)
2. deep stack trace disabled (you should get one long stack trace)


R/conditions.R
```diff
@@ -260,19 +266,84 @@ withLogErrors <- function(expr,
...
-  printStackTrace(cond, full = full, offset = offset)
-  lapply(rev(attr(cond, "deep.stack.trace", exact = TRUE)), function(st) {
-    message(
-      paste0(
-        "From earlier call:\n",
-        paste0(formatStackTrace(st, full = full, offset = offset), collapse = "\n"),
-        "\n"
-      )
+  
+  dfs <- mapply(1:length(stackTraceCalls), rev(stackTraceCalls), rev(stackTraceCallNames), rev(toShow), FUN = function(i, calls, nms, index) {
```
jcheng5: `seq_len` in case `stackTraceCalls` is length 0

Issue #1993, opened 2018-04-02
Incorrect search field breaks server-side selectize


carlganz:
This appears even after fixing #1992 
```{R}
library(shiny)

ui <- fluidPage(
  selectizeInput('group', NULL, NULL, multiple = TRUE, options = list(
    #### ADDED LINE####
    maxOptions = 5,    
    #### EDITED LINE####
    searchField = c('Sex', 'Age', 'Survived', 'WRONG'),
    
    placeholder = 'Select a category',
    
    # predefine all option groups
    optgroups = list(
      list(value = '1st', label = 'First Class'),
      list(value = '2nd', label = 'Second Class'),
      list(value = '3rd', label = 'Third Class'),
      list(value = 'Crew', label = 'Crew')
    ),
    
    # 'Class' is a field in Titanic2 created in server.R
    optgroupField = 'Class',
    
    optgroupOrder = c('1st', '2nd', '3rd', 'Crew'),
    
    # how to render the options (each item is a row in Titanic2)
    render = I("{
      option: function(item, escape) {
        return '<div>' + escape(item.Age) + ' (' +
                (item.Sex == 'Male' ? '&male;' : '&female;') + ', ' +
                (item.Survived == 'Yes' ? '&hearts;' : '&odash;') + ')' +
               '</div>';
      }
    }")
  )),
  
  verbatimTextOutput('row'),
  title = 'Using options groups for server-side selectize input'
)

server <- function(input, output, session) {
  
  Titanic2 <- as.data.frame(Titanic, stringsAsFactors = FALSE)
  Titanic2 <- cbind(Titanic2, value = seq_len(nrow(Titanic2)))
  Titanic2$label <- apply(Titanic2[, 2:4], 1, paste, collapse = ', ')
  updateSelectizeInput(session, 'group', choices = Titanic2, server = TRUE)
  
  output$row <- renderPrint({
    validate(need(
      input$group, 'Please type and search (e.g. Female)'
    ))
    Titanic2[as.integer(input$group), -(6:7)]
  })
}

shinyApp(ui, server)

```
If I try to search for any options other than the initial 5 options I get nothing. 
I have a simple fix for this and #1992 so I will make PR.

bborgesr:
We just merged some selectize fixes into master (https://github.com/rstudio/shiny/pull/1996). Can you check if it fixes your use case? If not, can you provide a smaller reproducible example?

carlganz:
Here is much clearer example:

```{R}
[...]

Issue #1992, opened 2018-04-02
selectizeJSON should be coercing values to lowercase


carlganz:
According to note [here](https://github.com/rstudio/shiny/blob/master/R/update-input.R#L700) data that is searched by server selectize is already in lower case, but some simple testing shows this is not the case.

MRE taken from this example: https://shiny.rstudio.com/gallery/option-groups-for-server-side-selectize.html

```{R}
library(shiny)

ui <- fluidPage(
  selectizeInput('group', NULL, NULL, multiple = TRUE, options = list(
    #### ADDED THIS ONE LINE####
    maxOptions = 5,
    placeholder = 'Select a category',
    
    # predefine all option groups
    optgroups = list(
      list(value = '1st', label = 'First Class'),
      list(value = '2nd', label = 'Second Class'),
      list(value = '3rd', label = 'Third Class'),
      list(value = 'Crew', label = 'Crew')
    ),
    
    # 'Class' is a field in Titanic2 created in server.R
    optgroupField = 'Class',
    
    optgroupOrder = c('1st', '2nd', '3rd', 'Crew'),
    
    # you can type and search in these fields in Titanic2
    searchField = c('Sex', 'Age', 'Survived'),
    
    # how to render the options (each item is a row in Titanic2)
    render = I("{
      option: function(item, escape) {
        return '<div>' + escape(item.Age) + ' (' +
                (item.Sex == 'Male' ? '&male;' : '&female;') + ', ' +
                (item.Survived == 'Yes' ? '&hearts;' : '&odash;') + ')' +
               '</div>';
      }
    }")
  )),
  
  verbatimTextOutput('row'),
  title = 'Using options groups for server-side selectize input'
)

server <- function(input, output, session) {
  
  Titanic2 <- as.data.frame(Titanic, stringsAsFactors = FALSE)
  Titanic2 <- cbind(Titanic2, value = seq_len(nrow(Titanic2)))
  Titanic2$label <- apply(Titanic2[, 2:4], 1, paste, collapse = ', ')
  updateSelectizeInput(session, 'group', choices = Titanic2, server = TRUE)
  
  output$row <- renderPrint({
    validate(need(
      input$group, 'Please type and search (e.g. Female)'
    ))
    Titanic2[as.integer(input$group), -(6:7)]
  })
}

shinyApp(ui, server)
```
[...]

PR #1994, opened 2018-04-02
Fixes for selectizeJSON

carlganz:
Addressing #1992 & #1993

### DIFF
--- a/R/update-input.R
+++ b/R/update-input.R
@@ -645,7 +645,6 @@ updateSelectizeInput <- function(session, inputId, label = NULL, choices = NULL,
 
   # server side updateSelectizeInput
   value <- unname(selected)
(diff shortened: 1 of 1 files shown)


bborgesr:
We've merged https://github.com/rstudio/shiny/pull/1996 to address regressions introduced by #1861. I think this overlaps with the problems you'd experienced. If there's another regression we missed, let us know. And thanks for bringing our attention to this! Thankfully, we hadn't released to CRAN yet, so we got to nip this one in the bud  :)

carlganz:
Looks like you took care of most the issues. FWIW I noticed that whenever the size of the data sent to selectize is smaller than maxOptions then `selectizeJSON` returns the entire data so `selectizeJSON` could skip all the actual searching whenever `nrow(data) <= query$maxop`

PR #1996, opened 2018-04-04
fix selectize capitalization (regression introduced by #1861)

### LINKED ISSUE #1995: Server-side selectize loses capitalization
For example, this app doesn't have capitalized options. But if `server = TRUE` is commented out, it works.

```R
ui <- fluidPage(
  selectizeInput("inSelect", "Select input", NULL)
)
  observe({
    updateSelectizeInput(session, "inSelect",
      choices = paste(c("Male", "male", "Female", "female")),
      server = TRUE
    )
  })
}
shinyApp(ui, server)
```

I think this was probably introduced by #1861. It might be the root cause of #1992 and #1993.

bborgesr:
Closes #1995.

### Repro
(check: first letter of each name should retain capitalization)

```r
library(shiny)

ui <- fluidPage(
  selectInput("select", "Select input", NULL)
)

  observe({
      choices = unique(babynames::babynames$name),
      server = TRUE
    )
  })
}

shinyApp(ui, server)

```

### DIFF
--- a/R/update-input.R
+++ b/R/update-input.R
@@ -651,18 +651,16 @@ updateSelectizeInput <- function(session, inputId, label = NULL, choices = NULL,
   # other objects return arbitrary JSON {x: , y: , foo: , ...}
   choices <- if (is.atomic(choices)) {
     # fast path
-    if(is.null(names(choices))) {
+    if (is.null(names(choices))) {
       lab <- as.character(choices)
     } else {
       lab <- names(choices)
       # replace empty names like: choices = c(a = 1, 2)
-      # int this case: names(choices) = c("a", "")
+      # in this case: names(choices) = c("a", "")
       # with replacement below choices will be: lab = c("a", "2")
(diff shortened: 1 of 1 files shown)


bborgesr:
The second commit fixes another problem introduced by #1861, since the `selected` argument is set as an attribute of `choice`s _before_ `choices` is reassigned. This needs to move down, which is what the second commit does. Interestingly, this regression is only present when the choice in not loaded, as dictated by the `maxOptions` param (in this case, "Sarah" is the 10th option, so the app doesn't show that in master, but it does with this fix, along with the capitalization).

### Repro
(with the master version of shiny)

```r
library(shiny)

ui <- fluidPage(
  selectizeInput("select", "Select input", NULL,
    options = list(maxOptions = 5))
)

  updateSelectizeInput(session, "select", selected = "Sarah", 
    choices = unique(babynames::babynames$name), server = TRUE)
}

shinyApp(ui, server)
```

PR #1997, opened 2018-04-04
Add renderCachedPlot()

wch:
This adds `renderCachedPlot()`, which is like `renderPlot()`, but caches the plots to disk. The cache can scoped to a session, to one run of an app (essentially, one call of `runApp`, which can service multiple user sessions), or the app author can specify a cache directory and control the lifetime of the cache.

The base size for cached plot is 400x400 pixels. It will allow other aspect ratios, in multiples of 1.2 by default, although the multiplier can be changed. This means that it will create and cache plots at aspect ratio of 1:1, 1:1.2, 1:1.44, and so on, as well as 1.2:1, 1.44:1, and so on.

It will also cache different sizes, in multiples of 1.2 by default. So for the 1:1 ratio it will cache images at 278x278, 333x333, 400x400, 480x480, 576x576, and so on.

Minimal usage example:

```R
library(shiny)
shinyApp(
  fluidPage(
    sidebarLayout(
      sidebarPanel(
        sliderInput("n", "Number of points", 4, 32, value = 8, step = 4)
      ),
      mainPanel(plotOutput("plot"))
    )
  ),
    output$plot <- renderCachedPlot({
        cat("renderCachedPlot user code\n")
        seqn <- seq_len(input$n)
      },
      cacheKeyExpr = { list(input$n) }
    )
  }
)
```

List of things to do in a separate PR:

* Use `touch` to for DiskCache to use LRU strategy on filesystems that don't support atime.
* Make it possible for the hosting environment to control the session-level cache, perhaps by setting an option that is a function which takes the session token.
* Make `renderCachedPlot` quosure-aware. Also impelment `installQuoFunction`.

### DIFF
--- a/R/render-cached-plot.R
+++ b/R/render-cached-plot.R
@@ -89,14 +89,14 @@
 #'   To use different settings for a session-scoped cache, you can call
 #'   \code{\link{shinyOptions}()} at the top of your server function. To use
 #'   the session-scoped cache, you must also call \code{renderCachedPlot} with
-#'   \code{scope="session"}. This will create a 20 MB cache for the session:
+#'   \code{cache="session"}. This will create a 20 MB cache for the session:
 #'   \preformatted{
 #'   function(input, output, session) {
 #'     shinyOptions(cache = memoryCache(size = 20e6))
 #'
 #'     output$plot <- renderCachedPlot(
 #'       ...,
-#'       scope = "session"
+#'       cache = "session"
 #'     )
 #'   }
 #'   }
@@ -277,7 +277,7 @@
 #'
 #'   output$plot <- renderCachedPlot(
 #'     ...,
-#'     scope = "session"
+#'     cache = "session"
 #'   )
 #' }
 #'
--- a/man/renderCachedPlot.Rd
+++ b/man/renderCachedPlot.Rd
@@ -128,14 +128,14 @@ will not have cache key collisions.
   To use different settings for a session-scoped cache, you can call
   \code{\link{shinyOptions}()} at the top of your server function. To use
   the session-scoped cache, you must also call \code{renderCachedPlot} with
-  \code{scope="session"}. This will create a 20 MB cache for the session:
+  \code{cache="session"}. This will create a 20 MB cache for the session:
   \preformatted{
   function(input, output, session) {
     shinyOptions(cache = memoryCache(size = 20e6))
 
     output$plot <- renderCachedPlot(
       ...,
-      scope = "session"
+      cache = "session"
     )
   }
   }
@@ -296,7 +296,7 @@ server <- function(input, output, session) {
 
   output$plot <- renderCachedPlot(
     ...,
-    scope = "session"
+    cache = "session"
   )
 }
 


R/render-cached-plot.R
```diff
@@ -0,0 +1,404 @@
...
+#'   possible cached images of widths 256, 320, 400, 500, 625, and so on, both
+#'   smaller and larger.
+#' @param res The resolution of the PNG, in pixels per inch.
+#' @param scope The scope of the cache. This can be \code{"app"} (the default),
+#'   \code{"session"}, or the path to a directory to store cached plots. See the
+#'   Cache Scoping section for more information.
+#'
+#' @export
+renderCachedPlot <- function(expr, cacheKeyExpr, cacheResetEventExpr = NULL,
+  baseWidth = 400, aspectRatioRate = 1.2, growthRate = 1.2, res = 72,
```
jcheng5: It'd be cool to replace baseWidth/aspectRatioRate/growthRate arguments with a single argument that's a function that takes width/height and returns width/height. Then you can have multiple functions that we ship that represent different policies.

R/render-cached-plot.R
```diff
@@ -0,0 +1,404 @@
...
+) {
+
+  # This ..stacktraceon is matched by a ..stacktraceoff.. when plotFunc
+  # is called
+  installExprFunction(expr, "func", env, quoted, ..stacktraceon = TRUE)
+  # This is so that the expr doesn't re-execute by itself; it needs to be
+  # triggered by the cache key (or width/height) changing.
+  isolatedFunc <- function() isolate(func())
+
+  cacheKey <- reactive(substitute(cacheKeyExpr), env = parent.frame(), quoted = TRUE)
```
jcheng5: Maybe add a label for the reactive

R/render-cached-plot.R
```diff
@@ -0,0 +1,404 @@
...
+    }
+  }
+
+  # Clear the cacheDir at the appropriate time. Use ignoreInit=TRUE because we
+  # don't want it to happen right in the beginning.
+  observeEvent(
+    substitute(cacheResetEventExpr), event.env = parent.frame(), event.quoted = TRUE,
+    ignoreInit = TRUE,
+    {
+      unlink(file.path(cacheDir(), "*.rds"))
```
jcheng5: Make safer for multiple processes

R/render-cached-plot.R
```diff
@@ -0,0 +1,404 @@
...
+        }
+
+        unlink(cacheDir(), recursive = TRUE)
+
+        # Recursively delete empty parent dirs, up to temp dir.
+        currentDir <- dirname(cacheDir())
+        while (currentDir != tempdir() &&
+               length(dir(currentDir, all.files = TRUE, no.. = TRUE)) == 0)
+        {
+          dirRemove(currentDir)
```
jcheng5: Make safer for multiple processes

R/shiny.R
```diff
@@ -491,6 +493,16 @@ ShinySession <- R6Class(
         return(defaultValue)
       return(result)
     },
+    withCurrentOutput = function(name, expr) {
```
jcheng5: Crap, I just realized you need a promise domain here. I can make that change.

R/render-cached-plot.R
wch: I seem to recall that this had something to do with the cache possibly being reset at the start of a new session, but I can't remember the details.

jcheng5:
Don't forget that for disk cache, when you write a file, write it to the intended path + a random suffix, then after that succeeds, rename it to the intended path (and if the latter fails then delete the former, and use the file that was written in the meantime instead).

R/cache-utils.R
```diff
@@ -0,0 +1,28 @@
...
+#' @export
+print.key_missing <- function(x, ...) {
+  cat("<Key Missing>\n")
+}
+
+
+validate_key <- function(key) {
+  if (grepl("[^a-z0-9]", key)) {
+    stop("Invalid key: ", key, ". Only lowercase letters and numbers are allowed.")
+  }
```
jcheng5: Look out for COM1, etc. Jenny should know.

R/cache-utils.R
```diff
@@ -0,0 +1,28 @@
...
+}
+
+#' @export
+print.key_missing <- function(x, ...) {
+  cat("<Key Missing>\n")
+}
+
+
+validate_key <- function(key) {
+  if (grepl("[^a-z0-9]", key)) {
```
jcheng5: Throw if key is not a 1 element character vector

jcheng5: Throw if key is empty

jcheng5: Throw if key is too long

R/cache-memory.R
```diff
@@ -0,0 +1,344 @@
...
+      private$log(paste0('get: key "', key, '" found'))
+      value <- private$cache[[key]]$value
+      self$prune()
+      value
+    },
+
+    set = function(key, value) {
+      private$log(paste0('set: key "', key, '"'))
+      validate_key(key)
+      if (!private$exec_missing && identical(value, private$missing)) {
```
jcheng5: Remove this check, there are valid reasons to store the missing value

R/cache-memory.R
```diff
@@ -0,0 +1,344 @@
...
+        stop("Attempted to store sentinel value representing a missing key.")
+      }
+
+      time <- as.numeric(Sys.time())
+      private$cache[[key]] <- list(
+        key = key,
+        value = value,
+        size = as.numeric(object.size(value)),  # Reported size is rough! See ?object.size.
```
jcheng5: Don't store if max_size is infinite

R/cache-memory.R
```diff
@@ -0,0 +1,344 @@
...
+    },
+
+    is_destroyed = function() {
+      is.null(private$cache)
+    }
+  ),
+
+  private = list(
+    cache = NULL,
+    meta = NULL,     # Metadata used for pruning
```
wch: Remove; no longer used.

R/cache-memory.R
```diff
@@ -0,0 +1,344 @@
...
+        }
+        rm(list = info$key[rm_idx], envir = private$cache)
+        info <- info[!rm_idx, ]
+    }
+      invisible(self)
+    },
+
+    size = function() {
+      # TODO: Validate this against metadata?
+      length(self$key())
```
jcheng5: Should be `keys()`

R/cache-memory.R
```diff
@@ -0,0 +1,344 @@
...
+      # 3. Remove objects if cache is too large.
+      if (sum(info$size) > private$max_size) {
+        cum_size <- cumsum(info$size)
+        rm_idx <- cum_size > private$max_size
+        if (any(rm_idx)) {
+          private$log(paste0("prune max_size: Removing ", paste(info$key[rm_idx], collapse = ", ")))
+        }
+        rm(list = info$key[rm_idx], envir = private$cache)
+        info <- info[!rm_idx, ]
+    }
```
jcheng5: Indentation

R/cache-memory.R
```diff
@@ -0,0 +1,344 @@
...
+    meta = NULL,     # Metadata used for pruning
+    max_age = NULL,
+    max_size = NULL,
+    max_n = NULL,
+    evict = NULL,
+    missing = NULL,
+    exec_missing = NULL,
+    logfile = NULL,
+
+    object_info = function() {
```
jcheng5: General comment: Consider the common code paths (getting, or setting without a prune actually removing items) and making them fast, by e.g. tracking state in cheap-to-consult fields.

R/cache-memory.R
```diff
@@ -0,0 +1,344 @@
...
+      time <- as.numeric(Sys.time())
+      timediff <- time - info$mtime
+      rm_idx <- timediff > private$max_age
+      if (any(rm_idx)) {
+        private$log(paste0("prune max_age: Removing ", paste(info$key[rm_idx], collapse = ", ")))
+      }
+
+      # Remove items from cache
+      rm(list = info$key[rm_idx], envir = private$cache)
+      info <- info[!rm_idx, ]
```
jcheng5: Put these two lines in the previous if clause

R/cache-memory.R
```diff
@@ -0,0 +1,344 @@
...
+
+      # Sort objects by priority, according to eviction policy.
```
jcheng5: This is only necessary if the cache is actually too big

R/cache-memory.R
```diff
@@ -0,0 +1,344 @@
...
+      }
+
+      # 2. Remove objects if there are too many.
+      if (nrow(info) > private$max_n) {
+        rm_idx <- seq_len(nrow(info)) > private$max_n
+        if (any(rm_idx)) {
+          private$log(paste0("prune max_n: Removing ", paste(info$key[rm_idx], collapse = ", ")))
+        }
+        rm(list = info$key[rm_idx], envir = private$cache)
+        info <- info[!rm_idx, ]
```
jcheng5: Previous two lines should go inside the if clause

jcheng5: Actually the if clause is unnecessary

R/cache-memory.R
```diff
@@ -0,0 +1,344 @@
...
+      }
+
+      # 3. Remove objects if cache is too large.
+      if (sum(info$size) > private$max_size) {
+        cum_size <- cumsum(info$size)
+        rm_idx <- cum_size > private$max_size
+        if (any(rm_idx)) {
```
jcheng5: This if is unnecessary

R/cache-memory.R
```diff
@@ -0,0 +1,344 @@
...
+        return(invisible)
+      }
+      private$log(paste0("destroy"))
+
+      private$cache <- NULL
+    },
+
+    is_destroyed = function() {
+      is.null(private$cache)
+    }
```
jcheng5: I'm ok if you want to remove `destroy`, but if you keep it, seems like it should be checked in the public-facing methods

R/cache-disk.R
```diff
@@ -0,0 +1,533 @@
...
+#'   information will be written. To log to the console, use \code{stdout()}.
+#'
+#' @export
+diskCache <- function(
+  dir = NULL,
+  max_size = 10 * 1024 ^ 2,
+  max_age = Inf,
+  max_n = Inf,
+  evict = c("lru", "fifo"),
+  destroy_on_finalize = NULL,
```
jcheng5: This should move to the end. Consider a default value of `"auto"`

R/utils.R
```diff
@@ -269,6 +269,50 @@ dirExists <- function(paths) {
...
+      stop("Error removing directory ", p, ".")
+    }
+  }
+}
+
+# If the path exists, this returns a canonical path (like normalizePath). If it
+# does not exist, it returns an absolute path, relative to the current dir. The
+# difference is that a canonical path follows symlinks and doesn't have any
+# `..`, while an absolute path here is simply one that starts with `/`.
+absolutePath <- function(path) {
```
jcheng5: Since we need this path to exist anyway, if the dir doesn't exist we can create it and then call `normalizePath` on it.

R/cache-disk.R
```diff
@@ -0,0 +1,533 @@
...
+      private$max_n               <- max_n
+      private$evict               <- match.arg(evict)
+      private$destroy_on_finalize <- destroy_on_finalize
+      if (private$evict == "lru" && !check_atime_support(private$dir)) {
+        # Another possibility for handling lack of atime support would be to
+        # create a file on disk that contains atimes. However, this would not
+        # be safe when multiple processes are sharing a cache.
+        warning("DiskCache: can't use eviction policy \"lru\" because filesystem for ",
+          private$dir, " does not support atime, or has low atime resolution. Using \"fifo\" instead."
+        )
```
jcheng5: It'd be nice to use `touch` (and treating mtime as atime) instead of just falling back to fifo.

R/cache-disk.R
```diff
@@ -0,0 +1,533 @@
...
+        private$evict             <- "fifo"
+      }
+      private$missing             <- missing
+      private$exec_missing        <- exec_missing
+      private$logfile             <- logfile
+
+      if (!dirExists(dir)) {
+        private$log(paste0("initialize: Creating ", dir))
+        dir.create(dir, recursive = TRUE, mode = "0700")
```
jcheng5: Or, inherit parent directory mode

jcheng5: Or take an option

R/cache-disk.R
```diff
@@ -0,0 +1,533 @@
...
+            stop("When `exec_missing` is true, `missing` must be a function that takes one argument.")
+          }
+          return(missing(key))
+        } else {
+          return(missing)
+        }
+      }
+
+      private$log(paste0('get: key "', key, '" found'))
+      self$prune()
```
jcheng5: Same as MemoryCache$get

R/cache-disk.R
```diff
@@ -0,0 +1,533 @@
...
+      value
+    },
+
+    set = function(key, value) {
+      private$log(paste0('set: key "', key, '"'))
+      self$is_destroyed(throw = TRUE)
+      validate_key(key)
+      if (!private$exec_missing && identical(value, private$missing)) {
```
jcheng5: Can be removed, see same comment in equivalent code in MemoryCache$set

R/cache-disk.R
```diff
@@ -0,0 +1,533 @@
...
+        stop("Attempted to store sentinel value representing a missing key.")
+      }
+
+      file <- private$key_to_filename(key)
+      temp_file <- paste0(file, "-temp-", shiny::createUniqueId(8))
```
jcheng5: Remove `shiny::` prefix

R/cache-disk.R
```diff
@@ -0,0 +1,533 @@
...
+      ref_object <- FALSE
+      tryCatch(
+        saveRDS(value, file = temp_file,
+          refhook = function(x) {
+            ref_object <<- TRUE
+            NULL
+          }
+        ),
+        error = function(e) {
+          save_error <<- TRUE
```
jcheng5: Delete `temp_file` if it exists

R/cache-disk.R
```diff
@@ -0,0 +1,533 @@
...
+      if (save_error) {
+        private$log(paste0('set: key "', key, '" error'))
+        stop('Error setting value for key "', key, '".')
+      }
+      if (ref_object) {
+        private$log(paste0('set: value is a reference object'))
+        warning("A reference object was cached in a serialized format. The restored object may not work as expected.")
+      }
+
+      file.rename(temp_file, file)
```
jcheng5: Move this to inside the `tryCatch`

R/cache-disk.R
jcheng5: Delete `temp_file`

R/cache-disk.R
```diff
@@ -0,0 +1,533 @@
...
+      # cache can be pruned to `max_n - n` and `max_size - size` before adding
+      # an object. Right now we prune after adding the object, so the cache
+      # can temporarily grow past the limits. The reason we don't do this now
+      # is because it is expensive to find the size of the serialized object
+      # before adding it.
+
+      private$log(paste0('prune'))
+      self$is_destroyed(throw = TRUE)
+
+      filenames <- dir(private$dir, "\\.rds$", full.names = TRUE)
```
jcheng5: Can you look into how many entries filesystems can handle in a single directory with reasonable performance? If necessary we can introduce a second or even third level of hierarchy.

R/cache-disk.R
```diff
@@ -0,0 +1,533 @@
...
+      files <- file.info(filenames)
```
jcheng5: Left off here on 7/16/2018

R/cache-disk.R
```diff
@@ -0,0 +1,533 @@
...
+      # 1. Remove any files where the age exceeds max age.
+      timediff <- as.numeric(Sys.time() - files[["mtime"]], units = "secs")
+      rm_idx <- timediff > private$max_age
+      if (any(rm_idx)) {
+        private$log(paste0("prune max_age: Removing ", paste(files$name[rm_idx], collapse = ", ")))
+      }
+      file.remove(files$name[rm_idx])
+
+      # Remove rows of files that were deleted.
+      files <- files[!rm_idx, ]
```
jcheng5: Move these two lines inside if clause

R/cache-disk.R
```diff
@@ -0,0 +1,533 @@
...
+      validate_key(key)
+      filename <- private$key_to_filename(key)
+
+      # Instead of calling exists() before fetching the value, just try to
+      # fetch the value. This reduces the risk of a race condition when
+      # multiple processes share a cache.
+      read_error <- FALSE
+      tryCatch(
+        {
+          value <- suppressWarnings(readRDS(filename))
```
jcheng5: If atime is not well-supported, you could do a `touch` here.

R/cache-disk.R
```diff
@@ -0,0 +1,533 @@
...
+        if (private$destroyed) {
+          stop("Attempted to use cache which has been destroyed:\n  ", private$dir)
+        }
+
+      } else {
+        private$destroyed
+      }
+    },
+
+    finalize = function() {
```
jcheng5: Let's not use the finalizer... scary.

R/server.R
```diff
@@ -594,6 +598,14 @@ runApp <- function(appDir=getwd(),
   )
   on.exit(options(ops), add = TRUE)
 
+  # Set up default disk cache for app. Default is 5 MB in size.
```
jcheng5: Comment is inaccurate

R/render-cached-plot.R
```diff
@@ -0,0 +1,601 @@
...
+              # recorded displaylist for the plot can't be serialized and
+              # restored properly within the same R session, so we NULL it out
+              # before saving. (The PNG can of course be saved and restored
+              # just fine.) Note that this was fixed in revision 74506
+              # (2e6c669), and should be in R 3.6, but we need it to work on
+              # older versions. Perhaps in the future we could do a version
+              # check and change caching behavior based on that.
+              result_copy <- result
+              if (!inherits(cache, "MemoryCache")) {
+                result_copy$recordedPlot <- NULL
```
jcheng5: Let's not include recordedPlot in the cached value. The cache key includes the size, so there isn't a case where you'd hit on the displaylist but not on a png.

R/render-cached-plot.R
```diff
@@ -0,0 +1,601 @@
...
+  # is called
+  installExprFunction(expr, "func", env, quoted, ..stacktraceon = TRUE)
+  # This is so that the expr doesn't re-execute by itself; it needs to be
+  # triggered by the cache key (or width/height) changing.
+  isolatedFunc <- function() isolate(func())
+
+  args <- list(...)
+
+  cacheKeyExpr <- substitute(cacheKeyExpr)
+  cacheKey <- reactive(cacheKeyExpr, env = parent.frame(), quoted = TRUE)
```
jcheng5: This isn't the complete cache key; the width/height/res/pixelratio at least are stirred in before actually accessing the cache. Maybe use a different name to clarify that?

R/cache-utils.R
wch: I think this is not necessary because every file has a `.rds` extension.

R/cache-disk.R
wch: I left it in there, but now it always defaults to FALSE. I was going to remove it and write a section in the docs about how to use a finalizer to clean up the disk, but I decided it would be simpler to leave this, and only change the default.

R/cache-disk.R
```diff
@@ -0,0 +1,533 @@
...
+    },
+
+    reset = function() {
+      private$log(paste0('reset'))
+      self$is_destroyed(throw = TRUE)
+      file.remove(dir(private$dir, "\\.rds$", full.names = TRUE))
+      invisible(self)
+    },
+
+    prune = function() {
```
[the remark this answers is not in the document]
wch: Done - now the pruning is throttled.

R/render-cached-plot.R
```diff
@@ -0,0 +1,568 @@
...
+#'
+#' }
+#' @export
+renderCachedPlot <- function(expr,
+  cacheKeyExpr,
+  sizePolicy = sizeGrowthRatio(width = 400, height = 400, growthRate = 1.2),
+  res = 72,
+  cache = "app",
+  ...,
+  env = parent.frame(), quoted = FALSE, outputArgs = list()
```
wch: Discard `env` and `quoted`; if users want to do that stuff, they can use rlang, for both `expr` and `cacheKeyExpr`. Make note of that in docs.

inst/www/shared/shiny.css
```diff
@@ -12,6 +12,11 @@ pre.shiny-text-output.noplaceholder:empty {
   height: 0;
 }
 
+.shiny-image-output img, .shiny-plot-output img {
```
wch: Change to something like `.shiny-image-output img.scalable`, and set the class on img tag.

jcheng5:
We'll need testing notes for @shalutiwari so she can approve before we can merge this.

wch:
Testing notes: There are two main categories of things that need to be tested: the user-facing plot caching, and the underlying caching layer (which is more general than plot caching).

For testing plot caching: The following app has a 2-second delay when rendering a plot. The initial plot will take 2 seconds; if you move the slider, it will take another two seconds; but if you move it back to a previously-used value, it should be instantaneous. If the app is opened in multiple browser windows, the cached plots should be shared. In other words, if a particular slider value is used in one browser, then in the other browser you move the slider to the same value, the plot should display instantly.

```R
shinyApp(
  fluidPage(
    sidebarLayout(
      sidebarPanel(
        sliderInput("n", "Number of points", 4, 32, value = 8, step = 4)
      ),
      mainPanel(plotOutput("plot"))
    )
  ),
    output$plot <- renderCachedPlot({
        seqn <- seq_len(input$n)
      },
      cacheKeyExpr = { list(input$n) }
      # , cache = "session"   # Uncomment this for session-level caching
      # , cache = diskCache()
    )
  }
)
```

There are two variants to test:

* If the `cache = "session"` line is uncommented, then the cache will not be shared across browsers.
* If the `cache = diskCache()` line is uncommented, then the cache will not be shared across browsers. Although this uses a disk cache instead of the default memory cache, the behavior should be the same as with `cache = "session"`


*****

For testing the caching objects, the main part that needs testing is concurrent use of the disk cache by multiple R processes.


If you start R in a terminal and run this, it will read and write to the cache, and print out logging information. The log will indicate that it occasionally prunes objects from the cache.

```R
library(shiny)
x <- 1:1e4
dc <- diskCache("/tmp/diskcache-test", max_size = 5e5, logfile = stdout())
while(1) {
  dc$set(sample(letters, 1), x)
  dc$get(sample(dc$keys(), 1))
}
```

For testing, you should start multiple R processes (3 or 4, for example) and run this exact same code in them. They will share the same directory for the cache, /tmp/diskcache-test. There should be no errors emitted as they run. (If there is an error, that R process will stop, and the logging messages will stop.)

While these R processes are reading and writing, calling `reset()` from one of them should not cause errors in the others, and calling `destroy()` should cause the others to stop. So if you run this code in another R session, the others should keep running just fine:

```R
library(shiny)
dc <- diskCache("/tmp/diskcache-test", max_size = 5e5, logfile = stdout())
dc$reset()
```

And (you can continue using the same session) this should cause the others to stop immediately:

```R
dc$destroy()
```

wikithink:
thanks a lot for the `renderCachedPlot()`,but I have a question, how to prevent caching when the plot is **blank** or **null**, thanks

PR #1999, opened 2018-04-07
Trigger shiny:value even if same data is received

### LINKED ISSUE #1978: shiny:value not triggered if same table data received
If a `renderTable` runs but sends the same data as before to the client, then `shiny:value` is not triggered again for the output. On the other hand `shiny:outputinvalidated` is triggered for the same table. This seems to cause issue in `shinycssloaders`:  https://github.com/andrewsali/shinycssloaders/issues/10

A small reproducible example (clicking the button shows in JS console that `shiny:value` is not re-triggered for the table) .

```
library(shiny)

# Define UI for application that draws a histogram
ui <- fluidPage(
  tags$script("$(document).on('shiny:value', function(event) {
  console.log('Value for: '+ event.target.id);
});
$(document).on('shiny:outputinvalidated', function(event) {
  console.log('Invalidated for: '+ event.target.id);
});"),
   # Application title
   
   sidebarLayout(
      sidebarPanel(
         actionButton("act","Invalidate")
      ),
      
      # Show a plot of the generated distribution
      mainPanel(
         plotOutput("distPlot"),
         tableOutput("distTable")
      ),
   )
)

     input$act
      x    <- faithful[, 2] 
      bins <- seq(min(x), max(x), length.out = 21)
      
   })
   
   output$distTable <- renderTable({
     input$act
     data.frame(a=1)
   })
}

# Run the application 

```

andrewsali:
This PR fixes #1978 . Without this PR, if the client receives the same data from the server, no `shiny:value` is triggered. On the other hand the client sees that the value was invalidated previously ( `shiny:outputinvalidated` is triggered), but unfortunately doesn't know when the old value is confirmed to be still valid.

The role of this PR is to bring back the symmetry: if `shiny:outputinvalidated` is triggered on an element, then eventually a `shiny:value` (or `shiny:error`) should also be triggered.


srcjs/shinyapp.js
```diff
@@ -331,18 +331,22 @@ var ShinyApp = function() {
...
     var evt = jQuery.Event('shiny:value');
     evt.name = name;
     evt.value = value;
     evt.binding = binding;
+
+    if (this.$values[name] === value) {
+      $(binding ? binding.el : document).trigger(evt);
+      return undefined;
+    }
+
```
jcheng5: The order is incorrect; the event shouldn't be triggered until after `this.$values[name] = value;` and `delete this.$errors[name];`. If for no other reason, errors in the trigger shouldn't stop those side effects.

srcjs/shinyapp.js
```diff
@@ -331,18 +331,22 @@ var ShinyApp = function() {
...
+    this.$values[name] = value;
+    delete this.$errors[name];
+
     $(binding ? binding.el : document).trigger(evt);
+
```
jcheng5: Triggered a second time? (Maybe I'm missing something)

srcjs/shinyapp.js
andrewsali: There are two cases, as in the version without the PR: a) client receives the same value as before, b) client receives a new value.

In case a) there is no need to assign the value (this should be more efficient as the DOM doesn't need to be updated) and there should be no errors to delete. This is the same behaviour as before, the only change here is that the event is now triggered (the event parameters need to be set up before).

srcjs/shinyapp.js
andrewsali: This is case b) when the value actually needs to be updated. The event needs to be triggered in this case as well.

andrewsali:
Thanks @jcheng5 , I have sent now the contributor agreement.

andrewsali:
Hi @jcheng5, is there anything holding back this PR?

jcheng5:
@andrewsali No, it's going in this week, probably today

andrewsali:
OK, thanks, that's much appreciated

PR #2004, opened 2018-04-10
Fix #2003: Long stack traces are truncated

### LINKED ISSUE #2003: Long stack traces are truncated
There seems to be a limit to how much you can write to stderr at once before R starts truncating.

Compare this:

```r
cat(file=stderr(), paste(rep(letters, 325), collapse=""))
```

to this:

```r
for (i in 1:325) cat(file=stderr(), paste(letters, collapse=""))
```

You can repro this in Shiny with this app:

```r
library(shiny)
library(promises)

options(shiny.fullstacktrace=TRUE, shiny.deepstacktrace=TRUE)

ui <- fluidPage(
  plotOutput("plot")
)

  output$plot <- renderPlot({
    promise_resolve(TRUE) %...>%
      {stop("hello")}
  })
}

shinyApp(ui, server)
```

The first call stack looks good but the second is truncated (the last line should begin with `1:`).

jcheng5:


R/conditions.R
```diff
@@ -349,8 +349,11 @@ printStackTrace <- function(cond,
             crayon::blue$bold(name)
           else
             crayon::white(name)
-        })
-      )))
+        }),
+        "\n"
+      )
+      for (line in formatted)
```
wch: Could this loop be replaced with the following?

```R
cat(file = stderr(), formatted, sep = "")
```

Issue #1962, opened 2018-02-28
fileInput drag and drop failed with rHandsontableOutputin the UI


dylancis:
It took me a while to pinpoint where the issue is coming from but I have now a reproducible example.
shiny fileInput component allows drag and drop file functionality. However this option is cancelling out while the ui contains a rHandsontableOutput widget.

Please check app.R
`
library(shiny)
library(rhandsontable)

ui <- fluidPage(

titlePanel("Uploading Files"),

sidebarLayout(

sidebarPanel(
  
  # Input: Select a file ---- this is where the issue is
  fileInput("file1", "Choose CSV File",
            multiple = TRUE,
            accept = c("text/csv",
                       "text/comma-separated-values,text/plain",
                       ".csv")),
  
  tags$hr(),
  
  checkboxInput("header", "Header", TRUE),
  
  radioButtons("sep", "Separator",
               choices = c(Comma = ",",
                           Semicolon = ";",
                           Tab = "\t"),
               selected = ","),
  
  radioButtons("quote", "Quote",
               choices = c(None = "",
                           "Double Quote" = '"',
                           "Single Quote" = "'"),
               selected = '"'),
  
  tags$hr(),
  radioButtons("disp", "Display",
               choices = c(Head = "head",
                           All = "all"),
               selected = "head")
  
),

mainPanel(
  
  tableOutput("contents"),
 # try to drag and drop any file in fileInput field with and without the line below
  rHandsontableOutput("hot")
)
)
)

server <- function(input, output) {

output$contents <- renderTable({
req(input$file1)

df <- read.csv(input$file1$datapath,
               header = input$header,
               sep = input$sep,
               quote = input$quote)

if(input$disp == "head") {
  return(head(df))
}
else {
  return(df)
}
})

}

shinyApp(ui, server)`

Looking at the developer tools from Chrome I see that every time I tried to drag a file into Shiny app get triggered:
![image](https://user-images.githubusercontent.com/4976433/36787626-b1a8e76c-1c82-11e8-830f-4d912e649951.png)

dylancis:
This is an annoying issue, an idea on how to fix it please?

bborgesr:
MRE

```r
library(shiny)

ui <- fluidPage(
  fileInput("file1", "Choose CSV File", multiple = TRUE,
    accept = c("text/csv", "text/comma-separated-values,text/plain", ".csv"))
  tableOutput("contents"),
  rhandsontable::rHandsontableOutput("hot")
)

[...]

PR #2005, opened 2018-04-10
Fix dragging and dropping in the presence of jQuery 3.0

alandipert:
Drag and drop relied on jQuery's [.size()](https://api.jquery.com/size/) method, which was deprecated in jQuery 1.8 and removed in 3.0.

The method works in Shiny's version of jQuery, 1.12.4. However, the feature breaks when htmlwidgets introduce jQuery >= 3.0, as happened in [#1962](https://github.com/rstudio/shiny/issues/1962).

This fix uses `.length` instead of `.size` and should make DnD more resilient in the presence of newer versions of jQuery.


alandipert:
@jcheng5 I updated NEWS and called this a bug fix, but later realized it might be considered an improvement/enhancement. What do you think?

jcheng5:
I don't have a preference

bborgesr:
Closes #1962

alandipert:
@jcheng5 approved and @shalutiwari completed QA, so merging

PR #2010, opened 2018-04-12
Fix #2000: Implicit calls to xxxOutput not working inside modules

### LINKED ISSUE #2000: Implicit calls to xxxOutput not working inside modules
I just answered [a question on SO](https://stackoverflow.com/questions/49721349/render-functions-dont-dispay-in-modaldialog-inside-shiny-module) regarding implicit calls to output functions (as defined by `markRenderFunction`) with shiny modules.

For some reason dynamic contents are not shown. Here is a minimal example

    ---
    runtime: shiny
    output: html_document
    ---

    This modal appears empty

    ```{r, echo = FALSE}
    secondModule <- function(input, output, sesison, trigger){
      observeEvent(
        trigger(),
        showModal(modalDialog(renderText({"Text 1"})))
      )
    }

    actionButton("btn", "show first modal")

    callModule(secondModule, "moduleId", reactive({input$btn}))
    ```

    This one does not

    ```{r, echo = FALSE}
    actionButton("btn2", "show second modal")

    observeEvent(
      input$btn2,
      showModal(modalDialog(renderText({"Text 2"})))
    )
    ```

The same behavior can be seen with `plotOutput` (See the SO link above).

The workaround I suggested was to use both `renderText`  and `textOutput` inside the module but I was wondering if this is actually a bug in the  implementation of `markRenderFunction` or if these functions are not meant to be used in such a "convoluted" way.

jcheng5:
I just forgot to have the implicit xxxOutput call respect namespaces.

See #2000 for the repro case. That code sample should be pasted into a new .Rmd file and run. With this patch applied, both buttons produce the same effect.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -43,11 +43,11 @@ This is a significant release for Shiny, with a major new feature that was nearl
 
 * Improved the error handling inside the `addResourcePath()` function, to give end users more informative error messages when the `directoryPath` argument cannot be normalized. This is especially useful for `runtime: shiny_prerendered` Rmd documents, like `learnr` tutorials. ([#1968](https://github.com/rstudio/shiny/pull/1968))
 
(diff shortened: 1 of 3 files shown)


shalutiwari:
@jcheng5 I would prefer it to be added in shiny-examples as I would not go in unit tests to run when doing regression testing.

PR #2013, opened 2018-04-12
renderPlot's ... args were being dropped

### LINKED ISSUE #2007: Investigate renderPlot ... arguments
We throw `...` away now, didn't we used to pass them through to the graphics device?

jcheng5:
Fixes #2007. No NEWS entry is necessary, since this bug was introduced as part of the async work.

**Testing notes:** Use the app here: https://github.com/rstudio/shiny-examples/tree/master/131-renderplot-args


jcheng5:
@wch I'm sure you're going to tell me how I'm doing do.call wrong! :)

PR #2018, opened 2018-04-17
Fix #928: allow inputs to trigger reactive flow even if the value of input hasn't changed

### LINKED ISSUE #928: Feature request: allow inputs to trigger reactive flow even if the value of input hasn't changed
Sometimes it's useful to trigger a reactive chain based on an input even if the input's value hasn't changed.

Simple silly example: suppose you have a shiny app with a few tabs and you want the user to be able to switch tabs by typing the corresponding number. For example, press the "2" key to switch to the second tab.  Here's a simple implementation that I would want to work

```
library(shiny)

jscode <- "
$(function(){ 
 $(document).keyup(function(e) {
 if (e.which >= 49 && e.which <= 57) {
 Shiny.onInputChange('numpress', e.which - 48);
 }
 });
})
"

runApp(shinyApp(
 ui = fluidPage(
 tags$script(HTML(jscode)),
 "Type a number to switch to that tab",
 tabsetPanel(
 id = "navbar",
 tabPanel("tab1", "Tab 1"),
 tabPanel("tab2", "Tab 2"),
 tabPanel("tab3", "Tab 3"),
 tabPanel("tab4", "Tab 4")
 )
 ),
 observe({
 if (is.null(input$numpress)) {
 return()
 }
 updateTabsetPanel(session, "navbar", sprintf("tab%s", input$numpress))
 })
 }
))
```

If you only switch tabs using the numbers on your keyboard, it works fine. But if you press "2" and then switch to a different tab using the mouse, and then press "2" again, the input won't be triggered so the tab won't switch.

I recently spoke with @jcheng5 and IIRC he agreed that supporting something like this could be useful. This is probably not a high priority but I wanted to make it an official request

jcheng5:
This PR introduces two big changes to the old, undocumented, but widely used `Shiny.onInputChange` API:

1. The function is now called `Shiny.setInputValue`; `onInputChange` sounds more like an event listener, not a setter. (The old function name will continue to work, but `setInputValue` will be the official, documented way to do it.)
2. You can pass a `{priority: "event"}` option for input values that want to be treated less like values to be monitored, and more like events that are not to be missed.

```javascript
Shiny.setInputValue("name", value, {priority: "event"});
```

Until now, Shiny has largely treated input values as a continuous stream of values that we sample at convenient times. If an input's value changed multiple times in quick succession, those changes [might be coalesced](https://github.com/rstudio/shiny/blob/cef1f3c7ee51269ada15b30613321475d7247d83/srcjs/input_rate.js#L184). If an input was set to the same value it already had, [it was a no-op](https://github.com/rstudio/shiny/blob/cef1f3c7ee51269ada15b30613321475d7247d83/srcjs/input_rate.js#L217).

#928 is a request to allow some inputs to be treated less like samples of a continuous stream, and more like discrete events, each of which is important. For example, clicking a point on a map might result in a `{lat, long}` input value being set; it's not only the lat/long values that are important, but also the fact that a click occurred at all. So we want two clicks at precisely the same location to always yield two distinct events, and for that to trigger an `observeEvent` on the server twice. But Shiny would see that the lat/long object is exactly the same the second time, and convert it to a no-op. In the past we've had to work around this by [adding nonces](https://github.com/rstudio/leaflet/blob/a4d20dae1a612463134e6b3201bb06757213002f/javascript/src/index.js#L142) to such values (which prevents deduplication/no-ops, but not coalescing--admittedly a much less common problem). This PR removes the need for that. It's not so much the client-side nonce-ing that's an inconvenience, more that the server side then needs to scrub that extra info away.

Previous to this commit, circumventing the client side dedupe logic was not enough. The
server side ReactiveValues object was also subject to deduping. With this commit, the
low-level ReactiveValues class's constructor now has a `dedupe` option, which defaults
to TRUE; the ReactiveValues used for a session's input has it turned to FALSE. I figure
if I had to work this hard to get the client to stop sending duplicates, and the input
values are only expected to ever be updated by the client, then there's really no reason
for server side deduping to be performed for this particular ReactiveValues object.

It would make sense as a future feature to also make deduping optional for user-created
reactiveValues and reactiveVal objects.

- [x] Get consensus with Shiny team on whether this is a good idea
- [ ] Add NEWS, documentation (where?)

### Testing notes

The following test app can be used to verify the behavior. If we decide to merge this PR then this app should be added to the shiny-examples repo.

Don't read the app's source code to understand what's going on, instead just run it and read the captions.

```r
library(shiny)

op <- options(digits.secs = 3)
onStop(function() {
  options(op)
})

testCaseUI <- function(expectation, explanation, code) {
  tagList(
    hr(),
    pre(code),
    p(
      strong(expectation),
      br(),
      explanation
    ),
    tags$button(
      class = "btn btn-default",
      onclick = code,
      "Run"
    )
  )
}

ui <- fluidPage(
  h2("Test cases for setInputValue priorities"),
  div(style = "max-width: 600px",
    testCaseUI(
      "Clicking the button should print a console message, but only the first time it's clicked.",
      "All subsequent calls are deduplicated.",
      "Shiny.setInputValue('one', 1);"
    ),
    testCaseUI(
      "Clicking the button should print a console message, every time.",
      "Events are never subject to deduplication.",
      "Shiny.setInputValue('two', 2, {priority: 'event'});"
    ),
    testCaseUI(
      "Clicking the button should print two console messages, every time.",
      "Events are not coalesced.",
      "Shiny.setInputValue('three', 3, {priority: 'event'});\nShiny.setInputValue('three', 3, {priority: 'event'});"
    ),
    testCaseUI(
      "Clicking the button should print one console message, every time.",
      "On the first click, the first (non-event) call coalesces into the second call; on subsequent clicks, it's ignored due to deduplication.",
      "Shiny.setInputValue('four', 4);\nShiny.setInputValue('four', 4, {priority: 'event'});"
    ),
    testCaseUI(
      "Clicking the button should print one console message, every time.",
      "The second (non-event) call is ignored due to deduplication.",
      "Shiny.setInputValue('five', 5, {priority: 'event'});\nShiny.setInputValue('five', 5);"
    ),
    testCaseUI(
      "Clicking the button should print one console message, every time.",
      "The first (non-event) call coalesces into the second call.",
      "Shiny.setInputValue('six', '6a');\nShiny.setInputValue('six', '6b', {priority: 'event'});"
    ),
    testCaseUI(
      "Clicking the button should print two console messages, every time.",
      "Because the event call happens first, the second call is not coalesced.",
      "Shiny.setInputValue('seven', '7a', {priority: 'event'});\nShiny.setInputValue('seven', '7b');"
    )
  )
)

  lapply(c("one", "two", "three", "four", "five", "six", "seven"), function(i) {
    observeEvent(input[[i]], {
      message("[", Sys.time(), "] ", i, ": ", input[[i]])
    })
  })
}

shinyApp(ui, server)
```

Deprecated version:

<details>

```r
library(shiny)

op <- options(digits.secs = 3)
onStop(function() {
  options(op)
})

ui <- fluidPage(
  p(
    "Clicking this button should print a console message, but only the first time it's clicked.",
    br(),
    tags$button(class = "btn btn-default",
      onclick = "Shiny.onInputChange('one', 1);",
      "One (Normal)"
    )
  ),
  p(
    "Clicking this button should print a console message, every time.",
    br(),
    tags$button(class = "btn btn-default",
      onclick = "Shiny.onInputChange('two', 2, {immediate: true});",
      "Two (Immediate)"
    )
  ),
  p(
    "Clicking this button should print two console messages, every time.",
    br(),
    tags$button(class = "btn btn-default",
      onclick = "Shiny.onInputChange('three', 3, {immediate: true}); Shiny.onInputChange('three', 3, {immediate: true});",
      "Three (Immediate, twice)"
    )
  ),
  p(
    "Clicking this button should print one console message, every time.",
    br(),
    tags$button(class = "btn btn-default",
      onclick = "Shiny.onInputChange('four', 4); Shiny.onInputChange('four', 4, {immediate: true});",
      "Four (Normal, then immediate)"
    )
  ),
  p(
    "Clicking this button should print one console message, every time.",
    br(),
    tags$button(class = "btn btn-default",
      onclick = "Shiny.onInputChange('five', 5, {immediate: true}); Shiny.onInputChange('five', 5);",
      "Five (Immediate, then normal)"
    )
  ),
  p(
    "Clicking this button should print one console message, every time.",
    br(),
    tags$button(class = "btn btn-default",
      onclick = "Shiny.onInputChange('six', '6a'); Shiny.onInputChange('six', '6b', {immediate: true});",
      "Six (Normal, then immediate; mixed values)"
    )
  ),
  p(
    "Clicking this button should print two console messages, every time.",
    br(),
    tags$button(class = "btn btn-default",
      onclick = "Shiny.onInputChange('seven', '7a', {immediate: true}); Shiny.onInputChange('seven', '7b');",
      "Seven (Immediate, then normal; mixed values)"
    )
  )
)

  observeEvent(input$one, {
    message("[", Sys.time(), "] One: ", input$one)
  })
  
  observeEvent(input$two, {
    message("[", Sys.time(), "] Two: ", input$two)
  })
  
  observeEvent(input$three, {
    message("[", Sys.time(), "] Three: ", input$three)
  })
  
  observeEvent(input$four, {
    message("[", Sys.time(), "] Four: ", input$four)
  })
  
  observeEvent(input$five, {
    message("[", Sys.time(), "] Five: ", input$five)
  })
  
  observeEvent(input$six, {
    message("[", Sys.time(), "] Six: ", input$six)
  })
  
  observeEvent(input$seven, {
    message("[", Sys.time(), "] Seven: ", input$seven)
  })
}

shinyApp(ui, server)
```

</details>


jcheng5:
Note that this sidesteps the grander ideas of "discrete" vs "continuous" reactivity. There are still potentially some cases that would be handled by having those notions formally baked into the reactive framework, but those are vastly less important than the use cases that will be taken care of by this change.

jcheng5:
I rewrote the PR description and example app.

wch:
It would be good to switch internal uses of `onInputChange` to `setInputValue`. Otherwise, LGTM.

daattali:
That's a great and very clear writeup @jcheng5 , and a very appreciated feature for some of us. Is `onInputChange()` going to be an alias for `setInputValue()`, so that I can just add the priority param to old code, or will it only be supported with the new function name?

Regarding where to add documentation--perhaps the https://shiny.rstudio.com/articles/ page can get an article page about how to send information from client to server, and this can be a small subsection in that page (IIRC there isn't currently any documentation on that website at all about onInputChange). I would also add to any documentation the fact that this is only available since shiny v1.1, and show how to check for the shiny version in javascript

shalutiwari:
Verified on master

PR #2022, opened 2018-04-18
Fix #2021: Memory leak with reactiveTimer and invalidateLater

### LINKED ISSUE #2021: Memory leak with reactiveTimer and invalidateLater
(Reported by Aaron Berg on behalf of a customer)

If a session ends while a timer (`reactiveTimer` or `invalidateLater`) is pending, the memory used by the session will be retained until the timer's next timeout. For apps that have very long timeouts (such as refreshing data on a dashboard app every 24 hours), lots of data, and frequent browser refreshes (or lots of visitors coming and going), this can lead to significant memory leakage.

The examples below demonstrate the problem, one for each type of timer. Reload the app many times in succession and you'll see that the RAM usage climbs with each refresh. If you wait until the 100 second timeouts start expiring, you'll see the RAM usage start to decline.

```r
library("shiny")

ui <- function(req) {
  basicPage(
    "Memory usage is ",
    capture.output(print(pryr::mem_used()))
  )
}

  session$userData$data <- runif(1000000)
  isolate(invalidateLater(100*1000))
}

shinyApp(ui,server)
```

```r
library("shiny")

ui <- function(req) {
  basicPage(
    "Memory usage is ",
    capture.output(print(pryr::mem_used()))
  )
}

  session$userData$data <- runif(1000000)
  reactiveTimer(100*1000)
}

shinyApp(ui,server)
```

jcheng5:
`reactiveTimer` and `invalidateLater` had mechanisms to ensure they wouldn't have any effect once their owning sessions ended. But they didn't have anything in place to ensure that their closures could be garbage collected. This is because the underlying TimerCallbacks class I implemented way way back in the day, did not provide any way to truly unregister something that was scheduled for execution. This PR corrects that oversight and modifies `reactiveTimer` and `invalidateLater` to take advantage of that.

### Testing notes

See #2021 for repro apps.

### DIFF
--- a/R/timer.R
+++ b/R/timer.R
@@ -42,6 +42,17 @@ TimerCallbacks <- R6Class(
 
       return(id)
     },
+    unschedule = function(id) {
+      toRemoveIndices <- .times$id %in% id
+      toRemoveIds <- .times[toRemoveIndices, "id", drop = TRUE]
+      if (length(toRemoveIds) > 0) {
+        .times <<- .times[!toRemoveIndices,]
+        for (toRemoveId in as.character(toRemoveIds)) {
+          .funcs$remove(toRemoveId)
+        }
+      }
+      return(id %in% toRemoveIds)
+    },
     timeToNextEvent = function() {
       if (dim(.times)[1] == 0)
         return(Inf)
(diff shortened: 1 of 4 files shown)


R/timer.R
```diff
@@ -42,6 +42,17 @@ TimerCallbacks <- R6Class(
...
+    unschedule = function(id) {
+      toRemoveIndices <- .times$id %in% id
+      toRemoveIds <- .times[toRemoveIndices, "id", drop = TRUE]
+      if (length(toRemoveIds) > 0) {
+        .times <<- .times[!toRemoveIndices,]
+        for (toRemoveId in as.character(toRemoveIds)) {
+          .funcs$remove(toRemoveId)
+        }
+      }
+      return(id %in% toRemoveIds)
```
schloerke: This matches the prior if statement.  Maybe a explicit return(TRUE) inside line 53 and return(FALSE) otherwise?

jcheng5: It's a vectorized result, so you can see which of the id's you passed in were actually found and removed.

jcheng5: Sorry, I guess I didn't make this explicit: you can pass in a vector multiple id's.

PR #2049, opened 2018-05-07
`textAreaInput()` doesn't work as expected for relative `width`

shrektan:
## Example Code

```r
library(shiny)
ui <- fluidPage(
  textAreaInput("text_area", "100% width text area", width = "100%"),
  hr(),
  p("compare the width to the horizontal line above")
)
  
}
shiny::shinyApp(ui, server)
```

## Before this PR

before

## After this PR

after


shrektan:
Being bite again... Isn't it a bug?

wch:
Am I correct that, if `width` is `50%`, then both the wrapper div and the textarea have the width set to 50%, and so the resulting overall width of the textarea is 25%?

shrektan:
@wch  Thanks for pointing out this. Now it should have worked expectedly.

### Example code

```r
library(shiny)
charornull <- function(x) {
  if (nzchar(x)) x else NULL
}
ui <- fluidPage(
  inputPanel(
    textInput('area_width', 'area-width', '50%'),
    textInput('area_height', 'area-height', ''),
    textInput('area_rows', 'area-rows', ''),
    textInput('area_cols', 'area-cols', '')
  ),
  uiOutput('ui_text_area')
)
  output$ui_text_area <- renderUI({
    textAreaInput("text_area", label = NULL, 
                  width = charornull(input$area_width),
                  height = charornull(input$area_height),
                  rows = charornull(input$area_rows),
                  cols = charornull(input$area_cols))
  })
}
shiny::shinyApp(ui, server)
```

### Output

image

shrektan:
@wch  Not to push... but this has been opened for two and half years already... 

Is there any potential drawbacks of this fix so it's hanged for so long ? 

I'm asking this is because I use a lot of this function and being unable to set the pct of the width forces me to define a seperate function every time, which is ugly and inconvinient... 

Would be highly appreciated if this issue can be fixed soon...

wch:
Sorry for the delay. I fixed the news locally and squash-merged it in c498b022.

Issue #2042, opened 2018-04-26
[Styling Recommendation] dateRangeInput dates within min/max not easily distinguishable from the rest


rich-iannone:
When using the date picker, it's difficult to visually distinguish between dates within `min` and `max` (the allowed dates) and all other dates. This may result in some confusion for users thinking they can select a date outside the allowed range.

The color values are different but perhaps not enough (two shades of dark gray, where disallowed dates are slightly lighter). Bolding may be an acceptable option for the allowed dates. The other design question is whether there should be two sets of styling for when either `min` and `max` have a date value, or, they are both `NULL` (i.e., styling for allowed/disallowed dates in the first case and a single default style for the second case).

Here is a screenshot of the rendered page from RStudio/Mac/Desktop: 

<img width="556" alt="dateinput" src="https://user-images.githubusercontent.com/5612024/39325571-6e78db5e-4947-11e8-9f4f-56dcd409c80a.png">

Here is the app itself:

```
# dateRangeInput with min/max dates
ui <- fluidPage(

  # Copy the line below to make a date range selector
  dateRangeInput(
    inputId = "dates",
    label = h3("Date range"),
    start = "2018-04-15",
    end = "2018-04-16",
    min = "2018-04-10",
    max = "2018-04-20"),

  hr(),
  fluidRow(column(4, verbatimTextOutput("value")))
)


server <- function(input, output) {

  # You can access the values of the widget (as a vector of Dates)
  # with input$dates, e.g.
  output$value <- renderPrint({ input$dates })
}

shinyApp(ui = ui, server = server)
```

wch:
Reopening this issue because it turns out that the fix we had was too specific -- it didn't pick up all possible kinds of disabled cells.

With the test app above, the disabled days are colored correctly:

![image](https://user-images.githubusercontent.com/86978/47588383-a4698500-d92b-11e8-82a4-6e4a07b7e753.png)


But when you zoom out, disabled months are not:

![image](https://user-images.githubusercontent.com/86978/47588428-c236ea00-d92b-11e8-8057-8f7e48fb6bde.png)


This is because the CSS class hierarchy is different for months (and years, decades, etc).

PR #2087, opened 2018-06-06
Addresses #2042: lighten dates outside of range for datepicker

tmastny:
Addresses #2042:

Before and After:
screen shot 2018-06-06 at 2 43 25 pm after

Tested with both `dateInput` and `dateRangeInput`.


inst/www/shared/datepicker/css/bootstrap-datepicker3.css
```diff
@@ -103,7 +103,7 @@
 .datepicker table tr td.disabled,
 .datepicker table tr td.disabled:hover {
   background: none;
-  color: #777777;
+  color: #AAA;
```
wch: I think it would be better to do this in shiny.css -- in general we try not to modify third-party library code, to make it easier to upgrade in the future.

PR #2091, opened 2018-06-07
Variable Select Input

schloerke:
For use within tidy eval.

Kitchen Variable Select Input Example: https://gist.github.com/schloerke/580ee1587e58ac1f367f00bb1acc22fc
Link to app: https://beta.rstudioconnect.com/connect/#/apps/3782/access

Two small examples:
```r
shinyApp(
  ui = fluidPage(
    varSelectInput("variable", "Variable:", mtcars),
    plotOutput("data")
  ),
    output$data <- renderPlot({
      ggplot2::qplot(!!input$variable, data = mtcars)
    })
  }
)

shinyApp(
  ui = fluidPage(
    varSelectInput("variable", "Variable:", mtcars, multiple = TRUE),
    tableOutput("data")
  ),
    output$data <- renderTable({
       if (length(input$variable) == 0) return(mtcars)
       mtcars %>% dplyr::select(!!!input$variable)
    }, rownames = TRUE)
  }
)
```


wch:
Do you mind deploying the app to rsc.radixu.com, or beta.rstudioconnect.com?

R/input-select.R
```diff
@@ -212,3 +212,121 @@ selectizeIt <- function(inputId, select, options, nonempty = FALSE) {
...
+
+
+
+
+
+
+
+#' Create a select list input control from a data.frame
+#'
+#' Create a select list that can be used to choose a single or multiple items
```
hadley: This seems to be a duplicate of the title? I think you can omit

R/input-select.R
```diff
@@ -212,3 +212,121 @@ selectizeIt <- function(inputId, select, options, nonempty = FALSE) {
...
 }
+
+
+
+
+
+
+
+
+#' Create a select list input control from a data.frame
```
hadley: Maybe "Select variables from a data frame" ?

R/input-select.R
```diff
@@ -212,3 +212,121 @@ selectizeIt <- function(inputId, select, options, nonempty = FALSE) {
...
+#'
+#' Create a select list that can be used to choose a single or multiple items
+#' from the column names of a data frame.
+#'
+#' The resulting output value will be returned as a symbol or a list of symbols
```
hadley: ... returned as a symbol (if multiple = FALSE) or a list of symbols (if multiple = TRUE)...

Or maybe make a bulleted list? Then you could explain `!!` vs `!!!` for each bullet too.

R/input-select.R
```diff
@@ -212,3 +212,121 @@ selectizeIt <- function(inputId, select, options, nonempty = FALSE) {
...
+#'
+#' # single selection
+#' shinyApp(
+#'   ui = fluidPage(
+#'     varSelectInput("variable", "Variable:", mtcars),
+#'     plotOutput("data")
+#'   ),
+#'   server = function(input, output) {
+#'     output$data <- renderPlot({
+#'       ggplot2::qplot(!!input$variable, data = mtcars)
```
hadley: I'd prefer to avoid `qplot()` if possible.

schloerke:
Thanks @hadley for the docs check! (updated details, title, and changed `qplot` to `ggplot`)

R/input-select.R
```diff
@@ -212,3 +212,130 @@ selectizeIt <- function(inputId, select, options, nonempty = FALSE) {
...
+    choices = choices,
+    selected = selected,
+    multiple = multiple,
+    selectize = selectize,
+    width = width,
+    size = size
+  )
+
+  # set the select tag class to be "symbol"
+  selectAttribs <- selectInputVal$children[[2]]$children[[1]]$attribs
```
jcheng5: Can you add a check here that `selectInputVal$children[[2]]$children[[1]]` is actually a `<select>`? (Either here, or in a unit test--and if a unit test, both `selectize=TRUE` and `FALSE`.) I just want to be sure we catch it if we slightly change the DOM structure of `selectInput`.

DESCRIPTION
```diff
@@ -90,8 +90,9 @@ Suggests:
 URL: http://shiny.rstudio.com
 BugReports: https://github.com/rstudio/shiny/issues
 Remotes:
+    tidyverse/ggplot2,
```
jcheng5: Does this mean we wouldn't be able to release Shiny to CRAN without a ggplot2 release first? If so, any idea of a timeline @hadley?

schloerke: to work with the `aes()` I believe we'll have to wait.  But that's only for the example.  

If there is a different example, we could use it instead to release sooner.

R/input-select.R
schloerke: I'll add a unit test

DESCRIPTION
hadley: We're in the middle of the release process, with submission scheduled for June 25.

hadley: (But equally it would be fine to eliminate that example — you can't replace it with anything particularly useful in released ggplot2)

PR #2094, opened 2018-06-08
Wrap dir.create in dir.exists

### LINKED ISSUE #2093: onBookmark fails with renderUI nested modules
https://github.com/rstudio/shiny/blob/3a2a3f21d411b6e8388a653a5a6072bffaffa7f9/R/shiny.R#L905

dir.create() returns FALSE when the directory already exists.

I have a veyr long convoluted Shiny Application with many modules within other modules that unfortunately I can't share, but bookmarking fails when trying to run on a module 3 modules down due to dir.create() returning FALSE when the directory already exists.  If the directory is not created I agree it should stop, but if it already exists then it should just proceed.

Can be fixed by wrapping the statement within dir.exists() first.

aslomko:
Wrap dir.create in dir.exists to fix unwanted stop if directory already exists
Fixes #2093

### DIFF
--- a/R/shiny.R
+++ b/R/shiny.R
@@ -901,9 +901,11 @@ ShinySession <- R6Class(
         # Create subdir for this scope
         if (!is.null(state$dir)) {
           scopeState$dir <- file.path(state$dir, namespace)
-          res <- dir.create(scopeState$dir)
-          if (res == FALSE) {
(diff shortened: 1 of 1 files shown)


wch:
I did a slightly different implementation in #2168.

PR #2099, opened 2018-06-14
Fix #2019: `updateSliderInput` changes formatting

### LINKED ISSUE #2019: updateSliderInput with different argument data type
_updateSliderInput_ does not acknowledge change of data type of arguments (min, max, value).
To run the [demo](https://gist.github.com/helgasoft/0353c54a3b50deceeb18afc1b0a7f460):
`runGist('https://gist.github.com/helgasoft/0353c54a3b50deceeb18afc1b0a7f460')`
 
Also an **enhancement** request - to allow **styling** of sliders generated from code, 
like `renderUI({ sliderInput(...`
Currently styling with themes, skins or custom CSS does not work with those on-the-fly generated sliders.

tmastny:
Fixes #2019. `updateSliderInput` now changes formatting when updating values from `date` to `datetime` and vice versa.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,7 +9,7 @@ shiny 1.1.0.9000
 
 * Addressed [#2042](https://github.com/rstudio/shiny/issues/2042): dates outside of `min`/`max` date range are now a lighter shade of grey to highlight the allowed range. [#2087](https://github.com/rstudio/shiny/pull/2087)
 
-* Addressed [#2019](https://github.com/rstudio/shiny/issues/2019): `updateInputSlider` now changes the slider formatting if the input type changes.
+* Resolved [#2019](https://github.com/rstudio/shiny/issues/2019): `updateInputSlider` now changes the slider formatting if the input type changes.[#2099](https://github.com/rstudio/shiny/pull/2099)
 
 ### Documentation Updates
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,6 +9,8 @@ shiny 1.1.0.9000
 
 * Addressed [#2042](https://github.com/rstudio/shiny/issues/2042): dates outside of `min`/`max` date range are now a lighter shade of grey to highlight the allowed range. [#2087](https://github.com/rstudio/shiny/pull/2087)
 
+* Addressed [#2019](https://github.com/rstudio/shiny/issues/2019): `updateInputSlider` now changes the slider formatting if the input type changes.
+
 ### Documentation Updates
 
 * Addressed [#1864](https://github.com/rstudio/shiny/issues/1864) by changing `optgroup` documentation to use `list` instead of `c`. [#2084](https://github.com/rstudio/shiny/pull/2084)
--- a/R/input-slider.R
+++ b/R/input-slider.R
@@ -86,24 +86,10 @@ sliderInput <- function(inputId, label, min, max, value, step = NULL,
                     version = "0.10.2.2")
   }
 
-  if (inherits(min, "Date")) {
-    if (!inherits(max, "Date") || !inherits(value, "Date"))
-      stop("`min`, `max`, and `value must all be Date or non-Date objects")
-    dataType <- "date"
+  dataType <- getSliderType(min, max, value)
 
-    if (is.null(timeFormat))
-      timeFormat <- "%F"
-
-  } else if (inherits(min, "POSIXt")) {
-    if (!inherits(max, "POSIXt") || !inherits(value, "POSIXt"))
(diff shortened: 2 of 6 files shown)


tmastny:
Minimal example:
```
library(shiny)

min_date <- as.Date("2018-05-30")
max_date <- as.Date("2018-06-11")

min_time <- as.POSIXlt("2018-05-30 12:13:00")
max_time <- as.POSIXlt("2018-06-11 12:13:00")

ui <- fluidPage(
  fluidRow(
    column(
      width = 3,
      offset = 1,
      actionButton("button1", "Switch"),
      sliderInput(
        "date",
        "Date Slider",
        min = min_date,
        max = max_date,
        value = c(min_date, max_date)
      )
    )
  )
)

  observeEvent(input$button1, {
    updateSliderInput(
      session,
      "date",
      min = min_time,
      max = max_time,
      value = c(min_time, max_time)
    )
  })
}


```

R/input-slider.R
```diff
@@ -86,24 +86,10 @@ sliderInput <- function(inputId, label, min, max, value, step = NULL,
...
-      stop("`min`, `max`, and `value must all be POSIXt or non-POSIXt objects")
-    dataType <- "datetime"
-
-    if (is.null(timeFormat))
-      timeFormat <- "%F %T"
-
-  } else {
-    dataType <- "number"
+  if (is.null(timeFormat)) {
+    timeFormat <- switch(dataType, date = "%F", datetime = "%F %T")
```
wch: Might be useful to have a comment about what happens when `dataType` is `"number"`.

NEWS.md
```diff
@@ -7,6 +7,8 @@ shiny 1.1.0.9000
 
 * Addressed [#2042](https://github.com/rstudio/shiny/issues/2042): dates outside of `min`/`max` date range are now a lighter shade of grey to highlight the allowed range. [#2087](https://github.com/rstudio/shiny/pull/2087)
 
+* Addressed [#2019](https://github.com/rstudio/shiny/issues/2019): `updateInputSlider` now changes the slider formatting if the input type changes.
```
wch: Please add a link to this PR.

srcjs/input_binding_slider.js
```diff
@@ -90,13 +122,26 @@ $.extend(sliderInputBinding, textInputBinding, {
         msg.from = data.value;
       }
     }
-    if (data.hasOwnProperty('min'))  msg.min   = data.min;
-    if (data.hasOwnProperty('max'))  msg.max   = data.max;
-    if (data.hasOwnProperty('step')) msg.step  = data.step;
+    var sliderFeatures = ['min', 'max', 'step'];
+    for (var feats of sliderFeatures) {
```
wch: Need to check browser support for `for - of`.

srcjs/input_binding_slider.js
```diff
@@ -90,13 +122,26 @@ $.extend(sliderInputBinding, textInputBinding, {
...
     if (data.hasOwnProperty('label'))
       $el.parent().find('label[for="' + $escape(el.id) + '"]').text(data.label);
 
+    var domElements = ['data-type', 'time-format', 'timezone'];
+    for (var elm of domElements) {
+      if (data.hasOwnProperty(elm))
+        $el.data(elm, data[elm]);
+    }
+
+    var dataType = $el.data('data-type');
```
wch: Should check other `receiveMessage` implementations to see if they use `$el.data()`.

tmastny: No they don't. Based on a keyword search no other `receiveMessage`s use `$el.data()`.

wch: We'll need to hang this state on the object somewhere, and I think maybe this is the best we can do. I can't think of any ill effects from doing this, so maybe we should just go with it.

tmastny: They is almost a way we can do it. Take a look this line: https://github.com/tmastny/shiny/blob/881fe0cfcea89c90d20b4c100fd4e16ed983286a/R/update-input.R#L431-L434

On `updateSliderInput` we recalculate `data-type` and `time-format` before sending the data to JS. The only state we have to maintain is the timezone.

The problem with the timezone is that we don't calculate it based on the POSIXl/ct, but with the `timezone` argument.

srcjs/input_binding_slider.js
tmastny: Support doesn't look good: 
JavaScript MDN: Internet Explorer note\ [supported](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of#Browser_compatibility)
I'll use traditional index iteration, but I could also just go back to how it was.

PR #2102, opened 2018-06-19
Resolves #1933: Serve-side selectize extended to all inputs

### LINKED ISSUE #1933: Server-side selectize doesn't work for some types of inputs
When server-side selectize is used, there are some types of inputs that don't work. They are illustrated by the test app at https://github.com/rstudio/shiny/pull/1861#issuecomment-337296766.

tmastny:
Resolves #1933.

Reproducible Example:
```
library(shiny)
bigvec <- paste0("a", 1:1e5)

test_set <- list(
  "Unnamed vector" = c(1, 2),
  "Unnamed list" = list(1, 2),
  "Nested list" = list(a = 1, B = list(B = 2, C = 4, 5), c = list(3)),
)

ui <- fluidPage(
  sidebarPanel(
  ),
  mainPanel(
    verbatimTextOutput("txt")
  )
)

      return(FALSE)
      return(FALSE)

    TRUE
  }

  observe({
    updateSelectizeInput(
      session,
      "select",
      choices = test_set[[input$set]],
      selected = NULL,
      server = input$server
    )
    cat("done\n")
  })

  output$txt = renderPrint({
    if (!validateSelection())
    input$select
  })
}

# not be as fast.
```

### DIFF
--- a/R/update-input.R
+++ b/R/update-input.R
@@ -668,7 +668,6 @@ updateSelectizeInput <- function(session, inputId, label = NULL, choices = NULL,
     data.frame(label = lab, value = choices, stringsAsFactors = FALSE)
   } else {
     # slow path
-
     list_names <- names(choices)
     if (is.null(list_names)) list_names <- ""
 
@@ -690,9 +689,25 @@ updateSelectizeInput <- function(session, inputId, label = NULL, choices = NULL,
         }
       }
 
-      data.frame(label = lab, value = as.character(choice), group = group)
+      list(
+        label = lab,
+        value = as.character(choice),
+        group = group
+      )
     }, SIMPLIFY = FALSE)
-    do.call(rbind, choice_list)
+
+
+    extract_vector <- function(x, name) {
+      vecs <- lapply(x, `[[`, name)
+      do.call(c, vecs)
+    }
+
+    data.frame(
+      label = extract_vector(choice_list, "label"),
+      value = extract_vector(choice_list, "value"),
+      group = extract_vector(choice_list, "group"),
+      stringsAsFactors = FALSE, row.names = NULL
+    )
   }
 
   value <- unname(selected)
--- a/R/update-input.R
+++ b/R/update-input.R
@@ -647,7 +647,7 @@ updateSelectizeInput <- function(session, inputId, label = NULL, choices = NULL,
   if (is.list(choices)) {
     # check if list is nested
     for (i in seq_along(choices)) {
-      if (is.list(choices[[i]]) || length(choices[[i]]) > 1) { # TODO: add ' || length(choices) > 1'
+      if (is.list(choices[[i]]) || length(choices[[i]]) > 1) {
         noOptGroup <- FALSE
         break()
       }
@@ -682,7 +682,7 @@ updateSelectizeInput <- function(session, inputId, label = NULL, choices = NULL,
       lab <- name
       if (lab == "") lab <- as.character(choice)
 
-      if (is.list(choice) || length(choice) > 1) { #TODO: add ' || length(choice) > 1'
+      if (is.list(choice) || length(choice) > 1) {
         group <- rep(name, length(choice))
         choice <- unlist(choice)
 
--- a/R/update-input.R
+++ b/R/update-input.R
@@ -644,17 +644,21 @@ updateSelectizeInput <- function(session, inputId, label = NULL, choices = NULL,
   }
 
   noOptGroup <- TRUE
-  for (i in 1:length(choices)) {
-    if (is.list(choices[[i]])) {
-      noOptGroup <- FALSE
-      break()
+  if (is.list(choices)) {
+    # check if list is nested
+    for (i in seq_along(choices)) {
+      if (is.list(choices[[i]]) || length(choices[[i]]) > 1) { # TODO: add ' || length(choices) > 1'
+        noOptGroup <- FALSE
+        break()
+      }
     }
   }
-
-  # convert a single vector to a data frame so it returns {label: , value: }
+  # convert choices to a data frame so it returns [{label: , value: , group: },...]
   choices <- if (is.atomic(choices) || noOptGroup) {
-    # fast path
-    if (is.list(choices)) choices <- unlist(choices)
+    # fast path for vectors and flat lists
+    if (is.list(choices)) {
+      choices <- unlist(choices)
+    }
     if (is.null(names(choices))) {
       lab <- as.character(choices)
     } else {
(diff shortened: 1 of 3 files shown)


srcjs/input_binding_select.js
```diff
@@ -55,9 +55,11 @@ $.extend(selectInputBinding, {
     if (data.hasOwnProperty('url')) {
       selectize = this._selectize(el);
       selectize.clearOptions();
+      selectize.clearOptionGroups();
       var thiz = this, loaded = false;
```
wch: I know you didn't do this, but can you change `thiz` to `self`?

srcjs/input_binding_select.js
```diff
@@ -55,9 +55,11 @@ $.extend(selectInputBinding, {
...
       selectize.settings.load = function(query, callback) {
         var settings = selectize.settings;
+        var thiz2 = this;
```
wch: Is `thiz2` really necessary?

srcjs/input_binding_select.js
```diff
@@ -111,7 +120,10 @@ $.extend(selectInputBinding, {
     var options = $.extend({
       labelField: 'label',
       valueField: 'value',
-      searchField: ['label']
+      searchField: ['label'],
+      optgroupField: 'group',
+			optgroupLabelField: 'group',
```
wch: Indentation: no tabs!

R/update-input.R
```diff
@@ -660,7 +668,46 @@ updateSelectizeInput <- function(session, inputId, label = NULL, choices = NULL,
...
+      list(
+        label = lab,
+        value = as.character(choice),
+        group = group
+      )
+    }, SIMPLIFY = FALSE)
+
+
+    extract_vector <- function(x, name) {
+      vecs <- lapply(x, `[[`, name)
```
wch: The existing implementation of `extract_vector ` was written to work when each list contained vectors of length >= 1. But if it's known that length == 1, and if the type is known ahead of time, then this could be replaced with `vapply`, and the `do.call` won't be necessary. (I think that the vectors are all char vectors, right?)

wch: Actually, never mind what I said. If I'm reading the code correctly now, the lists contained in `choice_list` can have variable length, in which case you need the existing `lapply` version.

tmastny: Yes, each list can contain vectors of length >= 1. For example, if `choices <- list(a = 1, B = list(B = 2, C = 4, 5), c = list(3))` then `choice_list` equals:
```
$a
$a$label
[1] "a"

$a$value
[1] "1"

$a$group
[1] ""


$B
$B$label
[1] "B" "C" "5"

$B$value
[1] "2" "4" "5"

$B$group
[1] "B" "B" "B"


$c
$c$label
[1] "3"

$c$value
[1] "3"

$c$group
[1] "c"
```
However, the vectors are all character vectors.

srcjs/input_binding_select.js
tmastny: Good catch. In fact, I can get rid of all of "thiz."

wch: I haven't looked at it closely yet, but you might still need `thiz`/`self`. (I'm assuming `thiz` was there for a reason.) You may already know this, but `this` in Javascript doesn't behave the way that it does in other languages.

For example, see https://gist.github.com/wch/b6ae0c7d88285f49c6b328c39f61e8da

tmastny: It looked like the original `thiz` was referring to the binding object, and was calling the `setValue` method. `thiz2` was referring to the selectize object (as per [this example](https://github.com/selectize/selectize.js/issues/151#issuecomment-111056161)) and was calling the `setValue` selectize method. 

However, in all cases I believe `thiz` was unnecessary as we could explicitly use the `selectize` variable declared on line 56.

tmastny:
For QA: I've created an app and shinytest file that iterates through selectize functionality: 
https://github.com/rstudio/shiny-examples/tree/selectize-server/135-selectize-server

`mytest-expected` in the `tests` directory was ran on my fork of Shiny. You can verify that the existing selectize functionality still works by running that test on the master branch. All tests with the `Server-side selectize` checkbox unchecked should pass on the master branch. 

The functionality tested includes
   - testing `selected = NULL` in `updateSelectizeInput` behaves as expected on the server-side.
      - the `selected` parameter has different behavior for single vs. multiple selectize, see `updateSelectizeInput` docs.
   - testing server-side selectize works with both single and multiple selectize by using the `multiple` parameter in `selectizeInput`
   - testing server-side selectize works on lists and builds optgroups

R/update-input.R
```diff
@@ -643,10 +643,18 @@ updateSelectizeInput <- function(session, inputId, label = NULL, choices = NULL,
     return(updateSelectInput(session, inputId, label, choices, selected))
   }
 
+  noOptGroup <- TRUE
+  for (i in 1:length(choices)) {
```
wch: Should use `seq_along` here.

R/update-input.R
```diff
@@ -643,10 +643,18 @@ updateSelectizeInput <- function(session, inputId, label = NULL, choices = NULL,
...
+    if (is.list(choices[[i]])) {
```
wch: Should also check for char vectors of length > 1. (?)

R/update-input.R
```diff
@@ -643,10 +643,18 @@ updateSelectizeInput <- function(session, inputId, label = NULL, choices = NULL,
...
+      noOptGroup <- FALSE
+      break()
+    }
+  }
+
   # convert a single vector to a data frame so it returns {label: , value: }
-  # other objects return arbitrary JSON {x: , y: , foo: , ...}
-  choices <- if (is.atomic(choices)) {
+  choices <- if (is.atomic(choices) || noOptGroup) {
```
wch: If `choices` is atomic, then we can skip the `noOptGroup` loop above.

R/update-input.R
```diff
@@ -643,10 +643,18 @@ updateSelectizeInput <- function(session, inputId, label = NULL, choices = NULL,
...
     # fast path
+    if (is.list(choices)) choices <- unlist(choices)
```
wch: Maybe add a comment that if it got here, it must be a flat list.

R/update-input.R
```diff
@@ -660,7 +668,46 @@ updateSelectizeInput <- function(session, inputId, label = NULL, choices = NULL,
     data.frame(label = lab, value = choices, stringsAsFactors = FALSE)
   } else {
     # slow path
-    as.data.frame(choices, stringsAsFactors = FALSE)
+    list_names <- names(choices)
+    if (is.null(list_names)) list_names <- ""
```
wch: Use `rep()` here instead of recycling, for clarity.

srcjs/input_binding_select.js
```diff
@@ -84,9 +85,17 @@ $.extend(selectInputBinding, {
             callback();
           },
           success: function(res) {
```
wch: Can you add a comment that shows an example the structure of `res`?

wch: Can you explain that `success` is called after the options are added, but before the groups, and that groups need to be added manually?

PR #2106, opened 2018-06-20
compile with grunt to get version in js code (v1.1.0.9000)

schloerke:
Compile and change `{{ VERSION }}` to be 1.1.0.9000 by calling `grunt` in tools/ folder


tools/Gruntfile.js
```diff
@@ -202,11 +202,23 @@ module.exports = function(grunt) {
...
+      }
+    }
+  );
+
   grunt.initConfig(gruntConfig);
 
   grunt.registerTask('default', [
     'newer:concat',
     'newer:string-replace',
+    'validateStringReplace',
```
wch: I think if you prefix this with `newer:`, it will also run when you do `grunt watch`

schloerke: watch task now executes the default task.  Before it did not run the version string replacement change.

wch: Ah, that makes sense. I'm glad you figured it out!

PR #2125, opened 2018-07-12
Fix plot interaction for scaled plots

wch:
This pull request builds on top of #1997 and should be merged only after that one. It adds support for accurate interaction information for plots that are scaled in the web browser. It also makes sure that the interaction information is correct when the img has a border and/or padding. It also adds support for brushes to resize themselves correctly when a plot is resized in the browser (note that this is distinct from the case when a new plot is sent to the browser -- brushes already were able to persist in that case).

Previously, the interaction code mapped between browser (CSS) pixel coordinates and data coordinates. Now, there is explicit mapping from CSS pixel coordinates to raw image pixel coordinates (that is, pixel coordinates on the unscaled PNG that is sent to the browser), and data coordinates.


srcjs/output_binding_image.js
```diff
@@ -312,90 +358,112 @@ imageutils.initCoordmap = function($el, coordmap) {
...
+  // many pixels. It's possible for an offset to be within more than one
+  // panel, because of the `expand` value. If that's the case, find the
+  // nearest panel.
+  coordmap.getPanelCss = function(offset_css, expand = 0) {
+    const offset_img = coordmap.scaleCssToImg(offset_css);
+    const x = offset_img.x;
+    const y = offset_img.y;
+
+    // Convert expand from css pixels to img pixels
+    const imgToCssRatio = findImgToCssScalingRatio($img);
```
jcheng5: Reverse this--`const cssToImgRatio = coordmap.cssToImgScalingRatio();` and change division below to multiplication

srcjs/output_binding_image.js
```diff
@@ -96,74 +93,91 @@ $.extend(imageOutputBinding, {
...
       opts.coordmap = [];
 
-    imageutils.initCoordmap($el, opts.coordmap);
+    // When the image loads, initialize all the interaction handlers. When the
+    // value of src is set, the browser may not load the image immediately,
+    // even if it's a data URL. If we try to initialize this stuff
+    // immediately, it can cause problems because we use .naturalWidth and
+    // .naturalHeight, but they're available only after the image loads.
+    $img.off("load.shiny-image-interaction");
+    $img.on("load.shiny-image-interaction", function() {
```
jcheng5: Investigate getting rid of this delay by having the server return the image width/height

srcjs/output_binding_image.js
```diff
@@ -90,94 +90,94 @@ $.extend(imageOutputBinding, {
...
 
-    if (!opts.coordmap)
-      opts.coordmap = [];
-
-    // When the image loads, initialize all the interaction handlers. When the
-    // value of src is set, the browser may not load the image immediately,
-    // even if it's a data URL. If we try to initialize this stuff
-    // immediately, it can cause problems because we use .naturalWidth and
-    // .naturalHeight, but they're available only after the image loads.
+    if (!opts.coordmap) {
```
wch: Need to test what happens if we have a plot without a valid coordmap, like a ggplot2 polar plot.

srcjs/output_binding_image.js
```diff
@@ -998,8 +998,15 @@ imageutils.createBrushHandler = function(inputId, $el, opts, coordmap, outputId)
...
-      brushInfoSender.immediateCall();
+      // Importing an old brush must happen after the image data has loaded
+      // and the <img> DOM element has the updated size. If importOldBrush()
+      // is called before this happens, then the css-img coordinate mappings
+      // will give the wrong result, and the brush will have the wrong
+      // position.
+      $el.find("img").one("load.shiny-image-interaction", function() {
+        brush.importOldBrush();
+        brushInfoSender.immediateCall();
+      });
```
jcheng5: Please add an example to shiny-examples that will repro if this regresses

wch:
Should file an issue to not null out click/hover/dblclick when image is scaled or a new one is sent from server. This should also be an option for `clickOpts()` and similar.

srcjs/output_binding_image.js
```diff
@@ -240,162 +259,221 @@ imageutils.initPanelScales = function(coordmap) {
...
-      shiftAmount = max - maxval;
-    } else if (minval < min) {
-      shiftAmount = min - minval;
-    }
+  // Given an offset in css pixels, return an object representing which panel
+  // it's in. The `expand` argument tells it to expand the panel area by that
+  // many pixels. It's possible for an offset to be within more than one
+  // panel, because of the `expand` value. If that's the case, find the
+  // nearest panel.
+  coordmap.getPanelCss = function(offset_css, expand = 0) {
```
jcheng5: No need to gather up all the panel objects that match--you can instead just keep track of the nearest one seen so far.

wch:
Add example to shiny-examples that uses CSS transforms. This will exercise a lot of parts of the code.

PR #2126, opened 2018-07-12
Add caching for eventReactive

wch:
This should be merged after #1997, because it builds on top of that pull request. It adds additional caching for `eventReactive`, similar to `renderCachedPlot`.

### DIFF
--- a/R/reactives.R
+++ b/R/reactives.R
@@ -2028,13 +2028,55 @@ observeEvent <- function(eventExpr, handlerExpr,
   invisible(o)
 }
 
+#' @section \code{eventReactive} caching:
+#'
+#'   Like regular \code{\link{reactive}} expressions, the most recent value of a
+#'   \code{eventReactive} is always cached. (Observers are not cached because
+#'   they are used for their side-effects, not their values.) If a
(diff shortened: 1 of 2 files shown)


jcheng5:
API issues are a little complicated... from in-person discussion with @wch:

```r
eventReactive(input$go, {
  # Do whatever
}, cache = "app")

# eventReactive without `key` may not be sufficient, here's an example where
# reactivity and cache key are distinct
cachedEventReactive(input$go, {
  # Do whatever
}, cache = "app", key = list(input$dataset))

# Here's one that's more like a regular reactive's semantics, but with caching.
# You end up having to `isolate()` the body, so that the arguments end up looking
# quite similar to `eventReactive`, but the intent is very different (explicit vs. implicit
# reactivity).
cachedReactive({
  # Something with input$dataset
  input$foo
}, cacheKey = list(input$dataset))


# Here's how you can manually implement caching reactives today.
cache <- DiskCache$new(...)

fun <- memoize(function(dataset) {
  isolate({
    # Do something with dataset
  })
}, cache)

reactive({
  fun(input$dataset)
})
```

schloerke:
Can be closed now that https://github.com/rstudio/shiny/pull/3094 is merged?

PR #2147, opened 2018-08-04
Fixes #174, allowing specific days of the week to be disabled.

### LINKED ISSUE #174: dateInput Request. Restrict to specific dates within min and max
I would like to have, say, only the Fridays clickable. So if one of the parameters could be a vector of allowable dates that would be v useful

nathancday:
Fixes #174, with a new parameter `daysofweekdisabled` available to `dataInput()`.

An example app:
```
library(shiny)

shinyApp(
  ui = fluidPage(
    dateInput("date1", "All Days:", daysofweekdisabled = NULL)
    dateInput("date2", "No Weekends:", daysofweekdisabled = c(0,6))
  ), 
  server = function(input, output, session) {}
)
```

Original code by Niko Eckerskorn (@kadrach). I made this PR because the original fork was 2 years behind `master` and the issue is still unresolved.


nathancday:
I forced push to my master and I'm not 100% that's what made TravicCI mad, but it was coincidental. Happy to adjust, just let me know the best path to resolution.

R/input-date.R
```diff
@@ -70,14 +74,22 @@
...
+#'             datesdisabled = c("2012-03-01", "2012-03-02"))
 #' )
 #'
 #' shinyApp(ui, server = function(input, output) { })
 #' }
 #' @export
 dateInput <- function(inputId, label, value = NULL, min = NULL, max = NULL,
-  format = "yyyy-mm-dd", startview = "month", weekstart = 0, language = "en",
+  format = "yyyy-mm-dd", startview = "month", weekstart = 0,
+  datesdisabled = NULL, daysofweekdisabled = NULL, language = "en",
```
wch: Can you move these arguments to the end? Even though this is the right place to put them conceptually, there could be existing code out there that is broken by this change.

wch:
Can you send a CLA, as described in https://github.com/rstudio/shiny/blob/master/CONTRIBUTING.md?

R/input-date.R
nathancday: Sure thing!

nathancday:
CLA sent. Updated argument ordering and added the same conditional formatting for `datesdisabled` that `min` and friends have.

Issue #2123, opened 2018-07-11
In click/hoverOpts, clip = FALSE throws a JS exception


tmastny:
Example:

```
library(shiny)
library(ggplot2)


shinyApp(
  ui = basicPage(
    fluidRow(
      column(
        width = 4,
        plotOutput(
          "plot",
          height=300,
          click = clickOpts(
            "plot_click",
            clip = FALSE
          )
        )
      ),
      column(
        width = 4,
        verbatimTextOutput("plot_clickinfo")
      )
    )
  ),
  server = function(input, output, session) {
    output$plot <- renderPlot({
      plot(cars$speed, cars$dist)
    })
    output$plot_clickinfo <- renderPrint({
      cat("Click:\n")
      str(input$plot_click)
    })
  }
)
```

PR #2157, opened 2018-08-13
Fix plot clipping issues and return pixel coordinates

### LINKED ISSUE #2079: Feature Request - hoverOpts - control units
I would like to request that an additional parameter be added to hoverOpts and plot interaction control functions interacting with plot_output and image_output. This parameter will control the units assigned to the x y values returned by input$plot_hover, input$plot_click etc values. Currently this is assumed based on the output function plot_output or image_output. with plot_output, the units are in the same units as the data, and with image_output the units are in pixels and based on location of the mouse.

I am using treemap data and I need to associate the mouse location with one of the boxes in the treemap. treemap software will calculate the coordinates of the boxes, however, when I plot a treemap using plot_output, i receive x/y coordinates in terms of the data and this is not useful. I would like to be able to define the units as "data" or "pixel" as a parameter of the hoverOpts function. This way I can get pixel location using the plot_output function.

I then can use the pixel coordinate together with the coordinates the treemap software calculates to identify which box the user has selected. With this I can then do complex follow up calculation for the user based on the box selected in the plot.

something like 
hoverOpts(id = NULL, delay = 300, delayType = c("debounce", "throttle"),
  clip = TRUE, nullOutside = TRUE, unitsType = c("data","pixels"))

tmastny:
This PR fixes plot clipping issues in #2123 and resolves #2079 by including pixel coordinates in the plot interaction output.

### DIFF
--- a/srcjs/output_binding_image.js
+++ b/srcjs/output_binding_image.js
@@ -515,7 +515,7 @@ imageutils.initCoordmap = function($el, coordmap) {
         exports.setInputValue(inputId, null);
         return;
       }
-
+      const coords = {};
       const offset_css = coordmap.mouseOffsetCss(e);
       // If outside of plotting region
       if (!coordmap.isInPanelCss(offset_css)) {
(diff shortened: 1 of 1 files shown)


srcjs/output_binding_image.js
```diff
@@ -525,12 +525,20 @@ imageutils.initCoordmap = function($el, coordmap) {
         }
         if (clip)
           return;
-      }
-      if (clip && !coordmap.isInPanelCss(offset_css)) return;
 
+        coords.x_px = coordmap.scaleCssToImg(offset_css).x;
+        coords.y_px = coordmap.scaleCssToImg(offset_css).y;
```
tmastny: What about a list `px` that contains `x` and `y`.

wch:
How about we send back:

* `coords_css`, which is a list with `x` and `y`
* `coords_img`, which is a list with `x` and `y`
* Rename `pixelratio` to `img_css_ratio`


There are some changes that will be needed in `image-interact.R` which I can take care of.

PR #2160, opened 2018-08-16
Use xxhash64 instead of sha256 for hash algorithm

wch:
On the advice of @jimhester and @hadley, this PR switches from SHA256 to xxHash64, because it's significantly faster:

```
> system.time(digest::digest(ggplot2::diamonds, "sha256"))
   user  system elapsed 
  0.042   0.002   0.046 
> system.time(digest::digest(ggplot2::diamonds, "xxhash64"))
   user  system elapsed 
  0.013   0.002   0.016 
```


jimhester:
Cool, good to see my old addition of xxHash support to digest paying dividends!

PR #2164, opened 2018-08-18
typo/clarify dox of reactiveValuesToList

MichaelChirico:
verb definitely missing, inferred "want" but "expect" could also fit


wch:
Thanks! Due to a merge conflict with the current version of shiny, I've implemented a similar change in 2d61709d.

PR #2166, opened 2018-08-23
Un-export knit_print methods from htmltools

wch:
This change un-exports the `knit_print` methods that are imported from htmltools. It requires rstudio/htmltools#108 to be merged first.


wch:
Note: this PR can wait until the updated version of htmltools is on CRAN. Then htmltools won't have to be added to Remotes.

alandipert:
We (me and @jcheng5) kicked this out of 1.2 and put in 1.3 so that @wch can give the green-light.

PR #2174, opened 2018-09-05
Add informative errors when xvar or yvar not in data for brushedPoints

AliciaSchep:
Currently, if attempting to use `nearPoints` or `brushedPoints` with ggplot where either x or y is a transformation and not just a raw data column, the functions will either fail silently without error or in the case of brushedPoints and `allRows = TRUE` give a cryptic error message about replacement having zero rows. While the documentation of brushedPoints mentions the need to use raw column names, the lack of an informative error message can make it hard to identify this issue when working on an app.

This PR proposes to detect such cases where the `xvar` or `yvar` is not in the data and give an informative error message. Perhaps additional text in error message about potential cause being ggplot2 and transformed variables would be even more informative.

Here is a minimal example app that upon brushing shows the cryptic error message with current shiny (and new more informative message with PR version):

```r
library(shiny)
library(ggplot2)

ui <- basicPage(
  plotOutput("plot1", brush = "plot_brush", height = 250),
  verbatimTextOutput("info")
)

  output$plot1 <- renderPlot({
    ggplot(mtcars, aes(x=log(wt), y=mpg)) + geom_point()
  })

  output$info <- renderPrint({
    brushedPoints(mtcars, input$plot_brush, allRows = TRUE)
  })
}

shinyApp(ui, server)
```

### DIFF
--- a/R/image-interact.R
+++ b/R/image-interact.R
@@ -1,3 +1,9 @@
+# Helper function for getting potentially transformed variable from input
+# data.frame to nearPoints or brushPoints
+eval_var_in_data <- function(var, data, envir) {
+  eval(parse(text = var), data, envir)
+}
+
 #' Find rows of data that are selected by a brush
 #'
 #' This function returns rows from a data frame which are under a brush used
@@ -86,18 +92,15 @@ brushedPoints <- function(df, brush, xvar = NULL, yvar = NULL,
   if (use_x) {
     if (is.null(xvar))
       stop("brushedPoints: not able to automatically infer `xvar` from brush")
-    if (!(xvar %in% names(df)))
-      stop("brushedPoints: `xvar` not in names of input")
     # Extract data values from the data frame
-    x <- asNumber(df[[xvar]])
+    x <- asNumber(eval_var_in_data(xvar, df, envir = parent.frame()))
     keep_rows <- keep_rows & (x >= brush$xmin & x <= brush$xmax)
   }
   if (use_y) {
     if (is.null(yvar))
       stop("brushedPoints: not able to automatically infer `yvar` from brush")
-    if (!(yvar %in% names(df)))
-      stop("brushedPoints: `yvar` not in names of input")
-    y <- asNumber(df[[yvar]])
+    # Extract data values from the data frame
+    y <- asNumber(eval_var_in_data(yvar, df, envir = parent.frame()))
     keep_rows <- keep_rows & (y >= brush$ymin & y <= brush$ymax)
   }
 
@@ -249,14 +252,9 @@ nearPoints <- function(df, coordinfo, xvar = NULL, yvar = NULL,
   if (is.null(yvar))
     stop("nearPoints: not able to automatically infer `yvar` from coordinfo")
 
-  if (!(xvar %in% names(df)))
-    stop("nearPoints: `xvar` not in names of input")
-  if (!(yvar %in% names(df)))
-    stop("nearPoints: `yvar` not in names of input")
-
   # Extract data values from the data frame
-  x <- asNumber(df[[xvar]])
-  y <- asNumber(df[[yvar]])
+  x <- asNumber(eval_var_in_data(xvar, df, envir = parent.frame()))
+  y <- asNumber(eval_var_in_data(yvar, df, envir = parent.frame()))
 
   # Get the coordinates of the point (in img pixel coordinates)
   point_img <- scaleCoords(coordinfo$x, coordinfo$y, coordinfo)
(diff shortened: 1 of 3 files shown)


jcheng5:
Hi @AliciaSchep, thanks for the PR! It looks like you haven't signed an RStudio CLA before--do you mind doing that? https://github.com/rstudio/shiny/blob/master/CONTRIBUTING.md Sorry for the inconvenience.

AliciaSchep:
@jcheng5 Have now emailed the CLA - thanks!

schloerke:
I really like exposing a better error message, but I believe this doesn't address the underlying cause.

What about replacing the calculations of `x` and `y` with an evaluation call?  In your example, the brush value coming has already been log transformed.  The values we're trying to subset against should be transformed as well, right?

```r
# somewhere in image-interact.R
eval_var_in_data <- function(var, data, envir) {
  eval(parse(text = var), data, envir)
}
```
```r
  # in brushedPoints()
  if (use_x) {
    if (is.null(xvar))
      stop("brushedPoints: not able to automatically infer `xvar` from brush")
    # Extract data values from the data frame from the calling frame of brushedPoints
    x <- asNumber(eval_var_in_data(xvar, df, parent.frame()))
    keep_rows <- keep_rows & (x >= brush$xmin & x <= brush$xmax)
  }
```
(Copied for all situations where `x` or `y` values are retrieved)


While normally I would would push for the `envir` param to be able to be set by the user, I believe it would only complicate things.  By evaluating with the parent frame, we can get access to custom functions...

```r
library(ggplot2)

ui <- basicPage(
  plotOutput("plot1", brush = "plot_brush", height = 250),
  verbatimTextOutput("info")
)

  barret <- function(...) log(...)
  output$plot1 <- renderPlot({
    ggplot(mtcars, aes(x=barret(wt), y=mpg)) + geom_point()
  })

  output$info <- renderPrint({
    brushedPoints(mtcars, input$plot_brush, allRows = TRUE)
  })
}

shinyApp(ui, server)
# evaluates on barret(wt) in the dataset mtcars and knows about the fn `barret()`
```

AliciaSchep:
thanks @schloerke - addressing the underlying issue would indeed be better than giving an error message!  I added the helper function & change you suggested, making the switch to calling `eval_var_in_data` for both xvar and yvar in both `nearPoints` and `brushedPoints`. I also updated the documentation for `xvar,yvar` and to remove the section about the limitation with computed variables.

Given that shiny already imports rlang, an alternative for `eval(parse(text = var), data, envir)` could be `rlang::eval_tidy(rlang::parse_expr(var), data = data, env = envir)` but I'm not sure there's any benefit

If these were new functions, I might think that having `xvar` and `yvar` take in expressions rather than strings might be preferable but given that the functions currently accept strings I would guess that would be a non-starter...

jcheng5:
It's not safe to eval xvar/yvar since they are (potentially malicious) input from the client. Correct me if I'm wrong, this situation can always be dealt with by mutating the data frame before passing it to ggplot? If so, then I think the original commit to provide a better error is the way to go.

schloerke:
@jcheng5 

True, eval is not safe, but this should be determined from the `x` and `y` `aes()` values directly.  Ah! Forgot about user input within an application.

`ggplot2` currently [uses `rlang::eval_tidy`](https://github.com/tidyverse/ggplot2/blob/484a2204dec494d9a8d6fadc804bd2213934fe63/R/layer.r#L222)  to calculate the data internally.

The _calculated_ variable can be fixed by making a new variable before making the plot and calling the new variable directly.  If we go that route, I'd like to know the missing variable name in the error message. ``"Can't find `xvar` ('ABC') in names of `df`"``

AliciaSchep:
@jcheng5 Makes sense re undesirability of `eval` in case of user input.  What about only doing `eval` in the case when `xvar` and `yvar` are inferred from the plot rather than provided by the user? Or is there still risk in that case?

jcheng5:
@AliciaSchep I believe there's still risk in that case. I'm not looking at the code right at this moment, but I believe even when `xvar`/`yvar` are inferred from the plot, that information is sent to the browser and then sent back to the server as part of the `input$<brush_id>` object. So in effect, it's all untrusted user content (where "user" in this case is not the Shiny app author--whom the R process always trust implicitly--but the user on the other end of the network connection).

jcheng5:
Looks good to me, and I can verify the CLA has been received. @schloerke if you have remaining objections let me know, otherwise let's merge.

jcheng5:
(Travis failure is the ggplot2 3.0.0 thing that is affecting all branches.)

AliciaSchep:
Thanks @jcheng5 for clarification re risk

PR #2176, opened 2018-09-05
Fix width option for textAreaInput

jreduardo:
The style in the `textarea` inherits the styles of the parent div. Since the parent div has a `shiny-input-container` class with a specified width, the width option in the `textAreaInput` of the percent type doesn't work. 

This PR modifies the width in the parent div to make the width options declared as percentage work.


wch:
Thanks for the PR. This looks like a duplicate of #2049.

PR #2179, opened 2018-09-10
Resolves #2172: Support data.frame as choices for selectize input

### LINKED ISSUE #2172: Custom label and html display in server-side selectize elements
In the current shiny CRAN version, it's possible to specify values, labels, and displayed html for server-side selectize elements separately. As described in 

https://shiny.rstudio.com/articles/selectize.html

Looks like this option was dropped in the current github version, which would be a pity. Here's an example working with the CRAN version:

```
library(shiny)

runApp(list(
  ui = bootstrapPage(
    selectizeInput("sel1", label = "Selectize 1", choices = NULL)
  ),
  
    valueStrs <- as.character(1:10)
    labelStrs <- paste("Label", valueStrs)
    htmlStrs <- paste0("<b>", valueStrs, "</b>")
    
    choiceFrame <- data.frame(value = valueStrs, label = labelStrs, 
                              html = htmlStrs, stringsAsFactors = FALSE)
    selected <- valueStrs[2]
    
    updateSelectizeInput(
      session, "sel1", server = TRUE, 
      choices = choiceFrame, 
      selected = selected, 
      
      options = list(render = I(
        '{ option: function(item, escape) {
            return "<div>Select " + item.html + "</div>";
            }
         }'
      ))
    )
    
    observe({
      print(input$sel1)
    })
    
  }
))

```

The dev version seems to interpret the choices as list and tries to group according to the column names (which is the documented behavior). One could specify some HTML as names(choices) and use a custom render function to display this (item.label). However, the escaped HTML is then also shown as label...

```
# test with github version
runApp(list(
  ui = bootstrapPage(
    selectizeInput("sel1", label = "Selectize 1", choices = NULL)
  ),
  
    valueStrs <- as.character(1:10)
    names(valueStrs) <- paste0("<b>", valueStrs, "</b>")
    selected <- valueStrs[2]
    
    updateSelectizeInput(
      session, "sel1", server = TRUE, 
      choices = valueStrs, 
      selected = selected, 
      
      options = list(render = I(
        '{ option: function(item, escape) {
            return "<div>Select " + item.label + "</div>";
            }
         }'
      ))
    )
    
    observe({
      print(input$sel1)
    })
    
  }
))
```

Is there a way to get back to the old behavior or to specify more data (columns) which could be used in the render function to generate HTML?

Thanks

landesfeind:
The support for data.frames is very helpful when providing custom renderer for layout of the drop-down options. As the internal data structure for the server-side selectize is a data.frame anyway, this is additionally convenient (see  #2172)

The test-app below is an extension of #2102 which initially dropped the support for data.frames.

```
library(shiny)
bigvec <- paste0("a", 1:1e5)
data_frame <- data.frame(
  label = c("Foo", "Bar"),
  value = c("foo", "bar"),
  html = c("F<span style=\"font-weight: bold\">oo</span>", "<span style=\"font-weight: bold\">B</span>ar")
)

test_set <- list(
  "Unnamed vector" = c(1, 2),
  "Unnamed list" = list(1, 2),
  "Nested list" = list(a = 1, B = list(B = 2, C = 4, 5), c = list(3)),
  "Data frame (server-side only)" = data_frame,
  "Data frame HTML-Label (server-side only)" = data_frame,
  "Data frame Custom Render (server-side only)" = data_frame,
)

ui <- fluidPage(
  sidebarPanel(
    checkboxInput("server", "Force server-side filtering"),
  ),
  mainPanel(
    verbatimTextOutput("txt")
  )
)

  use_server_side <- shiny::reactive({
    (!is.null(input$server) && input$server) ||
      grepl("server-side", input$set)
  })

  observe({

    options = list()
    if (grepl("Data frame HTML-Label", input$set)) {
      options = list(labelField = "html")
    }
    if (grepl("Custom Render", input$set)) {
      options[["render"]] = I("{
              option: function(item, escape) {
                return '<div>' + item.html +'</div>';
              }
          }")
    }

    updateSelectizeInput(
      session,
      "select",
      choices = test_set[[input$set]],
      selected = NULL,
      options = options,
      server = use_server_side()
    )
    cat("done\n")
  })

  output$txt = renderPrint({
    input$select
  })
}

# not be as fast.
shiny::runApp(shinyApp(ui, server), launch.browser = T)
```

### DIFF
--- a/R/update-input.R
+++ b/R/update-input.R
@@ -640,6 +640,8 @@ updateSelectizeInput <- function(session, inputId, label = NULL, choices = NULL,
     session$sendInputMessage(inputId, list(config = as.character(cfg)))
   }
   if (!server) {
+    if (is.data.frame(choices))
+      stop("Using a data frame for choices is only allowed for server-side filtering")
     return(updateSelectInput(session, inputId, label, choices, selected))
   }
 
@@ -670,6 +672,8 @@ updateSelectizeInput <- function(session, inputId, label = NULL, choices = NULL,
       lab[empty_names_indices] <- as.character(choices[empty_names_indices])
     }
     data.frame(label = lab, value = choices, stringsAsFactors = FALSE)
+  } else if (is.data.frame(choices)) {
+    choices
   } else {
     # slow path for nested lists/optgroups
     list_names <- names(choices)


schloerke:
@landesfeind I can not manually change the base merge branch. 😞 

Do you mind resubmitting the PR? Sorry for the hassle. Thank you

landesfeind:
Hi @schloerke, after the Shiny team ignored this +4 lines PR for 3 years, I kind of hassle to do it. It does not seem that there is any interest...
The code is dead simple and nothing "intellectual property"-worthy. In case someone would be interest, that person is free to just use it...

PR #2184, opened 2018-09-16
Temporary fix for test-plot-coordmap.R test

schloerke:
Removed ggplot2 remote as it's not needed anymore.

-------------------------------

Test now allows for different values of ggplot2 domain. **THIS IS A DUCT TAPE FIX** for the dev version of ggplot2.

Originally caused by tidyverse/ggplot2#2832

Need to wait for tidyverse/ggplot2#2821 to be merged.

Expecting a bug fix release between 3.0.0 and when this bug will be fixed, so can not compare against pkg versions. :-/

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -90,7 +90,6 @@ Suggests:
 URL: http://shiny.rstudio.com
 BugReports: https://github.com/rstudio/shiny/issues
 Remotes:
-    tidyverse/ggplot2,
     rstudio/httpuv
 Collate:
     'app.R'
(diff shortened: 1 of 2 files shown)


schloerke:
Looking at it with Winston, https://github.com/tidyverse/ggplot2/pull/2821 will rename the coord-trans range to backtransform_range, which should solve `summarise_layout` output, which calls `$range`

jcheng5:
To be clear, we shouldn't merge this, as the fix for this problem will land on the ggplot2 master branch. This problem never existed with any CRAN version of ggplot2; we were seeing CI errors because we had Remotes:tidyverse/ggplot2 (which I've now removed).

However, keeping this open to track the issue until the ggplot2 PR above is merged, then we can close this one.

Issue #2156, opened 2018-08-13
Fontawesome


zuzuzuzz:
Can you release the latest version of Fontawesome into dev?

rich-iannone:
@zuzuzuzz There are some official alternatives to try until Shiny proper gets a Fontawesome update. One option is to use the *fontawesome* package. It works well in Shiny, and it's updated to the latest free icons. The README (with a Shiny example) is at: https://github.com/rstudio/fontawesome.

nohlsen:
If you need icons in your valuebox you can do this:
```
infoBox(
  "title", 
  "value", 
  icon = tags$i(fa(name = "box", fill = "white")),
  color = "teal", fill =TRUE, width = 3
)
```

CalaoJuanPablo:
duplicate #1966

AveugleVisionnaire:
Thanks @rich-iannone. This save my day. Hovewer, it doesn't work on Microsft IE11. 
I'm using your suggeted alternative to display icon in my shinydashboard sidebar. It works fine with Chrome, Firefox and Edge. But on IE11 it only shows the icon instead of icon + menu item title.
Do you have any idea to fix this issue?
At least, can we have  menu item title displayed?

hlendway:
This isn't a great solution for icons in the menu bar because it doesn't pick up the proper styling. i.e. the icon is set at one color and not highlighted properly when selected or not selected.  Also the spacing between the icon and the menu item name is off.
![image](https://user-images.githubusercontent.com/25120859/45432295-239e4480-b66f-11e8-8aa5-a8c371f7938c.png)

jcheng5:
We've upgraded FA in Shiny on master, since the separate Font Awesome package raised more questions than we expected.

alandipert:
I'm closing this because the FA upgrade went out with Shiny 1.2, but feel free to re-open or make another issue if your needs remain unmet.

Issue #1966, opened 2018-03-03
Feature Request: Update font-awesome version


JesseVent:
https://github.com/rstudio/shiny/blob/174a1fe83437008bca69e48fe346470e5c09ecb9/R/bootstrap.R#L1547

I've noticed theres been huge changes the font-awesome icons in the last few versions they have released, the documentation for `icons()` is now is incorrect as most of the icons listed through their explorer will fail to load when using in shiny app.

Please update the font-awesome font files or the new svg implementation

nteetor:
Which link are you following to the list of Font Awesome icons? If one heads over to the Font Awesome 4.7 website, see https://fontawesome.com/v4.7.0/, the icons listed ought to all work with `icon()`. The updates in Font Awesome 5 are not trivial. For now the best solution may be to make sure Shiny's documentation points to the Font Awesome 4.7 website.

jcheng5:
https://www.drupal.org/project/fontawesome/issues/2930071

nzcoops:
Just adding a +1 for this request.
It gets confusing with the shiny help pages pointing to a page of icons where 90% of them are unusable with no clear indication (to new users) as to why some wouldn't/wouldn't work.

JesseVent:
I just committed some changes into a fork at https://github.com/JesseVent/shiny/commit/9ce6f87094ba5f6d4ca00455296a02d83454172d which appear to work, build passes all tests and CRAN checks. So could be something to look into - Unfortunately I couldn't think of a way to provide fallback support for 4.7.0 other than using a mapping document in a csv file.

Seems to work pretty well in the scenarios I tested, definitely renders in-app. Hope it helps.

```r
icon <- function(name, class = NULL, lib = "font-awesome", style="fas") {
  prefixes <- list(
    "font-awesome" = "fa",
    "glyphicon" = "glyphicon"
  )
  prefix <- prefixes[[lib]]

  # determine stylesheet
  if (is.null(prefix)) {
    stop("Unknown font library '", lib, "' specified. Must be one of ",
         paste0('"', names(prefixes), '"', collapse = ", "))
  }

  # Add fallback support for 4.7.0 by checking for legacy icons names and replacing with 5.2.0 names
  if ( prefix == "fa" && !is.null(name)) {
    mapfile <- system.file("www","shared","font-awesome", "fa-mapping.csv", package = "shiny")
    fa_lookup  <- read.csv(mapfile, stringsAsFactors = FALSE)
[...]

PR #2186, opened 2018-09-17
Upgrade FontAwesome to 5.3.1

alandipert:
- This upgrades FontAwesome to a new major (by itself, breaking) version, but
  is backwards compatible because we include the [v4-shims](https://fontawesome.com/how-to-use/on-the-web/setup/upgrading-from-version-4) CSS that maps
  old names to new.
- `icon` tags are now browsable
- This is a step toward full V5 adoption that doesn't require us to
  come up with a plan for deprecating V4 icon names.

(Related: #2156 and #1966)

Confirmed that the FontAwesome CSS and webfont licenses didn't change between 4 and 5, added CC license for .svgs we now also include.


R/bootstrap.R
```diff
@@ -1545,10 +1555,16 @@ icon <- function(name, class = NULL, lib = "font-awesome") {
   iconTag <- tags$i(class = iconClass)
 
   # font-awesome needs an additional dependency (glyphicon is in bootstrap)
-  if (lib == "font-awesome") {
-    htmlDependencies(iconTag) <- htmlDependency(
-      "font-awesome", "4.7.0", c(href="shared/font-awesome"),
-      stylesheet = "css/font-awesome.min.css"
+  if (lib %in% c("font-awesome", "fontawesome")) {
```
alandipert: Hm I'm not sure about this. Perhaps we should do this:

- If it's `font-awesome`, load the shim
- If it's `fontawesome`, don't load the shim

Then in subsequent versions of Shiny that ship V5 w/out the shim, specifying `font-awesome` is an error and `fontawesome` is the new deault `lib`

eawachtel:
I'm really confused on what the current status of fontawesome in shiny is?  I'm using shiny v 1.0.5 and have download the fontawesome 4.7 and placed in www/shared.  I can not get most icons to work.  I see the svg files for the icons I want, but they do not work.  Any help appreciated

jcheng5:
@eawachtel Please file a separate issue if you're having trouble with icons in 4.7.0. It'd be great if you could include a minimal reproducible example. Thanks!

eawachtel:
I most certainly can provide an example of my code.

I use the below quite a bit just calling out different icon names in the
fontawesome library
actionButton("closelogin", label = NULL, icon = icon("times", lib=
"font-awesome"))

This works fine.

but if I reference a different icon "gas-pump"
actionButton("closelogin", label = NULL, icon = icon("gas-pump", lib=
"font-awesome"))

I get nothing

alandipert:
@eawachtel I think `gas-pump` doesn't work because it was added in FontAwesome 5, and the currently-released version of Shiny only has FontAwesome 4.7. Once this PR is merged and we've released Shiny 1.2, `gas-pump` will work.

R/bootstrap.R
alandipert: A half-baked scheme for supporting 4 and 5 in a subsequent release, but there's no reason to do it now. Removed and simplified.

R/bootstrap.R
```diff
@@ -1545,10 +1555,16 @@ icon <- function(name, class = NULL, lib = "font-awesome") {
...
+    htmlDependencies(iconTag) <- list(
+      htmlDependency(
+        "fontawesome-all", "5.3.1", c(href="shared/fontawesome"),
+        stylesheet = "css/all.min.css"
+      ),
+      htmlDependency(
+        "fontawesome-v4-shims", "5.3.1", c(href="shared/fontawesome"),
+        stylesheet = "css/v4-shims.min.css"
+      )
```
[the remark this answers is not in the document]
alandipert: Makes sense. I did it 👍

jcheng5:
@alandipert Are we missing "CC BY 4.0 License" in the LICENSE file? It looks like it applies to the SVG data, which we appear to be bundling.

PR #2187, opened 2018-09-17
Fix spurious duplicate values being sent by selectInput

### LINKED ISSUE #2162: SelectInput sending changes to server when search box is empty
Referencing an old issue: https://github.com/rstudio/shiny/issues/404

As @jjallaire pointed out,

> (1) By default when the user empties out the select input Shiny sends no
changes to the server.

> (2) When the select input loses focus and is empty the last selection is
re-inserted into the control (again no changes are sent to the server).

This was the exact behaviour upto and including version 1.0.5.

When I updated to version 1.1 Shiny selectInput has started sending changes to server when the user hits backspace. This is causing all the downstream reactive dependancies to reexecute when the user hits backspace to clear the search box.

Reproducible example:

```
shinyApp(
  ui = fluidPage(
    selectInput("state", "Choose a state:",
                list(`East Coast` = c("NY", "NJ", "CT"),
                     `West Coast` = c("WA", "OR", "CA"),
                     `Midwest` = c("MN", "WI", "IA"))
    ),
    textOutput("result")
  ),
    output$result <- renderText({
      cat("Running...")
      paste("You chose", input$state)
    })
  }
)
```

When you hit backspace in shiny version < 1.1 `Running...` is not printed, however in version 1.1 it is printed everytime.

jcheng5:
Fixes #2162. I described the root cause in my [comment](https://github.com/rstudio/shiny/issues/2162#issuecomment-422190916).

The fix I made is to explicitly detect this condition in the selectize input binding, and suppress the event that (temporarily) sets the value to `""`. So the selectize input binding does cause an `"NY"` event to be raised when you backspace, but now it's correctly squelched by the client side dupe-detecting logic.

## Testing notes

Here is a slightly improved reprex of the original issue. Without this fix, you will see `"NY"` printed at first; and if you click into the select box and backspace, you'll see `"NY"` printed again. With this fix, you'll only see `"NY"` printed the first time. In no case should you be able to get it to show an empty string for the value.

```r
shinyApp(
  ui = fluidPage(
    selectInput("state", "Choose a state:",
      c("NY", "NJ", "CT")
    ),
    textOutput("result")
  ),
    output$result <- renderText({
      print(input$state)
      paste("You chose", input$state)
    })
  }
)
```

This example adds a "Please choose one" prompt, which causes `""` to now be a valid state. Both before and after this fix, backspacing should cause `""` to be printed at the console, and "Please choose one" should appear in the select box.

```r
shinyApp(
  ui = fluidPage(
    selectInput("state", "Choose a state:",
      c("Please choose one" = "", "NY", "NJ", "CT")
    ),
    textOutput("result")
  ),
    output$result <- renderText({
      print(input$state)
      paste("You chose", input$state)
    })
  }
)
```

This example is like the first one except it uses `selectizeInput` instead of `selectInput`; the defaults are slightly different, in that `selectizeInput` allows you to have nothing selected. So with and without the fix, you should see backspace cause `""` to be printed at the console, and clicking focus away from the select box control should leave it empty.

```r
shinyApp(
  ui = fluidPage(
    selectizeInput("state", "Choose a state:",
      c("NY", "NJ", "CT")
    ),
    textOutput("result")
  ),
    output$result <- renderText({
      print(input$state)
      paste("You chose", input$state)
    })
  }
)
```


srcjs/input_binding_select.js
```diff
@@ -115,7 +115,14 @@ $.extend(selectInputBinding, {
     $(el).trigger('change');
   },
   subscribe: function(el, callback) {
+    var thiz = this;
     $(el).on('change.selectInputBinding', function(event) {
```
alandipert: I think if you use an arrow function for the `change.selectInputBinding` callback then you don't need to explicitly capture `this` (just refer to `this` in arrow function).

jcheng5: Thanks, fixed. I thought we couldn't use ES2015 here but you're right, we can.

PR #2195, opened 2018-09-25
Fix custom selectize rendering

### LINKED ISSUE #2192: Selectize optgroup bug in shiny-examples 023-optgroup-server
https://github.com/rstudio/shiny-examples/tree/master/023-optgroup-server

jcheng5:
Fixes #2192. Two problems here:

1. It's not documented but apparently we supported data frames for
   choices in updateSelectInput/updateSelectizeInput (it doesn't
   appear to work correctly for selectInput/selectizeInput though).
   This was used in 023-optgroup-server as well as by the user who
   reported #2172.
2. The example in 023-optgroup-server was also counting on the
   default value of optgroupLabelField, which (starting post-Shiny
   v1.1) was being set to a new default of "group". That now won't
   happen unless optgroupField is also blank. I'm less confident
   about the ramifications of this change. The selectize docs with
   the relevant bits are here:
  https://github.com/selectize/selectize.js/blob/master/docs/usage.md#data_searching


jcheng5:
In particular I'm quite unsure whether we ever want to override all three `optgroup*` fields unless all three fields were missing in the first place.

jcheng5:
Modified version of @wch test app (we should add this to shiny-examples if it isn't already). My version adds a "data frame" option for the input.

```r
library(shiny)

bigvec <- paste0("a", 1:1e5)

test_set <- list(
  "Unnamed vector" = c(1, 2),
  "Unnamed list" = c(1, 2),
  "Nested list" = list(a = 1, B = list(B = 2), c = list(3)),
  "Big nested list (server-side only)" = nested_biglist,
  "Data frame (server-side only)" = data.frame(label = c("a", "B"), value = c(1, 2))
)

ui <- fluidPage(
  sidebarPanel(
  ),
  mainPanel(
    verbatimTextOutput("txt")
  )
)

      return(FALSE)
      return(FALSE)
    
    TRUE
  }
  
  observe({
      choices = test_set[[input$set]], selected = NULL, server = input$server
    )
    cat("done\n")
  })
  
  output$txt = renderPrint({
    if (!validateSelection())
    input$select
  })
}

# not be as fast.
```

alandipert:
shiny-examples PR that adds @jcheng5's app https://github.com/rstudio/shiny-examples/pull/101

PR #2198, opened 2018-09-26
Brush dims

schloerke:
Wait for image to be loaded in browser before initializing handlers (reverts similar behavior in 3354a47)

Default the height and width to the image clientHeight and clientWidth

Use raw image clientWidth and clientHeight instead of container clientWidth and clientHeight

Prevents being able to brush on non image areas


srcjs/output_binding_image.js
```diff
@@ -354,6 +364,10 @@ imageutils.initCoordmap = function($el, coordmap) {
     };
   }
 
+  // if no dim height and width values are found, set them to the image height and width
+  coordmap.dims.height = coordmap.dims.height || img.clientHeight;
+  coordmap.dims.width = coordmap.dims.width || img.clientWidth;
```
jcheng5: In cases where img has padding, `clientHeight`/`clientWidth` will include that padding. It'd be cool (very very low priority) to strip out that padding in this calculation so we can give more accurate CSS pixel level coordinates. But this would only happen today if you go out of your way to add padding to img.

PR #2200, opened 2018-09-27
Fix brushes not being properly cleared

### LINKED ISSUE #2197: Input brush isn't properly cleared when new brush is drawn
Observed in RStudio viewer 1.1.456 with v1.2-rc in https://github.com/rstudio/shiny-examples/tree/master/108-module-output

module

Interestingly, the issue doesn't occur in Chrome (with v1.2-rc), nor does it appear in the viewer with shiny v1.1

jcheng5:
Actually three separate issues addressed. Fixes #2197.

- brush.importOldBrush() was not being called anymore, due to it being
  registered as a load handler after the image was already loaded (this
  was a very recent regression, less than 24 hours old).
- Each time the brush changes, the plot is redrawn twice. This was
  because importing the old brush introduced floating point errors that
  led to a slightly different new brush being created.
- Sometimes the image's load event wasn't firing at all. This is due to
  behavior in WebKit where assigning an image's src to its existing
  value is a no-op.


jcheng5:
Should namespace `dragstart`

jcheng5:
Should use underscores as separators for event listener namespace

schloerke:
@jcheng5 Have added the changes after talking with winston.  I can not find a repro that breaks on this branch.  

I have implemented the two comments for namespacing `dragstart` and also making the namespace underscores.  

I have also changed the init `on()` method to a single `one()` method.  While we haven't seen any bad behavior, nice to be safe.

The last change was to `img.src`.  It was changed to `img.getAttribute("src")`.  Also, these src values will only be data64 files, never a relative link as it may only be produced from renderImage or renderPlot.

PR #2203, opened 2018-10-04
Fixes #2142 DnD file upload on Firefox 57+

### LINKED ISSUE #2142: Drag/drop fileInput gives NULL as filename under Ubuntu (ok on Win)
Latest Shiny from CRAN, same result from github as of today.

Ok on Edge Browser, Chrome. The following is for Firefox 61.0.1, also tested  with alle Extensions off.

- Use example from fileInput to test
- Using "Browse" button always works
- Drag/drop Works Ok under Windows
- Drag/drop gives NULL filename under Ubuntu/Shiny server 

Worked in previous versions (don't know exactly in which version it started to fail)

alandipert:
- Firefox 57+ appears not to fire a change event when the `files` field is modified,
  which prevented uploads from occuring. This commit triggers a change event manually
  and doesn't impact the functioning of other browsers.
- Namespaced events were used improperly in existing code which resulted in "jank"
  on Firefox. Namespaces shouldn't have been attached to events generated by the browser.
- The "drop" and "dragleave" handlers are now separate. This fixes a problem
  on Firefox where the drop event wasn't reliably changing the state of the input
  so it no longer glowed.


srcjs/input_binding_fileinput.js
```diff
@@ -309,72 +309,67 @@ $.extend(fileInputBinding, {
...
-    $el.off(".dragHover");
-  },
-  _enableDocumentEvents: function() {
-    let $doc = $("html");
-
-    this._enableDraghover($doc);
-    $doc.on({
-      "draghoverstart.fileDrag": e => {
-        $fileInputs.trigger("showZone.fileDrag");
+  _enableDraghover: function(el) {
```
jcheng5: I think it's important to add a comment here (or at least restore the old one) indicating why `_enableDraghover` is necessary in the first place--it wasn't obvious to me at all.

srcjs/input_binding_fileinput.js
```diff
@@ -309,72 +309,67 @@ $.extend(fileInputBinding, {
...
+          this._zoneOf($fileInputs).addClass(ACTIVE);
+        },
+        "draghover:leave.draghover": e => {
+          this._zoneOf($fileInputs).removeClass(ACTIVE);
+        },
+        "draghover:drop.draghover": e => {
+          this._zoneOf($fileInputs)
+            .removeClass(OVER)
+            .removeClass(ACTIVE);
+        }
```
jcheng5: Wow, I didn't know you could do that

srcjs/input_binding_fileinput.js
```diff
@@ -402,10 +397,13 @@ $.extend(fileInputBinding, {
       // (Chrome, Safari)
       $el.val("");
       el.files = e.originalEvent.dataTransfer.files;
+      // Recent versions of Firefox (57+, or "Quantum" and beyond) don't seem to
+      // automatically trigger a change event, so we trigger one manually here.
+      // On browsers that do trigger change, this operation appears to be
+      // idempotent, as el.files doesn't change between events.
+      $el.trigger("change");
```
jcheng5: Can you confirm that this doesn't result in the file being uploaded twice?

alandipert: I made an app with this in my `server` function:
```R
observeEvent(input$file1, {
  cat("file uploaded\n")
})
```
I saw `file uploaded` in the R console only once per drop.

srcjs/input_binding_fileinput.js
alandipert: Added explanatory comment in 87f16e3c3bbb5e55edaed4881cd820665412f5b8

jcheng5:
@alandipert BEFORE YOU COMMIT--can you rebase these changes onto `v1.2-rc` instead, and merge to that? Then we can bring back all of those fixes onto master at once.

alandipert:
@jcheng5 I fixed a tiny Grunt warning and then `merge --squashed` onto v1.2-rc via fcd09e2bae6d4ac2a1c5d006cbe0d315ab8a371b

PR #2211, opened 2018-10-16
Update datepicker

### LINKED ISSUE #2204: updateDateInput changes the initial date
Within R shiny, `dateInput("date", "date:", value = "2018-10-07"))` will always dispaly `2018-10-06` in the UI all the time.

**Below is my actual case:**
I set the server time zone to Sys.setenv(TZ=" Australia/Melbourne") in my R code.
2018-10-07 is the DayLight saving transition date in Australia.

I run the local test of my Shiny APP. However, when the client side updateDateinput with a value of 2018-10-07 from my DataBase, it immediately changed to 2018-10-06.

This has caused so much trouble for me. Not sure where this issue comes from, browser, Shiny Package?

I think the date component calculation from Shiny might have some problems.

wch:
This fixes #2204.


jcheng5:
The unminified copy of bootstrap-datepicker.js is gone?

jcheng5:
(I was hoping it would remain just so we could see diffs)

wch:
Previously, we copied over the .js file and then minified it ourselves, but in this PR I changed it to copy over the .min.js file. (The `htmlDependency` didn't offer an option to use the un-minified version.) 

Now that I look at it, we should still do it the previous way, because the minification process also adds in all the locale .js files. https://github.com/rstudio/shiny/blob/3cea5fb2/tools/Gruntfile.js#L143-L148

wch:
In a separate discussion, we decided to hold off on this update until after the 1.2 release.

Issue #2208, opened 2018-10-12
Broken font-awesome icon in shinydashboard


wch:
This is from rstudio/shinydashboard#297.

On `master` and `v1.2-rc`, the following app has a broken hamburger icon:

```R
library(shiny)
library(shinydashboard)

ui <- dashboardPage(
  dashboardHeader(),
  dashboardSidebar(),
  dashboardBody()
)

server <- function(input, output) { }

shinyApp(ui, server)
```

![image](https://user-images.githubusercontent.com/86978/46882973-11eac100-ce16-11e8-9395-8d1bfaf60e93.png)

We should look at this carefully, because it might be a sign that other users of FA icons could have problems

ismirsehregal:
Is this still part of the same issue?:

![image](https://user-images.githubusercontent.com/36849480/47362912-d681a980-d6d5-11e8-9583-c4daa5767742.png)

```
library(shiny)

ui <- fluidPage(
  p(),
  actionButton("Btn", "Btn", icon = icon("creative-commons-sampling", lib = "font-awesome"))
)

server <- function(input, output) {}

shinyApp(ui = ui, server = server)
```

wch:
@ismirsehregal Can you try with Shiny 1.2 RC?

```
devtools::install_github('rstudio/shiny', 'v1.2-rc')
```

ismirsehregal:
@wch same output using shiny     * 1.2.0   2018-10-23 Github (rstudio/shiny@2629e59)

ismirsehregal:
Thanks for investigating that fast again!

PR #2218, opened 2018-10-23
Fix Font-Awesome prefix classes

wch:
This fixes an issue raised in https://github.com/rstudio/shiny/issues/2208#issuecomment-432240206.

This PR:
* Adds a script for updating Font-Awesome. The script also extracts metadata related to the prefix class, "fa", "fab", "fas", or "far"  and saves in a data frame `fa_icons`. This data frame has a lookup table for the default prefix class to use. It preferentially uses "fa", then "fab", "fas", and then "far".
* Updates Font-Awesome to 5.4.1.
* Changes the `icon()` function to use the prefix class lookup table.

Here's part of the `fa_icons` data. The `default` column contains the CSS prefix class that will be used for each name:
image

Comments about the API:
* There is presently no way to choose which prefix class to use. It is determined by the lookup table in `fa_icons`. If the default is "fa", you only can have "fa". We may want to expose this now, as long as we're making the change to support V5. Users might find it surprising that some V5 icons don't look the way they expect.
* It always prefers the V4 shim version of an icon (with the `fa` prefix class) if available, but this is probably undesirable in the long run. We'll eventually want to default to the V5 version of an icon (with, say, `fab` or `fas`), and perhaps we should make that the default now.

The following text is copied from the update script, and it explains the situation in more detail:

> In Font-Awesome 5, there are four different CSS prefix classes: "fab", "fas", "far", and "fal" (which is only in the Pro version, so we won't use it here). These are for "brands", "solid", "regular", and "light", respectively. Each icon supports various combinations of these prefixes.
>
> The CSS prefix class for Font-Awesome 4 is "fa". Some of the icon names from V4 do not exist in V5, like "bank". Some of them do exist in V5, but the name now refers to a very different icon. This is true of "calendar". The v4-shims.css file will make "fa fa-calendar" (this is V4 syntax) display one icon (which is very similar to the V4 version), and "fas fa-calendar" (V5 syntax) display a different icon.
>
> In this version of Shiny, we decided to minimize visible changes for the same code. That means that if someone uses icon("calendar"), they will get "fa fa-calendar". In the future we may change it so that it defaults to "fas fa-calendar", which will result in the V5 icon.

### DIFF
--- a/tests/testthat/test-bootstrap.r
+++ b/tests/testthat/test-bootstrap.r
@@ -313,3 +313,18 @@ test_that("icon() provides correct Font-Awesome v4 and v5 output", {
     '<i class=" abc xyz"></i>'
   )
 })
+
+
+test_that("icon() provides correct output for glyphicon", {
+  # A name that's in glyphicon but not font-awesome
+  expect_identical(
+    as.character(icon("baby-formula", lib = "glyphicon")),
+    '<i class="glyphicon glyphicon-baby-formula"></i>'
+  )
+
+  # A name that's in both glyphicon and font-awesome
+  expect_identical(
+    as.character(icon("calendar", lib = "glyphicon")),
+    '<i class="glyphicon glyphicon-calendar"></i>'
+  )
+})


jcheng5:
@wch Did you see the test failures? I can repro it:

```r
navbarPage("App Title",
  tabPanel("Plot", icon = icon("bar-chart-o")),
  tabPanel("Summary", icon = icon("list-alt")),
  tabPanel("Table", icon = icon("table"))
)
```

wch:
Updated `fa_icons` data:

image

jcheng5:
Rather than sysdata.rda, perhaps have the script create a .csv--then we can diff

tools/updateFontAwesome.R
```diff
@@ -0,0 +1,142 @@
...
+      )
+    )
+  )
+
+
+fa_version <- version
+message("Writing fa_icons and fa_version to R/sysdata.rda")
+# Note: if in the future we add more objects to sysdata.rda, we'll have to do
+# this differently.
+save("fa_icons", "fa_version", file = find_package_root_file("R", "sysdata.rda"))
```
wch: `fa_icons` should be saved as CSV so we can easily see changes. Same for `fa_version`.

wch: I've added data-raw/fa_icons.csv and data-raw/fa_version.txt. They're both still saved in R/sysdata.rda, so that they're accessible within the package.

wch:
Holding off on merging until after 1.2 is out. Thoughts for the future:

* Add run-time validation for glyphicons as well.
* Add font-awesome and glyphicons picker app (not necessarily in this package).

PR #2257, opened 2018-11-11
fix dependency version since htmltools 0.3.6 is used

### LINKED ISSUE #2256: htmlDependency package parameter requires `htmltools` 0.3.6
https://github.com/rstudio/shiny/blob/5e2b40d3a92ea602d7187dafe49eba0d1e220cde/R/bootstrap.R#L1550-L1554

We should probably update the dependency on `htmltools` to be version >= 0.3.6, because per [this issue](https://community.rstudio.com/t/r-shinydashboard-instructions-doesnt-work/18000/2), it seems that `htmltools` 0.3.5 does not have the `package` parameter.

https://github.com/rstudio/shiny/blob/5e2b40d3a92ea602d7187dafe49eba0d1e220cde/DESCRIPTION#L73

colearendt:
Close #2256 

Obviously up for discussion if the use of the `package` parameter in `htmlDependency` should be avoided or if this was caused by some other issue (it was unclear to me in the `htmltools` repo when the `package` parameter was added... it looks like it was not present in 0.3.5). Otherwise, we can just bump up the version of `htmltools` that is used.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -70,7 +70,7 @@ Imports:
     jsonlite (>= 0.9.16),
     xtable,
     digest,
-    htmltools (>= 0.3.5),
+    htmltools (>= 0.3.6),
     R6 (>= 2.0),
     sourcetools,
     later (>= 0.7.2),


colearendt:
Random keyserver build fail. Restarting will probably pass, I expect. This is a trivial change! 

```
The command "sudo apt-key adv --keyserver ha.pool.sks-keyservers.net --recv-keys E298A3A825C0D65DFD57CBB651716619E084DAB9" failed and exited with 2 during .
```

jcheng5:
Thanks!

PR #2264, opened 2018-11-17
Add callback when creating a URL

nikolai-b:
I'd like to be able to shorten the URLs that we give to users as our state is quite long.  I wanted to add the ability to use a URL shorter or add the ability to customise the URL in someway.  I've not added documentation as I wasn't sure if this was something you'd be interested, if it is I'm happy to work a bit more on it.

Thanks.


wch:
I think that you could do this with something like :

```R
onBookmarked(function(url) {
    url <- doSomethingToUrl(url)
    showBookmarkUrlModal(url)
  }
)
```

PR #2284, opened 2018-12-17
Fix typo

chasemc:
"...use the JavaScript library selectize.js (https://github.com/selectize/selectize.js) ~~to~~ instead of the basic select..."


wch:
Thanks!

PR #2327, opened 2019-02-13
Exclude "session" from static path serving

### LINKED ISSUE #2325: File upload is throwing 400 error.
Run app 066-upload-file or 081-widgets-gallery and try uploading a file. It throws 400 error . 

Per Winston this could be related to httpuv static file serving code. 
screen shot 2019-02-07 at 2 16 47 pm

wch:
This fixes #2325. It depends on rstudio/httpuv#197.

To test:

```R
devtools::install_github("rstudio/httpuv@staticpath-exclude")
devtools::install_github("rstudio/shiny@staticpath-exclude")
```

Then run the test app from #2325. File uploads should now work.

image

### DIFF
--- a/R/app.R
+++ b/R/app.R
@@ -171,7 +171,9 @@ shinyAppDir_serverR <- function(appDir, options=list()) {
 
   wwwDir <- file.path.ci(appDir, "www")
   if (dirExists(wwwDir)) {
-    staticPaths <- list("/" = staticPath(wwwDir, indexhtml = FALSE, fallthrough = TRUE))
+    staticPaths <- list(
+      "/" = staticPath(wwwDir, indexhtml = FALSE, fallthrough = TRUE, exclude = "session")
+    )
   } else {
     staticPaths <- list()
   }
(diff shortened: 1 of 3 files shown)


wch:
Hm, it looks like this isn't sufficient for everything to work. For example, downloads don't work. The reason is here:

https://github.com/rstudio/shiny/blob/59221dfcf235d8bdfb53435d68f947fd87a43b5e/R/middleware-shiny.R#L61

This code rewrites the `PATH_INFO` for a request. If it starts out as `/session/d56638756cf46188e8f957c5886301a3/download/downloadData`, this code creates a subrequest with a `PATH_INFO` of `/download/downloadData`.

Excluding a `download` static path also could work for this specific case, but there are other targets like `uploadie` and [`dataobj`](https://github.com/rstudio/shiny/blob/59221dfcf235d8bdfb53435d68f947fd87a43b5e/R/shiny.R#L1882). Also, it seems likely that a user would want to use the `download` path for files to download.

Also, it looks like the reactlog stuff might also have a similar issue: https://github.com/rstudio/shiny/blob/59221dfcf235d8bdfb53435d68f947fd87a43b5e/R/middleware-shiny.R#L9-L24

R/app.R
```diff
@@ -90,13 +90,20 @@ shinyApp <- function(ui=NULL, server=NULL, onStart=NULL, options=list(),
...
+  )
+
   # Store the appDir and bookmarking-related options, so that we can read them
   # from within the app.
   shinyOptions(appDir = getwd())
   appOptions <- consumeAppOptions()
 
   structure(
     list(
+      staticPaths = staticPaths,
```
jcheng5: Why are we adding this hardcoded `staticPaths` to the shiny app object? I don't see where we are reading this?

PR #2343, opened 2019-02-27
addResourcePath: create staticPath object immediately. Fixes #2339

### LINKED ISSUE #2339: runApp fails if previous addResourcePath refers to now-deleted directory
```
dir.create("transient_directory")
addResourcePath("foo", "transient_directory")
unlink(recursive = TRUE, "transient_directory")
shiny::runExample("01_hello")
```

I encountered this after running `026-shiny-inline`; subsequent attempts to launch Shiny apps failed unless I restarted the process.

wch:
This fixes #2339. It creates the `staticPath` object immediately (which in turn calls `normalizePath`, which is where the error occurs when the path is missing), instead of doing it when the application starts.

### DIFF
--- a/R/server.R
+++ b/R/server.R
@@ -70,7 +70,7 @@ addResourcePath <- function(prefix, directoryPath) {
   }
 
   # .globals$resourcePaths persists across runs of applications.
-  .globals$resourcePaths[[prefix]] <- normalizedPath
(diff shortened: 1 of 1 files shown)


jcheng5:
@shalutiwari Here's a self-contained example I added to shiny-examples:
https://github.com/rstudio/shiny-examples/blob/rc-v1.3.0/148-addresourcepath-deleted/app.R

PR #2345, opened 2019-03-01
Short reactlog labels

schloerke:
We were shortening the value, never the label until within reactlog.  Now it's shortened to 250 chars for all shortened strings.

### DIFF
--- a/R/graph.R
+++ b/R/graph.R
@@ -512,11 +512,15 @@ MessageLogger = R6Class(
       self$reactCache[[reactObj$reactId]] <- reactObj
     },
     shortenString = function(txt, n = 250) {
-      if (nchar(txt) > n) {
-        paste0(substr(txt, 1, n - 3), "...")
-      } else {
-        txt
+      if (is.na(txt) || is.null(txt)) {
+        return("")
       }
+      if (nchar(txt) > n) {
+        return(
+          paste0(substr(txt, 1, n - 3), "...")
+        )
+      }
+      return(txt)
     },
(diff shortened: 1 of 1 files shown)


R/graph.R
```diff
@@ -511,7 +511,7 @@ MessageLogger = R6Class(
       if (identical(force, FALSE) && self$isNotLogging()) return(NULL)
       self$reactCache[[reactObj$reactId]] <- reactObj
     },
-    shortenString = function(txt, n = 100) {
+    shortenString = function(txt, n = 250) {
       if (nchar(txt) > n) {
```
cpsievert: Can `txt` be `NA` or `NULL`? If so, may want to wrap this in an `isTRUE()`

R/graph.R
```diff
@@ -511,12 +511,16 @@ MessageLogger = R6Class(
       if (identical(force, FALSE) && self$isNotLogging()) return(NULL)
       self$reactCache[[reactObj$reactId]] <- reactObj
     },
-    shortenString = function(txt, n = 100) {
+    shortenString = function(txt, n = 250) {
+      if (is.na(txt) || is.null(txt)) {
```
jcheng5: You might want to change this to `is.null(txt) && isTRUE(is.na(txt))`, otherwise you'll get a warning if `txt` is actually `NULL`.

schloerke: `&&` or `||` ?

jcheng5: Sorry, `||`, heh.

Issue #2329, opened 2019-02-15
networkD3 package and progress bars


baczychris:
Hi,
I have some issues with progress bars in Shiny and the package networkD3. If I try to use a progress bar and the networkD3 package, the bar shows up once but not in subsequent calls. But each time the slider is used the progess bar should appear, which is the case without the sankey diagram.
I tested in Chrome.
Here is a minimum example code:

```
require(shinyjs)
require(shiny)
require(networkD3)
require(jsonlite)
require(shinydashboard)

ui <- function(){
  fluidPage(
    fluidRow(
      column(width = 12, style = "margin: 0px; padding: 0px;",
             box(title = "Flow Diagram", status = "primary", solidHeader = F, width = 7,
                 sankeyNetworkOutput("sankeyFlow", width = "100%", height = "500px")
             )
      ), 
      sliderInput("numLanes", label = "Number of Top Lanes", min = 0, max = 30, value = 10)
    )
  )
}

server <- function(input, output, session) {

  observeEvent(c(input$numLanes), {
    withProgress(message = "Wait...", value = NULL, {
      Sys.sleep(1.0)
    })
  })

  output$sankeyFlow <- renderSankeyNetwork({
    print("sankey")

    withProgress(message = "Updating flow diagram...", value = NULL, {

      # Load energy projection data
      URL <- "https://cdn.rawgit.com/christophergandrud/networkD3/master/JSONdata/energy.json"
      Energy <- jsonlite::fromJSON(URL)

      nLanes <- input$numLanes

      plot <- sankeyNetwork(Links = Energy$links, Nodes = Energy$nodes, Source = "source",
                                  Target = "target", Value = "value", NodeID = "name",
                                  units = "TWh", fontSize = 12, nodeWidth = 30)
      return(plot)
    })
  })
}

shinyApp(ui = ui, server = server)
```
In my use case I use several progress bars and none are displayed if I include the flow diagram. 

Thanks, Christian

alandipert:
@baczychris thank you for the report, I think I was able to identify a fix.

cjyetman:
I think we found a workaround for this on our end (see https://github.com/christophergandrud/networkD3/issues/214), but I believe a fix here would be preferable... interested to see what fix you came up with.

alandipert:
This is finally fixed on master and should be available in the next Shiny release. Sorry for the delay.

Issue #1817, opened 2017-08-16
Server bookmarking does not work properly when plotting sankey chart using networkD3 package


deepjyot1991:
Browser doesn't display the bookmark link when bookmark button is clicked but bookmark folder is updated correctly. Below is the reproducible code for the issue.

```
library(shiny)
library(networkD3)

ui <- function(request) {
  fluidPage(
    textInput("txt", "Enter text"),
    checkboxInput("caps", "Capitalize"),
    sankeyNetworkOutput("plot_sankey"),
    bookmarkButton()
  )
}
server <- function(input, output, session) {
  output$plot_sankey <- renderSankeyNetwork({
    
    URL <- paste0(
      "https://cdn.rawgit.com/christophergandrud/networkD3/",
      "master/JSONdata/energy.json")
    Energy <- jsonlite::fromJSON(URL)
    # Plot
    sankeyNetwork(Links = Energy$links, Nodes = Energy$nodes, Source = "source",
                  Target = "target", Value = "value", NodeID = "name",
                  units = "TWh", fontSize = 12, nodeWidth = 30)
  })
}

shinyApp(ui, server, enableBookmarking = "server")
```
![image](https://user-images.githubusercontent.com/31074842/29375712-f780edb4-8283-11e7-8779-442d67ea4a03.png)

It works fine with other charts such as force network chart using the same package networkD3. Below is the code.

```
library(shiny)
library(networkD3)

ui <- function(request) {
  fluidPage(
    textInput("txt", "Enter text"),
    checkboxInput("caps", "Capitalize"),
    forceNetworkOutput("plot_network"),
    bookmarkButton()
  )
}
server <- function(input, output, session) {
  output$plot_network <- renderForceNetwork({
    
    # Load data
    data(MisLinks)
    data(MisNodes)
    
    # Plot
    forceNetwork(Links = MisLinks, Nodes = MisNodes,
                 Source = "source", Target = "target",
                 Value = "value", NodeID = "name",
                 Group = "group", opacity = 0.8)
  })
}

shinyApp(ui, server, enableBookmarking = "server")
```
![image](https://user-images.githubusercontent.com/31074842/29375662-cd528048-8283-11e7-9c7d-957e64c8a9dd.png)

alandipert:
@deepjyot1991 thank you for reporting this &mdash; I think we were finally able to identify a fix.

deepjyot1991:
Thank you for identifying a fix for the bug. I highly appreciate it :)

[...]

PR #2359, opened 2019-03-21
Fix #2348, #2329, #1817: bugs triggered by networkD3 sankey plot

### LINKED ISSUE #2348: [Feature Request] Eager versions of pollReactive and reactiveFileReader
The `pollReactive` and `reactiveFileReader` functions are, well, reactive, which means the `valueFunc` and `readFunc` functions are not run unless the reactive is needed by another observer or reactive.

I have a use-case where `valueFunc`/`readFunc` are relatively expensive, but observers that are dependent on these reactives need to be relatively fast.

In this case, it would be better to have an eager, observant version of `pollReactive` and `reactiveFileReader`, so that the idle time between `input$button` clicks can be used to run `valueFunc`/`readFunc`.

```
data <- reactiveFileReader(2000, session, filePath = "data.csv", readFunc = function(filePath) {
    ...
    # Takes ~1000ms to run
})

observe({
    # This block should only be triggered by input$button
    input$button
    use_data_somehow <- isolate(data())
    # user expectation is that this block runs effectively instantaneously (i.e. <100ms)

    # as it stands now, the first time input$button is clicked after a file
    # modification is detected, this block will take >1000ms to run
    # subsequent button clicks will be very fast (as desired), until a file change is detected again
})
```

### LINKED ISSUE #2349: Possible bug with dateRangeInput when plotting sankeyNetworks from networkD3 package
Bringing up the issue by @nkk36 from #1959.

Issues:

- date range input select box won't pop up; though manually typing values in min/max range does work (i.e., server reads it)

- found a workaround from SO (https://stackoverflow.com/questions/44132864/part-of-daterangeinput-ui-gets-hidden-under-the-shiny-dashboard-header), but solution is unstable; work/fail at different sessions.
   `tags$div(tags$style(HTML(".dropdown-menu{z-index:20000 !important;}")))`
- as per @jcheng5 original request in #1959, JS console shows no error when this happen

`reprex` from original post.

```
# Setup =================================
# Load Packages
library(networkD3)
library(shiny)
library(shinydashboard)


# UI =================================
ui = dashboardPage(
  dashboardHeader(title = "Title"),
  # UI Sidebar =================================
  dashboardSidebar(
    sidebarMenu(
      dateRangeInput(inputId = "dates",
                     label = "Select time period:",
                     start = "2000-01-01",
                     end = "2018-02-01",
                     min = "2000-01-01",
                     max = "2018-02-01",
                     format = "mm-yyyy")
    )
  ),
  # UI Body =================================
  dashboardBody(
    fluidRow(
      box(sankeyNetworkOutput(outputId = "sankey_diagram"), 
          width = 12)
            )
    )
)

# Server =================================
server = function(input, output, session){
  # Render sankey diagram =================================
  output$sankey_diagram = renderSankeyNetwork({
    
    
    t1 = data.frame(source = c(0,0,
                              1,1),
                   target = c(2,3,
                              2,3),
                   value = c(100, 25,
                             10, 20))
    
    t2 = data.frame(id = c(0,1,2,3),
                      name = c("Group 1", "Group 2", 
                               "Group 3", "Group 4"))
    
    s = sankeyNetwork(Links = t1,
                      Nodes = t2,
                      Source = "source",
                      Target = "target",
                      Value = "value",
                      NodeID = "name",
                      NodeGroup = "name",
                      fontSize = 12,
                      nodeWidth = 30,
                      iterations = 0,
                      colourScale = JS("d3.scaleOrdinal(d3.schemeCategory10);"))
    return(s)
    
  }) 
}
# Run Shiny App =================================
shinyApp(ui, server)
```

alandipert:
* All of these were caused by the presence of multiple body tags on the
page, which happened because networkD3's sankey plot generates SVGs
containing body tags via SVG's foreignObject tag
* In various places, the 'body' jQuery selector string is used under the
assumption there is only one 'body' tag on the page. The presence of
multiple 'body' tags breaks reliant code in strange ways.
* The fix in all cases was to make the body selector more specific:
'body:first' or document.body do the job.

Fixes the following:

* https://github.com/rstudio/shiny/issues/2349
* https://github.com/rstudio/shiny/issues/2329
* https://github.com/rstudio/shiny/issues/1817

Note that this has troubled networkD3 users for a long time: https://github.com/christophergandrud/networkD3/issues/214

The way Shiny behaves, that this PR addresses, is not strictly "wrong" &mdash; it should be safe to assume that in a given HTML page, only one `body` will appear. However, the existence of the problem is proof of how confusing `foreignObject` is (the [SVG spec example for `foreignObject`](https://www.w3.org/TR/SVG11/extend.html#ForeignObjectElement) demonstrates embedding `body` in an SVG!), and so it seems like a good policy for Shiny to do the robust (if technically unnecessary) thing.


wch:
We should have documentation and a script to update and patch bootstrap-datepicker:

* Add a script /tools/update-boostrap-datepicker.R that downloads, unzips, and applies patches
* Create patches and store them somewhere
* Document in /tools/README.md how to: update boostrap-datepicker (by running the script), and create a new patch

It can be like this stuff in httpuv, but better:
https://github.com/rstudio/httpuv/blob/master/tools/update_libuv.R
https://github.com/rstudio/httpuv/blob/master/src/README.md

alandipert:
Superseded by #2361

PR #2361, opened 2019-03-27
Fix #2348, #2329, #1817: bugs triggered by networkD3 sankey plot 

### LINKED ISSUE #2348: [Feature Request] Eager versions of pollReactive and reactiveFileReader

alandipert:
*This PR supersedes #2359 as it includes some new infrastructure for managing patches to `bootstrap-datepicker`*

## Cause and fix

* All of these were caused by the presence of multiple body tags on the page, which happened because networkD3's sankey plot generates SVGs containing body tags via SVG's foreignObject tag
* In various places, the `'body'` jQuery selector string is used under the assumption there is only one `body` tag on the page. The presence of multiple `body` tags breaks reliant code in strange ways.
* The fix in all cases was to make the selector more specific: `'body:first'` or `document.body` do the job.

## Issues that were fixed

* https://github.com/rstudio/shiny/issues/2349
* https://github.com/rstudio/shiny/issues/2329
* https://github.com/rstudio/shiny/issues/1817

## Further context

Note that this has troubled networkD3 users for a long time: https://github.com/christophergandrud/networkD3/issues/214

The way Shiny behaves, that this PR addresses, is not strictly "wrong" &mdash; it should be safe to assume that in a given HTML page, only one `body` will appear. However, the existence of the problem is proof of how confusing `foreignObject` is (the [SVG spec example for `foreignObject`](https://www.w3.org/TR/SVG11/extend.html#ForeignObjectElement) demonstrates embedding `body` in an SVG!), and so it seems like a good policy for Shiny to do the robust (if technically unnecessary) thing.

## New patch infrastructure

This PR also includes two new scripts, `tools/updateBootstrapDatepicker.R` and `tools/applyDatepickerPatches.R`, that can be used together to manage our own changes to the `bootstrap-datepicker` 3rd party library. A section was added to `tools/README.md` documenting their use.

In order to switch to this new method of applying patches, bootstrap-datepicker version 1.6.4 was re-imported, and a [previously committed change](0683b79fac9c7b256846f71d94cf47f1c2abffa7) was converted to a `.patch` and re-applied using the new tools.


alandipert:
@wch One problem with this approach that we may want to rectify is, if the bootstrap-datepicker repo disappears, we don't have a way to revert to a "clean import", since the version in git is the one we've patched.

We might consider adding the bootstrap-datepicker release zip file to our repo as part of the import process (for 1.6.4 it's 97k). What do you think?

NEWS.md
```diff
@@ -21,13 +21,12 @@ shiny 1.3.0
 
 * Fixed [#2308](https://github.com/rstudio/shiny/issues/2308): When restoring a bookmarked application, inputs with a leading `.` would not be restored. ([#2311](https://github.com/rstudio/shiny/pull/2311))
 
-* Fixed [#2305](https://github.com/rstudio/shiny/issues/2305), [#2322](https://github.com/rstudio/shiny/issues/2322), [#2351](https://github.com/rstudio/shiny/issues/2351): When an input in dynamic UI is restored from bookmarks, it would keep getting set to the same value. ([#2360](https://github.com/rstudio/shiny/pull/2360))
+* Fixed [#2349](https://github.com/rstudio/shiny/issues/2349), [#2329](https://github.com/rstudio/shiny/issues/2329), [#1817](https://github.com/rstudio/shiny/issues/1817): These were various bugs triggered by the presence of the [networkD3](https://christophergandrud.github.io/networkD3/) package's Sankey plot in an app. Impacted features included `dateRangeInput`, `withProgressBar`, and bookmarking ([#2359](https://github.com/rstudio/shiny/pull/2359))
```
wch: This replaced the other NEWS item?

tools/updateBootstrapDatepicker.R
```diff
@@ -1,47 +1,24 @@
...
+dest_dir  <- rprojroot::find_package_root_file("inst/www/shared/datepicker")
+tag       <- paste0("v", version)
+dest_file <- file.path(tempdir(), paste0("bootstrap-datepicker-", version, ".zip"))
+url       <- sprintf("https://github.com/uxsolutions/bootstrap-datepicker/releases/download/%s/bootstrap-datepicker-%s-dist.zip", tag, version)
+
+download.file(url, dest_file)
+unzip(
+  dest_file,
+  files = c(
+    "js/bootstrap-datepicker.js",
```
wch: Needs `locales` subdir also.

PR #2376, opened 2019-04-10
Preempt all reactlog computation if not logging

### LINKED ISSUE #2375: Reactlog performance regression with large objects
When an application sets a `reactiveValues` to a value that is a large list that takes a long time to print with `str()`, it can slow down the shiny application even when reactlog is disabled.

This is primarily an issue when the value is a nested list. With the value is a large flat list, `str()` by default limits the output to 99 items. However, when it's a nested list, each sublist prints up to 99 items, so the final output can be quite large, and quite slow.

Example:

```R
obj <- lapply(1:60, function(i) {
  lapply(1:60, function(j) {
    list(i, j)
  })
})

# Check how long it takes to call `str()` on the object
system.time({
  capture.output(str(obj))
})
#>   user  system elapsed 
#>  4.006   0.069   4.093 


library(profvis)
library(shiny)


profvis({
  runApp(
    shinyApp(
      
      ui = fluidPage(
        textOutput("txt")
      ), 
      
        r <- reactiveValues(x = 1)
        
        observe({
          r$x <- obj
        })
        
        output$txt <- renderText({
          length(r$x)
        })
      }
    )
  )
})
```

This takes about 6 seconds to display anything in the Shiny app. In Shiny 1.2.0, it is almost instantaneous.

The profile indicates that most of the time is spent in `rLog$defineKey`. (Unchecking "Hide internal function calls" in the profile shows useful information.)

image

schloerke:
Fixes #2375 

* Preempts all `reactlog` logging functions if logging is not taking place
* Will only `str` the top level and stop at 99 (default `str` value) elements.  This will prevent the recursive nature of `str` which would print forever when looking at lists of lists of lists of ..... (`bigO(n^k); n <= 99, k = nested_levels(x)`). Now it will only print the top level list (`bigO(n); n <= 99`).
* Keep a check when saving an entry to safeguard against future method forgetting to preempt

--------------

I've altered the example app to profile `capture.output({str(obj)})` to show how long the `str` method takes.

PR fix with no reactlog logging: (bottom right corner's `runApp` takes very little time)
Screen Shot 2019-04-10 at 11 50 45 AM

PR fix with reactlog logging: (bottom right corner's `runApp` has a `str` print method and takes a little bit longer than before, but does not take the **full** `str` print time)
Screen Shot 2019-04-10 at 11 51 16 AM

### DIFF
--- a/R/graph.R
+++ b/R/graph.R
@@ -201,7 +201,8 @@ RLog <- R6Class(
 
     valueStr = function(value, n = 200) {
       output <- try(silent = TRUE, {
-        utils::capture.output(utils::str(value))
+        # use max.level = 1 to deter against massive recursion
+        utils::capture.output(utils::str(value, max.level = 1))
       })
       outputTxt <- paste0(output, collapse="\n")
       msg$shortenString(outputTxt, n = n)
--- a/R/graph.R
+++ b/R/graph.R
@@ -227,7 +227,9 @@ RLog <- R6Class(
       isTRUE(getOption(private$option, FALSE))
     },
     notLogging = function() {
-      !self$isLogging()
+      # same as !self$isLogging()
+      # do not call self$isLogging for speed purposes
+      !isTRUE(getOption(private$option, FALSE))
     },
 
     define = function(reactId, value, label, type, domain) {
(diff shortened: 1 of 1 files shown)


schloerke:
Closing in favor of https://github.com/rstudio/shiny/pull/2345

PR #2384, opened 2019-04-11
createWebDependency() should resolve relative dependencies, closes #2383

### LINKED ISSUE #2383: createWebDependency can't handle relative htmlDependencies
Minimal example:

```r
library(shiny)
library(htmltools)
dep <- htmlDependency(
  name = "jquery",
  version = "1.0.0",
  src = c(file = "www/shared"),
  script = "jquery.min.js",
  package = "shiny"
)
renderDependencies(list(dep))
createWebDependency(dep)
```

```r
Error in value[[3L]](cond) : 
  Couldn't normalize path in `addResourcePath`, with arguments: `prefix` = 'jquery-1.0.0'; `directoryPath` = 'www/shared'
```

I noticed this in a shiny app that uses crosstalk and leaflet. The dev version of `crosstalk::crosstalkLibs()` now uses relative paths which ends up being a problem here

https://github.com/rstudio/leaflet/blob/44df7d18a2618e0aeecb4145d765e597ec65878b/R/utils.R#L201

cpsievert:
This resolves the relative path in a similar way to `htmltools::resolveDependencies()`

### DIFF
--- a/R/html-deps.R
+++ b/R/html-deps.R
@@ -25,6 +25,9 @@ createWebDependency <- function(dependency, scrubFile = TRUE) {
     stop("Unexpected non-html_dependency type")
 
   if (is.null(dependency$src$href)) {
+    if (!is.null(dependency$package) && !is.null(dependency$src$file)) {
+      dependency$src$file <- system.file(dependency$src$file, package = dependency$package)
+    }
     prefix <- paste(dependency$name, "-", dependency$version, sep = "")
     addResourcePath(prefix, dependency$src$file)
     dependency$src$href <- prefix


wch:
It seems strange that this version of `createWebDependency` takes an html_dependency object with `package` and a relative `src$file`, then returns an html_dependency object with `package` and an _absolute_ `src$file`.

I believe the input html_dependency object is valid, but the output one is not, because if it has `package`, the `src$file` should be relative, at least according to the docs for `htmlDependency`. Would it make sense to set `package` to `NULL` when the change from relative to absolute is made?

jcheng5:
Or maybe we don't touch `src$file` at all? Turn it absolute using a local variable instead?

PR #2386, opened 2019-04-13
Fix #2385: R Markdown documents containing subapps not rendering properly

### LINKED ISSUE #2385: R Markdown documents containing subapps not rendering properly
https://community.rstudio.com/t/shinyapp-does-not-render-plot-shiny-version-1-3-0/28343

The symptom is that JS and CSS assets return 404 in subapps.

Repro is in shiny-examples:
https://github.com/rstudio/shiny-examples/tree/master/156-subapps (needs to be launched with `rmarkdown::run("index.Rmd")`)

jcheng5:
#2385


R/server.R
```diff
@@ -69,8 +70,41 @@ addResourcePath <- function(prefix, directoryPath) {
...
+  # URLs are rewritten by R code (i.e. routeHandler) before they can be matched to
+  # a resource path.
+  .globals$resources[[prefix]] <- list(
+    directoryPath = normalizedPath,
+    func = staticHandler(normalizedPath)
+  )
+}
+
+resourcePathHandler <- function(req) {
+  if (!identical(req$REQUEST_METHOD, 'GET'))
```
wch: Just checking: when the original value of `req$REQUEST_METHOD` is `HEAD`, it would have been converted to GET by this point, right?

Untitled.Rmd
```diff
@@ -0,0 +1,23 @@
+---
```
wch: I think this file should be in shiny-examples, right?

hi.html
```diff
@@ -0,0 +1,2 @@
+<strong>Hello</strong>
```
wch: File should be removed, right?

NEWS.md
```diff
@@ -1,20 +1,18 @@
-shiny 1.3.1
+shiny 1.3.2
```
wch: I think we should use 1.3.1.9000, since it's possible there will be more changes. Also, update the DESCRIPTION?

jcheng5:
Heh, whoops. This is why it's a bad idea to make PRs late at night...

R/server.R
jcheng5: That's right, it's way up at the outer edge of the `call` method.

R/server.R
```diff
@@ -69,8 +70,41 @@ addResourcePath <- function(prefix, directoryPath) {
...
+
+  resInfo <- .globals$resources[[prefix]]
+  if (is.null(resInfo))
+    return(NULL)
+
+  suffix <- substr(path, 2 + len, nchar(path))
+
+  subreq <- as.environment(as.list(req, all.names=TRUE))
+  subreq$PATH_INFO <- suffix
+  subreq$SCRIPT_NAME <- paste(subreq$SCRIPT_NAME, substr(path, 1, 2 + len), sep='')
```
wch: I'm not sure I understand what this part is doing. What's `subreq$SCRIPT_NAME` before this line?

In general, it would be helpful for me to understand the code if it had comments with an example of the data that goes in and how it's transformed along the way.

jcheng5: I'll add some comments. You might also find this helpful: the last version of middleware.R when it was literately programmed. https://gist.github.com/jcheng5/ddb0f0e1330448d7676281edb292efe2 The part about `PATH_INFO` and `SCRIPT_NAME` is under the heading "Handler Routing".

jcheng5: Also, the Rook spec describes SCRIPT_NAME and PATH_INFO: https://github.com/jeffreyhorner/Rook/blob/master/README.md#the-environment

They originally came from the [CGI specification](https://tools.ietf.org/html/rfc3875).

A URL comes into a web server, that web server uses the URL to determine what static file to serve or what CGI script to run. If the latter, then there might also be "extra" path info after the URL of the script itself. For example, http://example.com/cgi-bin/script.pl/foo/bar, the SCRIPT_NAME would be `/cgi-bin/script.pl` and the PATH_INFO would be `/foo/bar`.

In httpuv, all HTTP requests (except static) are sent through the same `call()` codepath, so `SCRIPT_NAME` always starts out as `""` and `PATH_INFO` contains the entire path of the URL. In Shiny's middleware stack, as parts of the URL are "matched" by http handlers and then delegated to other nested http handlers, the parts of the URL that are matched are popped off the front of `PATH_INFO` and pushed onto the end of `SCRIPT_NAME`. This greatly simplifies the implementation of those nested http handlers, because otherwise every one of them would have to take an additional argument that says "ignore this prefix of `PATH_INFO`".

I don't think we ever actually use `SCRIPT_NAME` for anything, btw. But we might someday: e.g., a deeply nested `staticHandler` might decide that a URL that matches a directory but without a trailing slash, should receive a 301 response to add that trailing slash. The URL would then be `paste0(SCRIPT_NAME, PATH_INFO, "/")`.

jcheng5:
@wch bump

Issue #2387, opened 2019-04-13
sliderInput fires evens on each slider change


napoxog:
Hi I faced problem with shiny 1.3 run with R 3.5.1 and shinydashboard 0.7.1
If slider input is initialized with ```numeric``` class values and then updated with ```Date``` class values, the slider fires evens on each slider change.
This results in a very often reactive processing and so - slowing down the app.

Here is MRE:
```R
library(shiny)
library(shinydashboard)


Areas=c("Russia","USA")

#initialization with this produce an issue
initRange=c(1,2)    

#initialization with this doesn't produce an issue
#initRange=as.Date(c("2012-12-22","2017-08-09"))    


ui <- dashboardPage(#skin=shinytheme("simplex"),
  
  dashboardHeader(title = "Dashboard"),
  dashboardSidebar(
    h2("Basic dashboard"),
    sidebarMenu(
      menuItem("Status ", tabName = "dashboard", icon = icon("dashboard"))
    )
  ),
  dashboardBody(
    tabItems(
      # First tab content
      tabItem(tabName = "dashboard",
              title = "Page",
              selectInput("Area",label = "Area", choices = Areas),
              plotOutput("plot1", height = 400),
              sliderInput("slider_data",label = "Dates",min = initRange[1],max=initRange[2], value = initRange,dragRange = T,width = "100%")
      )
    )
  )
)

server <- function(input, output, session) {
  observeEvent(input$Area,{
    dRange=as.Date(c("2012-12-22","2017-08-09"))
    updateSliderInput(session, "slider_data", min = dRange[1], max = dRange[2], value = dRange)
  })
  output$plot1 <- renderPlot({
    print(input$slider_data)
  }) 
}

shinyApp(ui, server)
```

cpsievert:
A more minimal example:

```r
library(shiny)

initRange <- c(1, 2)

ui <- fluidPage(
  actionButton("update", "Update slider"),
  sliderInput(
    "slider",
    label = "Dates", 
    min = initRange[1], 
    max = initRange[2], 
    value = initRange, 
    dragRange = T
  ),
  verbatimTextOutput("output")
)

server <- function(input, output, session) {
  observeEvent(input$update,{
    dRange <- as.Date(c("2012-12-22","2017-08-09"))
    updateSliderInput(
      session, "slider", 
      min = dRange[1], 
      max = dRange[2], 
      value = dRange
    )
  })
  output$output <- renderPrint({
    print(input$slider)
  }) 
}

[...]

PR #2404, opened 2019-04-25
Fix issue with input rate policies

cpsievert:
See #2387 for a motivating example. In that example, the slider starts off with the correct `Debouncer` input rate policy, but when the slider is updated, it's policy changes to the default `Invoker` (i.e. immediate) rate policy. 

It seems this happens because `InputRateDecorators`'s `$ensureInit()` method is called with a `name` that contains *both* the input id and type. As a result, when the type of an input changes, it fails to look up the input rate policy and ends up defaulting to `Invoker()` instead.

### DIFF
--- a/srcjs/input_rate.js
+++ b/srcjs/input_rate.js
@@ -298,6 +298,9 @@ var InputRateDecorator = function(target) {
 };
 (function() {
   this.setInput = function(name, value, opts) {
+    const input = splitInputNameType(name);
+    name = input.name;
+
     this.$ensureInit(name);
 
     if (opts.priority !== "deferred")
--- a/srcjs/input_rate.js
+++ b/srcjs/input_rate.js
@@ -297,6 +297,10 @@ var InputRateDecorator = function(target) {
   this.inputRatePolicies = {};
 };
 (function() {
+  // Note the 'public' methods setInput() and setRatePolicy() are passed
+  // both the input name (i.e., inputId) and the input type (i.e., the
+  // result of InputBinding.getType()). However, the 'private' methods
+  // are passed just the input name (i.e., inputId).
   this.setInput = function(name_type, value, opts) {
     const {name: inputName} = splitInputNameType(name_type);
 
--- a/srcjs/input_rate.js
+++ b/srcjs/input_rate.js
@@ -189,8 +189,8 @@ var InputBatchSender = function(shinyapp) {
   this.lastChanceCallback = [];
 };
 (function() {
-  this.setInput = function(name, value, opts) {
-    this.pendingData[name] = value;
+  this.setInput = function(name_type, value, opts) {
+    this.pendingData[name_type] = value;
 
     if (!this.reentrant) {
       if (opts.priority === "event") {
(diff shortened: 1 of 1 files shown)


srcjs/input_rate.js
```diff
@@ -298,30 +298,37 @@ var InputRateDecorator = function(target) {
...
-      this.setRatePolicy(name, 'direct');
+    const input = splitInputNameType(name);
+
+    if (!(input.name in this.inputRatePolicies))
+      this.setRatePolicy(input.name, 'direct');
   };
   this.$doSetInput = function(name, value, opts) {
-    this.target.setInput(name, value, opts);
+    const input = splitInputNameType(name);
+    this.target.setInput(input.name, value, opts);
```
wch: I'm not sure this is the right thing to do here. IIUC, this means that the input decorator that `InputRateDecorator` wraps up won't see the input type -- it'll only get the input name.

srcjs/input_rate.js
```diff
@@ -298,30 +298,37 @@ var InputRateDecorator = function(target) {
 };
 (function() {
   this.setInput = function(name, value, opts) {
```
wch: As discussed in a meeting, this `name` should be renamed to `name_type` or `nameType` to make it clear that they accept strings like `"x:date"`, whereas other functions like `$doSetInput` only accept the name without type, as in `"x"`.

srcjs/input_rate.js
```diff
@@ -297,23 +297,27 @@ var InputRateDecorator = function(target) {
...
+    const {name: inputName} = splitInputNameType(name_type);
+
+    this.$ensureInit(inputName);
 
     if (opts.priority !== "deferred")
-      this.inputRatePolicies[name].immediateCall(name, value, opts);
+      this.inputRatePolicies[inputName].immediateCall(name_type, value, opts);
     else
-      this.inputRatePolicies[name].normalCall(name, value, opts);
+      this.inputRatePolicies[inputName].normalCall(name_type, value, opts);
```
cpsievert: @wch should `immediateCall()` and `normalCall()` be called with `name_type` or `inputName`?

wch: It invokes `$doSetInput`, so it should be called with `inputName`.

srcjs/input_rate.js
```diff
@@ -297,23 +297,27 @@ var InputRateDecorator = function(target) {
   this.inputRatePolicies = {};
 };
 (function() {
-  this.setInput = function(name, value, opts) {
-    this.$ensureInit(name);
+  this.setInput = function(name_type, value, opts) {
```
wch: It would help to have a comment explaining what `name_type` is.

cpsievert:
> Is it possible that having multiple inputs with the same ID, but different types, is a thing someone would do for some reason?

I don't think it's intended for the rate policy to depend on the input type, so I'm fairly certain that `this.inputRatePolicies` method was always meant to be using `inputId` (i.e., `name`) rather than both the name and the type.

PR #2406, opened 2019-04-26
Input label updating

cpsievert:
See rstudio/shiny-examples#140 for a test app of all the label updating functionality


alandipert:
Oh, I just realized, we should probably implement the inverse. That is, when you set `label = NULL` via `updateTextInput()`, we should remove the `label` tag if it exists. What do you think?

cpsievert:
Oh, yes, good point!

cpsievert:
@alandipert done in 5e8bc20 and here's an app to test the behavior 

```r
runApp(shinyApp(
  ui = fluidPage(
    textInput("test", NULL),
    actionButton("btn", "Button")
  ),
    
    label <- reactive({
      if (isTRUE(input$btn %% 2 == 0)) NULL else "My Label"
    })
    
    observeEvent(input$btn, updateTextInput(session, "test", label = label()))
  }
))
```

srcjs/input_binding_text.js
```diff
@@ -27,8 +27,27 @@ $.extend(textInputBinding, {
...
+    // If data.label exists, then we may need to insert a label
+    // tag into the DOM. If it doesn't exist, then the label tag
+    // should be removed to be more consistent with the behavior
+    // of `textInput(label = NULL)`
+    if (data.hasOwnProperty('label')) {
+
+      if (hasLabelTag) {
+        labelTag.text(data.label);
+      } else {
+        $('<label for="' + $escape(el.id) + '">' + data.label + '</label>').insertBefore(el);
```
wch: I think this `data.label` might need to be escaped. For example, what if its contents are `abc<br>xyz`?

srcjs/input_binding_text.js
```diff
@@ -27,8 +27,27 @@ $.extend(textInputBinding, {
...
+      }
+
+    } else {
```
wch: I don't think it will ever get to this `else` because in `updateTextInput`, it drops `NULL`s:

https://github.com/rstudio/shiny/blob/386078d4415da7bbb05b8d63874758d5f6b1672f/R/update-input.R#L39

wch: Oh actually, I was incorrect. It will encounter this code path, but it will get there even if the user doesn't mean to clear the label. If they call `updateTextInput(session, "foo", value = "bar")`, then it will clear the label.

wch:
A few thoughts:

* It would be good to implement this for other inputs, since (I think) they all have the same behavior.
* In other places, we've used `character(0)` to allow specifying an empty value. We may want to use that for clearing the label, assuming that doesn't already work.

Also:

* NEWS
* test example

srcjs/input_binding_text.js
cpsievert: Have a look at how `selectInput()` handles this with `character(0)`

cpsievert:
@wch 09388c9 implements this label updating logic across all inputs where it makes sense (I think).

See https://github.com/rstudio/shiny-examples/pull/140 for a test app

wch:
After some discussion, we are thinking of this:

* Always supply the `<label>` tag.
* If the label is NULL, then add a CSS class that results in `display: none`.
* On the JS side, the update functions should add/remove the CSS class as appropriate.

srcjs/input_binding_select.js
```diff
@@ -123,15 +123,7 @@ $.extend(selectInputBinding, {
...
 
-    if (data.hasOwnProperty('label')) {
-      let escaped_id = $escape(el.id);
-      if (this._is_selectize(el)) {
-        escaped_id += "-selectized";
-      }
-      $(el).parent().parent()
-        .find('label[for="' + escaped_id + '"]')
-        .text(data.label);
-    }
```
cpsievert: @wch any idea why `parent()` is called twice here but only once in the `getState()` method (above)? Maybe it should be only called once?

wch: I don't think `getState` has been used for a long time, so I wouldn't use that as a guide of the correct way to do things. Does a single `parent()` with both kinds of `selectInput`s -- those with `selectize=TRUE` and `selectize=FALSE`?

cpsievert: Oh, alright, a single parent appears to work for `selectInput()` but not `selectizeInput()`

R/input-utils.R
```diff
@@ -1,5 +1,13 @@
...
+  classes <- c(
+    if (is.null(label)) "shiny-label-null",
+    if (control) "control-label"
+  )
+  tags$label(
+    label,
+    class = if (!is.null(classes)) paste(classes, collapse = " "),
+    `for` = inputId
+  )
 }
```
cpsievert: Maybe we should always be supplying vontrol-label

cpsievert: Use `class` arg multiple times

NEWS.md
```diff
@@ -7,10 +7,10 @@ shiny 1.3.2.9000
 
 ### Bug fixes
 
+* Fixed [#868](https://github.com/rstudio/shiny/issues/868): If an input is initialized with a `NULL` label, it can now be updated with a string. Moreover, if an input label is initialized with a string, it can now be removed by updating with `label=character(0)` (similar to how `choices` and `selected` can be cleared in `updateSelectInput()`). ([#2406](https://github.com/rstudio/shiny/pull/2406))
```
wch: Minor formatting nit: there should be an empty line between each item (same below).

wch:
Just a minor note about NEWS; otherwise it should be good to merge.

Issue #1433, opened 2016-10-20
Brushing issue with facet_grid when scales="free",space="free"


jiayi9:
In R Shiny, when we brush points in ggplot, there is something wrong with the facet_grid function when scales="free",space="free".

For example, the following code works fine.

```
    D = data.frame(x1 = c("X","X","X","Y","Y","Y"),
               x2 = c("a","a","b","c","c","c"),
               y = rnorm(6)
               )

    library(shiny)
    library(ggplot2)
    server <- function(input, session, output) {
      vals <- reactiveValues( keeprows = FALSE)

      observeEvent(input$brush_1,{
        Res=brushedPoints(D,input$brush_1,allRows = TRUE)
        vals$keeprows = Res$selected_
      })

      output$table1 = renderPrint({
        brushedPoints(D,input$brush_1,allRows = TRUE)
      })

      output$x = renderPrint({
        vals$keeprows
      })


      output$plot_1 = renderPlot({
        set.seed(123)

        D$selected = vals$keeprows

        ggplot(D,aes(x=x1,y=y)) + geom_boxplot(outlier.shape = NA) +
          facet_grid(.~x2) +
          geom_jitter(aes(color=selected,size=selected),width=0.2)+
          scale_color_manual(values = c("black","red"),guide=FALSE)
      })
    }

    ui <- fluidPage(
      plotOutput("plot_1",brush = brushOpts("brush_1",resetOnNew = FALSE)),
      verbatimTextOutput("table1"),
      verbatimTextOutput("x")
    )

    shinyApp(ui = ui, server = server)

```

But when we set  scales="free",space="free", the selection goes wrong.

```
    D = data.frame(x1 = c("X","X","X","Y","Y","Y"),
               x2 = c("a","a","b","c","c","c"),
               y = rnorm(6)
               )
    library(shiny)
    library(ggplot2)


    server <- function(input, session, output) {


      vals <- reactiveValues( keeprows = FALSE)

      observeEvent(input$brush_1,{
        Res=brushedPoints(D,input$brush_1,allRows = TRUE)
        vals$keeprows = Res$selected_
      })


      output$table1 = renderPrint({
        brushedPoints(D,input$brush_1,allRows = TRUE)
      })



      output$x = renderPrint({
        vals$keeprows
      })


      output$plot_1 = renderPlot({
        set.seed(123)

        D$selected = vals$keeprows



[...]

PR #2410, opened 2019-04-29
Coordmap info should retain discrete limits

cpsievert:
See #1433 for some related discussion and motivating examples. It's worth noting some examples there use `geom_jitter()`, but a fix for jittering would be more involved than what's addressed here, so I filed a separate issue #2411

This PR address the fact that the current coordmap model assumes that the numeric (i.e., trained) brush domain maps directly back to the input data, which is problematic for discrete axes. I can think of at least two cases where it becomes a problem for ggplot2:

1. Anytime `scale_[x/y]_discrete(limits = ...)` is used. Since `limits` defines possible values of the scale as well as their order, it breaks the coordmap assumption that a brush domain that spans, say [0.8, 1.2], maps back to the first level of discrete variable.
2. Anytime you a discrete axis that is 'free' across multiple panels, for example:

```r
library(ggplot2)
p <- ggplot(mpg) +
    geom_point(aes(fl, cty), alpha = 0.2) +
    facet_wrap(~drv, scales = "free_x")
layout <- summarise_layout(ggplot_build(p))
layouts <- split(layout, seq_len(nrow(layout)))
lapply(layouts, function(lay) {
  rng <- lay$xscale[[1]]$range$range
  setNames(seq_along(rng), rng)
})
```

```r
$`1`
d e p r 
1 2 3 4 

$`2`
c d e p r 
1 2 3 4 5 

$`3`
e p r 
1 2 3 
```

Notice how, in panel 1, a value of `x=1` maps back to `"d"`, but in panel 2, that same x value maps back to `"c"`. The `asNumber()` function, used to do this 'inverse mapping' in both `brushedPoints()` and `nearPoints()`, currently does not consider the actual axis limits in this way, which is why you get an incorrect result when brushing in the app provided and discussed in https://github.com/rstudio/shiny/issues/1433#issuecomment-260043730:

```r
library(shiny)
library(ggplot2)
mtcars$wt2 <- cut(mtcars$wt,4)
mtcars$mpg2 <- cut(mtcars$mpg,4)

ui <- basicPage(
  plotOutput("plot1", brush = "plot_brush"),
  verbatimTextOutput("info")
)

  output$plot1 <- renderPlot({
    ggplot(mtcars, aes(wt2,mpg)) +
      geom_point(alpha = 0.5) +
      facet_grid(~mpg2, scales = "free_x")
  })
  output$info <- renderPrint({
    brushedPoints(mtcars, input$plot_brush)
  })
}

shinyApp(ui, server)
```

This PR works around the issue by attaching any discrete limits to the `coordmap.panels` info and using that information inside `brushedPoints()` and `nearPoints()`

### DIFF
--- a/R/image-interact.R
+++ b/R/image-interact.R
@@ -88,17 +88,14 @@ brushedPoints <- function(df, brush, xvar = NULL, yvar = NULL,
       stop("brushedPoints: not able to automatically infer `xvar` from brush")
     if (!(xvar %in% names(df)))
       stop("brushedPoints: `xvar` ('", xvar ,"')  not in names of input")
-    # Extract data values from the data frame
-    x <- asNumber(df[[xvar]], brush$domain$xrange)
-    keep_rows <- keep_rows & (x >= brush$xmin & x <= brush$xmax)
+    keep_rows <- keep_rows & within_brush(df[[xvar]], brush, "x")
   }
   if (use_y) {
     if (is.null(yvar))
       stop("brushedPoints: not able to automatically infer `yvar` from brush")
     if (!(yvar %in% names(df)))
       stop("brushedPoints: `yvar` ('", yvar ,"') not in names of input")
-    y <- asNumber(df[[yvar]], brush$domain$yrange)
-    keep_rows <- keep_rows & (y >= brush$ymin & y <= brush$ymax)
+    keep_rows <- keep_rows & within_brush(df[[yvar]], brush, "y")
   }
 
   # Find which rows are matches for the panel vars (if present)
@@ -281,8 +278,8 @@ nearPoints <- function(df, coordinfo, xvar = NULL, yvar = NULL,
     stop("nearPoints: `yvar` ('", yvar ,"')  not in names of input")
 
   # Extract data values from the data frame
-  x <- asNumber(df[[xvar]], coordinfo$domain$xrange)
-  y <- asNumber(df[[yvar]], coordinfo$domain$yrange)
+  x <- asNumber(df[[xvar]], coordinfo$domain$discrete_mapping$x)
+  y <- asNumber(df[[yvar]], coordinfo$domain$discrete_mapping$y)
 
   # Get the coordinates of the point (in img pixel coordinates)
   point_img <- coordinfo$coords_img
@@ -402,7 +399,17 @@ nearPoints <- function(df, coordinfo, xvar = NULL, yvar = NULL,
 #   ..$ y: NULL
 #  $ .nonce    : num 0.603
 
-
+# Helper to determine if data values are within the limits of
+# an input brush
+within_brush <- function(vals, brush, var = "x") {
+  var <- match.arg(var, c("x", "y"))
+  vals <- asNumber(vals, brush$domain$discrete_mapping[[var]])
+  # At least when a discrete mapping is relevant,
+  # it's possible for the data values to not be
+  !is.na(vals) &
+    vals >= brush[[paste0(var, "min")]] &
+    vals <= brush[[paste0(var, "max")]]
+}
 
 # Coerce various types of variables to numbers. This works for Date, POSIXt,
 # characters, and factors. Used because the mouse coords are numeric.
--- a/R/image-interact.R
+++ b/R/image-interact.R
@@ -278,8 +278,8 @@ nearPoints <- function(df, coordinfo, xvar = NULL, yvar = NULL,
     stop("nearPoints: `yvar` ('", yvar ,"')  not in names of input")
 
   # Extract data values from the data frame
-  x <- asNumber(df[[xvar]], coordinfo$domain$discrete_mapping$x)
-  y <- asNumber(df[[yvar]], coordinfo$domain$discrete_mapping$y)
+  x <- asNumber(df[[xvar]], coordinfo$domain$discrete_limits$x)
+  y <- asNumber(df[[yvar]], coordinfo$domain$discrete_limits$y)
 
   # Get the coordinates of the point (in img pixel coordinates)
   point_img <- coordinfo$coords_img
@@ -403,9 +403,10 @@ nearPoints <- function(df, coordinfo, xvar = NULL, yvar = NULL,
 # an input brush
 within_brush <- function(vals, brush, var = "x") {
   var <- match.arg(var, c("x", "y"))
-  vals <- asNumber(vals, brush$domain$discrete_mapping[[var]])
-  # At least when a discrete mapping is relevant,
-  # it's possible for the data values to not be
+  vals <- asNumber(vals, brush$domain$discrete_limits[[var]])
+  # It's possible for a non-missing data values to not
+  # map to the axis limits, for example:
+  # https://github.com/rstudio/shiny/pull/2410#issuecomment-488100881
   !is.na(vals) &
     vals >= brush[[paste0(var, "min")]] &
     vals <= brush[[paste0(var, "max")]]
--- a/R/image-interact.R
+++ b/R/image-interact.R
@@ -281,8 +281,8 @@ nearPoints <- function(df, coordinfo, xvar = NULL, yvar = NULL,
     stop("nearPoints: `yvar` ('", yvar ,"')  not in names of input")
 
   # Extract data values from the data frame
-  x <- asNumber(df[[xvar]])
-  y <- asNumber(df[[yvar]])
+  x <- asNumber(df[[xvar]], coordinfo$domain$xrange)
+  y <- asNumber(df[[yvar]], coordinfo$domain$yrange)
 
   # Get the coordinates of the point (in img pixel coordinates)
   point_img <- coordinfo$coords_img
@@ -406,8 +406,8 @@ nearPoints <- function(df, coordinfo, xvar = NULL, yvar = NULL,
 
 # Coerce various types of variables to numbers. This works for Date, POSIXt,
 # characters, and factors. Used because the mouse coords are numeric.
-asNumber <- function(x, map = NULL) {
-  if (length(map)) return(match(x, map))
+asNumber <- function(x, range = NULL) {
+  if (length(range)) return(match(x, range))
   if (is.character(x)) x <- as.factor(x)
   if (is.factor(x)) x <- as.integer(x)
   as.numeric(x)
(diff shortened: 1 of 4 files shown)


cpsievert:
Just realized ce6749b won't work for jittered points

cpsievert:
A minimal app to demonstrate it fixes #1433

```r
library(shiny)
library(ggplot2)
mtcars$wt2 <- cut(mtcars$wt,4)
mtcars$mpg2 <- cut(mtcars$mpg,4)

ui <- basicPage(
  plotOutput("plot1", brush = "plot_brush"),
  verbatimTextOutput("info")
)

  output$plot1 <- renderPlot({
    ggplot(mtcars, aes(wt2,mpg)) +
      geom_point(alpha = 0.5) +
      facet_grid(~mpg2, scales = "free_x")
  })
  output$info <- renderPrint({
    brushedPoints(mtcars, input$plot_brush)
  })
}

shinyApp(ui, server)
```

cpsievert:
And another app for demonstrating the fix for `nearPoints()`:

```r
library(shiny)
library(ggplot2)
mtcars$wt2 <- cut(mtcars$wt,4)
mtcars$mpg2 <- cut(mtcars$mpg,4)

ui <- basicPage(
  plotOutput("plot1", click = "plot_click"),
  verbatimTextOutput("info")
)

  output$plot1 <- renderPlot({
    ggplot(mtcars, aes(wt2,mpg)) +
      geom_point(alpha = 0.5) +
      facet_grid(~mpg2, scales = "free_x")
  })
  output$info <- renderPrint({
    nearPoints(mtcars, input$plot_click)
  })
}

shinyApp(ui, server)
```

R/render-plot.R
```diff
@@ -689,6 +700,16 @@ find_panel_info_non_api <- function(b, ggplot_format) {
...
+    # Remember the x/y range if it's a faceted
+    # plot with a free discrete axis. This is necessary
+    # to properly inverse map the numeric (i.e., trained)
+    # positions back to the data scale (#2410)
+    layout <- if (ggplot_format == "new") b$layout else b$panel$layout
+    if (nrow(layout) > 1) {
+      if (xscale$is_discrete()) domain$xrange <- xscale$range$range
+      if (yscale$is_discrete()) domain$yrange <- yscale$range$range
+    }
+
```
cpsievert: Remove this, only provide fix for >2.2.1

R/render-plot.R
```diff
@@ -570,6 +570,17 @@ find_panel_info_api <- function(b) {
       domain$bottom <- -domain$bottom
     }
 
+    # Remember the x/y range if it's a faceted
+    # plot with a free discrete axis. This is necessary
+    # to properly inverse map the numeric (i.e., trained)
+    # positions back to the data scale (#2410)
+    if (nrow(layout) > 1) {
+      free_x <- isTRUE(b$layout$facet$params$free$x)
```
wch: Instead accessing `b` directly, we should always access the `layout` object which was returned by `summarise_layout()` above. We'll probably need a change to ggplot2 to get the information that we want.

R/render-plot.R
```diff
@@ -570,6 +570,17 @@ find_panel_info_api <- function(b) {
...
+      free_y <- isTRUE(b$layout$facet$params$free$y)
+      if (free_x && xscale$is_discrete()) domain$xrange <- xscale$range$range
```
wch: `xrange` seems a little confusing, since we use `range` to mean something else in other places. Maybe `x_target` or `x_discrete_levels`?

It would also help to have an example of the data structures that come out of this.

wch: New thought: inside of  `domain`, it could look like:

```
#   .. ..$ domain    :List of 4
#   .. .. ..$ left  : num 1.32
#   .. .. ..$ right : num 5.62
#   .. .. ..$ bottom: num 9.22
#   .. .. ..$ top   : num 35.1
#           $ discrete_mapping
#              $ x: chr "level1" "level2" "level3"
#              $ y: chr "something1" "something2"
```

cpsievert:
@wch I just realized there are other cases where you need to retain `discrete_mapping`, even with one panel, for example:

```r
library(shiny)
library(ggplot2)
mtcars$wt2 <- cut(mtcars$wt,4)
mtcars$mpg2 <- cut(mtcars$mpg,4)

ui <- basicPage(
  plotOutput("plot1", brush = "plot_brush"),
  verbatimTextOutput("info")
)

  output$plot1 <- renderPlot({
    ggplot(mtcars, aes(wt2, mpg)) +
      geom_point(alpha = 0.5) +
      scale_x_discrete(limits = "(2.49,3.47]")
  })
  output$info <- renderPrint({
    brushedPoints(mtcars, input$plot_brush)
  })
}

shinyApp(ui, server)
```

For this reason, I'm starting to think we should always retain `panel.domain.discrete_mapping` (as long as there is at least one discrete axis) and inform shinytest users about the false positives differences it will generate

R/render-plot.R
cpsievert: In light of the fact that I've know realized we need to track this information is even more scenarios, I think we should always track the `discrete_mapping` which removes the need to have to check for free scales.

cpsievert:
Still left TODO:

- [ ] Write some tests
- [ ] Update the comments in render-plot.R#L283-400 that document the coordmap data structure

R/image-interact.R
```diff
@@ -281,8 +278,8 @@ nearPoints <- function(df, coordinfo, xvar = NULL, yvar = NULL,
     stop("nearPoints: `yvar` ('", yvar ,"')  not in names of input")
 
   # Extract data values from the data frame
-  x <- asNumber(df[[xvar]])
-  y <- asNumber(df[[yvar]])
+  x <- asNumber(df[[xvar]], coordinfo$domain$discrete_limits$x)
```
wch: Make sure to account for `NA`s here.

R/render-plot.R
```diff
@@ -995,3 +1027,22 @@ find_panel_ranges <- function(g, res) {
...
+#
+# Eventually, we may want to consider storing the entire ggplot2
+# object server-side and querying information from that object
+# as we need it...that's the only way we'll ever be able to
+# faithfully brush examples like this:
+# https://github.com/rstudio/shiny/issues/2411
+add_discrete_limits <- function(domain, scale, var = "x") {
+  var <- match.arg(var, c("x", "y"))
+  if (scale$is_discrete()) {
+    domain$discrete_limits[[var]] <- scale$limits %OR% scale$range$range
```
wch: It would help to have a comment explaining when `scale$limits` is present.

R/render-plot.R
```diff
@@ -995,3 +1027,22 @@ find_panel_ranges <- function(g, res) {
...
+# https://github.com/rstudio/shiny/pull/2410#issuecomment-488100881
+#
+# Eventually, we may want to consider storing the entire ggplot2
+# object server-side and querying information from that object
+# as we need it...that's the only way we'll ever be able to
+# faithfully brush examples like this:
+# https://github.com/rstudio/shiny/issues/2411
+add_discrete_limits <- function(domain, scale, var = "x") {
+  var <- match.arg(var, c("x", "y"))
+  if (scale$is_discrete()) {
```
wch: Should check how far back the `is_discrete` function is available from the scales package. If in not-too-old versions the function isn't present, then check for `is.function(scale$is_discrete)` before calling it.

tests/testthat/test-plot-coordmap.R
```diff
@@ -357,3 +368,51 @@ test_that("ggplot coordmap with various scales and coords", {
     sortList(list(left=-1, right=3, bottom=-2, top=4))
   )
 })
+
+
+test_that("ggplot coordmap maintains discrete limits", {
+  tmpfile <- tempfile("test-shiny", fileext = ".png")
+  on.exit(rm(tmpfile))
```
wch: `unlink`, not `rm`. And thanks for fixing the other places that have this same error. :)

tests/testthat/test-plot-coordmap.R
```diff
@@ -357,3 +368,51 @@ test_that("ggplot coordmap with various scales and coords", {
     sortList(list(left=-1, right=3, bottom=-2, top=4))
   )
 })
+
+
+test_that("ggplot coordmap maintains discrete limits", {
```
wch: It would also be good to have a test with explicitly-set limits.

tests/testthat/test-plot-coordmap.R
```diff
@@ -271,9 +271,20 @@ test_that("ggplot coordmap with various data types", {
...
+    right = 3,
+    bottom = 1,
+    top = 4,
+    discrete_limits = list(
+      x = letters[1:3],
+      y = LETTERS[1:4]
+    )
+  )
+
   expect_equal(
```
wch: If you require testthat 2.1.0, then you can use `expect_mapequal()`.

R/render-plot.R
cpsievert: `is_discrete` is actually defined in ggplot2, and has existed since at least 1.0.1.9000

https://github.com/tidyverse/ggplot2/commit/eea7fcdbb7b6d4d5de73941359e6ad6a67d44168#diff-8516aea4a3d39cf4082ac0268ba42aefR197

cpsievert: Same with `get_limits()` which seems like a safer (more official) way to about this https://github.com/tidyverse/ggplot2/blame/a92042d3f57e8500bd173131d6248e2466e015fd/R/scale-.r#L113

cpsievert:
TODO:

- [ ] Update news
- [ ] Update the categorical axes section of this article with a relevant example https://shiny.rstudio.com/articles/plot-interaction-advanced.html

R/image-interact.R
```diff
@@ -402,11 +399,23 @@ nearPoints <- function(df, coordinfo, xvar = NULL, yvar = NULL,
...
+  !is.na(vals) &
+    vals >= brush[[paste0(var, "min")]] &
+    vals <= brush[[paste0(var, "max")]]
+}
 
 # Coerce various types of variables to numbers. This works for Date, POSIXt,
 # characters, and factors. Used because the mouse coords are numeric.
-asNumber <- function(x) {
+asNumber <- function(x, range = NULL) {
+  if (length(range)) return(match(x, range))
```
cpsievert: Rename `range` to `limits` or `levels`

R/image-interact.R
cpsievert: `NA`s aren't an issue here (in `nearPoints()`) because we end up [using `which()`](https://github.com/rstudio/shiny/blob/6fed1c60ac018c84afd1b8aeba54f5a70ebbb37c/R/image-interact.R#L314) and subsetting by that resulting integer vector rather than a logical index

PR #2416, opened 2019-05-02
getSliderType() should be able to handle NULL min/max/value, fixes #2250

### LINKED ISSUE #2250: Error in passing NULL via updateSliderInput
The following in the documentation for updateSliderInput seems to no longer be true in the current version of Shiny:

> Any arguments with NULL values will be ignored; they will not result in any changes to the input object on the client.

For instance trying to update only the label of a slider previously worked with the following example, and should work according to the documentation:

```
## Only run this example in interactive R sessions
if (interactive()) {
  shinyApp(
    ui = fluidPage(
      sidebarLayout(
        sidebarPanel(
          p("The first slider controls the second"),
          sliderInput("control", "Controller:", min=0, max=20, value=10,
                      step=1),
          sliderInput("receive", "Receiver:", min=0, max=20, value=10,
                      step=1)
        ),
        mainPanel()
      )
    ),
      observe({
        val <- input$control
        # Update label
        updateSliderInput(session, "receive", label = "Updated")
      })
    }
  )
}
```

But now requires the following code, even though min and max default to NULL:

```
## Only run this example in interactive R sessions
if (interactive()) {
  shinyApp(
    ui = fluidPage(
      sidebarLayout(
        sidebarPanel(
          p("The first slider controls the second"),
          sliderInput("control", "Controller:", min=0, max=20, value=10,
                      step=1),
          sliderInput("receive", "Receiver:", min=0, max=20, value=10,
                      step=1)
        ),
        mainPanel()
      )
    ),
      observe({
        val <- input$control
        # Update label
        updateSliderInput(session, "receive", label = "Updated", min=0, max=20)
      })
    }
  )
}
```

The old version is preferable, and logical in my opinion, but if this is no longer possible an update to the documentation would be useful.

Cheers, Jon

cpsievert:
See #2250

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -1730,6 +1730,7 @@ createVarPromiseDomain <- function(env, name, value) {
 
 getSliderType <- function(min, max, value) {
   vals <- dropNulls(list(value, min, max))
+  if (length(vals) == 0) return("")
   type <- unique(lapply(vals, function(x) {
     if      (inherits(x, "Date"))   "date"
     else if (inherits(x, "POSIXt")) "datetime"
@@ -1738,7 +1739,7 @@ getSliderType <- function(min, max, value) {
   if (length(type) > 1) {
     stop("Type mismatch for `min`, `max`, and `value`. Each must be Date, POSIXt, or number.")
   }
-  if (length(type)) type[[1]] else NULL
+  type[[1]]
 }
 
 # Reads the `shiny.sharedSecret` global option, and returns a function that can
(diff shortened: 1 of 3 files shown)


R/utils.R
```diff
@@ -1738,7 +1738,7 @@ getSliderType <- function(min, max, value) {
   if (length(type) > 1) {
     stop("Type mismatch for `min`, `max`, and `value`. Each must be Date, POSIXt, or number.")
   }
-  type[[1]]
+  if (length(type)) type[[1]] else NULL
```
wch: Instead of this logic, it's simpler to return early after line 1732.

R/utils.R
```diff
@@ -1738,7 +1738,7 @@ getSliderType <- function(min, max, value) {
...
 }
```
cpsievert: Instead of checking here, exit early after `dropNulls()`

R/utils.R
wch: Also, it would be simpler and more consistent to return `""` instead of `NULL`.

PR #2429, opened 2019-05-08
Use fastmap as backing store for Map class

### LINKED ISSUE #2423: Memory leaks from renderPlot
This test app shows various ways `renderPlot()` and assigning to outputs can leak memory. (This is originally from https://community.rstudio.com/t/memory-leak-when-using-ggplot2-grobs-and-tabs-shiny-or-my-bug/30108/.) **Update:** I've also added examples with `renderText()`.

In the first case, `renderPlot()` is called in an observer and assigned to an output each time. It leaks about 76528 bytes each time.

In the second case, a `renderPlot()` is invalidated. This leaks 2352 bytes each time.

In the third case, `renderPlot()` is called in an observer, but it is simply discarded. This leaks 1480 bytes each time.

The next three cases are the same, but use `renderText()`. They leak 2352, 1168, and 1168 bytes, respectively.

```R
library(shiny)
library(pryr)

ui <- fluidPage(
  sidebarLayout(
    sidebarPanel(
      div(
        "In observer, call renderPlot() and assign to output", br(),
        actionButton(inputId = "replot_1", label = "Go")
      ),
      br(),
      div(
        "Invalidate renderPlot()", br(),
        actionButton(inputId = "replot_2", label = "Go" )
      ),
      br(),
      div(
        "In observer, call renderPlot() but don't assign to an output", br(),
        actionButton(inputId = "replot_3", label = "Go" )
      ),
      br(),
      div(
        "In observer, call renderText() and assign to output", br(),
        actionButton(inputId = "retext_1", label = "Go")
      ),
      br(),
      div(
        "Invalidate renderText()", br(),
        actionButton(inputId = "retext_2", label = "Go" )
      ),
      br(),
      div(
        "In observer, call renderText() but don't assign to an output", br(),
        actionButton(inputId = "retext_3", label = "Go" )
      ),
      br(),
      verbatimTextOutput("memory_info")
    ),
    mainPanel(
      plotOutput("plot_1"),
      plotOutput("plot_2"),
      textOutput("text_1"),
      textOutput("text_2")
    )
  )
)

  observeEvent(input$replot_1, {
    output$plot_1 <- renderPlot({
      plot(1:20, 1:20)
    })
  })

  output$plot_2 <- renderPlot({
    input$replot_2
    plot(1:20, 1:20)
  })

  observeEvent(input$replot_3, {
    renderPlot({
      plot(1:20, 1:20)
    })
  })

  observeEvent(input$retext_1, {
    output$text_1 <- renderText({
      "Text 1"
    })
  })

  output$text_2 <- renderText({
    input$retext_2
      "Text 2"
  })

  observeEvent(input$retext_3, {
    renderText({
      "Text 3"
    })
  })


  last_mem_used <- 0L
  output$memory_info <- renderText({
    input$replot_1
    input$replot_2
    input$replot_3
    input$retext_1
    input$retext_2
    input$retext_3
    gc()
    current_mem_used <- mem_used()
    res <- sprintf(
      "mem_used(): %d bytes\nChange: %d bytes",
      current_mem_used,
      current_mem_used - last_mem_used
    )
    last_mem_used <<- current_mem_used

    res
  })
}

```

wch:
This is still a work in progress. It replaces the environment-based `Map` class with one based on the [fastmap package](https://github.com/wch/fastmap). The reason for this is that environment-based maps can leak memory, whereas fastmap does not.

This partially fixes #2423. Before the change, the six test cases from that issue leak the following amounts of memory: 76528, 2824, 1480, 2352, 1168, and 1168 bytes (note that these numbers came out somewhat differently in this test run compared to when I tested it in #2423). After this PR, the test cases leak 76520, 256, 192, 512, 64, and 64 bytes.

**Update**: This should be good to merge now.


wch:
In another test, I disabled all reactlog-related stuff in reactives.R, and I found that the test cases leaked the following amounts of memory: 76160, 0, 0, 0, 0, and 0 bytes. So with fastmap and no reactlog code, the only problematic case is the one where a `renderPlot` is repeatedly assigned to an output value.

@schloerke I think we'll want to reinstate an updated version of #2376 to eliminate the reactlog-related leaks.

R/reactives.R
```diff
@@ -408,30 +405,29 @@ ReactiveValues <- R6Class(
...
         }
       }
 
-      dep.keys <- objects(
-        envir=.dependents,
-        pattern=paste('^\\Q', key, ':', '\\E', '\\d+$', sep=''),
-        all.names=TRUE
-      )
+      dep.keys <- .dependents$keys()
+      dep.keys <- dep.keys[grepl(paste('^\\Q', key, ':', '\\E', '\\d+$', sep=''), dep.keys)]
```
jcheng5: This could be `grep(value=TRUE)` I think?

wch:
TODO: Test `freezeReactiveValue`.

wch:
There seems to be a change in the order that reactives/observers execute.

Here's a test app:

```R
library(shiny)

ui <- fluidPage(
  actionButton("go", "Go"),
  verbatimTextOutput("txt"),
  verbatimTextOutput("log")
)

  log <- reactiveVal(character(0))
  log_append <- function(...) {
    isolate(log(c(
      log(),
      paste(..., sep = "")
    )))
  }

  observe({
    input$go
    log_append(input$go, ": observe")
  })

  output$txt <- renderText({
    log_append(input$go, ": renderText")
    paste("Number of button presses:", input$go)
  })
  
  output$log <- renderText({
    paste0(
      log(),
      collapse = "\n"
    )
  })
}

shinyApp(ui, server)
```

With shiny 1.3.2 in a new R session, the output looks like this after clicking 5 times:

image

When running the app a second time (still with 1.3.2), the order is consistent:

image


With this branch, the order switches every time the button is clicked:

image

Issue #2433, opened 2019-05-09
Subdirectories of www/ can clash with other static paths


cpsievert:
With the new static serving functionality, if a subdirectory of `www/` matches the name of another static path, that subdirectory will no longer be served. This has bit us in at least two places already 

https://community.rstudio.com/t/shiny-app-not-working-anymore/28971/2
https://github.com/rstudio/rmarkdown/pull/1579

I think at the very least we can do more to throw an informative warning when this happens, especially considering we never mentioned this 'breaking change' anywhere

PR #2434, opened 2019-05-09
Throw an informative warning if a subdirectory of www clashes with another static path

cpsievert:
See #2433

### DIFF
--- a/R/app.R
+++ b/R/app.R
@@ -544,7 +544,7 @@ knit_print.reactive <- function(x, ..., inline = FALSE) {
 # throw an informative warning if a www subdirectory conflicts
 # with another resourcePath
 wwwSubDirConflicts <- function(wwwdir) {
-  wwwSubDirs <- list.dirs(wwwDir, recursive = FALSE, full.names = FALSE)
+  wwwSubDirs <- list.dirs(wwwdir, recursive = FALSE, full.names = FALSE)
   resourcePrefixes <- c(
     names(.globals$resourcePaths),
     names(.globals$resources),
--- a/R/app.R
+++ b/R/app.R
@@ -169,7 +169,6 @@ shinyAppDir_serverR <- function(appDir, options=list()) {
 
   wwwDir <- file.path.ci(appDir, "www")
   if (dirExists(wwwDir)) {
-    wwwSubDirConflicts(wwwDir)
     staticPaths <- list("/" = staticPath(wwwDir, indexhtml = FALSE, fallthrough = TRUE))
   } else {
     staticPaths <- list()
@@ -323,7 +322,6 @@ shinyAppDir_appR <- function(fileName, appDir, options=list())
 
   wwwDir <- file.path.ci(appDir, "www")
   if (dirExists(wwwDir)) {
-    wwwSubDirConflicts(wwwDir)
     # wwwDir is a static path served by httpuv. It does _not_ serve up
     # index.html, for two reasons. (1) It's possible that the user's
     # www/index.html file is not actually used as the index, but as a template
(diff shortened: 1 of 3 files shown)


R/app.R
```diff
@@ -538,3 +540,31 @@ knit_print.reactive <- function(x, ..., inline = FALSE) {
...
+
+
+
+# throw an informative warning if a www subdirectory conflicts
+# with another resourcePath
+wwwSubDirConflicts <- function(wwwdir) {
+  wwwSubDirs <- list.dirs(wwwDir, recursive = FALSE, full.names = FALSE)
+  resourcePrefixes <- c(
+    names(.globals$resourcePaths),
+    names(.globals$resources),
```
wch: `names(.globals$resources)` shouldn't be necessary, since it should have the same content as `names(.globals$resourcePaths)`.

wch:
In https://github.com/rstudio/shiny/blob/547edd7e3261449c4fd4ddaef327fb6e3e95d0bc/R/server.R#L461, it would be good to do a pairwise check for conflicts on all the paths that are in `httpuvApp$staticPaths`. That way, we don't need to special case anything, and we'll also detect conflicts among `resourcePaths`, and between `resourcePaths` and `session` and `shared`.

R/server.R
```diff
@@ -459,6 +459,47 @@ startApp <- function(appObj, port, host, quiet) {
...
+    if (length(paths) < 2) return(NULL)
+    # ensure paths is a named character vector: c(resource_path = local_path)
+    local_paths <- vapply(paths, function(x) if (inherits(x, "staticPath")) x$path else x, character(1))
+    paths <- setNames(local_paths, names(paths))
+    # get all possible pairwise combinations of paths
+    pair_indicies <- utils::combn(length(paths), 2, simplify = FALSE)
+    lapply(pair_indicies, function(x) {
+      p1 <- paths[x[1]]
+      p2 <- paths[x[2]]
+      if (identical(names(p1), names(p2)) && (p1 != p2)) {
```
wch: `p1` and `p2` should never be the same, right?

R/server.R
```diff
@@ -459,6 +459,47 @@ startApp <- function(appObj, port, host, quiet) {
...
+  }
+
+  # check for conflicts in each pairwise combinations of resource mappings
+  checkResourceConflict <- function(paths) {
+    if (length(paths) < 2) return(NULL)
+    # ensure paths is a named character vector: c(resource_path = local_path)
+    local_paths <- vapply(paths, function(x) if (inherits(x, "staticPath")) x$path else x, character(1))
+    paths <- setNames(local_paths, names(paths))
+    # get all possible pairwise combinations of paths
+    pair_indicies <- utils::combn(length(paths), 2, simplify = FALSE)
```
wch: Super nitpick: should be spelled "indices". You can ignore this comment if you want though. :)

R/server.R
```diff
@@ -459,6 +459,47 @@ startApp <- function(appObj, port, host, quiet) {
...
+      )
+    }
+  }
+
+  # check for conflicts in each pairwise combinations of resource mappings
+  checkResourceConflict <- function(paths) {
+    if (length(paths) < 2) return(NULL)
+    # ensure paths is a named character vector: c(resource_path = local_path)
+    local_paths <- vapply(paths, function(x) if (inherits(x, "staticPath")) x$path else x, character(1))
+    paths <- setNames(local_paths, names(paths))
```
wch: This line shouldn't be necessary, right? That is, `vapply` should preserve names, I think.

R/server.R
```diff
@@ -459,6 +459,47 @@ startApp <- function(appObj, port, host, quiet) {
     ),
     .globals$resourcePaths
   )
+
+  # throw an informative warning if a subdirectory of the
+  # app's www dir conflicts with another resource prefix
+  wwwDir <- httpuvApp$staticPaths[["/"]]$path
+  if (length(wwwDir)) {
+    wwwSubDirs <- list.dirs(wwwDir, recursive = FALSE, full.names = FALSE)
```
wch: It might make sense to use `recursive = TRUE` here, since a sub-sub-directory might conflict with a static path.

PR #2436, opened 2019-05-10
Add issue templates

alandipert:
These were adapted from rstudio/rstudio and will give us a new issue page like this one: https://github.com/rstudio/rstudio/issues/new/choose

I also adapted these wiki pages, linked from the new issue templates:

* [Writing Good Bug Reports](https://github.com/rstudio/shiny/wiki/Writing-Good-Bug-Reports)
* [Writing Good Feature Requests](https://github.com/rstudio/shiny/wiki/Writing-Good-Feature-Requests)

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -4,46 +4,39 @@ about : Report a bug in Shiny.
 ---
 
 <!--
-
-Before you file an issue, we would appreciate it if you would upgrade to the latest version of Shiny from CRAN and confirm that the problem persists.
-
-# To install latest shiny from CRAN:
-install.packages("shiny")
-# To update all your R packages:
-update.packages(ask = FALSE, checkBuilt = TRUE)
-# Please remember to restart R after updating packages.
-
 This issue tracker is for bugs and feature requests in the Shiny package. If you're having trouble with Shiny Server or a related package, please file an issue in the appropriate repository. 
 
 If you're having trouble with shinyapps.io, and you have a paid account (Starter, Basic, Standard, or Pro), please file a support ticket via https://support.rstudio.com. If you have a Free account, please post to the RStudio Community with the shinyappsio tag: https://community.rstudio.com/tags/shinyappsio.
 
 Finally, if you are an RStudio customer and are having trouble with one of our Pro products, get in touch with our support team at support@rstudio.com.
 
+Before you file an issue, please upgrade to the latest version of Shiny from CRAN and confirm that the problem persists.
+
+# First, restart R.
+# To install latest shiny from CRAN:
+install.packages("shiny")
+
 See our guide to writing good bug reports for further guidance: https://github.com/rstudio/shiny/wiki/Writing-Good-Bug-Reports. The better your report is, the likelier we are to be able to reproduce and ultimately solve it.
 -->
 
 ### System details
 
-    Shiny Version           :
-    OS Version              : 
-    R Version               : 
-    Browser Version         :
-    RStudio Edition/Version : <!-- If applicable -->
+Browser Version: <!-- If applicable -->
+
+Output of `sessionInfo()`:
+
+```
+# sessionInfo() output goes here
+```
 
 ### Example application *or* steps to reproduce the problem
 
-<!-- A minimal, self-contained example app demonstrating the problem is extremely helpful to us. Thank you in advance for taking the time to create one. For guidance on how to do so, please see https://github.com/rstudio/shiny/wiki/Writing-Good-Bug-Reports -->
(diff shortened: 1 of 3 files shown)


schloerke:
@alandipert Looks great! 

Should we include any links to https://community.rstudio.com/t/shiny-debugging-and-reprex-guide/10001 for reproducible shiny examples? (or steal parts from it)

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,43 @@
+---
+name  : Bug report
+about : Report a bug in Shiny.
+---
+
+<!--
+
+Shiny v1.3.2 is now available on CRAN. If you're currently using an older version of Shiny, we would appreciate it if you could update to Shiny v1.3.2 and confirm whether the problem still persists.
+
```
cpsievert: It'd be nice not to have to update this after every CRAN release. Perhaps we could say something like:


> Before filing an issue, please make sure your issue has not already been fixed or filed elsewhere. That is, please test your [reproducible example](https://community.rstudio.com/t/shiny-debugging-and-reprex-guide/10001) holds with recent R package version:

```r
# To install latest shiny from CRAN
install.packages("shiny")
# To update all your R packages
update.packages(ask = FALSE, checkBuilt = TRUE)
# please remember to restart R after updating packages
```

alandipert: Unfortunately I can't format it that nicely since the instructional text lives in an HTML comment, but I agree with the spirit of what you propose 👍

alandipert:
@schloerke I edited the [wiki page](https://github.com/rstudio/shiny/wiki/Writing-Good-Bug-Reports) and added a link to your post. Does the wiki page look good to you? Feel free to edit further if think it can be improved.

.github/ISSUE_TEMPLATE/bug_report.md
wch: It would also be good to tell users to restart R before installing/updating packages, because on Windows, upgrading a package can get messed up if the package is currently loaded.

wch: Another thought: I think that asking users to update all their packages is a bit too onerous. I think it's better if we take it out, even if it means it results in some false positives.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,49 @@
...
+If you're having trouble with shinyapps.io, and you have a paid account (Starter, Basic, Standard, or Pro), please file a support ticket via https://support.rstudio.com. If you have a Free account, please post to the RStudio Community with the shinyappsio tag: https://community.rstudio.com/tags/shinyappsio.
+
+Finally, if you are an RStudio customer and are having trouble with one of our Pro products, get in touch with our support team at support@rstudio.com.
+
+See our guide to writing good bug reports for further guidance: https://github.com/rstudio/shiny/wiki/Writing-Good-Bug-Reports. The better your report is, the likelier we are to be able to reproduce and ultimately solve it.
+-->
+
+### System details
+
+    Shiny Version           :
```
wch: I think it's better to have people call `sessionInfo()` and paste it here. It's both more informative and easier to do. (I know this because when I report bugs on the RStudio IDE, I get annoyed at having to look up versions of everything and enter them in.)

I also think we shouldn't have people fill out Browser Version and RStudio Version -- those are extra hurdles that are relevant in a relatively small proportion of cases.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,49 @@
...
+    OS Version              : 
+    R Version               : 
+    Browser Version         :
+    RStudio Edition/Version : <!-- If applicable -->
+
+### Example application *or* steps to reproduce the problem
+
+<!-- A minimal, self-contained example app demonstrating the problem is extremely helpful to us. Thank you in advance for taking the time to create one. For guidance on how to do so, please see https://github.com/rstudio/shiny/wiki/Writing-Good-Bug-Reports -->
```
wch: I think we should copy the content from the "Creating Shiny reprexes" from https://community.rstudio.com/t/shiny-debugging-and-reprex-guide/10001 to the wiki page. I don't think that people are going to follow two links to get there.

Also, I think that it would be good if the wiki page itself were more concise because I don't think many people will want to take the time to read all of it.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,49 @@
...
+
+```R
+# Minimal, self-contained example app goes here
+```
+
+### Describe the problem in detail
+
+### Describe the behavior you expected
```
wch: In my experience filing issues on projects that have these two sections (describe the problem, and describe the expected behavior), I always feel kind of disconcerted when I encounter the second section, because usually the description of the problem includes a description of the expected behavior. Then I try to fill out the expected behavior section and feel frustrated with trying to fit a bug report that was perfectly sensible into the template.

PR #2460, opened 2019-05-28
Disable dragging of plots with any interactions enabled

### LINKED ISSUE #1393: Should disable image dragging if any plot interactions are enabled
If just clicking is enabled, it's distracting and annoying if you click and drag a little bit and the browser thinks you want to drag the image to a folder. (Currently, when brushing is enabled, the image dragging is disabled.)

### LINKED ISSUE #2223: Plot interaction: when click is enabled, dragging image should be disabled

wch:
Closes #1393, closes #2223.

Test app:

```R
library(shiny)

shinyApp(
  ui = fluidPage(
    p("The plot below doesn't have any interactions. It should be draggable."),
    plotOutput("plot_no_interact", height = 250, width = 250),
    
    p("The plot below allows click interactions. It should not be draggable."),
    plotOutput("plot_interact", height = 250, width = 250, hover = "plot_click")
  ),
    output$plot_no_interact <- renderPlot({
      plot(cars)
    })
    output$plot_interact <- renderPlot({
      plot(cars)
    })
  }
)
```

### DIFF
--- a/srcjs/output_binding_image.js
+++ b/srcjs/output_binding_image.js
@@ -139,6 +139,8 @@ $.extend(imageOutputBinding, {
       // Register the various event handlers
       // ----------------------------------------------------------
       if (opts.clickId) {
+        imageutils.disableDrag($el, $img);
+
         var clickHandler = imageutils.createClickHandler(opts.clickId,
           opts.clickClip, opts.coordmap);
         $el.on('mousedown2.image_output', clickHandler.mousedown);
@@ -151,6 +153,8 @@ $.extend(imageOutputBinding, {
       }
 
       if (opts.dblclickId) {
+        imageutils.disableDrag($el, $img);
+
         // We'll use the clickHandler's mousedown function, but register it to
         // our custom 'dblclick2' event.
         var dblclickHandler = imageutils.createClickHandler(opts.dblclickId,
@@ -162,6 +166,8 @@ $.extend(imageOutputBinding, {
       }
 
       if (opts.hoverId) {
+        imageutils.disableDrag($el, $img);
+
         var hoverHandler = imageutils.createHoverHandler(opts.hoverId,
           opts.hoverDelay, opts.hoverDelayType, opts.hoverClip,
           opts.hoverNullOutside, opts.coordmap);
@@ -173,13 +179,7 @@ $.extend(imageOutputBinding, {
       }
 
       if (opts.brushId) {
-        // Make image non-draggable (Chrome, Safari)
-        $img.css('-webkit-user-drag', 'none');
-        // Firefox, IE<=10
-        $img.on('dragstart.image_output', function() { return false; });
-
-        // Disable selection of image and text when dragging in IE<=10
-        $el.on('selectstart.image_output', function() { return false; });
+        imageutils.disableDrag($el, $img);
 
         var brushHandler = imageutils.createBrushHandler(opts.brushId, $el, opts,
           opts.coordmap, outputId);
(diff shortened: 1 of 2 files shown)


wch:
Note that the generated JS resources aren't part of this PR, so they will need to be rebuilt after merging.

jcheng5:
There's a `"draggable"` attribute for img, can we just add that in R?

wch:
We can't do it from R because the `img` tag is created in JavaScript. We could do `$img.attr('draggable', false)` in JS, but then it doesn't quite work right in Firefox -- the image can be selected.



And apparently if you try to disable selection in Firefox, then `draggable="false"` stops working: https://bugzilla.mozilla.org/show_bug.cgi?id=1376369

I seem to recall trying a bunch of different things when this code was originally written and settled on what's here. We know it works so I'd prefer to stick with it.

srcjs/output_binding_image.js
```diff
@@ -225,6 +225,18 @@ outputBindings.register(imageOutputBinding, 'shiny.imageOutput');
...
+  $img.css('-webkit-user-drag', 'none');
+
+  // Firefox, IE<=10
+  // First remove existing handler so we don't keep adding handlers.
+  $img.off('dragstart.image_output');
+  $img.on('dragstart.image_output', function() { return false; });
+
+  // Disable selection of image and text when dragging in IE<=10
+  $el.on('selectstart.image_output', function() { return false; });
+};
```
cpsievert: Is this `$el` defined? `yarn build` gives me 

```
../srcjs/output_binding_image.js
  241:3  warning  '$el' is not defined  no-undef

✖ 1 problem (0 errors, 1 warning)
```

wch: Shoot, I copied and pasted that part and forgot to change it. I'll fix it.

srcjs/output_binding_image.js
```diff
@@ -162,6 +166,8 @@ $.extend(imageOutputBinding, {
       }
 
       if (opts.hoverId) {
+        imageutils.disableDrag($img);
```
jcheng5: Did you mean for hover to disable drag?

jcheng5: (It's fine if you did)

srcjs/output_binding_image.js
```diff
@@ -225,6 +225,18 @@ outputBindings.register(imageOutputBinding, 'shiny.imageOutput');
...
+  // Make image non-draggable (Chrome, Safari)
+  $img.css('-webkit-user-drag', 'none');
+
+  // Firefox, IE<=10
+  // First remove existing handler so we don't keep adding handlers.
+  $img.off('dragstart.image_output');
+  $img.on('dragstart.image_output', function() { return false; });
+
+  // Disable selection of image and text when dragging in IE<=10
+  $el.on('selectstart.image_output', function() { return false; });
```
jcheng5: Do you also need to remove this handler first, the way you did for `dragstart.image_output`?

wch: Yup, I was being sloppy with this part :(

cpsievert:
What's the argument for having draggable plot images in the first place? Is it so that they can be drag-and-dropped into some other consumer widget? Are there other use cases? 

Also, what's the argument for disabling drag for any interaction mode? It's seems possible that someone might want draggability if hover and/or click are enabled...

wch:
There's no particular reason for having draggable plots; that's just how images in general work in web browsers, so for the regular non-interactive case, it makes sense to just keep that behavior. The typical use case is that users can drag and drop to a file outside of the browser.

The reason for disabling drag when interactions are enabled is that when clicking is enabled, people often click and just drag without thinking about it (I do this) and then you end up dragging the image, which just gets in the way. When dragging is disabled, if users want to save the file, it's still possible by right-clicking and saving.

cpsievert:
Can't figure how to officially approve, but LGTM 👍

PR #2497, opened 2019-06-14
Add sep argument to renderText

### LINKED ISSUE #2469: renderText() should expose sep argument
This would make the connection to `cat()` more clear, and make it (slightly) easier to return multiline outputs (by setting `sep = "\n"`)

trestletech:
### Questions

I'd rather defer the `sep` default to `cat` rather than having to specify it myself in `renderText`. Is there a more idiomatic way to specify the `sep` passthrough to `renderText`? If I don't provide a default, I get a "missing with no default" error. I could check with `missing` but I'll still end up having to provide my own default in `renderText`. Seem unlikely that `cat` will change its default, so it's probably just an academic concern, but it made me feel a little icky.

### Behavior

Screen Shot 2019-06-14 at 10 28 06 AM
Screen Shot 2019-06-14 at 10 27 53 AM

Closes #2469

### DIFF
--- a/man/renderText.Rd
+++ b/man/renderText.Rd
@@ -5,7 +5,7 @@
 \title{Text Output}
 \usage{
 renderText(expr, env = parent.frame(), quoted = FALSE,
-  outputArgs = list())
+  outputArgs = list(), sep)
 }
 \arguments{
 \item{expr}{An expression that returns an R object that can be used as an
@@ -19,6 +19,9 @@ is useful if you want to save an expression in a variable.}
 \item{outputArgs}{A list of arguments to be passed through to the implicit
 call to \code{\link{textOutput}} when \code{renderText} is used in an
 interactive R Markdown document.}
+
+\item{sep}{A separator passed to \code{cat} to be appended after each
+element.}
 }
 \description{
 Makes a reactive version of the given function that also uses
--- a/man/renderText.Rd
+++ b/man/renderText.Rd
@@ -5,7 +5,7 @@
 \title{Text Output}
 \usage{
 renderText(expr, env = parent.frame(), quoted = FALSE,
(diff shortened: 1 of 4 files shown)


man/renderText.Rd
```diff
@@ -5,7 +5,7 @@
 \title{Text Output}
 \usage{
 renderText(expr, env = parent.frame(), quoted = FALSE,
-  outputArgs = list())
+  outputArgs = list(), sep)
```
wch: I think this needs to be re-documented?

trestletech:
Great catch! Just pushed

tests/testthat/test-text.R
```diff
@@ -18,6 +18,17 @@ test_that("renderPrint and renderText behavior is correct", {
                'foo')
   expect_equal(isolate(renderText({ invisible("foo") })()),
                'foo')
+  expect_equal(isolate(renderText({ c("hello", "world") })()),
+               'hello world')
+
+  expect_equal(isolate(renderText({ c("hello", "world") }, sep=", ")()),
```
hadley: I wonder if it’s worth suppressing the reactive context error by default inside of tests so that you don’t have to put isolate everywhere.

tests/testthat/test-text.R
```diff
@@ -18,6 +18,17 @@ test_that("renderPrint and renderText behavior is correct", {
...
+               'hello, world')
+  # Really just measuring the behavior of `cat` when it receives a vector of
+  # separators, but probably worth covering.
+  expect_equal(isolate(renderText({ c("hello", "there", "world", "repeats") },
```
hadley: Tbh I think you should make this an error rather than allowing very loose base behaviour to flow upwards.

PR #2502, opened 2019-06-17
Define optgroup when using

### LINKED ISSUE #2474: Improve optgroup documentation
`selectInput()` says

> List of values to select from. If elements of the list are named, then that name rather than the value is displayed to the user. This can also be a named list whose elements are (either named or unnamed) lists or vectors. If this is the case, the outermost names will be used as the "optgroup" label for the elements in the respective sublist. This allows you to group and label similar choices. See the example section for a small demo of this feature.

It's not clear what an optgroup is. 

I think this is an easy fix: first say that this allows you to group and label similar choices, and that this is called an optgroup, because it produces an `<optgroup>` html tag.

trestletech:
Closes #2474

### DIFF
--- a/R/input-select.R
+++ b/R/input-select.R
@@ -15,12 +15,12 @@
 #'
 #' @inheritParams textInput
 #' @param choices List of values to select from. If elements of the list are
-#'   named, then that name rather than the value is displayed to the user.
-#'   This can also be a named list whose elements are (either named or
-#'   unnamed) lists or vectors. If this is the case, the outermost names
-#'   will be used as the "optgroup" label for the elements in the respective
-#'   sublist. This allows you to group and label similar choices. See the
-#'   example section for a small demo of this feature.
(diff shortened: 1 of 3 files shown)


R/input-select.R
```diff
@@ -15,12 +15,12 @@
 #'
 #' @inheritParams textInput
 #' @param choices List of values to select from. If elements of the list are
-#'   named, then that name rather than the value is displayed to the user.
-#'   This can also be a named list whose elements are (either named or
-#'   unnamed) lists or vectors. If this is the case, the outermost names
-#'   will be used as the "optgroup" label for the elements in the respective
-#'   sublist. This allows you to group and label similar choices. See the
-#'   example section for a small demo of this feature.
+#'   named, then that name -- rather than the value -- is displayed to the user.
```
hadley: Em-dashes (which is want you want for a parenthetical), should be `---`

PR #2504, opened 2019-06-18
Clarify interaction between width/height and CSS/templates

### LINKED ISSUE #2465: cols doesn't change width of textAreaInput?
```
ui <- fluidPage(
  textAreaInput("a", "Label", rows = 3, cols = 40),
  textAreaInput("b", "Label", rows = 3, cols = 80),
  textAreaInput("c", "Label", rows = 3, cols = 120)
)

```

image

trestletech:
Closes #2465 

Feedback on wording welcome!


R/input-textarea.R
```diff
@@ -6,11 +6,15 @@
...
+#'   \code{80}. This argument will only take effect if there is not a CSS
+#'   \code{width} rule defined for this element; such a rule could come from the
+#'   \code{width} argument of this function or from a containing page layout
+#'   such as \code{\link{fluidPageLayout}}.
 #' @param rows The value of the visible character rows of the input, e.g.
-#'   \code{6}. If used with \code{height}, \code{height} will take precedence in
-#'   the browser's rendering.
+#'   \code{6}. This argument will only take effect if there is not a CSS
+#'   \code{height} rule defined for this element; such a rule could come from
+#'   the \code{height} argument of this function or from a containing page
```
hadley: I think I'd remove this — there's much less likely to be a height css rule, than a width css rule.

Alternatively, you might consider `@param rows,cols` or maybe `@param rows,width` to better link the sources of the width.

Or maybe explicitly list the places the size can come from? e.g. The width of the text area can be specified in three ways: with the `cols` argument (specified in number of units), the `width` argument (specified in CSS units), or elsewhere in the CSS (for example the `fluidPage()` layout automatically sets the width to `100%`.

trestletech: Good points. Unfortunately, we inherit `width` via an `@inheritParams`, so that would take some refactoring if we wanted to combine (unless I'm missing some feature there).

I reverted rows to make it sound more likely to work since the default styling doesn't screw it up.

PR #2506, opened 2019-06-18
tham -> than

cpsievert:


wch:
There was a merge conflict due to the switch to Markdown, so I did the same change in 0fc3b90e.

Issue #2505, opened 2019-06-18
Generate docs and JS in CI


trestletech:
We should generate Roxygen docs and JS docs in CI. We can then compare the committed code against the generated versions to see if there are differences. That will allow us to automatically fail our CI pipeline if a proposed PR forgot to update some generated asset.

hadley:
I think you should hold of on doing anything for roxygen2, since we're working on a general approach.

trestletech:
Ah, good to know. I think I've got a solution in https://github.com/rstudio/shiny/pull/2507 but we'll gladly roll over to the official one when it's available.

PR #2507, opened 2019-06-18
Test for Roxygen doc & JS changes in Travis

trestletech:
This work checks to make sure that the Roxygen docs are current in proposed changes. If the rendered docs are stale, it fails the Travis build with an informative message.

Right now this takes multiple minutes, but my understanding is that Travis doesn't cache changes into the packages directory until they're deployed via master. So I believe that once this is merged we should see all the built packages from installing `devtools` and `roxygen2` will become much faster.

Related to #2505 

Happy output from Travis: https://travis-ci.org/rstudio/shiny/jobs/547498661

```
Updating shiny documentation
Writing NAMESPACE
Loading shiny
Writing NAMESPACE
No difference detected; Roxygen docs are current.
```

Sad output from Travis: https://travis-ci.org/rstudio/shiny/jobs/547497580

```
Updating shiny documentation
Writing NAMESPACE
Loading shiny
Writing textAreaInput.Rd
Writing NAMESPACE
 M man/textAreaInput.Rd
Please generate the Roxygen documentation and commit the updates.
The above files changed when we generated the Roxygen documentation. This most often occurs when a user changes the Roxygen documentation in an R file but doesn't regenerate the documentation before committing.
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,12 +6,13 @@ matrix:
       r_packages:
         - devtools
         - roxygen2
-      script:
-        - ./tools/checkDocsCurrent.sh
+      script: ./tools/checkDocsCurrent.sh
     - name: "Javascript check"
       language: node_js
       cache: yarn
       script: ./tools/checkJSCurrent.sh
+      node_js:
+        - "10"
     - name: "Old Release Check"
       r: oldrel
     - name: "Current Release Check"
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,10 @@ r:
   - devel
 sudo: false
 cache: packages
-
+after_script:
+  - git reset --hard && git clean -ffdx
+  - Rscript -e "install.packages(c('devtools', 'roxygen2'))" || travis_terminate 1
+  - ./tools/checkDocsCurrent.sh || travis_terminate 1
 notifications:
   email:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,11 +1,26 @@
...
-  - release
-  - devel
+matrix:
+  include:
+    - name: "Roxygen check"
+      r: release
+      r_packages:
+        - devtools
+        - roxygen2
+      script: ./tools/checkDocsCurrent.sh
```
jimhester: One option would be to put this `before_script`, then you could run this along with the current release check, which would save on a build. The other option would be to do this as a separate job up front before any of the R specific version checks are run...

trestletech:
@jcheng5 I believe I've handled your suggestions, but -- again -- my Bash is pretty limited. Let me know if you see anything funky.

.travis.yml
trestletech: Our concern was that doing this in-line with the R tests would pollute our tests with a bunch of other packages. I originally had it at the end of our tests, but it felt silly to do it three times redundantly -- especially now that we have a test for yarn/JS stuff in there, too.

Let me know if you think I'm missing something or if there's a cleaner way to flesh out this matrix.

PR #2522, opened 2019-07-05
Fix reactivePoll leak

### LINKED ISSUE #1548: reactivePoll leaks the polling observer
(I am going to use reactiveFileReader as an example, since it directly uses reactivePoll and is easier to use as a demonstration)

Imagine this use case:
```
reactiveReader <- reactiveFileReader(100, session, somePath, readLines)
observer <- observe({
   content <- reactiveReader()
   ## do some magic stuff with content here
   if (!is.null(observer)) {
      observer$destroy()
      observer <- NULL
      reactiveReader <- NULL
   }
})
```

After executing this code, the programmer will probably think that the reactiveReader is properly cleaned-up and the polling mechanism has been properly disposed of.

Wrong.

Thanks to the leaked observer on line [1316](https://github.com/rstudio/shiny/blob/19623694f585c8e7a8cf2c38e831a6752e5520c6/R/reactives.R#L1316), the polling function will keep getting executed and as far as I know there is no way to actually cancel it.

If you have a long lasting session, this can and will cause a lot of trouble :)

wch:
This fixes #1548. It allows `reactivePoll` objects to clean up when there are no longer any references to them.

Note that there must no longer be any references to the `reactivePoll` object for it to be cleaned up. If there's still a reference to the object, even if that reference isn't consumed by any other reactives/observers, that's not sufficient for it to be cleaned up; in other words, it will keep firing.

Example app:

```R
library(shiny)
shinyApp(
  fluidPage(
    p("Value of ", code("count")),
    verbatimTextOutput("count"),
    p("Value of ", code("i()")),
    verbatimTextOutput("i"),
    actionButton("replace", "Replace count()")
  ),
    i <- reactiveVal(0)
    count <- reactivePoll(1000, session,
      checkFunc = function() {
        isolate({
          i(i() + 1)
          i()
        })
      },
      valueFunc = function() {
        isolate(i())
      }
    )

    output$count <- renderText({
      count()
    })
    output$i <- renderText({
      i()
    })

    observeEvent(input$replace, {
      count <<- reactive("count() has been replaced")
      gc()
    })
  }
)
```

Here's how it behaves before this PR:

And after this PR:

Note that before the change, the replacement reactive, which has the value `"count() has been replaced"`, shows up in `output$count`, but after the change, it doesn't. This is because, before the change, the invalidation keeps happening on the timer and causes the `renderText` to re-execute; but after the change, the invalidation stops happening after `count()` is replaced, so it does not cause the `renderText` to re-execute.

Here's a test that works outside of a Shiny app:

```R
library(shiny)
i <- 0
count <- reactivePoll(100, NULL,
  checkFunc = function() {
    i <<- i + 1
    i
  },
  valueFunc = function() i
)

observe({
  message(count())
})
while(i < 3) {
  Sys.sleep(0.1)
  shiny:::timerCallbacks$executeElapsed()
  shiny:::flushReact()
}

rm(count)
gc()

# If the reactivePoll was cleaned up, then running the following should no
# longer increment i.
Sys.sleep(0.2)
shiny:::timerCallbacks$executeElapsed()
shiny:::flushReact()
Sys.sleep(0.2)
shiny:::timerCallbacks$executeElapsed()
shiny:::flushReact()

# This is the crucial part
stopifnot(i == 3)
```

Update: With the final version of this PR, in test app, `i()` will go one number higher than `count` -- what's important is that `i()` stops at some point. Also, in the test code for outside of a Shiny app, it should check for `stopifnot(i == 3)` (this is in the unit tests in the PR).

### DIFF
--- a/R/reactives.R
+++ b/R/reactives.R
@@ -1733,7 +1733,8 @@ reactivePoll <- function(intervalMillis, session, checkFunc, valueFunc) {
 
   rv <- reactiveValues(cookie = isolate(checkFunc()))
 
-  observe({
+  o <- observe({
+    cat(".")
     rv$cookie <- checkFunc()
     invalidateLater(intervalMillis(), session)
   })
@@ -1745,6 +1746,18 @@ reactivePoll <- function(intervalMillis, session, checkFunc, valueFunc) {
     valueFunc()
 
   }, label = NULL)
+
+  # When no one holds a reference to this object anymore, destroy and remove the
+  # observer so that it doesn't keep firing, and hold onto resources.
+  reg.finalizer(attr(re, "observable"), function(e) {
+    message("finalizing reactive")
+    o$destroy()
+    rm(o, envir = parent.env(environment()))
+  })
+
+  # So that the observer and finalizer function don't (indirectly) hold onto a
+  # reference to `re` and thus prevent it from getting GC'd.
+  on.exit(rm(re))
 
   return(re)
 }
--- a/R/reactives.R
+++ b/R/reactives.R
@@ -1733,7 +1733,17 @@ reactivePoll <- function(intervalMillis, session, checkFunc, valueFunc) {
 
   rv <- reactiveValues(cookie = isolate(checkFunc()))
 
+  re_finalized <- FALSE
+
   o <- observe({
+    # When no one holds a reference to the reactive returned from
+    # reactivePoll, destroy and remove the observer so that it doesn't keep
+    # firing and hold onto resources.
+    if (re_finalized) {
+      o$destroy()
+      rm(o, envir = parent.env(environment()))
+    }
+
     rv$cookie <- checkFunc()
     invalidateLater(intervalMillis(), session)
   })
@@ -1746,11 +1756,8 @@ reactivePoll <- function(intervalMillis, session, checkFunc, valueFunc) {
 
   }, label = NULL)
 
-  # When no one holds a reference to this object anymore, destroy and remove the
-  # observer so that it doesn't keep firing, and hold onto resources.
-  safe_finalizer(attr(re, "observable"), function(e) {
-    o$destroy()
-    rm(o, envir = parent.env(environment()))
+  reg.finalizer(attr(re, "observable"), function(e) {
+    re_finalized <<- TRUE
   })
 
   # So that the observer and finalizer function don't (indirectly) hold onto a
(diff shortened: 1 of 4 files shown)


R/reactives.R
```diff
@@ -1715,7 +1715,8 @@ reactivePoll <- function(intervalMillis, session, checkFunc, valueFunc) {
 
   rv <- reactiveValues(cookie = isolate(checkFunc()))
 
-  observe({
+  o <- observe({
+    cat(".")
```
jcheng5: ?

R/reactives.R
```diff
@@ -1728,6 +1729,18 @@ reactivePoll <- function(intervalMillis, session, checkFunc, valueFunc) {
 
   }, label = NULL)
 
+  # When no one holds a reference to this object anymore, destroy and remove the
+  # observer so that it doesn't keep firing, and hold onto resources.
+  reg.finalizer(attr(re, "observable"), function(e) {
+    message("finalizing reactive")
+    o$destroy()
```
jcheng5: I'm scared of this... can we do (a lot) less work inside the finalizer, preferably all in C?

How about we make our own version of reg.finalizer, that's built on top of reg.finalizer but instead of invoking your callback inline with reg.finalizer, it just schedules your callback to be invoked at a more convenient moment? (The callback would need to not access the object being finalized)

R/reactives.R
wch: Oh yeah, I should have mentioned that this is meant to be a draft PR. That's there for debugging.

R/reactives.R
wch: I don't think we need to use C to schedule things to happen later. (If we do use C, that would be new for Shiny.) When do you have in mind for running the actual finalizers?

wch: Would `later(callback, 0)` be a reasonable solution?

wch: Note: the original `reg.finalizer` call has been replaced with `safe_finalizer` here.

tests/testthat/test-reactivity.r
```diff
@@ -1333,31 +1333,33 @@ test_that("reactivePoll doesn't leak observer (#1548)", {
...
+    Sys.sleep(0.05)
+    shiny:::timerCallbacks$executeElapsed()
+    shiny:::flushReact()
+  }
 
-  expect_equal(i, 3L)
+  # It is possible in the future we will have a better method which will remove
+  # the observer without requiring the one extra run. If that happens, the
+  # expected value will be 3 instead of 4.
+  expect_equal(i, 4L)
```
jcheng5: It's the future! (I think the `4L` needs to be `3L` now; it was the missing `return()` that was causing the one extra run.)

jcheng5: Also the unit test is failing because of this.

PR #2523, opened 2019-07-06
Support reading of Shiny outputs

jcheng5:
Until now, you've been able to assign to the `output` object (i.e. `output$table <- renderTable(...)`) but we throw an error if you try to read from it (i.e. `print(output$plot)`). The reason for this is because I didn't want Shiny beginners to think they could read the current plot.

However, with shinymeta we now have a need for reading outputs. Although shinymeta enables an important scenario, this is admittedly a pretty advanced, niche use; so this PR adds the functionality, but you have to enable it with a global option `shiny.allowoutputreads` (undocumented for now).

I've implemented this as an R option instead of a `shinyOption` because we explicitly do not want to tie the lifetime of the option to an app, but instead to the R process lifetime (I am setting this option in shinymeta's `.onLoad`).

### DIFF
--- a/R/shiny.R
+++ b/R/shiny.R
@@ -1088,6 +1088,9 @@ ShinySession <- R6Class(
         # will be attached to the observer after it's created.
         outputAttrs <- attr(func, "outputAttrs", TRUE)
 
+        # Save this for getOutput purposes
+        outputAttrs$renderFunc <- func
+
         funcFormals <- formals(func)
         # ..stacktraceon matches with the top-level ..stacktraceoff.., because
         # the observer we set up below has ..stacktraceon=FALSE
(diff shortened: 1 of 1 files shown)


jcheng5:
@wch Can I ask you to take a look at this early next week? I'd love to have it merged to master before useR so shinymeta can stop relying on a specific Shiny branch.

wch:
One more thought: it would be good to document `shiny.allowoutputreads` in the `shiny-options` help page.

jcheng5:
@wch Would it be bad to keep it undocumented? 😬 I couldn't think of a legitimate use for this outside of shinymeta.

wch:
Well, we do have other weird things documented there which no one should normally use... But if you feel strongly about it, I'm OK with not documenting it.

PR #2524, opened 2019-07-11
selectInput: improve factor handling

### LINKED ISSUE #2515: Shiny gives error if named selectInput choices are a factor
Selection inputs in Shiny allow you to provide a named vector of choices so that the names are shown to the user but the values are what are passed to input. It was designed for the choices to be a character vector, but it's easy to accidentally pass it a named factor vector. If you do you get an extremely confusing error.

It seems like selectInput should either:
* convert the vector from factor to character with as.character, or
* provide a more clear error.

Here is an example showing the issue.

Thanks!

```r

library(shiny)

choices <- letters

choices <- factor(choices) # comment this line out and the app works
# if the choices are a factor instead of character, shiny gives the error:
# Error in (function (choice, name)  : 
#   All sub-lists in "choices" must be named.

names(choices) <- LETTERS

ui <- fluidPage(selectInput("example", "Choose", choices))
server <- function(input, output) {}

```

alandipert:
This work was motivated by #2515, which demonstrated that unnamed factors work as expected, but named factors did not.

1. `listify()` was restructured into a generic function and a series of S3 methods in order to make clearer how various types are handled. Methods were added for all previously-supported classes, and support for named factors was added.
1. `selectInput()` documentation now mentions factor handling.
1. `makeNamed()` was renamed to `ensureNamed()` to indicate better that it adds names only if they are not present.
1. Additional unit tests were added to check factor handling, and existing unit tests were improved to exercise handling of `complex` and `logical` vectors.

## Open Questions

1. Should `selectInput()` produce a warning if a factor is used as an option list or otherwise included in the list of options?


R/input-utils.R
```diff
@@ -90,31 +90,39 @@ generateOptions <- function(inputId, selected, inline, type = 'checkbox',
...
+listify <- function(x) UseMethod("listify", x)
+
+#' @export
+listify.list <- function(x) ensureNamed(lapply(x, listify))
+#' @export
+listify.NULL <- listify.list
+
+#' @export
+listify.character <- function(x) {
+  if (length(x) == 1 && is.null(names(x))) x else as.list(ensureNamed(x))
```
schloerke: `... else ensureNames(as.list(x))` ?

Is the goal to return a list with every element's name as `''` or to return a list with no names?

alandipert: In that case, return a list with every element's name as `''` (unless it's already named)

R/input-utils.R
```diff
@@ -90,31 +90,39 @@ generateOptions <- function(inputId, selected, inline, type = 'checkbox',
...
+# If a list/vector is unnamed, give it blank names
+ensureNamed <- function(x) {
+  if (is.null(names(x))) names(x) <- character(length(x))
+  x
+}
+
+# Take a vector or list, and convert to list. Also, if any children are
+# vectors with length > 1, convert those to list. If the list is unnamed,
+# convert it to a named list with blank names.
+listify <- function(x) UseMethod("listify", x)
```
schloerke: My gut reaction is this should be rewritten to be a `if` / `else`.  

* It wouldn't need to be exported to work.
* It could handle odd s3 cases better (things that don't have atomic classes).

Setup: 
```
if `is.list`, `is.null`
  "code for `is.list`"
else if `is.numeric`, `is.complex`, `is.logical`, `is.factor`
  "code for `is.numeric`"
else 
  "code for `is.character`"
```

R/input-utils.R
schloerke: That's my interpretation of https://github.com/rstudio/shiny/pull/2524/files#diff-c851d180beb8024d6b46a73dab4756d9L113 .  `makeNamed(as.list(val))` vs `as.list(ensureNamed(x))`

I like your version better.  "Set all the names to the character values if names does not exist"

R/input-utils.R
alandipert: You know, now that it has all the `@export` noise, I have to agree. I swear, it was looking pretty tight before all that 😄

wch: Since the `listify` generic isn't exported from the package, you don't need to `@export` every method, and therefore nothing will be added to the NAMESPACE file. That said, I also lean toward having a single function with if-else here.

wch:
One thought -- we should make sure there are tests for lists (and atomic vectors) where the list is named, but some or all of the names are `""`. For example:

```R
x <- list(A="a", "b", "c")
x
#> $A
#> [1] "a"
#> 
#> [[2]]
#> [1] "b"
#> 
#> [[3]]
#> [1] "c"
attributes(x)
#> $names
#> [1] "A" ""  ""

y <- x[2:3]
y
#> [[1]]
#> [1] "b"
#> 
#> [[2]]
#> [1] "c"
attributes(y)
#> $names
#> [1] "" ""

# Contrast y with:
z <- list("b", "c")
z
#> [[1]]
#> [1] "b"
#> 
#> [[2]]
#> [1] "c"
attributes(z)
#> NULL
```

alandipert:
@wch there's a test for that case: https://github.com/rstudio/shiny/blob/0d7b14969a6deb52d2b1fd9413aae5e78e23b6a7/tests/testthat/test-bootstrap.r#L89

R/input-utils.R
```diff
@@ -90,31 +90,32 @@ generateOptions <- function(inputId, selected, inline, type = 'checkbox',
   div(class = "shiny-options-group", options)
 }
 
+# If a list/vector is unnamed, give it blank names
+ensureNamed <- function(x) {
```
wch: It turns out we already have a function that does this, so we should use that:
https://github.com/rstudio/shiny/blob/250303790c1121a0f8de7bf929e786696e93ab25/R/utils.R#L176-L183

I like your implementation better than the one that currently exists for `asNamedVector`, though.

R/input-utils.R
```diff
@@ -90,31 +90,32 @@ generateOptions <- function(inputId, selected, inline, type = 'checkbox',
...
-        listify(val)
-      else if (length(val) == 1 && is.null(names(val)))
-        as.character(val)
-      else
-        makeNamed(as.list(val))
-    })
-
-    makeNamed(res)
+  } else {
+    listify(stats::setNames(as.character(x), names(x)))
```
wch: It would help to have a comment saying that this is the code path where factors will end up (otherwise it's non-obvious what types take this path and we may forget in the future).

PR #2525, opened 2019-07-15
Add noscript to basic template

carlganz:
Since Shiny requires Javascript to work (I believe), this small bit added to basic template should provide a warning in unlikely scenario a user tries to view a Shiny application with Javascript disabled.

Feel free to disregard


wch:
@carlganz Thanks for the thinking about this. However, I think it goes without saying these days that web apps need JavaScript to work.

PR #2540, opened 2019-08-02
Fix selectInput/selectizeInput handling character(1) options

alandipert:
Addresses a bug found by Shalu in testing #2524 that manifested with the example 013-selectize.

## Notes

Refactoring this code took a *lot* more time than expected. The original version was complex for at least these reasons (or, excuses):

* `choicesWithNames()` was recursive
* `listify()`, called within `choicesWithNames()`, was itself recursive
* `listify()` dispatched internally on combinations of type, named-ness, and length in a single if/else block

The recursion was needless because no other function that depend on `choicesWithNames()` need to process arbitrarily-deep input trees. `selectizeInput()` requires a tree of depth = 2 to process groups, but that's it.

I believe the principal enhancement with this change is the elimination of recursion. Now, it's obvious the 2 broad categories of input accepted: a flat choice list, or a choice list with some groups, each of which will be processed as a flat choice list.

### DIFF
--- a/R/input-utils.R
+++ b/R/input-utils.R
@@ -155,9 +155,7 @@ processGroupedChoices <- function(choices) {
 # Takes a vector/list/factor, and adds names (same as the value) to any entries
 # without names. Coerces all leaf nodes to `character`.
 choicesWithNames <- function(choices) {
-  if (length(choices) == 0) {
-    choices
-  } else if (hasGroups(choices)) {
+  if (hasGroups(choices)) {
     processGroupedChoices(choices)
   } else {
     processFlatChoices(choices)
--- a/R/input-utils.R
+++ b/R/input-utils.R
@@ -125,10 +125,12 @@ processFlatChoices <- function(choices) {
 
 # Processes a "nested" set of choices, or a collection of choices that contains
 # one or more named groups of related choices and zero or more "flat" choices.
-# choices should be named, and any choice group must have a non-empty name.
-# Empty names of remaining "flat" choices are replaced with that choice's value
-# coerced to a character.
+# choices should be a named list, and any choice group must have a non-empty
+# name. Empty names of remaining "flat" choices are replaced with that choice's
+# value coerced to a character.
 processGroupedChoices <- function(choices) {
+  # We assert choices is a list, since only a list may contain a group.
+  stopifnot(is.list(choices))
   choices <- mapply(function(name, choice) {
     choiceIsGroup <- isGroup(choice)
     if (choiceIsGroup && name == "") {
--- a/R/input-utils.R
+++ b/R/input-utils.R
@@ -90,21 +90,22 @@ generateOptions <- function(inputId, selected, inline, type = 'checkbox',
   div(class = "shiny-options-group", options)
 }
 
-# Take a vector or list, and convert to named list. Also, if any children are
+# Take a vector/list/factor, and convert to named list. Also, if any children are
 # vectors with length > 1, convert those to list. If the list is unnamed,
 # convert it to a named list with blank names.
-listify <- function(x) {
+#' @importFrom stats setNames
+listify <- function(x, child = FALSE) {
   if (is.list(x) || is.null(x)) {
-    asNamedVector(lapply(x, listify))
-  } else if (is.character(x)) {
-    if (length(x) == 1 && is.null(names(x))) {
-      x
-    } else {
-      as.list(asNamedVector(x))
-    }
+    # List children are processed and a named list is returned
+    asNamed(lapply(x, listify, TRUE))
+  } else if (child && is.null(names(x)) && length(x) == 1) {
+    # Unnamed children of length = 1 are considered leaves and are converted to
+    # a character(1)
+    as.character(x)
   } else {
-    # Can get here if x is a factor.
-    listify(stats::setNames(as.character(x), names(x)))
+    # Anything else is converted to a (possibly named) character vector, then a
+    # named list
+    asNamed(setNames(as.list(as.character(x)), names(x)))
   }
 }
 
--- a/R/input-utils.R
+++ b/R/input-utils.R
@@ -109,8 +109,7 @@ setDefaultNames <- function(x) {
   x
 }
 
-# Makes a character vector out of x in a way that preserves names if x is a
-# factor.
+# Makes a character vector out of x in a way that preserves names.
 asCharacter <- function(x) {
   stats::setNames(as.character(x), names(x))
 }
--- a/R/input-utils.R
+++ b/R/input-utils.R
@@ -90,93 +90,58 @@ generateOptions <- function(inputId, selected, inline, type = 'checkbox',
   div(class = "shiny-options-group", options)
 }
 
-# This pass accepts a compound structure and returns a "normalized" structure
-# that simplifies subsequent passes.
-#
-# Input: A tree consisting of named and unnamed lists and atomic vectors.
-#
-# Output: A tree consisting of possibly-named lists and unnamed character(1)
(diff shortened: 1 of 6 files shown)


R/utils.R
```diff
@@ -173,8 +173,8 @@ anyUnnamed <- function(x) {
 }
 
 
-# Given a vector/list, returns a named vector (the labels will be blank).
-asNamedVector <- function(x) {
```
wch: A historical note: the reason this function was called `asNamedVector` is that a list is a kind of vector. For example:

```R
vector(mode = "list")
```

That said, I agree that the name `asNamed` is a little clearer, because it avoids the ambiguity of what "vector" means in R -- sometimes it means atomic vector, sometimes it has the broader meaning which includes lists.

R/input-utils.R
```diff
@@ -90,21 +90,25 @@ generateOptions <- function(inputId, selected, inline, type = 'checkbox',
   div(class = "shiny-options-group", options)
 }
 
-# Take a vector or list, and convert to named list. Also, if any children are
+# Take a vector/list/factor, and convert to named list. Also, if any children are
 # vectors with length > 1, convert those to list. If the list is unnamed,
 # convert it to a named list with blank names.
-listify <- function(x) {
+listify <- function(x, child = FALSE) {
```
wch: I'm finding the logic for this function really hard to follow and I want to make sure that it's well documented so that in the future we understand what's going on here. This function has always been surprisingly difficult to understand -- I've spent over an hour trying to fully understand it, and many hours in the past trying to understand previous versions of it, so clearly something isn't right.

I think this function currently does a mix of different things. This is what I think it does -- please feel free to correct me if this isn't right.

The input can be one of:
* Unnamed vector (usually character, but other types are OK)
* Named vector (usually character, but other types are OK)
* Unnamed factor
* Named factor
* Unnamed list
* Named list

If the input is a list, then each element of the list can be any of those types. However, it doesn't recurse past that; there can't be any grandchild nodes. (Note: I just realized that this function currently allows grandchildren, but the shiny input functions which call this function do not make use of grandchildren.)

For top-level inputs, this function does the following for each type.

* Unnamed vector: Converts to named list where names are same as values.
* Named vector: Converts to named list.
* Unnamed factor: Converts to named list where names are same as values.
* Named factor: Converts to named list.
* Unnamed list: Adds names and recurses. (Note: how does it choose the names if the children aren't single elements?)
 Named list: Recurses.

For children, this function does the following for each type:

* Unnamed vector:
  * If length 1: Converts to char vector (does not add names).
  * If length >1: Converts to named list where names are same as values.
* Named vector: Converts to named list.
* Unnamed factor:
  * If length 1: Converts to char vector (does not add names).
  * If length >1: Converts to named list where names are same as values.
* Named factor: Converts to named list.
* Unnamed list: ??
* Named list: ??

I've put "??" in some places because I haven't figured out exactly what happens

At any rate, I think that all this complexity has led to the bugs that we've had with this function, and we probably should do this a different way, like having multiple functions that each do a simpler thing to the input. (Though if you can describe what's happening in a way that's simple to understand and can put it in a comments, that would also be a reasonable way to deal with this.)

alandipert: To answer your question, named and unnamed list children become named lists (first branch).

In a previous attempt to clarify this function I made it generic, but maybe that wasn't the home run I hoped because type is only one of the three things the function dispatches on. Depth and named-ness are the others. It was also significantly longer.


That grandchildren can't be is a great observation. That simplifies the concept of depth and illuminates an opportunity to throw an error instead of process deeper.

R/input-utils.R
```diff
@@ -90,41 +90,93 @@ generateOptions <- function(inputId, selected, inline, type = 'checkbox',
...
+# and FALSE otherwise.
+branchHasNames <- function(branch) {
+  stopifnot(is.list(branch))
+  ns <- names(branch)
+  !is.null(ns) && !length(ns[ns == ""])
+}
+
 # Takes a vector or list, and adds names (same as the value) to any entries
 # without names. Coerces all leaf nodes to `character`.
 choicesWithNames <- function(choices) {
```
wch: Maybe have a `maxDepth` argument here.

tests/testthat/test-bootstrap.r
```diff
@@ -69,6 +69,11 @@ test_that("Repeated names for selectInput and radioButtons choices", {
 
 
 test_that("Choices are correctly assigned names", {
+  # Empty character vector
+  expect_identical(
+    choicesWithNames(c("")),
```
wch: I think this should test a non-empty string, because `""` isn't a name that will work with Shiny.

R/input-utils.R
```diff
@@ -90,41 +90,93 @@ generateOptions <- function(inputId, selected, inline, type = 'checkbox',
...
+  }
+}
+
+# Computes the depth of a tree given by the isBranch and children functions.
+# - isBranch() should return TRUE if the given tree is a branch and FALSE
+#   otherwise
+# - children() should return a branch's children as a list.
+getDepth <- function(tree, isBranch = is.list, children = identity, depth = 0) {
+  if (isBranch(tree)) {
+    max(sapply(children(tree), getDepth, isBranch, children, depth + 1))
```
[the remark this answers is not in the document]
alandipert: note to self: vapply with numeric(1) -- type of each element

R/input-utils.R
```diff
@@ -90,41 +90,58 @@ generateOptions <- function(inputId, selected, inline, type = 'checkbox',
...
+# the corresponding value coerced to a character(1)
+setDefaultNames <- function(x) {
+  x <- asNamed(x)
+  emptyNames <- names(x) == ""
+  names(x)[emptyNames] <- as.character(x)[emptyNames]
+  x
+}
+
+# Makes a character vector out of x in a way that preserves names if x is a
+# factor.
```
wch: The comment needs a tweak: this actually is important for preserving names for all types of vectors (not just factors), because `as.character` always drops names.

R/input-utils.R
```diff
@@ -90,41 +90,77 @@ generateOptions <- function(inputId, selected, inline, type = 'checkbox',
...
-    } else {
-      as.list(asNamedVector(x))
-    }
-  } else {
-    # Can get here if x is a factor.
-    listify(stats::setNames(as.character(x), names(x)))
-  }
+# True when a choice list item represents a group of related inputs.
+isGroup <- function(choice) {
+  length(choice) > 1 || !is.null(names(choice))
```
wch: Is this condition correct for a leaf node where it's a named list/vector with an empty name?

```R
x <- list(A="a", "b")[2]
x
#> [[1]]
#> [1] "b"
dput(x)
structure(list("b"), .Names = "")
```

alandipert: In the existing code it's considered a group, so the condition is correct in that respect. I can see a good argument for changing that behavior, though. That object is leafier than it is groupy.

tests/testthat/test-bootstrap.r
alandipert: Note to self: add comment explaining seemingly-pointless test

R/input-utils.R
```diff
@@ -90,41 +90,58 @@ generateOptions <- function(inputId, selected, inline, type = 'checkbox',
...
-    choicesWithNames(choice)
-  }, SIMPLIFY = FALSE)
-
-  # default missing names to choice values
-  missing <- names(choices) == ""
-  names(choices)[missing] <- as.character(choices)[missing]
-
-  choices
+  if (length(choices) == 0) {
+    choices
```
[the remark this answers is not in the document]
alandipert: Note to self: need to port this https://github.com/rstudio/shiny/blob/master/R/input-utils.R#L115-L116

PR #2545, opened 2019-08-07
Fix selected dynamic tab in deployed environments

### LINKED ISSUE #2116: appendTab and insertTab do not work correctly on Connect
To see the problem:

* Visit https://beta.rstudioconnect.com/content/3819/
* Click the "Add 'Dynamic' Tab" button
* Click on the newly-created tab
* Click the button two more times

Clicking on the second and third new tabs will have the same content. The problem is that the tab IDs have `NaN`. 

image

They get `NaN` because the pattern that Shiny is looking for is a relative URL like `href="#tab-1093-1"`, not an absolute one like `href="https://beta.rstudioconnect.com/content/3819/_w_9c84ffcd/#tab-1093-1"`.

This problem does not happen when running locally. The problem likely has to do with a `<base>` tag in the HTML of a deployed app, added by shiny-server-client.

This is the code for the app:

```R
options(shiny.minified = FALSE)

ui <- fluidPage(
  sidebarLayout(
    sidebarPanel(
      actionButton("add", "Add 'Dynamic' tab")
    ),
    mainPanel(
      tabsetPanel(id = "tabs",
        tabPanel("Hello", "This is the hello tab")
      )
    )
  )
)
  n <- 0
  observeEvent(input$add, {
    appendTab(inputId = "tabs",
      {
        n <<- n + 1
        tabPanel(
          "Dynamic",
          paste("Content for dynamic tab", n)
        )
      }
    )
  })
}

shinyApp(ui, server)
```

schloerke:
Fixes #2116

This fixes https://github.com/rstudio/shiny/issues/2116 but it does not fix the underlying issue.

Underlying issue: When tabs are clicked on, the relative urls are upgraded to full urls.  I can not determine where that is happening.  For now, we will remove everything that is before the hash.

My guess is that bootstrap is doing this.


schloerke:
Bootstrap is not. Best to just remove the url and retrieve the relative url only.  Original bug caused by shiny-client. Can't remove it.  Best to just patch it as the only time we are getting to this code path is to change tabs.

PR #2547, opened 2019-08-08
Automatically load 'helpers' in R/ directory at runtime

trestletech:
Sources files ending in `.r` (case-insensitive) found in the `R/` directory (case-sensitive) when starting the app. If using `app.R`, the helpers are sourced before `app.R`. If using `server.R`, the `global.R` file is source first, then the helpers, then your `server.R` and `ui.R` files (when needed).

#### With `server.R` and `ui.R`:

```
+ global env (global.R)
|
+-+ helpers env (R/ files)
| |
| +-- ui.R env
| |
| +-- server.R env
```

The only change here is the introduction of a new environment in between the child ui and server environments into which the helpers are loaded.

#### With `app.R`

```
+ global env
|
+-+ helpers env (R/ files)
| |
| +-- app.R
```

### Breaking changes

 - We automatically source all scripts in the top level of R/ when an app is run

### Testing Notes

1. Try running the shiny example 168 against Shiny installed from CRAN. I have this app [here](https://github.com/rstudio/shiny-examples/compare/jeff/helpers2) (not merged, since it doesn't work against Shiny on master). You'll see that it fails with an error like:

```
ERROR: could not find function "counterButton"
```

2. Install this branch: `remotes::install_github("jeff/feature/helpers")`
3. Restart R
4. Rerun example app 168 and confirm that the app now runs. You should see a button that you can click on to increment a counter. If that works, that demonstrates that the R file in the R/ directory was automatically loaded.


R/app.R
```diff
@@ -288,18 +295,37 @@ initAutoReloadMonitor <- function(dir) {
   obs$destroy
 }
 
+# Loads in all helpers in the R/ directory of the app
+# From `list.files`:
+# > The files are sorted in alphabetical order, on the full path
+loadHelpers <- function(appDir, envir=globalenv()){
+  helpersDir <- file.path(appDir, "R")
+  if (!file.exists(helpersDir)){
+    # We may be on a case-sensitive filesystem and have a dir named r/
```
hadley: I don't think you should allow this, as packages require `R/` not `r/`

R/app.R
```diff
@@ -288,18 +295,37 @@ initAutoReloadMonitor <- function(dir) {
...
+    helpersDir <- file.path(appDir, "r")
+  }
+  helpers <- list.files(helpersDir, pattern="\\.[rR]$", recursive=TRUE, full.names=TRUE)
```
hadley: I'd also suggest not loading recursively, again to increase compatibility with package code loading.

trestletech: Ah, I didn't realize packages were non-recursive.

Do you feel strongly on this point? I feel like large codebases will require multiple directories to organize content (the Angular structure comes to mind [e.g.](https://scotch.io/tutorials/angularjs-best-practices-directory-structure)). 

You could obviously accomplish that with R files at the base that `source` the inner directories, but that feels odd to me. Do people often have R files nested inside directories within `R/` that would be problematic if they were sourced?

trestletech: Oh, counter-point to my previous comment: if we want to maintain compatibility with golem (i.e. allow users to painlessly flip this structure over to using an R package as the organizational construct), then supporting something that R packages don't support would be problematic.

Maybe we just do the top-level for now and then if/when packages support nested directories, we can follow suit.

hadley: :+1:

R/app.R
```diff
@@ -288,18 +293,53 @@ initAutoReloadMonitor <- function(dir) {
...
+#'   [list.files]). `global.R` is evaluated before the supporting R files in the
+#'   `R/` directory.
+#' @param appDir The application directory
+#' @param renv The environmeny in which the files in the `R/` directory should
+#'   be evaluated.
+#' @param globalrenv The environment in which `global.R` should be evaluated. If
+#'   `NULL`, `global.R` will not be evaluated at all.
+#' @export
+loadSupport <- function(appDir, renv=new.env(parent=globalenv()), globalrenv=globalenv()){
+  if (!is.null(globalrenv)){
```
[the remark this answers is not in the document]
trestletech: I use it in the `app.R` case to signify that we don't want to consider `global.R`. https://github.com/rstudio/shiny/pull/2547/files#diff-51c87956ac3c179689d4ababd8f62bfcR386

I had considered that making an additional, explicit parameter but decided to consolidate it here.

R/app.R
```diff
@@ -288,18 +293,53 @@ initAutoReloadMonitor <- function(dir) {
   obs$destroy
 }
 
+#' Load an app's supporting R files
+#'
+#' Loads all of the supporting R files of a Shiny application. Specifically,
+#' this function loads any top-level supporting `.R` files in the `R/` directory
+#' adjacent to the app and a `global.R` file.
```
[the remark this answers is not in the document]
trestletech: Correct. I'll just spell those out.

R/app.R
```diff
@@ -137,12 +137,30 @@ shinyAppFile <- function(appFile, options=list()) {
   shinyAppDir_appR(basename(appFile), appDir, options = options)
 }
 
+#' Determine which renv to use to source the helpers.
+#' Currently returns NULL (disabling autoloading of R/*.R) unless
+#' the user opts in via the sets the option `shiny.autoload.r=TRUE`.
+#' @noRd
+getREnv <- function(se=sharedEnv) {
```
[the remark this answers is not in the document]
trestletech: 🤦‍♂️ forgot how parameters work for a minute there...

BetweenTwoTests:
I do agree with the idea and intent of auto loading `R` folder. It makes sense with R semantics.
My only worry is that if this default setting is in the dev branch already, then the next update will surprise some of the existing project maintainers. We are all R nerds, so we can't wait to meticulously read the latest package update docs as soon as they come out... but not many people do :)

Updating package version is expected to have breaking changes, but I feel like this implicit default is somewhat "hidden". A scenario I can think of is that a user uses R/ folder to have different data versions loaded for one-off analysis while maintaining Shiny app to display some pegged data version. If variable shadowing happens, then the Shiny app will still work (but with the wrong data). If the app is complicated or the end user is not the app developer, then this will go unnoticed for long time.

Perhaps doing a staggered update with notifications like how tidyverse packages notifies future breaking changes might be good? I know that this is open source, so these kinds of issues are expected. I think this is a bigger issue of communication between open source project and programmers. There is no answer to that question, but I am still concerned about the "wrong data" scenario that I described above. Those types of bugs are "undetectable"

trestletech:
Just a couple of small updates here: 

 1. I just opened https://github.com/rstudio/shiny/pull/2686 which adds messaging to make this change more prominent for people who would be affected.
 2. We're still gathering feedback, but the plan is to push this forward. We do see the value in rolling this out in stages over multiple releases, but worry that people who skip multiple releases would still be impacted. Since there's not a bulletproof way to do this without impacting anyone, the argument is that our best shot at causing the least damage possible is to release this all at once and make a big deal about it on our blog and Twitter (along with adding prominent messages in the R process itself as done in ^). We hope that this will make folks aware of the change and minimize any fallout.

We're still multiple months away from having all of that in place or cutting the next release to CRAN, but just wanted to keep you in the loop.

PR #2548, opened 2019-08-09
More supports on class, style, and other attributes for UI

atusy:
This PR makes following changes as written in NEWS.md

* Added `class` and `...` arguments to `*Output` functions, so that these functions
  can take more controlls on the outputs' containers (generally `div` or `span`).

* Added `style` arguments to `plotOutput` and `imageOutput`, so that addtional CSS
  styles can be applied to them.

* Resolved inconsitency in `class` argument among `downloadButton()`, `downloadLink()` and `icon()`.
  The `class` argument can be a character vector for `downloadLink()`, but not for others.
  Now all of them accepts a vector.

# Usage example

To apply `display: inline-block` to plots to arrange them,
currently they have to be wrapped by additional `div` (see `ui`).

This PR removes additional `div` (see `ui_new`).

```
library(shiny)
library(ggplot2)

p <- ggplot2::qplot(1, 1)

  output$a <- output$b <- output$c <- renderPlot(p)
}

# ui's are wrapped by div
ui <- fluidPage(
  plotOutput("a", width = 200, height = 200) %>% tags$div(style = "display: inline-block"),
  plotOutput("b", width = 200, height = 200) %>% tags$div(style = "display: inline-block"),
  plotOutput("c", width = 200, height = 200) %>% tags$div(style = "display: inline-block")
)

# ui's does not have to be wrapped by div anymore
ui_new <- fluidPage(
  plotOutput("a", width = 200, height = 200, style = "display: inline-block"),
  plotOutput("b", width = 200, height = 200, style = "display: inline-block"),
  plotOutput("c", width = 200, height = 200, style = "display: inline-block")
)

# compare
runGadget(ui, server)
runGadget(ui_new, server)

```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,6 +7,16 @@ shiny 1.3.2.9001
 
 ### Improvements
 
+* Added `class` and `...` arguments to `*Output` functions, so that these functions
+  can take more controlls on the outputs' containers (generally `div` or `span`).
+
+* Added `style` arguments to `plotOutput` and `imageOutput`, so that addtional CSS
+  styles can be applied to them.
+
+* Resolved inconsitency in `class` argument among `downloadButton()`, `downloadLink()` and `icon()`.
+  The `class` argument can be a character vector for `downloadLink()`, but not for others.
+  Now all of them accepts a vector.
+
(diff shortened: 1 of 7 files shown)


atusy:
Note:  I have signed to contributor agreement when I contributed to `bookdown`

Issue #2461, opened 2019-05-30
All input functions should define what they yield


hadley:
Currently they have a `@returns` section which describe what the function returns. But I think they all also need a section that describes what they yield in the server function. `fileUpload()` puts this in the "details" section, but I think it would be more useful to have a consistent section name across all functions.

PR #2551, opened 2019-08-13
[WIP] Define what all input functions yield for server

leslie-huang:
Enhancement to address #2461 

@schloerke

### DIFF
--- a/R/input-action.R
+++ b/R/input-action.R
@@ -37,6 +37,10 @@
 #' }
 #'
 #' @seealso [observeEvent()] and [eventReactive()]
+#'
+#' @section Server value:
+#' Returns an integer, initialized as 0 and incremented with each press.
+#'
 #' @export
 actionButton <- function(inputId, label, icon = NULL, width = NULL, ...) {
 
--- a/R/input-action.R
+++ b/R/input-action.R
@@ -39,7 +39,7 @@
 #' @seealso [observeEvent()] and [eventReactive()]
 #'
 #' @section Server value:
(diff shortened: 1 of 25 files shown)


R/input-action.R
```diff
@@ -37,6 +37,10 @@
 #' }
 #'
 #' @seealso [observeEvent()] and [eventReactive()]
+#'
+#' @section Server value:
+#' An integer, initialized as 0 and incremented with each press.
```
jcheng5: It might be worth noting that this `0` is slightly different than any other `0` value, in that it is considered falsy (see `?req`). Basically, `req(input$btn)` on an unpressed button should throw, while `req(input$num)` where `input$num==0` should not.

R/input-daterange.R
```diff
@@ -70,6 +70,10 @@
 #'
 #' shinyApp(ui, server = function(input, output) { })
 #' }
+#'
+#' @section Server value:
+#' A vector of two `Date` objects.
```
jcheng5: Should link to `base::Date`

R/input-select.R
```diff
@@ -72,6 +72,10 @@
 #'   }
 #' )
 #' }
+#'
+#' @section Server value:
+#' A vector of character strings, usually with a length of 1, with the value of the selected items.
```
jcheng5: If nothing is selected, what happens? (when `multiple=TRUE`, or `Choose an option`=""`)

trestletech:
Migrating to #2600 since we can't push to this branch.

PR #2585, opened 2019-09-05
Add a test runner

trestletech:
This adds a `runTests()` test runner to Shiny that behaves much like `R CMD check` -- it invokes all the top-level R files in the `tests/` directory next to your application. This can be used to write different types of tests (e.g. testthat unit tests, shinytest functional tests). To do so, you'd place all your specific tests in nested directories inside of `tests/` and in the top-level of the `tests/` directory, just have one the "test runners" that invoke these tests. 

By mimicking `R CMD check`, we remain compatible with golem in which a Shiny app is literally an R package.

## Open questions

- should we create a nice `print` method for these test output classes in this PR or do it later?
- What should be captured for each file? The initial description suggested a pass/fail boolean, but I now have the error or NA.

## Validation notes

There are a few scenarios that we can validate here (note that most, if not all of these, are also covered in automated tests in this PR).

### Happy Path

Check out an app like https://github.com/rstudio/shiny-examples/compare/jeff/test?expand=1 that has multiple types of tests. Run `shiny::testApp()` against that application. If all tests pass, you should see a result like:

```
$result
[1] TRUE

$files
$files$testthat.R
[1] NA

$files$shinytest.R
[1] NA
```

### Broken Tests

If one of the tests is failing, you'll see those `TRUE`s above switch to `FALSE`. e.g. make a change to your Shiny app that you know will break a shinytest, or add a line like `expect_equal(2,3)` to your testthat unit tests to make them fail. You should see `result` flip to `FALSE` and the specific file that failed will also show `FALSE` in the `files` list.

### No tests

If an app has no `tests/` directory, you'll see an error like:

```
> shiny::testApp("../001-hello/")
Error in shiny::testApp("../001-hello/") : 
  No tests directory found: ../001-hello//tests
```

If it has a directory but it's empty or has no R files in it, you should see a message that no test runners were found, and the result will be `NA`:

```
> shiny::testApp("../001-hello/")
No test runners found in ../001-hello//tests
$result
[1] NA

$files
list()

attr(,"class")
[1] "shinytestrun"
```


R/test.R
```diff
@@ -0,0 +1,107 @@
...
+
+#' Runs the tests associated with this Shiny app
+#'
+#' Sources the `.R` files in the top-level of `tests/` much like `R CMD check`.
+#' These files are typically simple runners for tests nested in other
+#' directories under `tests/`.
+#'
+#' @param appDir The base directory for the application.
+#' @param filter If not `NULL`, only tests with file names matching this regular
+#'   expression will be executed passed to `grepl`. Matching is performed on
```
wch: I think "passed to `grepl`" should be removed?

R/test.R
```diff
@@ -0,0 +1,107 @@
...
+  if (all(isST)){
+    # just call out to shinytest
+    # We don't need to message/warn here since shinytest already does it.
+    if (!requireNamespace("shinytest", quietly=TRUE) ){
+      stop("It appears that the .R files in ", testsDir,
+           " are all shinytests, but shinytest is not installed.")
+    }
+
+    if (getOption("shiny.autoload.r", FALSE)) {
+      warning("You've enabled `shiny.autoload.r` but this is not supported yet with shinytest.")
```
wch: This message occurs if the user has explicitly set `options(shiny.autoload.r=T)` in this R session, but is using Shiny 1.4.0, where it defaults to `F`, right? I think that for the upcoming version of Shiny, where it defaults to `T`, then this message shouldn't come up.

If that's right, then this message should be tweaked a bit to explain what's going on. We could also make shinytest pass along this R session's value for `shiny.autoload.r` by default, but it would be simpler to just get users to upgrade their Shiny version.



This message occurs if the user has explicitly set `options(shiny.autoload.r=T)` in this R session, but is using Shiny 1.4.0, where it defaults to `F`, right? I think that for the upcoming version of Shiny, where it defaults to `T`, then this message shouldn't come up.

If that's right, then this message should be tweaked a bit to explain what's going on. We could also make shinytest pass along this R session's value for `shiny.autoload.r` by default, but it would be simpler to just get users to upgrade their Shiny version.

R/test.R
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    sares <- shinytest::testApp(appDir)
+    res <- list()
+    lapply(sares$results, function(r){
+      e <- NA_character_
+      if (!r$pass){
```
wch: I'm not sure if I'm interpreting this logic right, but if a shinytest script fails for this app, does that mean that `"Unknown shinytest error"` will always be the recorded result?

R/test.R
```diff
@@ -0,0 +1,107 @@
...
+        e <- simpleError("Unknown shinytest error")
+      }
+      res[[r$name]] <<- e
+    })
+    return(structure(list(result=all(is.na(res)), files=res), class="shinytestrun"))
+  }
+
+  testenv <- new.env(parent=globalenv())
+  renv <- new.env(parent=testenv)
+  if (getOption("shiny.autoload.r", FALSE)) {
```
wch: I think this should default to `TRUE` now.

DESCRIPTION
```diff
@@ -90,9 +90,13 @@ Suggests:
     ggplot2,
     reactlog (>= 1.0.0),
     magrittr,
+    shinytest,
     yaml,
     future,
     dygraphs
+Remotes:
```
wch: These Remotes aren't needed, right?

R/test.R
trestletech: Correct. I didn't see any more detailed information exposed back from shinytest so I just left it at that.

DESCRIPTION
trestletech: I have no idea where these are coming from. When I git blame it's leading me to https://github.com/rstudio/shiny/commit/3b04c642aeaeb3da4159f823983f07fb173c78b1 ?

 I guess I'll just delete them.

trestletech:
FYI all appveyor builds have been failing to load the curl package, so those are both borked. And Travis is failing on the roxygen one because we switched to build from GitHub master and they made some changes that alter formatting so I can either regenerate and pull those in (which would add a lot of noise to the diff) or leave it for now.

We might be better off pinning to a specific version of Roxygen on GitHub so we don't have to go through this as they iterate. Obviously, the long term is that we'd end up back on the CRAN version but we have to wait until they release the R6 stuff to CRAN.

Issue #2584, opened 2019-09-05
Too-chatty addResourcePath messages


jcheng5:
On Windows 10, R 3.6.1, shiny from master.

When running 063-superzip I get:

```
The resource path 'crosstalk-1.0.0' used to point to C:/Users/jcheng/Documents/R/win-library/3.6/crosstalk/www, but it now points to C:\Users\jcheng\Documents\R\win-library\3.6\crosstalk\www. If your app doesn't work as expected, you may want to choose a different prefix name.
The resource path 'dt-core-1.10.16' used to point to C:/Users/jcheng/Documents/R/win-library/3.6/DT/htmlwidgets/lib/datatables, but it now points to C:\Users\jcheng\Documents\R\win-library\3.6\DT\htmlwidgets\lib\datatables. If your app doesn't work as expected, you may want to choose a different prefix name.
The resource path 'jquery-1.11.3' used to point to C:/Users/jcheng/Documents/R/win-library/3.6/crosstalk/lib/jquery, but it now points to C:\Users\jcheng\Documents\R\win-library\3.6\crosstalk\lib\jquery. If your app doesn't work as expected, you may want to choose a different prefix name.
The resource path 'crosstalk-1.0.0' used to point to C:/Users/jcheng/Documents/R/win-library/3.6/crosstalk/www, but it now points to C:\Users\jcheng\Documents\R\win-library\3.6\crosstalk\www. If your app doesn't work as expected, you may want to choose a different prefix name.
The resource path 'dt-core-1.10.16' used to point to C:/Users/jcheng/Documents/R/win-library/3.6/DT/htmlwidgets/lib/datatables, but it now points to C:\Users\jcheng\Documents\R\win-library\3.6\DT\htmlwidgets\lib\datatables. If your app doesn't work as expected, you may want to choose a different prefix name.
The resource path 'jquery-1.11.3' used to point to C:/Users/jcheng/Documents/R/win-library/3.6/crosstalk/lib/jquery, but it now points to C:\Users\jcheng\Documents\R\win-library\3.6\crosstalk\lib\jquery. If your app doesn't work as expected, you may want to choose a different prefix name.
The resource path 'crosstalk-1.0.0' used to point to C:/Users/jcheng/Documents/R/win-library/3.6/crosstalk/www, but it now points to C:\Users\jcheng\Documents\R\win-library\3.6\crosstalk\www. If your app doesn't work as expected, you may want to choose a different prefix name.
[...]

PR #2586, opened 2019-09-05
Suppress resource path changes by default

cpsievert:
See #2584 (introduced by https://github.com/rstudio/shiny/pull/2459)

### DIFF
--- a/R/server.R
+++ b/R/server.R
@@ -85,25 +85,25 @@ addResourcePath <- function(prefix, directoryPath) {
     }
   )
 
-  # Often times overwriting a resource path is "what you want",
-  # but sometimes it can lead to difficult to diagnose issues
-  # (e.g. an implict dependency might set a resource path that
-  # conflicts with what you, the app author, are trying to register)
-  # Note that previous versions of shiny used to warn about this case,
-  # but it was eventually removed since it caused confusion (#567).
-  # It seems a good comprimise is to throw a more information message.
-  if (getOption("shiny.resourcePathChanges", FALSE) &&
-      prefix %in% names(.globals$resourcePaths)) {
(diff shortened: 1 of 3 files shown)


R/server.R
```diff
@@ -92,7 +92,8 @@ addResourcePath <- function(prefix, directoryPath) {
   # Note that previous versions of shiny used to warn about this case,
   # but it was eventually removed since it caused confusion (#567).
   # It seems a good comprimise is to throw a more information message.
-  if (prefix %in% names(.globals$resourcePaths)) {
+  if (getOption("shiny.resourcePathChanges", FALSE) &&
```
wch: For options we generally separate words with `.`

wch: Although now that I look at it, we're pretty inconsistent with it :/

wch:
This option seems so specific that it seems unlikely that someone will realize to turn it if they have problems. Maybe we should have a more general option that enables various diagnostic/debugging messages. But we could defer that discussion until the after the upcoming release.

In the short term we do need to change the default behavior so that it's quieter.

R/server.R
cpsievert: Yea, I went with camel case since this is about `addResourcePath()`

cpsievert:
Ok, so maybe I'll comment this whole thing out for the time being?

https://github.com/rstudio/shiny/blob/1c8071a96fbc026fd8c6738b64cc4d6e6a7905ed/R/server.R#L95-L106

jcheng5:
I wonder if it'd make sense to have `createWebDependency` use a different code path? Perhaps even using a reserved URL prefix. We're unlikely to care about `createWebDependency` collisions, I think? It just means a new/different `jquery-1.1` or whatever is being provided.

PR #2588, opened 2019-09-05
Add try-catch to ion.rangeSlider

### LINKED ISSUE #2587: Error when updating slider value in PhantomJS
When testing an app with shinytest (which uses) PhantomJS, the current master version of Shiny throws an error when setting the value of the slider. This started happening after we updated to jQuery 3 (#2557).

The error happens when ion.rangeSlider calls `.removeProp("tabindex")` here:
https://github.com/rstudio/shiny/blob/ff5377da9e9c4fe904081b44a6d29e5703b43fe7/inst/www/shared/ionrangeslider/js/ion.rangeSlider.js#L1726

According to https://api.jquery.com/removeProp/ :
> With some built-in properties of a DOM element or window object, browsers may generate an error if an attempt is made to remove the property.

The `tabIndex` property is built-in, and most browsers apparently do not throw an error if you try to remove it, but PhantomJS does throw an error.

In the previous version of jQuery that we used (1.12.4), it wrapped the removal in a try-catch:
 https://github.com/jquery/jquery/blob/1.12.4/src/attributes/prop.js#L16-L26

However, in the current version that we're using (3.4.1), the try-catch is gone:
https://github.com/jquery/jquery/blob/3.4.1/src/attributes/prop.js#L18-L22

I think the solution is to modify our copy of ion.rangeSlider so that it wraps the `removeProp()` with a try-catch.

wch:
Closes #2587.


trestletech:
Just tested locally and what was broken on master works on this PR.

tools/updateIonRangeSlider.R
```diff
@@ -1,35 +1,57 @@
...
+# Then applies patches from tools/ion.rangeSlider-patches.
 
 # The version of ion.rangeSlider that we use is modified from the base version
 # in two ways:
 # * In our version, mouse events on the slider are not propagated to lower
 #   layers (#711, #1630).
 # * We include a custom skin for Shiny.
+# * When .update() is called, the `.removeProp("tabindex")` is wrapped in
+#   try-catch. This is because this line results in an error with jQuery 3 and
+#   PhantomJS. (#2587)
```
cpsievert: Minor nitpick: I think I'd prefer if these bullets appeared where the patches are applied, and have them reference the patch filenames

tools/updateIonRangeSlider.R
```diff
@@ -1,35 +1,57 @@
...
-    # 'source'd via R console
-    return(normalizePath(sys.frames()[[1]]$ofile))
-  }
-}
+# =============================================================================
+# Download library
+# =============================================================================
+# TODO: In the future, use a tag instead of a bare commit hash. See
+# updateBootstrapDatepicker.R for an example.
+version   <- "f0e1116b065c3d091a9c5f04477aaf8894974e2a"
```
cpsievert: Minor nitpick: could maybe add some historical context as to why we're using that specific commit?

tools/updateIonRangeSlider.R
```diff
@@ -43,6 +35,15 @@ file.copy(file.path(unzip_dir, "img"), dest_dir, recursive = TRUE)
 # =============================================================================
 # Apply patches
 # =============================================================================
+# The version of ion.rangeSlider that we use is modified from the base version
+# in two ways:
```
cpsievert: two -> three

wch: Oops, I'll fix it.

PR #2593, opened 2019-09-10
Showcase filenames

### LINKED ISSUE #1926: Showcase mode won't display files with spaces in the name
Hi there,

Found a little bug with shiny showcase mode - the code tabpanel won't display R scripts with spaces in the name. I've published a minimal example here: https://cwthom94.shinyapps.io/showcase-mode-issue/

Repo with code here: https://github.com/cwthom/shiny-showcase-issue

Possibly this isn't a bug but I couldn't find it documented anywhere so here you go :)

Huge thanks for shiny!
Chris

cpsievert:
(Best to review commit by commit)

The first commit fixes #1926 

The second commit fixes what seems to be an undocumented problem when running an app file that is named something other than `app.R` in showcase mode (the app file should be 'active', that is, shown by default).

Will link to a testing app shortly.


R/showcase.R
```diff
@@ -105,12 +105,11 @@ navTabsDropdown <- function(files) {
...
     with(tags,
-      div(class=paste("tab-pane",
+      div(class=paste0("tab-pane",
+                      # TODO: what if the app filename is something else?
                       if (tolower(file) %in% c("app.r", "server.r")) " active"
-                      else "",
-                      sep=""),
-          id=paste(gsub(".", "_", file, fixed=TRUE),
-                   "_code", sep=""),
+                      else ""),
```
trestletech: Nit: I found this logic a little hard to follow. It may be idiomatic and I haven't seen it yet, but I would have found an `ifelse` more readable (or just pulling it out into a separate variable elsewhere)

R/showcase.R
```diff
@@ -81,7 +81,7 @@ appMetadata <- function(desc) {
 navTabsHelper <- function(files, prefix = "") {
   lapply(files, function(file) {
     with(tags,
-      li(class=if (tolower(file) %in% c("app.r", "server.r")) "active" else "",
+      li(class=if (tolower(file) %in% c("app.r", "server.r", tolower(.globals$appFile))) "active" else "",
```
trestletech: Is there a reason to still have these hardcoded `app.R` and `server.r` included in the list? Would those be covered by `.globals$appFile`?

trestletech: Also does this logic permit multiple files to have class `active`? i.e. if you run an app with `app.R` and `server.R`, I believe `runApp` will give precedence to `server.R`, but this logic would highlight them both as active, right?

May not be a bug worth fixing in this PR, but if that's true we should at least track it IMHO.

cpsievert: All good points. At the moment, `.globals$appFile` will only be populated if `shinyAppFile()` is used, but we could also populate it in `shinyAppDir()` to avoid hard-coding and guessing wrong.

R/showcase.R
cpsievert: I tend to avoid `ifelse()` for coercion reasons. That being said, I don't like the nested `if` `else` logic here, but the rest of `showcase.R` is littered with that pattern so I feel like that could be for another PR

R/showcase.R
cpsievert: Done in 4b2e6fc

R/showcase.R
```diff
@@ -81,7 +81,7 @@ appMetadata <- function(desc) {
 navTabsHelper <- function(files, prefix = "") {
   lapply(files, function(file) {
     with(tags,
-      li(class=if (tolower(file) %in% c("app.r", "server.r", tolower(.globals$appFile))) "active" else "",
+      li(class=if (file %in% .globals$appFile) "active" else "",
```
trestletech: Nit: If `$appFile` is just length-one would it be more clear to just use `identical` or some such instead of `%in%`?

R/showcase.R
trestletech: Fair point! Let's leave it as-is, then.

R/app.R
```diff
@@ -121,8 +121,10 @@ shinyAppDir <- function(appDir, options=list()) {
   appDir <- normalizePath(appDir, mustWork = TRUE)
 
   if (file.exists.ci(appDir, "server.R")) {
+    .globals$appFile <- "server.R"
```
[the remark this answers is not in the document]
cpsievert: Would `shinyOptions()` be a decent way to ensure the filename is local to each app? Looks like we already do that to set/get the app directory here https://github.com/rstudio/shiny/blob/4b2e6fc38604f85760290aaf4fb3417d1f966dee/R/app.R#L95

PR #2595, opened 2019-09-10
soft-deprecate bootstrapPage() and basicPage()

### LINKED ISSUE #2594: Hide bootstrapPage docs
In a similar spirit to #2452, we should probably do more to make `bootstrapPage()` and `basicPage()` less discoverable and more obvious that they're soft-deprecated.

cpsievert:
Fixes #2594


hadley:
@jcheng5 I think it's low risk to add soft deprecation, even at the end of a release cycle, because it doesn't actually change behaviour (just documentation).

PR #2600, opened 2019-09-11
Define what all input functions yield for server

trestletech:
Enhancement to address #2461

This is a continuation of #2551 (but I don't have access to push to that branch, so... here we are!)

### DIFF
--- a/R/input-action.R
+++ b/R/input-action.R
@@ -39,7 +39,10 @@
 #' @seealso [observeEvent()] and [eventReactive()]
 #'
 #' @section Server value:
-#' An integer, initialized as 0 and incremented with each press.
+#' An integer, initialized to `0` and incremented with each press. This `0` is special
+#' in that it's considered "falsy", meaning that it will cause [req()] to fail. This
+#' special behavior is only manifested with `actionButton`; traditional [numericInput()]s
+#' will not cause `req` to fail when --- even when set to `0`.
 #'
 #' @export
 actionButton <- function(inputId, label, icon = NULL, width = NULL, ...) {
--- a/R/input-action.R
+++ b/R/input-action.R
@@ -37,6 +37,10 @@
 #' }
 #'
 #' @seealso [observeEvent()] and [eventReactive()]
+#'
+#' @section Server value:
+#' Returns an integer, initialized as 0 and incremented with each press.
+#'
 #' @export
 actionButton <- function(inputId, label, icon = NULL, width = NULL, ...) {
 
--- a/R/input-action.R
+++ b/R/input-action.R
@@ -39,7 +39,7 @@
 #' @seealso [observeEvent()] and [eventReactive()]
 #'
 #' @section Server value:
-#' Returns an integer, initialized as 0 and incremented with each press.
+#' An integer, initialized as 0 and incremented with each press.
 #'
 #' @export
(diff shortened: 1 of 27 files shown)


R/input-slider.R
```diff
@@ -70,6 +70,10 @@
 #' # Complete app with UI and server components
 #' shinyApp(ui, server)
 #' }
+#'
+#' @section Server value:
+#' A number, or a vector of two numbers in the case of slider range.
```
hadley:
```suggestion
#' A number, or in the case of slider range, a vector of two numbers.
```

R/input-date.R
```diff
@@ -86,6 +86,10 @@
 #'
 #' shinyApp(ui, server = function(input, output) { })
 #' }
+#'
+#' @section Server value:
+#' A date-time of class `Date`.
```
cpsievert:
```suggestion
#' A [Date] vector of length 1.
```

R/input-daterange.R
```diff
@@ -70,6 +70,10 @@
 #'
 #' shinyApp(ui, server = function(input, output) { })
 #' }
+#'
+#' @section Server value:
+#' A vector of two [Date] objects.
```
cpsievert:
```suggestion
#' A [Date] vector of length 2.
```

R/input-numeric.R
```diff
@@ -24,6 +24,10 @@
 #' }
 #' shinyApp(ui, server)
 #' }
+#'
+#' @section Server value:
+#' The input value as a `double`.
```
cpsievert:
```suggestion
#' A [numeric] vector of length 1.
```


It's possible for the input value to be an integer

```r
library(shiny)
ui <- fluidPage(numericInput("x", "Hey", 3L))
  observe(print(str(input$x)))
}
shinyApp(ui, server)
```

```r
Listening on http://127.0.0.1:4368
 int 3
```

cpsievert: Actually, I guess it's more correct to say `A length 1 vector with a [mode] of `"numeric"``

R/input-action.R
```diff
@@ -37,6 +37,13 @@
 #' }
 #'
 #' @seealso [observeEvent()] and [eventReactive()]
+#'
+#' @section Server value:
+#' An integer, initialized to `0` and incremented with each press. This `0` is special
+#' in that it's considered "falsy", meaning that it will cause [req()] to fail. This
+#' special behavior is only manifested with `actionButton`; traditional [numericInput()]s
+#' will not cause `req` to fail when --- even when set to `0`.
+#'
```
cpsievert: I think it's important to mention the connection to event handlers here...I'd do something like this:

```
#' An integer of class `"shinyActionButtonValue"`. This class differs from 
#' ordinary integers in that a value of 0 is considered "falsy". This implies two things:
#' * Event handlers (e.g., [observeEvent()], [eventReactive()]) won't execute on initial load.  
#' * Input validation (e.g., [req()], [need()]) will fail on initial load.
```

R/input-numeric.R
hadley: I don't think it's a good idea to talk about the "mode" of a vector since "mode" is mostly an S concept. I think a numeric vector of length 1 is fine (although I wouldn't link to `numeric()` since that's an alias for `double()` which isn't the meaning you want to imply here)

PR #2602, opened 2019-09-11
Bump later version requirement

wch:
This should be merged after https://github.com/r-lib/later/pull/104 is merged. It is not strictly required for shiny to work properly, but it will help users get the updated version of later.


cpsievert:
We can remove httpuv from remotes now, right?

PR #2619, opened 2019-09-17
Better checking for implicit globals in JS

wch:
Previously, the eslint rules allowed accidentally assigning to properties of `window` (like `window.name`) with the following:

```js
name = 123;
```

With this change, it will treat that assignment as an error. If you really want to assign to `window.name`, you will have to do:

```js
window.name = 123;
```


wch:
In the future we should add a check that the JS version number matches the DESCRIPTION. cc @schloerke

PR #2621, opened 2019-09-17
Use pageX/pageY instead of offsetX/offsetY

### LINKED ISSUE #2620: plotOutput()'s double-click input value isn't set in Firefox
To reproduce: Try double-clicking in Firefox with 093-plot-interaction-basic. The input value tied to double-click events is not set.

This issue is a regression from shiny CRAN (v.1.3.2) and the problem was introduced by upgrading jQuery in https://github.com/rstudio/shiny/commit/9a1f7cba6877230dbeb758f32c490f986103cea6. It looks as though the `pending_e.offsetX/Y` values here no longer contain the correct information

https://github.com/rstudio/shiny/blob/02f7a4fdc9976b27e4a522c88c9d483421c46fec/srcjs/output_binding_image.js#L706-L711

Fortunately, I think we can fix the problem by using the more official `pageX`/`pageY` instead to determine if the second click is "too far away"

cpsievert:
Closes #2620


jcheng5:
Are the mousedown2 and dblclick2 events private to Shiny? Or are they designed for users to access? @wch

PR #2625, opened 2019-09-19
Fix updateSelectizeInput(), fixes #2624

### LINKED ISSUE #2624: Regression in selectizeInput handling of nested lists
Nested lists not handled correctly anymore. Minimal repro is here, both select inputs should have the same choices:

```
library(shiny)

choices <- list(a = 1, B = list(B = 2), c = list(3))

ui <- bootstrapPage(
  selectizeInput("x1", "x", choices = choices),
  selectizeInput("x2", "x", choices = NULL)
)

  updateSelectizeInput(session, "x2", choices = choices, server = TRUE)
}

shinyApp(ui, server)
```

Until #2540, these both looked like the second version.

alandipert:
This PR fixes #2624 by aligning the choice processing behavior of `selectizeInput()`/`selectInput()` and `updateSelectizeInput()`/`updateSelectInput()`.

* In previous releases of Shiny, named `choices` of length 1 that were unnamed lists were considered groups, not leaves. `list(x=list(123))` is an example. My changes in #2540 altered that behavior so unnamed lists of length 1 were considered leaves, not groups. The bug was revealed during testing of `updateSelectizeInput()`, because its choice processing implementation considered them groups, just like the original `choicesWithNames()` code did before I rewrote it.
  * To fix this, I changed the definition of `isGroup()` to include lists and added a test.
* I noticed that empty lists and vectors were considered leaves and not groups. I modified `isGroup()` to also include them and added tests.
* I noticed that lists containing unnamed groups would result in a different error than the previous code produced. I added an `asNamed()` in `processGroupedChoices()` to ensure `mapply()` would work; the function inside is what produces the correct error. I added tests for this.
* I noticed that a few special cases of `NA` and `NULL` were untested. I added tests.

## Notes

Note that a previous iteration of this PR included [a replacement](https://gist.github.com/alandipert/7da41616548def947fdca7feca6424a5) for `updateSelectizeInput()`'s choice processing code that built on #2540 instead of implementing choice processing separately.

Unfortunately, that code necessitated 2 walks of the `choices` tree, and so was twice as slow. We are highly sensitive to performance in this case as users select `server = TRUE` specifically for performance reasons as `choices` is large.

In order to compete with the existing `updateSelectizeInput()` code, `choicesWithNamesRectangle()` would have needed to be rewritten as a single pass of `choices`, which we deemed too significant an undertaking given the urgency of a 1.4.0 release.

So, instead of doing that, we reduced the scope of this PR to a small change of the new code and the addition of tests that bring it to feature parity with the code inside `updateSelectizeInput()`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -26,7 +26,7 @@ shiny 1.4.0
 
 * Added `resourcePaths()` and `removeResourcePaths()` functions. ([#2459](https://github.com/rstudio/shiny/pull/2459))
 
-* Resolved [#2515](https://github.com/rstudio/shiny/issues/2515): `selectInput()` now deals appropriately with named factors. ([#2524](https://github.com/rstudio/shiny/pull/2524))
+* Resolved [#2515](https://github.com/rstudio/shiny/issues/2515): `selectInput()` and `selectizeInput()` now deal appropriately with named factors. Note that `updateSelectInput()` and `updateSelectizeInput()` **do not** yet handle factors; their behavior is unchanged. ([#2524](https://github.com/rstudio/shiny/pull/2524), [#2540](https://github.com/rstudio/shiny/pull/2540), [#2625](https://github.com/rstudio/shiny/pull/2625))
 
 * Resolved [#2433](https://github.com/rstudio/shiny/issues/2433): An informative warning is now thrown if subdirectories of the app's `www/` directory are masked by other resource prefixes and/or the same resource prefix is mapped to different local file paths. ([#2434](https://github.com/rstudio/shiny/pull/2434))
 
--- a/R/input-utils.R
+++ b/R/input-utils.R
@@ -92,7 +92,10 @@ generateOptions <- function(inputId, selected, inline, type = 'checkbox',
 
 # True when a choice list item represents a group of related inputs.
 isGroup <- function(choice) {
-  length(choice) > 1 || !is.null(names(choice))
+  is.list(choice) ||
+    !is.null(names(choice)) ||
+    length(choice) > 1 ||
+    length(choice) == 0
 }
 
 # True when choices is a list and contains at least one group of related inputs.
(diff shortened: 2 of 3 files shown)


NEWS.md
```diff
@@ -26,7 +26,7 @@ shiny 1.3.2.9001
 
 * Added `resourcePaths()` and `removeResourcePaths()` functions. ([#2459](https://github.com/rstudio/shiny/pull/2459))
 
-* Resolved [#2515](https://github.com/rstudio/shiny/issues/2515): `selectInput()` now deals appropriately with named factors. ([#2524](https://github.com/rstudio/shiny/pull/2524))
+* Resolved [#2515](https://github.com/rstudio/shiny/issues/2515): `selectInput()` and `selectizeInput()` now deal appropriately with named factors. Note that `updateSelectInput()` and `updateSelectizeInput()` **do not** yet handle factors; their behavior is unchanged. ([#2525](https://github.com/rstudio/shiny/pull/2524))
```
wch: The URL should point to 2525.

wch:
This PR should be targeted toward the rc-v1.40 branch instead of master.

NEWS.md
```diff
@@ -26,7 +26,7 @@ shiny 1.4.0
 
 * Added `resourcePaths()` and `removeResourcePaths()` functions. ([#2459](https://github.com/rstudio/shiny/pull/2459))
 
-* Resolved [#2515](https://github.com/rstudio/shiny/issues/2515): `selectInput()` now deals appropriately with named factors. ([#2524](https://github.com/rstudio/shiny/pull/2524))
+* Resolved [#2515](https://github.com/rstudio/shiny/issues/2515): `selectInput()` and `selectizeInput()` now deal appropriately with named factors. Note that `updateSelectInput()` and `updateSelectizeInput()` **do not** yet handle factors; their behavior is unchanged. ([#2525](https://github.com/rstudio/shiny/pull/2525))
```
wch: Oops, this should reference 2524 and 2625.

Issue #2630, opened 2019-09-19
In some cases user-code may not have access to initShiny methods


cpsievert:
As of jQuery 3.0, [document-ready handlers are now asynchronous](https://jquery.com/upgrade-guide/3.0/#breaking-change-document-ready-handlers-are-now-asynchronous), which means it's possible for the methods in `initShiny` to not be available to downstream user-code because of this:

https://github.com/rstudio/shiny/blob/02f7a4fdc9976b27e4a522c88c9d483421c46fec/srcjs/init_shiny.js#L511-L515

For a minimal example of where this becomes of problem:

```r
library(shiny)
library(leaflet)
shinyApp(
  ui = leaflet() %>% addTiles(),
  server = function(input, output, session) {}
)
```

Before jQuery 3.0, `initShiny` would be called before leaflet's javascript executes; however, after jQuery 3.0, `initShiny` is called after leaflet's javascript, so not all of Shiny's methods (e.g., `onInputChange`) is not available to leaflet

cpsievert:
And also, to help clarify when this becomes an issue, I believe this problem surfaces whenever you have a statically rendered object in the UI that's expecting to be able to use the `initShiny` methods. Here's a plotly minimal example as well:

```r
library(shiny)
shinyApp(
  ui = plotly::plot_ly(x = 1, y = 1),
  server = function(input, output, session) {}
)
```

jcheng5:
After investigation with @cpsievert @schloerke @alandipert, the crux of the issue is that `$(fn)` used to invoke `fn` synchronously from a `DOMContentLoaded` handler, but now, a `setTimeout` is inserted between `DOMContentLoaded` and `fn`.

In other words, jQuery 1.x was like:

```javascript
function $(callback) {
  document.addEventListener("DOMContentLoaded", callback);
}
```

but jQuery 3.x is like:

```javascript
function $(callback) {
  document.addEventListener("DOMContentLoaded", () => {
    setTimeout(callback);
  }
}
```

Meanwhile, Shiny (both 1.3.2 and 1.4.0) has code like this:

```javascript
$(function() {
  setTimeout(init_shiny, 1);
});
```

While htmlwidgets+leaflet has code that is (effectively) like this:

```javascript
document.addEventListener("DOMContentLoaded", function() {
  setTimeout(updateBounds, 1);
});
```

where it's important for `init_shiny` to be called before `updateBounds`.

Three options we discussed for fixing this:

[...]

PR #2632, opened 2019-09-20
Opt-in to jQuery 1.12.1

cpsievert:
This allows users to opt-in to jQuery 1.12.1 by doing

```r
options(shiny.jquery.version = 1)
```

It will also message users that have old version of htmlwidgets installed to upgrade to version 1.5 which will have a shim for the breaking behavior described in https://github.com/rstudio/shiny/issues/2630

https://github.com/ramnathv/htmlwidgets/pull/345


R/shinyui.R
```diff
@@ -42,9 +42,28 @@ renderPage <- function(ui, connection, showcase=0, testMode=FALSE) {
...
+    if (version == 3) {
+      return(htmlDependency(
+        "jquery", "3.4.1",
+        c(href = "shared"),
+        script = "jquery.min.js"
+      ))
+    }
+    if (version == 1) {
+      return(htmlDependency(
+        "jquery", "1.12.1",
```
wch: Isn't the latest 1.x version 1.12.4?

NEWS.md
```diff
@@ -1,6 +1,10 @@
 shiny 1.4.0
 ===========
 
+## Library updates
+
+* Resolved [#2554](https://github.com/rstudio/shiny/issues/2554): Upgraded jQuery from v.1.12.1 to v3.4.1 and bootstrap from v3.3.7 to v3.4.1. ([#2557](https://github.com/rstudio/shiny/pull/2557)). Since the jQuery upgrade may introduce breaking changes to user code, there is an option to switch back to the old version by setting `options(shiny.jquery.version = 1)`. If you're using an `htmlTemplate()`  (or similar) that references the `shared/jquery[.min].js` file that ships with the **shiny** package, in order to downgrade, you'll have to change that filepath to `shared/legacy/jquery[.min].js`.
```
wch: Should be 1.12.4 instead of 1.12.1.

We don't need to mention the `htmlTemplate()` function here -- typically, it's used with `headContent()`, which allows shiny to automatically inject dependencies like jQuery and shiny.js. It's only when there's an actual index.html that has ` <script src="shared/jquery.min.js"></script>` that the path would need to be changed.

R/globals.R
```diff
@@ -25,6 +25,28 @@ register_s3_method <- function(pkg, generic, class, fun = NULL) {
...
+register_upgrade_message <- function(pkg, version) {
+
+  msg <- sprintf(
+    "This version of Shiny is designed to work with '%s' >= %s.
+    Please upgrade via install.packages('%s').",
+    pkg, version, pkg
+  )
+
+  # emit startup message of the package is already loaded
+  if (pkg %in% loadedNamespaces()) {
```
jcheng5: Don't you need to check the version number?

R/shinyui.R
```diff
@@ -42,9 +42,28 @@ renderPage <- function(ui, connection, showcase=0, testMode=FALSE) {
...
+        c(href = "shared/legacy"),
+        script = "jquery.min.js"
+      ))
+    }
+    stop("Unsupported version of jQuery: ", version)
+  }
+
   shiny_deps <- list(
     htmlDependency("json2", "2014.02.04", c(href="shared"), script = "json2-min.js"),
```
jcheng5: Maybe we can get rid of this in Shiny v1.5? What even is this, a JSON polyfill?

R/globals.R
cpsievert: whoops, yep

NEWS.md
cpsievert: whoops, yep, must have got it from the jquery UI version

R/shinyui.R
cpsievert: Yep, will fix

R/shinyui.R
cpsievert: Good question, I wondered the same thing

R/globals.R
```diff
@@ -25,6 +25,32 @@ register_s3_method <- function(pkg, generic, class, fun = NULL) {
...
+    packageStartupMessage(msg)
+  }
+
+  # Always register hook in case package is later unloaded & reloaded
```
wch: I think this comment needs a little more context so it doesn't throw us off in the future. The common case is that the package is loaded later (and not unloaded and reloaded). The fact that we're doing it unconditionally means that it will emit the message in case the package is unloaded and reloaded, but seems incidental, because it's much less common, and the user would have already seen the message.

R/globals.R
```diff
@@ -25,6 +25,32 @@ register_s3_method <- function(pkg, generic, class, fun = NULL) {
   )
 }
 
+register_upgrade_message <- function(pkg, version) {
+  # Do nothing if the package isn't installed or
+  # meets the version requirement
+  if (system.file(package = pkg) == "") return(NULL)
+  if (packageVersion(pkg) >= as.character(version)) return(NULL)
```
wch: Instead of `as.character`, `package_version` would be more correct. But `Ops.numeric_version` will coerce the right had side to a numeric version object anyway, so I don't think there's a need to have either as.character` or `package_version`.

R/shinyui.R
wch: I think it's for IE8, so we can get rid of it.

R/globals.R
cpsievert: That makes sense. I copied that comment from `register_s3_method()`...do you want that updated as well?

R/globals.R
cpsievert: Oh, interesting, thanks!

jcheng5: Sorry, I don't believe this is the right place to check; don't we want to check right before each of the two `packageStartupMessage(msg)` lines?

R/globals.R
wch: Yes, that would be good, thanks. `git blame` says that one is my fault :)

R/globals.R
jcheng5: Much better, thanks. I'm running some spot tests now, if it all looks good I'll approve

R/shinyui.R
cpsievert: I filed an issue here https://github.com/rstudio/shiny/issues/2633

NEWS.md
```diff
@@ -1,6 +1,10 @@
 shiny 1.4.0
 ===========
 
+## Library updates
```
jcheng5:
```suggestion
## Breaking changes
```

jcheng5:
Other than the one suggested change I filed ("Library updates" -> "Breaking changes"), LGTM! Thanks for jumping on this, @cpsievert!

PR #2634, opened 2019-09-23
Add reactive values from list

### LINKED ISSUE #2511: Suggestion: Add a "reactiveValuesFromList" function that mirrors "reactiveValuesToList"
We have a way to convert a reactiveValues list into a regular list with `reactiveValuesToList()`. It would be make sense IMO to have an opposite function `reactiveValuesFromList()` that takes a regular list and creates a `reactiveValues` list from it.

For testing purposes, I use this often, to save the state of a complex reactiveValues variable and read it back into the app at a later time, and it'd be easier if such a function existed. I'm not only asking for myself, I feel like having this symmetry in functions makes sense.

wfjvdham:
resolves #2511


wfjvdham:
I signed it and send it to jj@rstudio.com. But here is it again.

wch:
@wfjvdham We are switching to a new, automated system for CLA's, which unfortunately requires everyone to agree to the CLA using the online system, even if they have already sent in a CLA in the past. This new system will make things easier for future contributors; apologies that you'll have to do it again.

PR #2639, opened 2019-09-25
skip call stack unit tests if promises is outdated

cpsievert:


cpsievert:
Actually, this won't be necessary as we already require promises >= 1.1.0 in the DESCRIPTION

cpsievert:
Ahhh, this won't be necessary as shiny 1.4 will require promises 1.1.0

PR #2650, opened 2019-09-30
Run Travis checks on more R versions

wch:


.travis.yml
```diff
@@ -13,10 +13,16 @@ matrix:
       script: ./tools/checkJSCurrent.sh
       node_js:
         - "10"
-    - name: "Old Release Check"
-      r: oldrel
-    - name: "Current Release Check"
```
trestletech: Is it worth keeping this one around in case we forget to manually update? Maybe we could have this instead of 3.6 for now and then when we upgrade to 3.7 we'd just add 3.6 to our list of old instances to test on?

.travis.yml
```diff
@@ -13,10 +13,16 @@ matrix:
...
-      r: release
+    - name: "3.2"
+      r: 3.2
+    - name: "3.3"
```
trestletech: Electrons are cheap these days, but I'm wondering if we need to test all the incremental versions in between. What if we just tested on 3.2, current, and devel? Are the others necessary?

.travis.yml
wch: Sure, that sounds reasonable.

wch:
Looks like it's failing on R 3.2 and 3.3, essentially because on those versions:

```R
identical(function(x) x, function(x) x)
#> FALSE
```

But on R >=3.4:

```R
identical(function(x) x, function(x) x)
#> TRUE
```

I'll change the test to use `expect_equal` instead of `expect_identical`.

PR #2651, opened 2019-10-03
Use pkgdown for reference page generation

alandipert:
Shiny's [website](https://shiny.rstudio.com) isn't generated by [pkgdown](https://pkgdown.r-lib.org/), but previously we embedded a partial `staticdocs` (original name for pkgdown) configuration that was used as part of the rstudio/shiny-dev-center build.

Part of the (manual) process for that project's build is generation of the pkgdown function reference HTML, which is incorporated into the enclosing Jekyll-based site.

This PR is the first of a 2-step process to move away from `staticdocs` for generating the function reference page on the website. This step is:

1. Embed a `_pkgdown.yml` that names references in the new style. `_pkgdown.yml` supersedes the way staticdocs was configured via `index.r`.
1. Update the relevant unit test that ensures all exported functions are listed in `_pkgdown.yml` file.

The second step is https://github.com/rstudio/shiny-dev-center/pull/128


tests/testthat/test-pkgdown.R
```diff
@@ -0,0 +1,39 @@
+context("pkgdown")
+
+get_exported <- function() {
+  if ('pkgload' %in% installed.packages() && pkgload::is_dev_package('shiny')) {
```
wch: The `pkgload::` is causing a WARNING in R CMD check. I don't think we should add it as a Suggested package, so we'll need a different way of detecting whether the package was loaded via pkgload/devtools.

tests/testthat/test-pkgdown.R
```diff
@@ -0,0 +1,39 @@
...
+                       "pageWithSidebar", "headerPanel", "shiny.appobj",
+                       "deprecatedReactives")
+
+  ## This test ensures that every documented topic is included in
+  ## staticdocs/index.r, unless explicitly waived by specifying it
+  ## in the known_unindexed variable above.
+  missing <- setdiff(all_topics, c(known_unindexed, indexed_topics))
+  unknown <- setdiff(c(known_unindexed, indexed_topics), all_topics)
+  expect_equal(length(missing), 0,
+    info = paste("Functions missing from _pkgdown.yaml:\n",
```
wch: `s/yaml/yml`

tests/testthat/test-pkgdown.R
```diff
@@ -0,0 +1,39 @@
...
+    # We're testing an installed package, possibly for R CMD check
+    unique(unname(readRDS("../../shiny/help/aliases.rds")))
+  }
+}
+
+get_indexed <- function(f = system.file('_pkgdown.yml', package = 'shiny')) {
+  unlist(lapply(yaml::yaml.load_file(f)$reference, function(x) x$contents))
+}
+
+test_that("All man pages have an entry in _pkgdown.yml", {
```
wch: We should use `skip_on_cran()` for this set of tests.

tests/testthat/test-pkgdown.R
wch: yaml vs. yml here (you fixed the other one)

PR #2659, opened 2019-10-08
Make shiny.autoload.r default to TRUE

wch:
Do not merge yet -- this still needs a NEWS item, which I'll add after Shiny 1.4.0 is on CRAN and the RC branch has been merged.

### DIFF
--- a/R/app.R
+++ b/R/app.R
@@ -147,7 +147,7 @@ shinyAppDir_serverR <- function(appDir, options=list()) {
   # change on disk, or are created, or are removed.
 
   # In an upcoming version of shiny, this option will go away and the new behavior will be used.
-  if (getOption("shiny.autoload.r", FALSE)) {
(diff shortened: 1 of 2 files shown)


R/app.R
```diff
@@ -350,7 +356,7 @@ shinyAppDir_appR <- function(fileName, appDir, options=list())
   fullpath <- file.path.ci(appDir, fileName)
 
   # In an upcoming version of shiny, this option will go away and the new behavior will be used.
```
trestletech: Stale comment.

riccardoporreca:
@wch, is it too late to mention this in the NEWS.md for shiny 1.5.0?

From a comment above it looks that was the intention and if still possible I believe it would indeed deserve an explicit entry (at least it would have helped me a lot figuring out an issue I had after updating shiny :smiley: )

PR #2665, opened 2019-10-15
Best-effort task scheduling through the session

trestletech:
Currently a no-op change, as the session just passes through to the global `scheduleTask ` implementation. But this allows us to mock the method for testing.

## QA Validation

Run apps that have (shiny example numbers in parenthesis)...

 - `invalidateLater` (034)
 - `reactivePoll` (059)
 - `reactiveTimer` (142)

### DIFF
--- a/R/timer.R
+++ b/R/timer.R
@@ -119,25 +119,3 @@ scheduleTask <- function(millis, callback) {
     invisible(timerCallbacks$unschedule(id))
   }
 }
-
-#' Get a scheduler function for scheduling tasks. Give priority to the
-#' session scheduler, but if it doesn't exist, use the global one.
-#' @noRd
-defineScheduler <- function(session){
-  if (!is.null(session) && !is.null(session$scheduleTask)){
-    return(session$scheduleTask)
-  }
-  scheduleTask
-}
-
-
-#' Get the current time a la `Sys.time()`. Prefer to get it via the
-#' `session$now()` function, but if that's not available, just return the
-#' current system time.
-#' @noRd
-getTime <- function(session){
-  if (!is.null(session) && !is.null(session$now)){
-    return(session$now())
-  }
-  Sys.time()
-}
--- a/R/timer.R
+++ b/R/timer.R
@@ -96,3 +96,15 @@ scheduleTask <- function(millis, callback) {
     invisible(timerCallbacks$unschedule(id))
   }
 }
+
+#' Get a scheduler function for scheduling tasks. Give priority to the
+#' session scheduler, but if it doesn't exist, use the global one.
+#' @noRd
+defineScheduler <- function(session){
+  if (!is.null(session)){
+    if (!is.null(session$scheduleTask)){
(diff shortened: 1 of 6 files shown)


R/timer.R
```diff
@@ -1,20 +1,22 @@
 # Return the current time, in milliseconds from epoch, with
 # unspecified time zone.
-now <- function() {
+getNow <- function() {
   as.numeric(Sys.time()) * 1000
 }
 
-TimerCallbacks <- R6Class(
+BaseTimerCallbacks <- R6Class(
   'TimerCallbacks',
```
wch: This name should be changed to `BaseTimerCallbacks`

R/timer.R
```diff
@@ -86,6 +88,40 @@ TimerCallbacks <- R6Class(
...
+  public = list(
+    # Empty constructor defaults to the getNow implementation
+    initialize = function() {
+      super$initialize(self$now)
+    },
+    now = function(){
+      return(private$time)
+    },
+    elapse = function(millis){
+      private$time <<- private$time + millis
```
wch: If you're doing assignment with `private$xyz`, you don't need `<<-`.

R/timer.R
```diff
@@ -86,6 +88,40 @@ TimerCallbacks <- R6Class(
   )
 )
 
+TimerCallbacks <- R6Class(
```
wch: Is this class necessary? It appears to do exactly the same thing as the superclass (except that you can't pass in anything to the `initialize` method).

R/timer.R
```diff
@@ -96,3 +132,29 @@ scheduleTask <- function(millis, callback) {
...
+  scheduleTask
+}
+
+
+#' Get the current time a la `Sys.time()`. Prefer to get it via the
+#' `session$now()` function, but if that's not available, just return the
+#' current system time.
+#' @noRd
+getTime <- function(session){
+  if (!is.null(session)){
```
wch: Minor nit: I think it's slightly clearer to have a single `if` statement with `&&`, since there's no complicated logic inside.

R/timer.R
trestletech: Fair point. I had in my head that I wanted to keep the contract for TimerCallbacks consistent. But I now realize that it's only internal, so who cares. I'll simplify.

R/timer.R
```diff
@@ -96,3 +119,25 @@ scheduleTask <- function(millis, callback) {
     invisible(timerCallbacks$unschedule(id))
   }
 }
+
+#' Get a scheduler function for scheduling tasks. Give priority to the
+#' session scheduler, but if it doesn't exist, use the global one.
+#' @noRd
+defineScheduler <- function(session){
```
wch: Don't need this; can always use `session$scheduleTask` instead.

trestletech: Make this and the now function prefixed with a `.` and add both to the base session. Then get rid of these helpers and always use the version dangling off the session.

Add comments to the method definition in R6, too.

PR #2675, opened 2019-10-18
Introduce a MockShinySession object

trestletech:
Will be used for automated testing.

There are two commits here which touch different things: 

 - 4d2ff80 - introduces a bunch of new code for this MockSession and a bunch of tests that confirm that its behavior matches the contract we have around what a session should be.
 - 616ae99 - is the only risky change I see here to existing code. It flips the time comparison for firing events to be `<=` instead of `<`. I don't see any reason why this wouldn't be safe when I looked at the code, but I did get a little nervous about the possibility that there's some initialization to `0` that would change when we include equality in the comparison. But I didn't see any such code path.

(EDIT as discovered in the third commit here: `Sys.time()` isn't actually all that precise on Windows: "On Windows it increments in clock ticks (usually 1/60 of a second) reported to millisecond accuracy." So it's actually not implausible that this change would alter some behavior -- but we're talking executing a 60th of a second earlier.)

## Work Omitted

 - Make `getCurrentOutputInfo()` work.

> Might not be worth addressing right now (maybe in a separate PR), but we'll probably want to wrap the call to func() in private$withCurrentOutput(). This is so the getCurrentOutputInfo() function will work. That function allows code in the output, or in any reactives that it calls, to introspect and find out information about the output that's being executed. The information is generated by https://github.com/rstudio/shiny/blob/005295f/R/shiny.R#L1298-L1337

## QA Validation

Run apps that have (shiny example numbers in parenthesis)...

 - `invalidateLater` (034)
 - `reactivePoll` (059)
 - `reactiveTimer` (142)


R/mock-session.R
```diff
@@ -0,0 +1,283 @@
...
+
+      # create output
+      out <- .createOutputWriter(self)
+      class(out) <- "shinyoutput"
+      self$output <- out
+
+      # Create a read-only copy of the inputs reactive.
+      self$input <- .createReactiveValues(private$.input, readonly = TRUE)
+    },
+    onFlush = function(fun, once) {
```
wch: Should `once` have a default, like the `onFlush` from `ShinySession`? Same for `onFlushed` below.

R/mock-session.R
```diff
@@ -0,0 +1,283 @@
...
+
+      fileData <- readBin(file, 'raw', n=bytes)
+      b64 <- rawToBase64(fileData)
+      return(paste('data:', contentType, ';base64,', b64, sep=''))
+    },
+
+    setInputs = function(...) {
+      vals <- list(...)
+      # TODO: is there really not a way to access `names` from inside an lapply?
+      lapply(names(vals), function(k){
```
wch: The way this is usually done is:

```R
mapply(names(vals), vals, function(name, value) {
  private$.input$set(name, value)
})
```

If the return value matters, you may twiddle with the `SIMPLIFY` and `USE.NAMES` arguments, but I don't think that's the case here.

R/mock-session.R
```diff
@@ -0,0 +1,283 @@
...
+#   users are currently assigning into clientData? That would not work as expected here.
+#' @noRd
+#' @export
+`$.mockclientdata` <- function(x, name) {
+  if (name == "allowDataUriScheme") { return(TRUE) }
+  if (name == "pixelratio") { return(1) }
+  if (name == "url_protocol") { return("http:") }
+  if (name == "url_hostname") { return("mocksession") }
+  if (name == "url_port") { return(1234) }
+  if (name == "url_pathname") { return("/mockpath") }
```
wch: Is there a reason these particular values are used for the path, hash, search, etc?

R/mock-session.R
trestletech: Yup. Good catch. I'll make that change.

R/mock-session.R
```diff
@@ -0,0 +1,283 @@
...
+        private$timer$elapse(t)
+
+        # timerCallbacks must run before flushReact.
+        private$timer$executeElapsed()
+        private$flush()
+      }
+
+      private$timer$elapse(msLeft)
+
+      # TODO: needed? We're guaranteed to not have anything to run given the above loop, right?
```
wch: Can the code below be relevant if a scheduled event schedules another event?

R/mock-session.R
```diff
@@ -0,0 +1,283 @@
...
+      # Contract is to return Sys.time, which is seconds, not millis.
+      private$timer$getElapsed()/1000
+    },
+
+    defineOutput = function(name, value, label) {
+      obs <- observe({
+        # We could just stash the promise, but we get an "unhandled promise error". This bypasses
+        prom <- NULL
+        tryCatch({
+          v <- value(self, name) #TODO: I'm not clear what `name` is supposed to be
```
wch: `name` is the name of the output. For example for `output$foo <- ...`, the `name` would be `"foo"`.

R/mock-session.R
```diff
@@ -0,0 +1,283 @@
...
+      private$timer$executeElapsed()
+      private$flush()
+    },
+
+    .now = function() {
+      # Contract is to return Sys.time, which is seconds, not millis.
+      private$timer$getElapsed()/1000
+    },
+
+    defineOutput = function(name, value, label) {
```
wch: There are a bunch of tricky things that the real `defineOutput` does that I think the the mock version will also need to do:

* Reactive domain: There needs to be a reactive domain in order for the observer to be automatically cleaned up. (I'm not sure if that's already happening or not.)
* Overwriting an output should destroy the previous one.
* I think that `value` should be renamed to `func`, for consistency with the real `defineOutput`.
* Might not be worth addressing right now (maybe in a separate PR), but we'll probably want to wrap the call to `func()` in ` private$withCurrentOutput()`. This is so the `getCurrentOutputInfo()` function will work. That function allows code in the output, or in any reactives that it calls, to introspect and find out information about the output that's being executed. The information is generated by https://github.com/rstudio/shiny/blob/005295f/R/shiny.R#L1298-L1337

R/mock-session.R
```diff
@@ -0,0 +1,283 @@
...
+      obs <- observe({
+        # We could just stash the promise, but we get an "unhandled promise error". This bypasses
+        prom <- NULL
+        tryCatch({
```
wch: I think it would be a little clearer if the `tryCatch` only wrapped the `v <- value(self, name)`, and not the `if-else` stuff that checks whether `v` is a promise, since that stuff can't really error.

R/mock-session.R
```diff
@@ -0,0 +1,283 @@
...
+          v <- value(self, name) #TODO: I'm not clear what `name` is supposed to be
+          if (!promises::is.promise(v)){
+            # Make our sync value into a promise
+            prom <- promises::promise(function(resolve, reject){ resolve(v) })
```
wch: I'm a little wary of converting all non-promises to promises... I can't think of a way that it would break things in this case, but we should be careful that it doesn't introduce timing issues and error handling.

R/mock-session.R
```diff
@@ -0,0 +1,281 @@
...
+            list(val = NULL, err = e)
+          })
+      })
+      private$outs[[name]] <- list(obs = obs, func = value, promise = NULL)
+    },
+
+    getOutput = function(name) {
+      # Unlike the real outputs, we're going to return the last value rather than the unevaluated function
+      if (is.null(private$outs[[name]]$promise)) {
+        stop("The test referenced an output that hasn't been defined yet: output$", name)
```
wch: Am I correct that if `private$outs[[name]]` is not NULL, but the `$promise` part is NULL, then the output has been defined, but the observer has not yet run?

R/mock-session.R
```diff
@@ -0,0 +1,281 @@
...
+    onBookmark = function(fun) {
+      warning("Bookmarking isn't meaningfully mocked in MockShinySession")
+    },
+    onBookmarked = function(fun) {
+      warning("Bookmarking isn't meaningfully mocked in MockShinySession")
+    },
+    doBookmark = function() {
+      warning("Bookmarking isn't meaningfully mocked in MockShinySession")
+    },
+    onRestore = function(fun) {},
```
wch: I think `onRestore` and `onRestored` should give warnings as well.

R/mock-session.R
```diff
@@ -0,0 +1,281 @@
...
+      # Make promise return
+      v <- extract(private$outs[[name]]$promise)
+      if (!is.null(v$err)){
+        stop(v$err)
+      } else {
+        v$val
+      }
+    },
+
+    registerDataObj = function(name, data, filterFunc) {},
```
wch: Should the all the no-op functions give warnings?

trestletech: Perhaps. My mentality here was that there are some methods that you might expect to work (e.g. bookmarking), but there are others that are pretty meaningless in a mock session (e.g. sendCustomMessage). 

I didn't want to litter the console with a bunch of warnings for every module that uses sendCustomMessage when (hopefully) everybody already understands that there's no client to send a message to in this setup. Effectively my goal was to warn for any method that I think we might implement in the future until it's implemented.

But I could be convinced that it might be more consistent to always warn. The other thing I'd considered was an opt-out on the session constructor (or an option?) that would allow you to suppress these warnings. If we had that escape hatch, I don't think I'd be so opposed to always warning, though it still feels a little weird to me.

R/mock-session.R
trestletech: This may just fold into the conversation on the other comment, but my logic for not warning here is that -- in my head -- people wouldn't expect onRestore/d to be invoked ever since there's no client to do the restoring. So defining an `onRestore` method in a module isn't harmful or broken -- it's just rightly a no-op.

R/mock-session.R
trestletech: Yup.  just fixed that and added a test to make that more explicit.

trestletech:
Just pushed a commit with a few fixes. A couple of notes:

1. I have more extensive tests around get/defineOutput that will come in in the next PR, so those are intentionally pretty light right now.
2. In my latest commit I bring in two functions that I hadn't defined in this PR to synchronously block on a promise. We might have better approaches to doing that or -- even better -- be able to leverage something in an imported package instead of rolling our own.

R/mock-session.R
trestletech: Talked live and decided to stick with this for now for simplicity. Will revisit if we find edge cases.

R/mock-session.R
trestletech: Spoke live. Points 2, 3 are handled in recent commits. Point 1 will fall to the caller (in a subsequent PR). Point 4 will be a follow-on PR to come later. Will prioritize as necessary based on feedback of people who hit that.

R/mock-session.R
```diff
@@ -1,3 +1,31 @@
+# Promise helpers taken from:
+#   https://github.com/rstudio/promises/blob/master/tests/testthat/common.R
+# Block until all pending later tasks have executed
+wait_for_it <- function() {
+  while (!later::loop_empty()) {
+    later::run_now()
```
wch: I'm pretty sure the `run_now` and `Sys.sleep` can be replaced with `run_now(0.1)`. I'm not sure why it was written that way in promises. @jcheng5 can you enlighten us?

R/mock-session.R
```diff
@@ -282,6 +282,9 @@ MockShinySession <- R6Class(
     getTestSnapshotUrl = function(input=TRUE, output=TRUE, export=TRUE, format="json") {},
     ns = function(id) {
       paste0("mock-session-", id) # TODO: does this need to be more complex/intelligent?
+    },
+    flushReact = function(){
+      shiny:::flushReact()
```
wch: Don't need `shiny:::`

R/mock-session.R
jcheng5: I’m sure you’re right. I just threw those helpers together.

PR #2682, opened 2019-10-24
Introduce integration testing framework

trestletech:
This PR introduces an integration testing framework that facilitates the testing of reactives inside of a Shiny server function or a Shiny module. The two entry-points are `testModule` and `testServer` which take a module or a shiny application object, respectively.

The vignette included in this PR best captures the functionality and how to use this. The initial version is published here: https://beta.rstudioconnect.com/content/6874/integration-testing.html

Intentionally omitted from this PR is any inline documentation on these new functions or their associated objects. I'm still debating whether this would be best served by an article in the dev center, something roxygenized, more vignettes, or some combination. I'll start work on how to document and expose this stuff next. ([Preview](https://github.com/rstudio/shiny/compare/jeff/int-test...jeff/int-test-docs?expand=1))

## Validation

This PR finally surfaces some of the internal work that we've been doing for a few weeks now. So there's a little more surface area that we could test here. However, any edge cases that we've discovered around weird reactivity or timing issues are already covered in unit tests, so I don't think there will be too much exploring required.

Take a look at `testServer` and `testModule` and pass in a few applications or modules as discussed in the vignette (linked above). If everything there works, I think we may be good for now. Once the documentation for all this lands, I'll want to get more feedback on usability and approachability, but that's intentionally omitted right now.


.gitignore
```diff
@@ -10,3 +10,4 @@ shinyapps/
 README.html
 .*.Rnb.cached
 tools/yarn-error.log
+vignettes/*.R
```
wch: Why do these files need to be ignored?

R/test-module.R
```diff
@@ -0,0 +1,117 @@
+
+
+#' Test a shiny module
+#' @param module The module under test
```
wch: "under test" is a little confusing -- "The module to test" seems clearer to me.

.gitignore
trestletech: I have no idea. Generating vignettes seems to create a .R and a .html file in the vignettes directory. I cargo-culted this from dplyr (https://github.com/tidyverse/dplyr/blob/master/.gitignore#L12) and have just been manually ignoring/deleting the generated HTML file... 🙈

If you know better, I'm happy to revise.

R/test-module.R
```diff
@@ -0,0 +1,117 @@
...
+  # Capture the environment from the module
+  # Inserts `session$env <- environment()` at the top of the function
+  fn_body <- body(module)
+  fn_body[seq(3, length(fn_body)+1)] <- fn_body[seq(2, length(fn_body))]
+  fn_body[[2]] <- quote(session$env <- environment())
+  body(module) <- fn_body
+
+  # Substitute expr for later evaluation
+  if (!is.call(expr)){
+    expr <- substitute(expr)
```
wch: Isn't this already handled by `testModule`?

R/test-module.R
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  # Create a mock session
+  session <- MockShinySession$new()
+
+  # Parse the additional arguments
+  args <- list(...)
```
wch: These lines can be consolidated with:

```R
args <- list(..., input = session$input, output = session$output, session = session)
```

R/test-module.R
trestletech: Good catch! Dead code that I forgot to pull out.

DESCRIPTION
```diff
@@ -89,7 +90,9 @@ Suggests:
     ggplot2,
     reactlog (>= 1.0.0),
     magrittr,
-    yaml
+    yaml,
+    future,
```
wch: These packages aren't needed anymore, right?

trestletech: They're still used in `test-test-module`, actually.

PR #2686, opened 2019-10-25
Message when autoloading R/ files

trestletech:
Adds a message on app startup when automatically sourcing R files from the R/ dir to make this change more obvious. Also adds a way to opt-out using a magical file instead of just the option. This will enable environments like RStudio Connect which don't allow setting of options before the app is run to opt-out of this behavior.

Note that https://rstd.io/shiny-autoload currently just points to this PR. Once we have a real article, we'll want to update that.

## Validation

`remotes::install_github("rstudio/shiny")`

 - Should not show any additional message when loading a Shiny application that doesn't have an `R/` dir or if it's empty.
 - Should show the message `Automatically loading X .R files found in the R/ directory.\nSee https://rstd.io/shiny-autoload for more info.` if there is an R directory with files in it. See https://github.com/rstudio/shiny-examples/tree/master/168-supporting-r-dir
 - If you add a file named `R/_disable_autoload.R` (case-insensitive) to an application that has an R directory with content in it (as in the test above), you should see that the files no longer load automatically and you get the following message: `R/_disable_autoload.R detected; not loading the R/ directory automatically`


R/app.R
```diff
@@ -340,8 +339,21 @@ loadSupport <- function(appDir, renv=new.env(parent=globalenv()), globalrenv=glo
   }
 
   helpersDir <- file.path(appDir, "R")
+
+  disabled <- list.files(helpersDir, pattern="_disable_autoload\\.r$", recursive=FALSE, ignore.case=TRUE)
```
wch: I think the pattern should start with `^`.

trestletech: Touche! Fixed.

wch:
Any particular reason why the file ends with `.R`?

trestletech:
I'm shooting for compatibility with R packages. I assume the file will get loaded as a .R file. I confirmed that R CMD check doesn't whine about a file name starting with an underscore, but I wasn't sure how it would react to some .txt file in the R dir.

wch:
Ah, the package compatibility thing makes sense.

PR #2689, opened 2019-10-30
bs4 compatibility

cpsievert:
Requires https://github.com/rstudio/htmltools/pull/145

For examples, see 

https://github.com/rstudio/bs4/blob/master/demo/elements.R
https://github.com/rstudio/bs4/blob/master/demo/elements2.R


R/bootstrap-layout.R
```diff
@@ -231,8 +231,12 @@ column <- function(width, ..., offset = 0) {
     stop("column width must be between 1 and 12")
```
cpsievert: Maybe we should consider a `bs4::column()` that doesn't require `width` to take advantage of the new auto-widths

jcheng5: Yes, I started working on that. Also wanted a way to specify different widths for different breakpoints (sm, lg, etc)

nteetor: There are equal width columns and also the auto width (width of content) columns. I referred to these values as `"equal"` and `"content"` in yonder, but I could see changing the values for the sake of continuity across packages. 

As for breakpoints I opted for `c(sm = 1, lg = "auto", default = 12)`, where `default` could also be `xs`.

R/input-utils.R
```diff
@@ -1,7 +1,7 @@
 shinyInputLabel <- function(inputId, label = NULL) {
   tags$label(
     label,
-    class = "control-label",
+    class = "control-label col-form-label",
```
jcheng5: Maybe we want to shim this instead??

R/showcase.R
```diff
@@ -80,10 +80,16 @@ appMetadata <- function(desc) {
...
+    isActive <- tolower(file) %in% c("app.r", "server.r")
+    withTags(
+      li(
+        class = if (isActive) "active",
+        a(
+          href = paste("#", gsub(".", "_", file, fixed = TRUE), "_code", sep = ""),
+          "data-toggle" = "tab",
+          paste0(prefix, file)
+        )
+      )
```
jcheng5: Looks to me like this logic didn't change after all?

srcjs/input_binding_fileinput.js
```diff
@@ -177,6 +177,8 @@ $.extend(FileUploader.prototype, FileProcessor.prototype);
   };
   this.$setError = function(error) {
     this.$bar().toggleClass('progress-bar-danger', (error !== null));
+    // bs4 forward compatibility
+    this.$bar().toggleClass('bg-danger', (error !== null));
```
jcheng5: It's weird that there isn't a `text-danger`, isn't it? https://getbootstrap.com/docs/4.0/utilities/colors/#background-color

Maybe we should set `progress-bar-danger` in bs3compat, using `color-yiq($danger)` or whatever to set the `color`; or else maybe we should introduce `.text-danger`.

R/input-file.R
```diff
@@ -120,7 +121,7 @@ fileInput <- function(inputId, label, multiple = FALSE, accept = NULL,
     tags$div(
       id=paste(inputId, "_progress", sep=""),
       class="progress progress-striped active shiny-file-input-progress",
-      tags$div(class="progress-bar")
+      tags$div(class="progress-bar progress-bar-striped")
```
jcheng5: Consider just getting rid of the stripes. They look a bit dated.

jcheng5: Make sure that we don't need to support animated stripes (i.e. indeterminate state)

R/input-utils.R
jcheng5: I actually think this might be completely wrong... the bs4 docs only use this for horizontal form labels.

srcjs/input_binding_fileinput.js
jcheng5: https://github.com/rstudio/bootstraplib/pull/30

R/input-file.R
cpsievert: I'm not sure this is a great idea for the `Progress` API...it seems like there should be some visual cue for indeterminate state

jcheng5: We actually don't have indeterminate state for the Progress API. We just hide the progress bar if we don't have a number.

jcheng5:
@cpsievert I think this is ready. I'll probably squash-merge, there are so many commits (and shiny.js/shiny.min.js builds) that didn't end up being part of the final commit.

R/input-file.R
cpsievert: Maybe I don't understand what you mean by indeterminate state...I was thinking of a long running computation, like:

```r
progress <- Progress$new(session, min=1, max=3)
progress$set(message = 'Calculation in progress', detail = 'This may take a while...')
progress$set(value = 2)
Sys.sleep(5)
progress$set(value = 3)
progress$close()
```

PR #2690, opened 2019-10-30
Datepicker: make disabled months and years lighter gray

### LINKED ISSUE #2042: [Styling Recommendation] dateRangeInput dates within min/max not easily distinguishable from the rest

### LINKED ISSUE #2628: Disabled months and years in dateInput/dateRangeInput are too dark
Disabled day numbers are colored #aaa, but when you zoom out, disabled months and years are colored `#777`, which makes it much less obvious that they are disabled. 

Running 037-date-and-date-range from shiny-examples, the dates look like this:

image

However, months look like this:
image

wch:
Closes #2042, closes #2628.

This makes disabled months and years a lighter gray, which makes it more obvious that they're disabled.

When running shiny-examples/037-date-and-date-range/:

Before:

image

After:

image

Test notes:

Run shiny-examples/037-date-and-date-range/, and make sure that disabled years and months are light gray.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,6 +9,8 @@ shiny 1.4.0.9000
 
 ### Minor new features and improvements
 
+* Fixed [#2042](https://github.com/rstudio/shiny/issues/2042), [#2628](https://github.com/rstudio/shiny/issues/2628): In a `dateInput` and `dateRangeInput`, disabled months and years are now a lighter gray, to make it easier to see that they are disabled. ([#2690](https://github.com/rstudio/shiny/pull/2690))
+
 ### Bug fixes
 
 * Fixed [#2606](https://github.com/rstudio/shiny/issues/2606): `debounce()` would not work properly if the code in the reactive expression threw an error on the first run. ([#2652](https://github.com/rstudio/shiny/pull/2652))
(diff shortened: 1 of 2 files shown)


wch:
Good idea. I've made that change.

PR #2694, opened 2019-11-04
Expand testServer/Module docs

trestletech:
Decided to defer to the dev center since we decided to keep the vignette out of the package. Hoping to avoid having too much duplicated content, but happy to revise.


R/test-module.R
```diff
@@ -10,6 +15,29 @@
...
+#'     input$x * 2
+#'   })
+#'   output$txt <- renderText({
+#'     paste0("I am ", myreactive())
+#'   })
+#' }
+#'
+#' testModule(module, {
+#'   session$setInputs(x = 1)
+#'   # You're alos free to use third-party
```
wch: Typo: "alos"

PR #2696, opened 2019-11-05
fileupload.R: change upload temp filename to a unique id

thomasbrr:
This would enable the use of several fileInput() on the same page.


wch:
Can you please file an issue describing the problem and provide a reproducible example? That would be helpful for others to evaluate this pull request.

PR #2713, opened 2019-11-26
.testModule(): remove args parameter fixes #2709

### LINKED ISSUE #2709: testModule ignores args argument
@jcheng5 pointed out that `.testModule()` currently ignores the `args` argument, which is a bug. It should also perhaps use `rlang::list2()` to construct an argument list instead of `base::list()` so that `!!!`-style splicing can be used by consumers.

Relatedly, it might be possible to simplify `.testModule()` internals by using rlang stuff to metaprogram instead of manual AST manipulation.

alandipert:
This PR does 2 things:

1. **Improvement**: Simplifies synthesis of a modified module function body by leveraging `rlang::expr()` and `!!!`
1. **Improvement**: Uses `rlang::list2()` instead of `list()` to incorporate `...` into the module's arg list. Consequently, the dots are now *dynamic*, and the `args` argument was removed. Docs and tests around dynamic dots were added.

I didn't edit NEWS.md since this fixes a bug in an unreleased feature.

### DIFF
--- a/R/test-module.R
+++ b/R/test-module.R
@@ -26,11 +26,8 @@
 #'   })
 #' }
 #'
-#' # !!/:= and !!! from rlang are used below to programmatically splice
-#' # these into the testModule() argument list.
-#' multiplier_arg_name = "multiplier"
-#' more_args <- list(prefix = "I am ")
-#'
+#' # Basic Usage
+#' # -----------
 #' testModule(module, {
 #'   session$setInputs(x = 1)
 #'   # You're also free to use third-party
@@ -42,6 +39,19 @@
 #'   session$setInputs(x = 2)
 #'   stopifnot(myreactive() == 4)
 #'   stopifnot(output$txt == "I am 4")
+#'   # Any additional arguments, below, are passed along to the module.
+#' }, multiplier = 2)
+#'
+#' # Advanced Usage
+#' # --------------
+#' multiplier_arg_name = "multiplier"
+#' more_args <- list(prefix = "I am ")
+#' testModule(module, {
+#'   session$setInputs(x = 1)
+#'   stopifnot(myreactive() == 2)
+#'   stopifnot(output$txt == "I am 2")
+#'   # !!/:= and !!! from rlang are used below to splice computed arguments
+#'   # into the testModule() argument list.
 #' }, !!multiplier_arg_name := 2, !!!more_args)
 #' @export
 testModule <- function(module, expr, ...) {
--- a/R/test-module.R
+++ b/R/test-module.R
@@ -49,16 +49,17 @@ testModule <- function(module, expr, args, ...) {
 .testModule <- function(module, expr, args, ...) {
   # Capture the environment from the module
   # Inserts `session$env <- environment()` at the top of the function
-  fn_body <- body(module)
-  fn_body[seq(3, length(fn_body)+1)] <- fn_body[seq(2, length(fn_body))]
-  fn_body[[2]] <- quote(session$env <- environment())
-  body(module) <- fn_body
+  body(module) <- rlang::expr({
+    session$env <- environment()
+    !!!body(module)
+  })
 
   # Create a mock session
   session <- MockShinySession$new()
 
   # Parse the additional arguments
-  args <- list(..., input = session$input, output = session$output, session = session)
+  args <- if (missing(args)) list() else args
+  args <- append(rlang::list2(..., input = session$input, output = session$output, session = session), args)
 
   # Initialize the module
   isolate(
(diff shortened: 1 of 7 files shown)


R/test-module.R
```diff
@@ -49,16 +49,17 @@ testModule <- function(module, expr, args, ...) {
...
+    session$env <- environment()
+    !!!body(module)
+  })
 
   # Create a mock session
   session <- MockShinySession$new()
 
   # Parse the additional arguments
-  args <- list(..., input = session$input, output = session$output, session = session)
+  args <- if (missing(args)) list() else args
```
wch: Using `missing()` is generally not a good idea because it makes it hard to programmatically call a function. I think it would be better to modify the signature of `testModule` so that it has `args = NULL`.

R/test-module.R
```diff
@@ -49,16 +49,17 @@ testModule <- function(module, expr, args, ...) {
...
+  args <- append(rlang::list2(..., input = session$input, output = session$output, session = session), args)
```
wch: Instead of `append(x, y)`, it's more idiomatic to do `c(x, y)` (which is what `append` does under the hood anyway).

R/test-module.R
wch: A couple more thoughts:

* `args` should be renamed to `args_` to avoid name collisions (if the module itself takes `args`).
* We should validate that `args_` is either NULL, or a list with all elements named. The reason to check for all named elements is because if any are not named, the order that they'll be passed to the module function will be opaque to the user.

wch: New thoughts:

* Get rid of `args`, since the user can use `!!!` to splice in lists of args.
* Provide an example of using `!!!`.
* In the documentation, it says that `...` are "additional named arguments", but that should be changed to say "additional arguments".
* See what happens if user provides input/output/session vars.

tests/testthat/test-test-module.R
```diff
@@ -4,6 +4,14 @@ library(promises)
...
 plan(multisession)
 
+test_that("testModule passes args", {
+  module <- function(input, output, session, someArg) {
+    expect_false(missing(someArg))
+    expect_equal(someArg, 123)
+  }
+  testModule(module, {}, args = list(someArg = 123))
+})
+
```
wch: Note: Add a test with `!!!`, and with named/unnamed arg combinations.

R/test-module.R
```diff
@@ -37,28 +42,28 @@
 #'   session$setInputs(x = 2)
 #'   stopifnot(myreactive() == 4)
 #'   stopifnot(output$txt == "I am 4")
-#' })
+#' }, !!multiplier_arg_name := 2, !!!more_args)
```
wch: I  think there should be two examples with `testModule`. The first one wouldn't use `!!` or `!!!` and simply pass `multiplier = 2`.

The next one could use `!!` and `!!!` (and there should be a comment saying this is for advanced users). Otherwise normal users will see this example and get confused or scared off, and by something that is only used in rare occasions.

tests/test-helpers/app4-both/R/upper.R
```diff
@@ -0,0 +1 @@
+upperHelper <- 'abc'
```
wch: Is this left over from running this? https://github.com/rstudio/shiny/blob/e84beffee39d540d3e891faafb9f333af32deca4/tests/testthat/test-app.R#L54

If it is, then that test block should have an `on.exit` that removes the file. (It's OK to put it in this PR)

tests/testthat/test-test-module.R
```diff
@@ -4,6 +4,31 @@ library(promises)
...
+  }
+
+  # Test with !!! to splice in a whole named list constructed with base::list()
+  moreArgs <- list(someArg = 123)
+  testModule(module, {}, !!!moreArgs)
+
+  # Test by splicing in a list constructed with rlang::list2() that uses := to
+  # splice in an argument name
+  argName <- "someArg"
+  moreArgs <- rlang::list2(!!argName := 123)
```
wch: I think this test is effectively the same as the previous one, because when `list2()` is called here, it resolves the `!!!`.

Maybe this test should instead use `!!` in the call to `testModule`?

R/test-module.R
alandipert: I tried to clean this up, let me know if it's what you had in mind.

tests/test-helpers/app4-both/R/upper.R
alandipert: I went with ye olde `testthat::teardown()` for this instead of `on.exit()`

R/test-module.R
wch: Looks good!

wch:
Fixes #2709.

Issue #2712, opened 2019-11-26
Using `testModule` breaks when module calls another


akgold:
```R
module <- function(input, output, session) {
  callModule(module2, "module2")
}
module2 <- function(input, output, session) {
}
testModule(module)
```
Error output
```
Error in session$makeScope(id) : attempt to apply non-function
```

alandipert:
Thank you for the report. Like #2708 I think this could be because `MockShinySession` is missing a method, in this case `makeScope()`.

alandipert:
@akgold with @wch's guidance I think I was able to fix this, and the fix is on the `test-nested-modules` branch if you have a moment to try it out.

I'd also be curious to know how you ran into this problem, and if nested modules are something you've worked with before. I'd like to add a test for this, and ideally the test would be based on a usage or pattern that exists in the wild.

Thanks again for the report, and in advance for the additional context.

akgold:
Hi @alandipert , 

Getting a different error now: 

```r
remotes::install_github("rstudio/shiny", "test-nested-modules")
library(shiny)

module <- function(input, output, session) {
  
  output$var <- 3
  callModule(module2, "module2")
}

module2 <- function(input, output, session) {
  output$var2 <- 2
}

testModule(module, {
  print(output$var)
})
```
```
Error in func(self, name) : could not find function "func"
```

This is a use pattern from a real app. In this case, the use pattern was having users submit a form. I have a module that houses the form itself, which calls a nested module when the form is submitted. In this case the submit logic is complicated enough that I wanted to be able to separate it out from the form.

I have another nested module (in a different app) that's conceptually similar. The app displays information about a class, and I house the logic for the various updates that are allowed in a separate module.

alandipert:
@akgold thanks for that example. I think the error you saw is because a statement like `output$var <- 3` would need to be something like `output$var <- renderText(3)` to work, generally. 

In a regular old Shiny module, doing that would normally produce an error message like this:
```
 Error in .subset2(x, "impl")$defineOutput(name, value, label) : 
  Unexpected numeric output for var 
```

However, since `testModule()` and various new code paths are involved, you're seeing a lame error message. I pushed a commit that should improve the error message.

akgold:
Beautiful! This works nicely on my toy example now. Will be working on a real implementation after the holiday.

PR #2714, opened 2019-11-27
Add MockShinySession$makeScope()

alandipert:
This PR adds a missing method to `MockShinySession` and ports several checks from `ShinySession$defineOutput()` to `MockShinySession$defineOutput()`.

- [x] Implement
- [x] Test

## Nested modules

I took some time on this one to explore the case for nesting modules, to see if a test could be added that was representative of some expected usage, or usage pattern.

@akgold was kind enough to share his nested module usage, and we concluded that using `callModule()` in a module function doesn't provide any affordance that a regular function call does not. However, it should definitely work, and it didn't until this PR, which adds `MockShinySession$makeScope()`.

I ultimately decided to keep the nested module test short and sweet.

## defineOutput() checks

In the #2712 comments Alex ran into a confusing error message resulting from the fact that `MockShinySession$defineOutput()` was missing various checks performed by `ShinySession`. So, I ported the checks from ShinySession over to MockShinySession.

## QA Notes

It should now be possible to use `callModule()` from within a module function without error, and a unit test was added to ensure this.

Assigning to an output in a module function called by `testModule()` or `testServer()` should now also produce an error. For example, `output$var <- 123` should fail (`output$var <- renderText("123")` is the correct way). A unit test was added for this as well.

There should be no other behavior changes to `testModule()`/`testServer()`.

### DIFF
--- a/tests/testthat/test-test-module.R
+++ b/tests/testthat/test-test-module.R
@@ -500,7 +500,7 @@ test_that("testModule works with nested modules", {
   })
 })
 
-test_that("Assigning an output in a module function with a non-function errors", {
+test_that("assigning an output in a module function with a non-function errors", {
   module <- function(input, output, session) {
     output$someVar <- 123
   }
--- a/tests/testthat/test-test-module.R
+++ b/tests/testthat/test-test-module.R
@@ -500,6 +500,14 @@ test_that("testModule works with nested modules", {
   })
 })
 
+test_that("Assigning an output in a module function with a non-function errors", {
+  module <- function(input, output, session) {
+    output$someVar <- 123
+  }
+
+  expect_error(testModule(module, {}), "^Unexpected")
+})
+
 test_that("testServer works", {
   # app.R
   testServer({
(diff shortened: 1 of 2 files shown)


tests/testthat/test-test-module.R
```diff
@@ -485,6 +485,29 @@ test_that("accessing a non-existant output gives an informative message", {
   })
 })
 
+test_that("testModule works with nested modules", {
+  outerModule <- function(input, output, session) {
+    output$someVar <- renderText("rendered text")
+    expect_equal(callModule(innerModule, "innerModule"), "a return value")
+  }
+
+  innerModule <- function(input, output, session) {
```
wch: Maybe pass in a reactive expression `r` like this:

```R
outerModule <- function(input, output, session) {
    output$someVar <- renderText("rendered text")
    r <- reactive({ input$x + 1 })
    module_val <- callModule(innerModule, "innerModule", r)

    module_val
  }

  innerModule <- function(input, output, session, r) {
    new_string <- reactive({
      paste("a return value:", r())
    })

    new_string
  }

  testModule(outerModule, {
    session$setInputs(x = 1)

    expect_equal(output$someVar, "rendered text")
    expect_equal(session$returned(), "a return value: 2")
  })
```

PR #2717, opened 2019-12-02
Add label to file reload observer

akgold:
Closes rstudio/reactlog#58


schloerke:
(adding for auto close)

Fixes https://github.com/rstudio/reactlog/issues/58

PR #2718, opened 2019-12-03
Allow for `shown` method to be called directly on document.body

schloerke:
Found from https://github.com/rstudio/learnr/issues/308 where changing slides in a slidy does not trigger shiny to detect output elements' visibility.  (This issue will be addressed in https://github.com/rstudio/rmarkdown/pull/1717)

From https://api.jquery.com/on/ :

> If selector is omitted or is null, the event handler is referred to as direct or directly-bound. The handler is called every time an event occurs on the selected elements, whether it occurs directly on the element or bubbles from a descendant (inner) element.
>
> When a selector is provided, the event handler is referred to as delegated. The handler is not called when the event occurs directly on the bound element, but only for descendants (inner elements) that match the selector.

I would like to be able to call
```js
$(document.body).trigger("shown");
```

and not be required to call

```js
$(document.body).children().first().trigger("shown");
```

Possible changeup:
- [ ] Add a `Shiny.redetect()` method that calls `$(document.body).children().first().trigger("shown");` to avoid adding the proposed listeners.


srcjs/init_shiny.js
```diff
@@ -421,10 +421,20 @@ function initShiny() {
   });
 
   // This is needed for Bootstrap 2 compatibility and for non-Bootstrap
-  // related shown/hidden events (like conditionalPanel)
+  //   related shown/hidden events (like conditionalPanel)
+  // Execute on all body inner elements
   $(document.body).on('shown.sendImageSize', '*', sendImageSize);
-  $(document.body).on('shown.sendOutputHiddenState hidden.sendOutputHiddenState', '*',
-               sendOutputHiddenState);
+  $(document.body).on(
```
wch: If you replace `$(document.body)` with `$(document)`, then you can remove the duplicated calls below (the ones without the `'*'`. I think that should work fine without any drawbacks, but it wouldn't hurt to check.

schloerke: Checked.  It wouldn't hurt as we have the fns debounced, but it's called twice.

```js
$(document).on("schloerke", "*", function(){console.log("schloerke")})
// jQuery.fn.init [document, context: document]
$(document).on("schloerke.wch", "*", function(){console.log("schloerke.wch")})
// jQuery.fn.init [document, context: document]
$(document.body).trigger("schloerke")
// VM1091:1 schloerke
// VM1125:1 schloerke.wch
// VM1091:1 schloerke
// VM1125:1 schloerke.wch
// jQuery.fn.init [body.api.jquery.single.single-post.postid-461.single-format-standard.single-author.singular, context: body.api.jquery.single.single-post.postid-461.single-format-standard.single-author.singular]
```

Even then, I'd like to clean up the trigger request to `$(document).trigger("shown")`... which makes me think it should be a standardized Shiny.js method.

schloerke:
Closing this PR as the core philosophy of the PR wrong.  Just trigger `shown` on an element of concern.

------------------------------

What I had originally wanted was `Shiny.redetect()`, but adding to the public API is not a good option at this point.  

Adding a listener to document.body is ok, but I was afraid of the unintended side effects that may occur.  

The original problem came from https://github.com/rstudio/rmarkdown/pull/1717, however I am able to find a slide element to `trigger` the `'shown'` method.  Therefore, the request of wanting to call a `redetect()` method on `document` (or `document.body`) does not seem valid anymore.

To keep things safe, we should leave the API as is.

PR #2721, opened 2019-12-09
Use a frozen lock file within CI

schloerke:
I ran into yarn.lock being updated in this repo and/or another repo.  The updated lock file would cause a failure.

-------------------------------

https://yarnpkg.com/lang/en/docs/cli/install/
> If you need reproducible dependencies, which is usually the case with the continuous integration systems, you should pass --frozen-lockfile flag.

https://yarnpkg.com/lang/en/docs/cli/install/#toc-yarn-install-frozen-lockfile
> Don’t generate a yarn.lock lockfile and fail if an update is needed.

### DIFF
--- a/tools/checkJSCurrent.sh
+++ b/tools/checkJSCurrent.sh
@@ -3,7 +3,7 @@
 set -e
 
 # Run JS build process
-(cd "$(dirname "$0")" && yarn && yarn build)
+(cd "$(dirname "$0")" && yarn --frozen-lockfile && yarn build)
 
 if [ -n "$(git status --porcelain)" ]
 then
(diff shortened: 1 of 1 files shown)


wch:
Does `yarn` update the yarn.lockfile by default? That seems to make the "lock" part of it pointless.

schloerke:
Yes.

`npm` does this now by default as well. `npm` introduced `npm ci` to install what the lockfile describes.  I think yarn should do the same, even if it's just a wrapper around `yarn --frozen-lockfile`.

https://medium.com/better-programming/npm-ci-vs-npm-install-which-should-you-use-in-your-node-js-projects-51e07cb71e26

Issue #2732, opened 2019-12-20
Inline markdown support


alandipert:
It's possible to support inline markdown in a limited sense: https://tailrecursion.com/wondr/posts/shiny-inline-markdown.html

@hadley suggested adding something like it to Shiny on Twitter: https://twitter.com/hadleywickham/status/1208110661533876258

It's definitely nice to be able to write prose in markdown in an app without needing a separate markdown file, but there are a few things we'd need to address before adding to Shiny:

1. Indentation
    - R doesn't have multiline string syntax
    - Markdown is sensitive to leading whitespace (code blocks)
    - The proof-of-concept on my blog works by removing leading whitespace by regex, so code blocks are impossible to write unless every line of the string starts at the beginning of every line, but that looks really weird if the string is nested in other tags.
    - Maybe there's some trick to making a multiline string operator in R that I just don't know?
        - __Hadley pointed out [glue::trim()](https://glue.tidyverse.org/reference/trim.html) in comments__
    - On Windows newlines are `\r\n`, so if we go the regex route, the replacement behavior would have to be based on the format of the source file. Considering files on Windows being edited in RStudio might have Unix line endings, this doesn't seem like it would be as straightforward as checking the current platform.
1. Need to promote`markdown` from `Suggests` to `Imports`
    - __Maybe use [commonmark](https://github.com/jeroen/commonmark) instead of markdown per Hadley in comments__

hadley:
Two thoughts:

* Probably want to implement something like `glue::trim()`

* Might be worth using commonmark instead of markdown package.

alandipert:
`glue::trim()` looks great, any reason not to pull in `glue`?

PR #2737, opened 2020-01-04
Add shiny::markdown() for inline Markdown

alandipert:
This is a stab at implementing https://github.com/rstudio/shiny/issues/2732 and incorporates implementation suggestions by @hadley.

The implementation is very small, but this does bring in the commonmark and glue packages as dependencies (but no transitive dependencies). 

- commonmark is widely used as a dependency of roxygen2 but is unlikely to be installed on user machines not geared toward package development.
- glue is part of the tidyverse, and so probably likelier to exist on user machines.
    - I prototyped a pure-R implementation of glue::trim(), but maintaining and testing it seems onerous (I include it at the bottom). It's also wildly inefficient compared to glue's C version.

## Open questions

* Should this be an addition to htmltools instead of to Shiny?
* Currently input strings are trimmed with `glue::trim()` but are not passed to `glue::glue()`, which would both trim them and support interpolation. If the strings were passed to `glue::glue()`, it's not obvious to me how additional arguments to `glue::glue()` could be passed, since additional arguments are currently passed as commonmark::markdown_html() arguments.

## Remaining dev tasks

- [x] Update NEWS
- [x] Add tests

## `trim()` in R

```R
trim1 <- function(txt) {
  stopifnot(is.character(txt) && length(txt) == 1)
  # Remove leading and trailing whitespace
  txt <- gsub("^\\W*", "", txt)
  txt <- gsub("\\W*$", "", txt)
  # Determine and strip leading indentation
  lines <- strsplit(txt, "\n", fixed = TRUE)[[1]]
  rest <- tail(lines, -1)
  indents <- vapply(
    gregexpr("^\\W+", x),
    function (m) attr(m, "match.length"),
    numeric(1)
  )
  stripped <- gsub(paste0("^\\W{", min(indents), "}"), "", lines)
  paste0(stripped, collapse = "\n")
}
```

## QA Notes

1. Install Shiny from Github with `remotes::install_github("rstudio/shiny")`
2. Create a new single-file Shiny app in the RStudio IDE with `File > New File > Shiny Web App...`
3. At the top right of the generated `app.R` tap pane, press the "Run App" button. A window containing the Shiny demo app should pop up.
4. Close the demo app window and go back to the `app.R` tab.
5. Between lines 29 and 30, insert the following code:
```
markdown("# Some markdown
* This
* Is
* __MARKDOWN__
"),
```
6. Save `app.R` and press the "Run App" button again.
7. The demo app window should appear again. This time, you should see the markdown code you added rendered properly in the app, and it should look similar to the image below:

Screen Shot 2020-04-07 at 8 39 18 PM

8. If you make the following changes and save and run the app, the Markdown output should remain the same:

* Spaces before `#` in the line `# Some markdown` 
* Empty or blank lines before `# Some markdown`
* Empty or blank lines between `# Some markdown` and `* This`
* Empty or blank lines after the line `* __MARKDOWN__`
9. Generally, any markdown you can think to try should work and its rendered appearance should not surprise you.


DESCRIPTION
```diff
@@ -79,7 +79,9 @@ Imports:
     crayon,
     rlang (>= 0.4.0),
     fastmap (>= 1.0.0),
-    withr
+    withr,
```
hadley: If you wanted to keep the dependencies light, you could make put these in suggests. Then you'd only need to install if you wanted to use `markdown()`.

R/shiny.R
```diff
@@ -2294,3 +2294,50 @@ ShinyServerTimingRecorder <- R6Class("ShinyServerTimingRecorder",
...
+#'
+#'     [Links](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) work;
+#'     so does *emphasis*.
+#'
+#'     To see more of what's possible, check out [commonmark.org/help](https://commonmark.org/help).
+#'     ")
+#' )
+markdown <- function(mds, ...) {
+  args <- rlang::list2(glue::trim(mds), ...)
+  html <- do.call(commonmark::markdown_html, args)
```
hadley: You could simplify the previous two lines to:

```R
html <- rlang::exec(commonmark::markdown_html, glue::trim(mds), ...)
```

R/shiny.R
```diff
@@ -2294,3 +2294,50 @@ ShinyServerTimingRecorder <- R6Class("ShinyServerTimingRecorder",
...
+#' This function accepts a character vector of
+#' [Markdown](https://en.wikipedia.org/wiki/Markdown)-syntax text and returns a
+#' [htmltools::tagList()] of HTML tags that may be in included in an
+#' application's UI.
+#'
+#' Prior to interpretation as Markdown, leading whitespace is trimmed from text
+#' with [glue::trim()]. This makes it possible to insert Markdown and for it to
+#' be processed correctly even when the call to `markdown()` is indented.
+#'
+#' Markdown rendering is performed by [commonmark::markdown_html()]. Additional
```
hadley: I wonder if it's best to default to `extensions = TRUE`, since that gives gfm, which I think most people will be familiar with.

R/shiny.R
```diff
@@ -2294,3 +2294,50 @@ ShinyServerTimingRecorder <- R6Class("ShinyServerTimingRecorder",
...
+#' arguments to `markdown()` are passed as arguments to `markdown_html()`
+#'
+#' @param mds A character vector of Markdown source to convert to HTML.
```
hadley: How are multiple elements combined?

DESCRIPTION
alandipert: @schloerke, @wch, and myself discussed adding the dependency more today, and also tossed around the idea of adding this feature to `htmltools` instead of to Shiny. We concluded Shiny is the right place and that Imports is preferable to Suggests.

R/shiny.R
alandipert: HTML is concatenated

alandipert:
I was experimenting with this more today and I'm not sure `glue::trim()` is handling blank lines in the way we need. Here's an example:

```
s <- glue::trim("
                # Title
                
                This is a paragraph, here's a list of things:
                
                1. foo
                1. bar
                ")
```
That results in the following:
```
[1] "# Title\n\n                This is a paragraph, here's a list of things:\n\n                1. foo\n1. bar"
```
Notice all the leading spaces before the paragraph. That indentation produces `<pre>` instead of `<p>` when we run it through markdown:

```
md <- commonmark::markdown_html(s)

[1] "<h1>Title</h1>\n<pre><code>            This is a paragraph, here's a list of things:\n\n            1. foo\n</code></pre>\n<ol>\n<li>bar</li>\n</ol>\n"
```

I experimented with the reference implementation of trim from [PEP 257](https://www.python.org/dev/peps/pep-0257/#handling-docstring-indentation) that's mentioned in the `glue::trim()` docs, and it appears to handle blank lines the way I'd expect: https://repl.it/repls/StudiousLightsalmonLicense

hadley:
Probably worth a bug report to glue; it's never worked quite the way I expected either.

alandipert:
Blocked on https://github.com/tidyverse/glue/pull/162 or similar

alandipert:
Yay, I'll do it.

NEWS.md
```diff
@@ -13,6 +13,8 @@ shiny 1.4.0.9001
 
 * The new `moduleServer` function provides a simpler interface for creating and using modules. ([#2773](https://github.com/rstudio/shiny/pull/2773))
 
+* `markdown()` is a new function for writing Markdown with Github extensions directly in Shiny UIs. Markdown rendering is performed by the [commonmark](https://github.com/jeroen/commonmark) package. ([#2737](https://github.com/rstudio/shiny/pull/2737))
```
wch, line 16: Should start with something like "Resolved #2732:".

DESCRIPTION
```diff
@@ -79,7 +79,9 @@ Imports:
...
+    commonmark (>= 1.7),
+    glue (>= 1.3.1)
```
wch, line 84: I think this should be 1.3.2: https://github.com/tidyverse/glue/commit/2e3a4e01e088c3e9a702480e616905e5d1336495

PR #2740, opened 2020-01-17
Reporting of output styles (colors and fonts)

cpsievert:
Adds the ability to obtain computed CSS styles of output containers via `getCurrentOutputInfo()` within any `renderPlot()`, `renderImage()`, or any output binding with a `.shiny-report-theme` class. Currently the computed styles include `bg` (background-color), `fg` (color), `accent` (link-color), `font` (font-family), and `renderedFamily` (the actual font-family chosen by the browser....this information is only available if [`FontFaceSet.check()`](https://developer.mozilla.org/en-US/docs/Web/API/FontFaceSet/check) is available). For example:

```r
library(shiny)

shinyApp(
  fluidPage(
    tags$style(HTML("#info {color: #FFFFFF; background-color: #000000}")),
    tags$style(HTML("#info a {color: #00FF00}")),
    h3("Below is computed styles for the info container"),
    tagAppendAttributes(
      textOutput("info"), 
      class = "shiny-report-theme"
    )
  ),
  function(input, output) {
    output$info <- renderText({
      info <- getCurrentOutputInfo()
      jsonlite::toJSON(
        list(
          bg = info$bg(), 
          fg = info$fg(), 
          accent = info$accent(), 
          font = info$font()
        ), 
        auto_unbox = TRUE
      )
    })
  }
)
```

Screen Shot 2020-05-06 at 12 15 49 PM

Note that after this change, `htmlwidgets::shinyWidgetOutput()` will [gain a `reportTheme` argument](https://github.com/ramnathv/htmlwidgets/pull/361) to make it easier for htmlwidgets developers to include the `.shiny-report-theme` class (and, thus, access styles with `getCurrentOutputInfo()`).

In addition, a new `shiny.useragg` option was added to use `ragg::agg_png()` over `Cairo::CairoPNG` and non-quartz `png()` (if **ragg** is installed). For now, this option will default to `FALSE`, but will likely default to `TRUE` in a future release.

## Testing notes

Install `remotes::install_github("rstudio/shiny#2740")`, then run the example app above and confirm that you see the same information as the screenshot. That's ok, if you see differences in the `font` information, but everything else (i.e., `bg`, `fg`, `accent`) should be the same.

By the way, there's a fully automated testing app here https://github.com/rstudio/shinycoreci-apps/pull/47


jcheng5:
I'm seeing an error when trying to use this branch with a regular (non-autocolor) plot:

```
Warning: Error in <-: replacement has length zero
  130: maybe_set_palette
  129: drawPlot
  115: <reactive:plotObj>
   99: drawReactive
   86: origRenderFunc
   85: output$plot
    5: runApp
    3: print.shiny.appobj
    1: source
```

R/render-plot.R
```diff
@@ -266,24 +321,171 @@ drawPlot <- function(name, session, func, width, height, pixelratio, res, ...) {
...
+  do.call(lattice::trellis.par.set, params)
+}
+
+maybe_set_palette <- function(fg) {
+  p <- palette()
+  if (is_default_pallete(p)) {
+    p[[1]] <- fg
+    palette(p)
+  }
+  p
```
jcheng5: Is this supposed to return the old palette, or the modified one? A comment where this is called seems to indicate it's the former, but the implementation seems to be doing the latter.

cpsievert: Oops, yea, it's supposed to return the old one, thanks for fixing.

srcjs/init_shiny.js
```diff
@@ -288,6 +288,53 @@ function initShiny() {
...
+  // Compute the color property of an a tag, scoped within the element
+  function getComputedLinkColor(el) {
+    let a = document.createElement("a");
+    a.href = "/";
+    let div = document.createElement("div");
+    div.style.position = "absolute";
+    div.style.top = "-1000px";
+    div.appendChild(a);
+    el.appendChild(div);
+    let link = el.querySelector("a");
```
jcheng5: I don't think you need to do this, just use the `a` variable

srcjs/init_shiny.js
```diff
@@ -288,6 +288,53 @@ function initShiny() {
...
+    let linkColor = window.getComputedStyle(link).getPropertyValue("color");
+    el.removeChild(el.lastElementChild);
```
jcheng5:
```suggestion
    el.removeChild(div);
```

srcjs/init_shiny.js
```diff
@@ -288,6 +288,53 @@ function initShiny() {
...
+    return bgColor;
+  }
+
+  // Compute the color property of an a tag, scoped within the element
+  function getComputedLinkColor(el) {
+    let a = document.createElement("a");
+    a.href = "/";
+    let div = document.createElement("div");
+    div.style.position = "absolute";
+    div.style.top = "-1000px";
```
jcheng5: I would set `left`, `width`, and `height` as well, just so there's no question about how it will lay out.

jcheng5: Actually to be really safe you could make the inline CSS properties `!important`; you have to use this syntax (TIL): `div.style.setProperty("position", "absolute", "important");`

R/render-plot.R
```diff
@@ -266,26 +315,366 @@ drawPlot <- function(name, session, func, width, height, pixelratio, res, ...) {
...
+
+  # At this point we should be dealing with an accent color...
+  # If accent is NA though, then the user has specified to NOT change it
+  if (is.na(accent)) color else accent
+}
+
+
+mix_colors <- function(bg, fg, amount) {
+  if (!length(bg) || !length(fg)) return(NULL)
+  mid_color <- grDevices::colorRamp(c(bg, fg), alpha = TRUE)(amount)
```
cpsievert: Look into whether we can/should do better than a linear interpolation between these two colors

jcheng5: FWIW, for the `bs_theme_base_colors` impl, I tried switching to `space="Lab"` and it immediately complained about `alpha=TRUE`, that was enough to make me leave it as linear.

R/imageutils.R
```diff
@@ -1,17 +1,25 @@
...
   # If quartz is available, use png() (which will default to quartz).
   # Otherwise, if the Cairo package is installed, use CairoPNG().
   # Finally, if neither quartz nor Cairo, use png().
   if (capabilities("aqua")) {
     pngfun <- grDevices::png
+  } else if ((getOption('shiny.useragg') %OR% TRUE) &&
+             nchar(system.file(package = "ragg"))) {
+    pngfun <- ragg::agg_png
+    args$background <- args$background %OR% args$bg
+    args$bg <- NULL
```
cpsievert, line 13: Especially when it comes to text rendering quality, `quartz()` seems preferable to `ragg::agg_png()`. 

When it comes to `ragg` vs Cairo, the answer is [it depends](https://ragg.r-lib.org/articles/ragg_quality.html) on what you're interested in, so I think it makes sense to introduce a new `shiny.useragg` option (maybe this should default to `FALSE` for now?)

R/imageutils.R
```diff
@@ -1,17 +1,25 @@
 startPNG <- function(filename, width, height, res, ...) {
+  args <- rlang::list2(filename=filename, width=width, height=height, res=res, ...)
+
   # If quartz is available, use png() (which will default to quartz).
   # Otherwise, if the Cairo package is installed, use CairoPNG().
   # Finally, if neither quartz nor Cairo, use png().
   if (capabilities("aqua")) {
     pngfun <- grDevices::png
+  } else if ((getOption('shiny.useragg') %OR% TRUE) &&
+             nchar(system.file(package = "ragg"))) {
```
hadley, line 10: BTW since you have an rlang dep, you could use `rlang::is_installed()`.

cpsievert: Just curious, what's the rationale for using `requireNamespace()` instead of `system.file()` (in `is_installed`)?

hadley: I don't think it's a particularly strong motivation, but `requireNamespace()` also actually checks that package can be loaded.

R/imageutils.R
```diff
@@ -1,17 +1,28 @@
 startPNG <- function(filename, width, height, res, ...) {
+  args <- rlang::list2(filename=filename, width=width, height=height, res=res, ...)
+
+  if (is.null(args$bg)) {
+    args$bg <- getCurrentOutputInfo()[["bg"]] %OR% "white"
+  }
```
cpsievert, line 6: This is technically a breaking change, but maybe it's a good breaking change?

R/imageutils.R
cpsievert: Ok, thanks, we'll probably do this in batch in another PR, so I'll keep as is for now for consistency sake

R/imageutils.R
```diff
@@ -28,37 +35,46 @@ startPNG <- function(filename, width, height, res, ...) {
...
-#' @param ... Arguments to be passed through to [grDevices::png()].
+#' @param ... Arguments to be passed through to `device`.
 #'   These can be used to set the width, height, background color, etc.
+#' @param device A function that opens a png graphics device (e.g., [png()]).
 #' @export
 plotPNG <- function(func, filename=tempfile(fileext='.png'),
-                    width=400, height=400, res=72, ...) {
-  dv <- startPNG(filename, width, height, res, ...)
+                    width=400, height=400, res=72, ...,
+                    device=getShinyOption("shiny.device", shinyDevice())) {
```
cpsievert, line 76: `shiny.plotDevice`

R/imageutils.R
```diff
@@ -4,14 +4,50 @@ startPNG <- function(filename, width, height, res, ...) {
...
+      args$background <- args$bg
+    }
+    args$bg <- NULL
+  }
+
+  # Let showtext know about the resolution (for rendering custom fonts)
+  # https://github.com/yixuan/showtext/issues/33
+  if (rlang::is_installed("showtext")) {
+    showtext::showtext_opts(dpi = res)
+  }
```
cpsievert, line 48: Restore this option on exit

R/imageutils.R
```diff
@@ -4,14 +4,50 @@ startPNG <- function(filename, width, height, res, ...) {
...
     pngfun <- grDevices::png
   }
 
-  pngfun(filename=filename, width=width, height=height, res=res, ...)
+  args <- rlang::list2(filename=filename, width=width, height=height, res=res, ...)
+
+  # The order of priority for the bg color is:
+  # (1) User specified bg via `renderPlot()`
+  # (2) bg option was set via thematic
+  if (is.null(args$bg) && rlang::is_installed("thematic")) {
```
cpsievert, line 25: Add comment about how this is really only relevant to `Cairo::CairoPNG`

R/imageutils.R
```diff
@@ -4,14 +4,50 @@ startPNG <- function(filename, width, height, res, ...) {
...
+
+  # Handle both bg and background device arg
+  # https://github.com/r-lib/ragg/issues/35
+  fmls <- names(formals(pngfun))
+  if (("background" %in% fmls) && (!"bg" %in% fmls)) {
+    if (is.null(args$background)) {
+      args$background <- args$bg
+    }
+    args$bg <- NULL
+  }
```
cpsievert, line 42: Maybe scope to the ragg case?

srcjs/init_shiny.js
```diff
@@ -288,6 +288,90 @@ function initShiny() {
...
+    }
+    return bgColor;
+  }
+
+  // Compute the color property of an a tag, scoped within the element
+  function getComputedLinkColor(el) {
+    let a = document.createElement("a");
+    a.href = "/";
+    let div = document.createElement("div");
+    div.style.position = "absolute";
```
cpsievert, line 320: Can be removed

R/imageutils.R
wch, line 25: Change to a check if thematic is loaded.

R/imageutils.R
```diff
@@ -4,14 +4,50 @@ startPNG <- function(filename, width, height, res, ...) {
...
+
+  # Let showtext know about the resolution (for rendering custom fonts)
+  # https://github.com/yixuan/showtext/issues/33
+  if (rlang::is_installed("showtext")) {
```
wch, line 46: Let's try to minimize the probability of side effects, in this case, loading showtext:

- This is only needed for `png(type = "cairo")`, not ragg, or Cairo package
- Move into `drawPNG`

R/shiny.R
```diff
@@ -1337,6 +1337,26 @@ ShinySession <- R6Class(
...
+
+      fg <- paste0("output_", name, "_fg")
+      if (fg %in% cd_names()) {
+        tmp_info$fg <- htmltools::parseCssColors(self$clientData[[fg]])
+      }
+
+      accent <- paste0("output_", name, "_accent")
+      if (accent %in% cd_names()) {
+        tmp_info$accent <- htmltools::parseCssColors(self$clientData[[accent]])
+      }
```
cpsievert, line 1353: `mustWork = FALSE`

R/imageutils.R
cpsievert: Decided against this...could be useful if and when we support a general device function

srcjs/init_shiny.js
```diff
@@ -288,6 +288,89 @@ function initShiny() {
...
+      }
+    }
+
+    return font;
+  }
+
+  $('.shiny-image-output, .shiny-plot-output, .shiny-report-theme').each(function() {
+    var id = getIdFromEl(this);
+    initialValues['.clientdata_output_' + id + '_bg'] = getComputedBgColor(this);
+    initialValues['.clientdata_output_' + id + '_fg'] = window.getComputedStyle(this).getPropertyValue("color");
```
cpsievert, line 369: Could compute style once and pass to `getComputedFont()` and `getComputedBgColor()`

R/imageutils.R
```diff
@@ -4,14 +4,45 @@ startPNG <- function(filename, width, height, res, ...) {
   # Finally, if neither quartz nor Cairo, use png().
   if (capabilities("aqua")) {
     pngfun <- grDevices::png
+  } else if ((getOption('shiny.useragg') %OR% FALSE) &&
+             nchar(system.file(package = "ragg"))) {
```
wch, line 8: We should make this into a function, like `is_installed`.

R/shiny.R
```diff
@@ -1337,6 +1337,35 @@ ShinySession <- R6Class(
         }
       }
 
+      bg <- paste0("output_", name, "_bg")
+      if (bg %in% cd_names()) {
+        tmp_info$bg <- htmltools::parseCssColors(
+          self$clientData[[bg]] %OR% "",
```
wch, line 1343: A comment about the `NULL` and `NA` handling would help.

R/graph.R
```diff
@@ -1,6 +1,9 @@
 is_installed <- function(package, version) {
```
wch, line 1: `version` should default to `NULL`, then have a `NULL` check below.

wch: Also, please move to utils.R

R/shiny.R
```diff
@@ -1337,6 +1337,39 @@ ShinySession <- R6Class(
         }
       }
 
+      # parseCssColors() currently errors out if you hand it any NAs
+      # This'll make sure we're always working with a string (and if
+      # that string isn't a valid CSS color, will return NA)
+      # https://github.com/rstudio/htmltools/issues/161
+      bg <- paste0("output_", name, "_bg")
+      if (bg %in% cd_names()) {
+        tmp_info$bg <- htmltools::parseCssColors(
```
cpsievert, line 1346: Wrap in `reactive()`. Add comment about why (ability to take a partial dependency). Also. look into doing the parsing via an `InputHandler`

R/shiny.R
```diff
@@ -1319,6 +1319,10 @@ ShinySession <- R6Class(
 
       # If we don't already have width for this output info, see if it's
       # present, and if so, add it.
+
+      # Note that all the following clientData values (which are reactiveValues)
+      # are wrapped in reactive() so that users can take a dependency on particular
+      # output info (i.e., just depend on width/height, or just depend on bg, fg, etc)
```
wch, line 1325: Can you add: "If this function just returned a list of values from `self$clientData`, than anything that calls this function would take a reactive dependency on all of these values."

cpsievert: How about:

> To put it another way, if `getCurrentOutputInfo()` simply returned a list of values from self$clientData, than anything that calls `getCurrentOutputInfo()`would take a reactive dependency on all of these values.

R/imageutils.R
```diff
@@ -4,14 +4,43 @@ startPNG <- function(filename, width, height, res, ...) {
   # Finally, if neither quartz nor Cairo, use png().
   if (capabilities("aqua")) {
     pngfun <- grDevices::png
-  } else if ((getOption('shiny.usecairo') %OR% TRUE) &&
-      nchar(system.file(package = "Cairo"))) {
+  } else if ((getOption('shiny.useragg') %OR% FALSE) && is_available("ragg")) {
```
wch, line 7: A few things:

* This can be first in the if-else, so that Mac users can use ragg if they really want.
* We don't document this option since it's just for experimentation for now, and will be superseded by a different one in the future. It also should have a comment stating as much.
* We can change to `getOption('shiny.useragg', default = FALSE)` here.

R/imageutils.R
```diff
@@ -36,14 +65,16 @@ startPNG <- function(filename, width, height, res, ...) {
...
-#' quartz (via [grDevices::png()]), then [Cairo::CairoPNG()],
+#' quartz (via [grDevices::png()]), then [ragg::agg_png()], then [Cairo::CairoPNG()],
 #' and finally [grDevices::png()]. This is in order of quality of
 #' output. Notably, plain `png` output on Linux and Windows may not
 #' antialias some point shapes, resulting in poor quality output.
 #'
-#' In some cases, `Cairo()` provides output that looks worse than
-#' `png()`. To disable Cairo output for an app, use
-#' `options(shiny.usecairo=FALSE)`.
+#' In some rare cases, `ragg::agg_png()` provides output that looks worse than
```
wch, line 73: Can remove all the documentation about ragg.

R/shiny-options.R
```diff
@@ -110,6 +110,9 @@ getShinyOption <- function(name, default = NULL) {
 #' \item{shiny.usecairo (defaults to `TRUE`)}{This is used to disable graphical rendering by the
 #'   Cairo package, if it is installed. See [plotPNG()] for more
 #'   information.}
+#'  \item{shiny.useragg (defaults to `TRUE`)}{This is used to disable graphical rendering by the
```
wch, line 113: Remove this.

srcjs/init_shiny.js
```diff
@@ -288,6 +288,89 @@ function initShiny() {
...
   });
+
+  function getComputedBgColor(el) {
+    if (!el) {
+      // Top of document, can't recurse further
+      return null;
+    }
+
+    let props = window.getComputedStyle(el);
+    let bgColor = props.getPropertyValue("background-color");
```
wch, line 299: Safer to use our `getStyle()` function

srcjs/init_shiny.js
cpsievert: We did some testing and `getComputedStyle()` appears to be very fast...on the order of 30-40ms for 10,000 calls to it

Issue #2743, opened 2020-01-17
onSessionEnded callback can be executed by next running app


wch:
The following app displays the value of `x` and has a button that will stop the app.  The value of `x` gets initialized by the `onStart` function. There's an `onSessionEnded` callback which increments `x`.

We'd expect the app to display `x: 1` each time it runs.

```R
library(shiny)

x <- NULL
app <- shinyApp(
  fluidPage(
    textOutput("txt"),
    actionButton("close", "Close app")
  ),
  function(input, output) {
    output$txt <- renderText({
      paste0("x: ", x)
    })
    
    observeEvent(input$close, {
      message("close button was clicked")
      stopApp("close button")
    })

    onSessionEnded(function() {
      message("Executing onSessionEnded callback.")
      x <<- x + 1
    })
  },
  onStart = function() {
    x <<- 1
  }
)
```

When you run it like this, with one `runApp` after the other, the app displays `x: 1` on the first run, and then `x: 2` on the second. Also note the strange order that the diagnostic messages are printed out:

```R
{
  runApp(app)
  runApp(app)
}
#> Listening on http://127.0.0.1:7789
#> close button was clicked
#> 
#> Listening on http://127.0.0.1:7789
#> Executing onSessionEnded callback.
#> close button was clicked
#> [1] "close button"
#> Executing onSessionEnded callback.
```

Adding some `run_now()` calls in between makes it display `x: 1` both times.

```R
{
  runApp(app)
  for (i in 1:10) later::run_now()
  runApp(app)
}
#> Listening on http://127.0.0.1:7789
#> close button was clicked
#> Executing onSessionEnded callback.
#> 
#> Listening on http://127.0.0.1:7789
#> close button was clicked
#> [1] "close button"
#> Executing onSessionEnded callback.
```

The problem seems to be that, without the `run_now()` calls after the first app, the `onSessionEnded` callback can bleed into the next app.

cc @schloerke

PR #2749, opened 2020-01-29
Fix for close ShinySession on on.exit

galachad:
The onSessionEnd is probably triggered by `later` loop and is never trigger on `SIGINT` signal. There is a difference between interactive and not interactive R process.  In not interactive R process after global onStop callback, the R process is terminated immediately.

The minimal example:
```R
library(shiny)

ui <- fluidPage(

)

  onSessionEnded(function() {
    print("close ShinySession session")
  })
}

onStop(function() {
  # Custom code that was used before this PR.
  lapply(shiny:::appsByToken$keys(), function(session) {
    session <- shiny:::appsByToken$get(session)
    if (isFALSE(session$closed)) session$wsClosed()
  })
})

shinyApp(ui, server)
```

The gif example shows the difference.
Peek 2020-01-09 10-11

If I can do some more in this case or you need more clarification, please let me know.

### DIFF
--- a/R/shiny.R
+++ b/R/shiny.R
@@ -2166,6 +2166,13 @@ flushPendingSessions <- function() {
 }
 
 .globals$onStopCallbacks <- Callbacks$new()
+# Close ShinySession on.exit event (SIGINT signal)
+.globals$onStopCallbacks$register(function() {
+  lapply(appsByToken$keys(), function(session) {
+    session <- appsByToken$get(session)
+    if (isFALSE(session$closed)) session$wsClosed()
+  })
+})
 
 #' Run code after an application or session ends
 #'


R/shiny.R
```diff
@@ -2166,6 +2166,13 @@ flushPendingSessions <- function() {
 }
 
 .globals$onStopCallbacks <- Callbacks$new()
+# Close ShinySession on.exit event (SIGINT signal)
+.globals$onStopCallbacks$register(function() {
+  lapply(appsByToken$keys(), function(session) {
+    session <- appsByToken$get(session)
+    if (isFALSE(session$closed)) session$wsClosed()
```
shrektan: isFALSE() is a function introduced by R3.5.0 so it's better to use `identical(session$closed, FALSE)`

galachad: that's true, thanks for catching this.

wch:
I think that #2743 may have the same root cause as the issue you're describing.

In your command-line script, if, instead of using this PR, you add this after the call to `runApp()`, does the `onSessionEnded` callback execute properly?

```R
for (i in 1:10) later::run_now()
```

galachad:
@wch I totally agree, it's even better to trigger later loop on stop callback.
It's not really solved #2743 issue because the second runApp is executed before later loop. In my opinion it will be good to add `run_now` at the beginning of runApp.
```
runApp(app)
later::run_now()
runApp(app)
```

galachad:
I was to fast, I check twice and it does not execute callback properly.

PR #2753, opened 2020-02-06
Add {min/max}viewmode parameters to dateInput and dateRangeInput

### LINKED ISSUE #1365: dateInput minViewMode option

davidchall:
Fixes #1365.

Added `minviewmode` and `maxviewmode` as new parameters to `dateInput()` and `dateRangeInput()`. This exposes the underlying arguments of [Bootstrap Datepicker](https://bootstrap-datepicker.readthedocs.io/en/latest/options.html#maxviewmode). The `minviewmode` parameter makes it possible to restrict date selection to a month, a year, a decade, or a century.


Cormochamelion:
I would be very happy to have the ability to only select months/years. What is currently preventing this from being merged? Only the CLA?

PR #2758, opened 2020-02-13
Fix broken timer tests

schloerke:
Adjusted MS time value by a factor of 1000 to compare to seconds.

Also, for some reason...

``` r
testthat::expect_lte(as.numeric(Sys.time()), Sys.time())
#> Error in `-.POSIXt`(act$val, exp$val): can only subtract from "POSIXt" objects
```

<sup>Created on 2020-02-13 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

So switching test style to 

``` r
testthat::expect_true(as.numeric(Sys.time()) <= Sys.time())
```

<sup>Created on 2020-02-13 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

### DIFF
--- a/tests/testthat/test-timer.R
+++ b/tests/testthat/test-timer.R
@@ -72,11 +72,11 @@ test_that("mockableTimer works", {
 })
 
 test_that("getDomainTimeMs works", {
-  start <- Sys.time()
+  start <- as.numeric(Sys.time()) * 1000
   t1 <- getDomainTimeMs(NULL)
   t2 <- getDomainTimeMs(list())
   t3 <- getDomainTimeMs(list(.now = function(){456}))
-  end <- Sys.time()
+  end <- as.numeric(Sys.time()) * 1000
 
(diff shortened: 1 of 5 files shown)


tests/testthat/test-timer.R
```diff
@@ -78,10 +78,10 @@ test_that("getDomainTimeMs works", {
   t3 <- getDomainTimeMs(list(.now = function(){456}))
   end <- Sys.time()
```
wch: I think it would be clearer with:

```R
as.numeric(Sys.time()) * 1000
```

The `as.numeric` here makes it clear what the value is that will be compared later -- this way, `start`, `end`, `t1`, etc. all have the same type.

wch: Sorry, I wasn't clear: I to store all the times as a number, as in `start <- as.numeric(Sys.time()) * 1000`

PR #2764, opened 2020-02-19
Tweak runTests() output format

alandipert:
Currently `runTests()` returns a structure with a member that varies in type between a result value, NA, or an error object.

This PR normalizes the format of the return value of `runTests()` at the cost of some redundancy. The goal is to simplify consumption by downstream tools interested in test outcomes.

## Dev TODO

- [x] Add test(s)
- [x] Improve documentation
- [x] Add to NEWS.md


alandipert:
CI is currently failing with the following `R CMD check` NOTE:
```
* checking installed package size ... NOTE
  installed size is 11.1Mb
  sub-directories of 1Mb or more:
    R     1.6Mb
    www   8.3Mb
```

schloerke:
That is a standard/consistent `NOTE` for shiny

R/test.R
```diff
@@ -1,3 +1,27 @@
...
+#' @param file Name of the test runner file, a character vector of length 1.
+#' @param pass Whether or not the test passed, a logical vector of length 1.
+#' @param result Value (wrapped in a list) obtained by evaluating `file` or `NA`
+#'   if no value was obtained, such as with `shinytest`.
+#' @param error Error, if any, (and wrapped in a list) that was signaled during
+#'   evaluation of `file`.
+#'
+#' @return A 1-row data frame representing a single test run. `result` and
+#'   `error` are "list columns", or columns that may contain list elements.
+result_row <- function(file, pass, result, error) {
```
schloerke: If this function is not exported for public use, let's add a `#' @noRd` tag to prevent the man file from being produced.

R/test.R
```diff
@@ -19,6 +43,16 @@ isShinyTest <- function(text){
...
+#' @return A data frame classed with the supplemental class `"shinytestrun"`.
+#'   The data frame has the following columns:
+#'
+#' | **Name** | **Type** | **Meaning** |
+#' | :-- | :-- | :-- |
+#' | `file` | `character(1)` | File name of the runner script in `tests/` that was sourced. |
+#' | `pass` | `logical(1)` | Whether or not the runner script signaled an error when sourced. |
+#' | `result` | any or `NA` | The return value of the runner, or `NA` if `pass == FALSE`. |
+#' | `error` | any or `NA` | The error signaled by the runner, or `NA` if `pass == TRUE`. |
+#'
```
schloerke: Cool!

NEWS.md
```diff
@@ -7,6 +7,10 @@ shiny 1.4.0.9001
 
 ### New features
 
+* `runTests()` is a new function that behaves much like R CMD check. `runTests()` invokes all of the top-level R files in the tests/ directory inside an application, in that application's environment. ([#2585](https://github.com/rstudio/shiny/pull/2585))
```
wch: I think it would be good to also have a link to this PR.

R/test.R
```diff
@@ -64,16 +99,10 @@ runTests <- function(appDir=".", filter=NULL){
...
-    res <- list()
-    lapply(sares$results, function(r){
-      e <- NA_character_
-      if (!r$pass){
-        e <- simpleError("Unknown shinytest error")
-      }
-      res[[r$name]] <<- e
-    })
-    return(structure(list(result=all(is.na(res)), files=res), class="shinytestrun"))
+    return(do.call(rbind, lapply(shinytest::testApp(appDir)[["results"]], function(r) {
```
wch: Instead of calling `shinytest::testApp`, can we make this just use the tests/shinytest.R script? The idea is that we wouldn't hard-code each type of testing, but let the user provide scripts to run each type of test.
For example: https://github.com/rstudio/shinycoreci-apps/blob/master/apps/041-dynamic-ui/tests/shinytest.R

This may

alandipert: We surely can, are you suggesting that we don't attempt to detect or special-case shinytest at all? If so, 👍

R/test.R
```diff
@@ -64,16 +99,10 @@ runTests <- function(appDir=".", filter=NULL){
       warning("You've disabled `shiny.autoload.r` via an option but this is not passed through to shinytest. Consider using a _disable_autoload.R file as described at https://rstd.io/shiny-autoload")
```
wch: It would help if at line 90 (`if (all(isST))`) there was a comment that explained how we get into that code path -- that is, when the tests/ subdirectory uses the old structure, where it contains shinytest scripts.

R/test.R
alandipert: Discussed in person with @wch and we realized this branch was for dealing with old shinytest behavior that populated tests/ with shinytest-specific code.

R/test.R
alandipert: Turns out this was already illuminated, just higher up in the roxygen. I added a pointer to the code.

PR #2766, opened 2020-02-22
Support shiny.autoreload even when there are errors

jcheng5:
Previously, `options(shiny.autoreload = TRUE)` had limited usefulness because any error in the UI would cause autoreload to stop working, or any error which caused a session disconnected. This is because the autoreload functionality was built around Shiny sessions, so a live and working session had to be active to initiate an autoreload.

This PR solves the problem by opening a dedicated websocket connection for file monitoring, and using this connection both for successful UI loads and for error pages.

Also, error pages look quite different, they need to be HTML now instead of the previous text/plain httpuv default.

### DIFF
--- a/R/middleware.R
+++ b/R/middleware.R
@@ -311,16 +311,32 @@ HandlerManager <- R6Class("HandlerManager",
         },
         call = .httpServer(
           function (req) {
-            withCallingHandlers(withLogErrors(handlers$invoke(req)),
-              error = function(cond) {
-                sanitizeErrors <- getOption('shiny.sanitize.errors', FALSE)
-                if (inherits(cond, 'shiny.custom.error') || !sanitizeErrors) {
-                  stop(cond$message, call. = FALSE)
-                } else {
-                  stop(paste("An error has occurred. Check your logs or",
-                             "contact the app author for clarification."),
-                       call. = FALSE)
+            hybrid_chain(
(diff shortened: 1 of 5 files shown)


R/middleware.R
```diff
@@ -311,16 +311,32 @@ HandlerManager <- R6Class("HandlerManager",
...
+                    content_type = "text/html",
+                    content = as.character(htmltools::htmlTemplate(
+                      system.file("template", "error.html", package = "shiny"),
+                      message = conditionMessage(err)
+                    ))
+                  )
                 }
+              ),
+              function(resp) {
+                maybeInjectAutoreload(resp)
```
wch: Will this only inject the autoreload stuff into the index page for an app? I just want to make sure it doesn't get injected into any other pages that happen to get served up.

jcheng5: I think it might be for every page with `</head>`. I was thinking it couldn't just be `/` because of Shiny Rmd. I don't think adding the script everywhere is harmful though? (Keep in mind this is just for development use)

jcheng5: Actually I guess we could move to this wherever we insert `<script src="shiny.js">` or whatever?

jcheng5: Wait, sorry, the whole reason for putting it here is it needs to apply to error message pages too.

Issue #2297, opened 2019-01-09
Bookmarking: unnecessarily unrecoverable error when query string contains more params than expected


daattali:
If a bookmarkable app is opened with a bookmarkable URL that contains an extra parameter that the bookmark system doesn't know about, then **none** of the inputs get set to their values and the following error message appears `Error in RestoreContext initialization: Failed to parse URL parameter "value"`.

I think Shiny should be a little more forgiving when this happens, and still set all the values that it's able to. Whether or not to notify the user about the extra values is another question as well (I think it should be quiet).

This can happen from many situations, for example if an app is updated and an input gets removed, if the app developer or end user change the URL, or if a URL is shared using a third party application that automatically adds a parameter (for example, links in marketing emails or from social media often get a parameter added). None of these cases should cause the bookmark engine to fail.

Example:

```
library(shiny)

ui <- function(req) {
  fluidPage(
    bookmarkButton(),
    textInput("text", "text", "text")
  )
}

server <- function(input, output, session) {
  
}

shinyApp(ui, server, enableBookmarking = "url", options = list(port = 3900))
```

Run this app, and navigate to `http://127.0.0.1:3900/?_inputs_&text="texttest"`. This works. Now navigate to `http://127.0.0.1:3900/?_inputs_&text="texttest"&foo=bar`. The bookmarking doesn't work and an error is shown.

Potentially related to #1650

alandipert:
I like this idea.

daattali:
I would make the argument this isn't "low" priority because this issue causes bookmarking to simply *not work at all* if an app is ever shared using a medium that adds its own parameters. For example, I have an app that I sent to my friend on facebook using Facebook Messenger (the standard facebook chat). Clicking on the link in my message, Facebook appends `&fbclid=XXXXXXXXXXXXXXX` to the URL, which makes my app's bookmarking useless.

alandipert:
I agree. I think it even qualifies as a bug.

daattali:
Update: my initial assessment of the exact bug was wrong.

Shiny won't fail if I add more parameters, as long as those parameters get parsed properly by shiny. So if I add `&foo="bar"` that would be fine, but adding `&foo=bar` causes the whole bookmark system to fail. 

The good news is that at least this means bookmarks are not going to break just I removed inputs from my UI, but it does still mean that any external query params will break it.

relund:
[...]

PR #2770, opened 2020-02-28
fix bookmarking bug #2297: don't break all bookmarking system if some URL params don't parse correctly

daattali:
This should certainly require some more testing, but it passed some basic manual tests I've tried.

The idea is that for any URL query params that don't parse to JSON, instead of erroring out, simply ignore them.

### DIFF
--- a/R/bookmark-state.R
+++ b/R/bookmark-state.R
@@ -346,15 +346,21 @@ RestoreContext <- R6Class("RestoreContext",
       values <- parseQueryString(valueStr, nested = TRUE)
 
       valuesFromJSON <- function(vals) {
-        mapply(names(vals), vals, SIMPLIFY = FALSE,
+        valsParsed <- mapply(names(vals), vals, SIMPLIFY = FALSE,
           FUN = function(name, value) {
             tryCatch(
               safeFromJSON(value),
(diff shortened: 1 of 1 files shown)


R/bookmark-state.R
```diff
@@ -346,16 +346,22 @@ RestoreContext <- R6Class("RestoreContext",
...
-        mapply(names(vals), vals, SIMPLIFY = FALSE,
+        valsParsed <- mapply(names(vals), vals, SIMPLIFY = FALSE,
           FUN = function(name, value) {
             tryCatch(
               safeFromJSON(value),
               error = function(e) {
-                stop("Failed to parse URL parameter \"", name, "\"")
+                # If a value did not parse from JSON successfully, return NA
+                # because that's a unique value that can't be obtained from JSON
+                NA
```
wch: I think you'll need a different sentinel value, since it is possible to get `NA`:
 
```R
safeFromJSON('[null]')
#> [1] NA
```

daattali: @wch I suppose NA_character_ would be unachievable with jsonlite::fromJSON. But perhaps this is just too hacky and not the correct approach.

PR #2772, opened 2020-02-28
Small changes to testModule()/testServer() semantics

alandipert:
This PR makes a few changes to `.testModule()`, the private function that powers `testModule()` and `testServer()`. The changes should make the behavior of `testModule()` and `testServer()` more intuitive.

1. **`.testModule()` returns a meaningful value.** Previously, `.testModule()` returned a meaningless value derived from implementation detail. Now, `.testModule()` returns the value the test expression evaluates to.
1. **Test expression is masked by the module environment**. Previously, the test expression was evaluated in the dynamic environment obtained from inside the module function after invoking it and stored as `session$env`. The parent of this environment included the lexical environment in which the module function was defined. Therefore, `session$env` and the environment in which `testModule()`/`testServer()` were called were potentially disjoint. Now, `.testModule()` *masks* the dynamic environment surrounding any `testModule()/testServer()` call with `session$env`. This ensures that definitions in the test environment are available within the test expression, regardless of where the module was defined. While this behavior is generally more intuitive, it gives rise to two cases that may require mitigation:
    1. **Module lexical environment required for testing** When the full lexical environment of the module is required it can be referred to explicitly in test expressions with `session$env`. 
    1. **Variables in module shadow testing dynamic environment** When the `session$env` mask shadows definitions in the dynamic testing environment, tidy eval unquote (`!!`/`!!!`) can be used to splice shadowed definitions into the test expression. A contrived example of this would be if for some reason a module defined an `expect_equal` variable. This could be circumvented in tests by referring to `!!expect_equal` in the test expression.

### DIFF
--- a/R/test-module.R
+++ b/R/test-module.R
@@ -55,53 +55,60 @@
 #' }, !!multiplier_arg_name := 2, !!!more_args)
 #' @export
 testModule <- function(module, expr, ...) {
-  expr <- substitute(expr)
-  .testModule(module, expr, ...)
+  .testModule(
+    module,
+    quosure = rlang::enquo(expr),
+    dots = rlang::list2(...),
+    env = rlang::caller_env()
+  )
 }
 
 #' @noRd
 #' @importFrom withr with_options
-.testModule <- function(module, expr, ...) {
-  # Capture the environment from the module
-  # Inserts `session$env <- environment()` at the top of the function
+.testModule <- function(module, quosure, dots, env) {
+  # Modify the module function locally by inserting `session$env <-
+  # environment()` at the beginning of its body. The dynamic environment of the
+  # module function is saved so that it may be referenced after the module
+  # function has returned. The saved dynamic environment is the basis for the
+  # `data` argument of tidy_eval() when used below to evaluate `quosure`, the
+  # test code expression.
   body(module) <- rlang::expr({
-    session$env <- environment()
+    session$env <- base::environment()
     !!!body(module)
   })
 
-  # Create a mock session
   session <- MockShinySession$new()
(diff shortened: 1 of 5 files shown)


R/test-module.R
```diff
@@ -55,53 +56,58 @@
...
+  # was saved when the module function was invoked. flush is not needed before
+  # entering the loop because the first expr executed is `{`.
   isolate({
     withReactiveDomain(
       session,
       withr::with_options(list(`shiny.allowoutputreads`=TRUE), {
-        eval(expr, new.env(parent=session$env))
+        rlang::eval_tidy(
+          quosure,
+          data = rlang::as_data_mask(as.list(session$env)),
```
lionel-: My only concern with the masking approach is that it is only viable when the mask is shallow and distinct from the lexical envs. This is the case here because (IIUC) the mask is a clone of the execution env and doesn't include any parent environment of the module.

Are there cases where a module is a closure, i.e. where it depends on variables defined in one of the parents of the module function? In that case, these closure modules won't work properly with this approach.

alandipert: I think that case would be unusual, but if necessary users could access enclosed variables in parent scopes with e.g. `get("foo", session$env)`

tests/testthat/test-test-module.R
```diff
@@ -556,10 +583,43 @@ test_that("testServer works", {
...
 })
 
+test_that("testModule allows lexical environment access through session$env", {
+  m <- local({
+    a_var <- 123
+    function(input, output, session) {
+      b_var <- 321
+    }
+  })
+  expect_error(a_var, "not found")
```
wch: I think it's not safe to use the `"not found"` pattern, because the error message in other locales will be different.

PR #2773, opened 2020-03-03
Add moduleServer function

wch:
This adds `moduleServer`, which provides for a nicer interface for modules.


jcheng5:
Does this have any implications for `testModule`?

wch:
Yup, we’re reworking testModule to work with this.

PR #2775, opened 2020-03-03
Updates for new grid in r-devel

cpsievert:
Workaround for https://github.com/wch/r-source/commit/f9b8a42eb2052790cfef3119a8598fe59c5f730c#diff-f9381d3501833f2cbd164596fdd3dc2eR448-R449


R/render-plot.R
```diff
@@ -889,6 +889,15 @@ find_panel_info_non_api <- function(b, ggplot_format) {
   })
 }
 
+# Use public API for getting the unit's type (grid::unitType(), added in R 4.0)
+# https://github.com/wch/r-source/blob/f9b8a42/src/library/grid/R/unit.R#L179
+getUnitType <- function(u) {
+  if (getRversion() >= "4.0.0") {
```
wch: Safer to not use a version check, and check for the existence of `unitType` directly. Also, can use `getFromNamespace`.

PR #2783, opened 2020-03-05
moduleServer testing

alandipert:


R/modules.R
```diff
@@ -121,23 +121,40 @@ createSessionProxy <- function(parentSession, ...) {
 #'
 #' @export
 moduleServer <- function(id, module, session = getDefaultReactiveDomain()) {
+  if (inherits(session, c("MockShinySession"))) session$isModuleServer <- TRUE
```
wch: Instead of adding `session$isModuleServer` here and checking it in `callModule`, I think it would be clearer if the modifications to `module` were done in this function. You could skip the call to `callModule` in the case where `session` is a `MockShinySession`.

I also still feel a weird about embedding this test-related code into the `moduleServer` or `callModule` functions, but I don't see a better way to do it at the moment.

R/modules.R
```diff
@@ -121,23 +121,40 @@ createSessionProxy <- function(parentSession, ...) {
...
-    module(childScope$input, childScope$output, childScope, ...)
-  })
+  if (inherits(session, "MockShinySession") && session$isModuleServer) {
+    # If the module is under test *and* was called by moduleServer(), modify the
+    # module function locally by inserting the equivalent of `session$env <-
+    # environment()` at the beginning of its body. A similar operation is
+    # performed by .testModule() if the module is *not* called through
+    # moduleServer() but is under test. See .testModule() for details.
+    body(module) <- rlang::expr({
+      base::assign("env", base::environment(), envir = !!session)
```
wch: Can this line just be the same as the one down below? The other one is a more idiomatic way of doing it.

R/test-module.R
```diff
@@ -64,33 +64,55 @@ testModule <- function(module, expr, ...) {
   )
 }
 
+isOldModule <- function(func) {
```
wch: It would be helpful to have a comment explaining the purpose of this function (and how it works).

R/modules.R
```diff
@@ -121,23 +121,40 @@ createSessionProxy <- function(parentSession, ...) {
...
+      !!!body(module)
+    })
+    isolate(
```
wch: Is this `isolate(withReactiveDomain(..., with_options` stuff necessary? That stuff is also present in test-module.R, line 106.

R/test-module.R
```diff
@@ -64,33 +64,55 @@ testModule <- function(module, expr, ...) {
...
+  stopifnot(is.function(func))
+  required <- c("input", "output", "session")
+  declared <- names(formals(func))
+  setequal(required, intersect(required, declared))
```
wch: A bit more idiomatic to do something like:

```R
all(required %in% declared)
```

alandipert:
Continued by #2807

PR #2793, opened 2020-03-12
Add class `'function'` to `reactive` and `reactiveVal` objects

### LINKED ISSUE #2787: class of reactive should include function
Because a reactive **is a** function 

``` r
library(shiny)
class(reactive(1))
#> [1] "reactiveExpr" "reactive"
```

<sup>Created on 2020-03-11 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

schloerke:
Fixes #2787


wch:
Can you add a NEWS item?

PR #2795, opened 2020-03-16
Use roxygen2 reexport setup for htmltools and fastmap

schloerke:
Related: https://github.com/r-lib/pkgdown/issues/1251

When submitting to CRAN, there is often an argument mismatch between the submitted documentation of shiny and the real, underlying functions.  However, we build our shiny documentation using pkgdown, which requires fully defined man files to exist.

To get around this, we will submit a plain/simple "These files are reexported by shiny", internal man file.  Internal man files are not required to have argument definitions and each function will still have some man file that points to the original man location.

**Example:** `dplyr` [reexports `tibble::data_frame`](https://github.com/tidyverse/dplyr/blob/ac2e33d8759ba20cfbb71346ec86ef6ee0953d38/R/reexport-tibble.r#L3-L5) which produces a [`reexports.Rd` man file](https://github.com/tidyverse/dplyr/blob/ac2e33d8759ba20cfbb71346ec86ef6ee0953d38/man/reexports.Rd#L7).

So that `shiny` and `shiny-dev-center` know of the same information, all reexports are put in `./inst/_reexports.json`.

The handling of the final documentation will be done `shiny-dev-center` where all missing `man` files will be collected in the `./man` folder and `./man/reexports.Rd` will be removed.  This allows for shiny to have full documentation, which can be reproduced at any time. (Ex: `htmltools` or `fastmap` updates)


inst/_pkgdown.yml
```diff
@@ -193,7 +193,7 @@ reference:
       - onStop
       - diskCache
       - memoryCache
-      - reexports
+      - key_missing
```
wch: Shouldn't `key_missing` be in  `reexports`?

And should `reexports` still be here, or is it not necessary because of its functions have `@keywords internal` now?

schloerke: This is for the website only.  For the website, we should link to `fastmap::key_missing`.

(`reexports` has been removed)

schloerke: Before this PR, `reexports` was just `fastmap::key_missing`

tests/testthat/test-pkgdown.R
```diff
@@ -15,20 +15,31 @@ get_indexed <- function(f = system.file('_pkgdown.yml', package = 'shiny')) {
 }
 
 test_that("All man pages have an entry in _pkgdown.yml", {
-  skip_on_cran()
+  # skip_on_cran()
```
wch: I think we can still skip on CRAN -- this will still be tested in CI, which we do all the time. Also, if we skip on CRAN and only test when using `load_all()` (though I'm not sure if that's possible), then we don't need to put `_reexports.json` in `inst/`.

schloerke: put `_reexports.json` in `./tools`?

wch: Sounds good to me.  I'm not sure if there's a `skip_` function that means "skip if package loaded without `load_all`", though. You might have to get creative for this. See how `pkgload::dev_meta` is implemented to see how to detect if a package was loaded with devtools/pkgload.

schloerke: 👍

schloerke: Moved the test to the `./tools/documentation` folder.  This script is then called within `checkCurrentDocs.sh`. Like before, the test will fail if the pkgdown information is not correct.

.travis.yml
```diff
@@ -6,7 +6,7 @@ matrix:
       r_packages:
         - devtools
         - rprojroot
-      script: ./tools/checkDocsCurrent.sh
+      script: ./tools/documentation/checkDocsCurrent.sh
```
wch: Can we run this when running `devtools::test()` locally?

schloerke:
* Added a `testthat` test. 
* Added a `skip_on_cran()`. 
* Fixed GitHub API issue caused by missing PAT.

The only failures are travis R3.2 and R3.3 (Rcpp related).

Merging...

Issue #2730, opened 2019-12-20
actionButton should have class argument


hadley:
To match `downloadButton()`. (It would automatically paste `btn` to beginning of the class hierarchy)

daattali:
I like the argument for consistency. 

I would argue that `downloadButton` should **dropp** the class argument instead of `actionButton` **adding** it. `class` is a standard attribute for these tags anyway, even without making it an explicit argument, we would still be able to call these functions in exactly the same way and I don't think it'd have any effect. None of the other possible attributes get their own parameters, I think `class` should be no different.

nteetor:
I would second the removal of `class` from `downloadButton` as opposed to adding a formal argument to `actionButton`. `actionButton` already informally(?) supports additional classes through `...`. `downloadButton`, too, passes along `...` to the underlying builder function, the `class` argument is redundant. From the blame history, `class` was always part of `downloadButton`, `...` was added later with `class` left as a formal argument.

hadley:
Removing `class` would mean that any attempt to set the class would override the default, which I don’t think is the correct behaviour as it would require the user to know the full class specification in order to do any customisation. 

`class` is the second most important HTML attribute (after `id` which is already an argument) and I think it would be appropriate to have it available and documented for every input and output control.

daattali:
I'm not able to reach a computer and I very well may be wrong, but I think it might still keep the default classes. The same way that action buttons dont have a class argument right now and setting a class keeps the default ones

nteetor:
I agree we couldn't expect a user to know or specify the default button classes. Fortunately, the htmltools builder functions concatenate duplicate attributes.

``` r
library(htmltools)

btn <- tags$button(
  class = "btn btn-default",
  class = "btn-custom"
)

print(btn)
#> <button class="btn btn-default btn-custom"></button>
```

<sup>Created on 2019-12-27 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

This feature could be used to ensure the default classes are included even if the formal `class` argument is dropped. `actionButton` currently does this. There is no `class` argument, but if one specifies a custom class the default classes are preserved and are ahead of the custom classes.

``` r
library(shiny)

ab <- actionButton(
  inputId = "count",
  label = "Label",
  class = "btn-custom"
)

print(ab)
#> <button class="btn btn-default action-button btn-custom" id="count" type="button">Label</button>
```

[...]

PR #2806, opened 2020-03-27
Update actionButton example

schloerke:
Fixes https://github.com/rstudio/shiny/issues/2730 (again)

cc @daattali


daattali:
Awesome, I think showing people that these classes exist and can be added at no extra cost is going to help many. 

Perhaps consider changing "Information Button" to "Information Link" for consistency :)

PR #2807, opened 2020-03-31
moduleServer/testServer overhaul

alandipert:
This PR aligns the feature formerly known as "integration testing" with the new `moduleServer()` function and usage patterns.

## Changes

Previously, the integration testing feature consisted of two entrypoints: `testModule()` and `testServer()`. Now, there's only one entrypoint, `testServer()`. `testServer()` can be used to test either the server functions in established Shiny apps, or module functions.

In addition to the consolidation of these functions, there is a slight semantic change corresponding to the change in definition of a Shiny module. Previously, the environment captured from within module functions was used to mask the surrounding dynamic environment of the `testServer()`. The combination of the module's dynamic environment mask and the environment surrounding the call to `testServer()` defined the environment in which the `expr` argument to `testServer()` was evaluated.

Now, because "new-style" modules involve at least two functions -- and two environments of interest, with respect to testing -- these two function environments are combined, with the inner-most environment taking precedence. The new semantics are demonstrated in `tests/testthat/test-test-server-scope.R`.


R/test-server.R
```diff
@@ -110,39 +110,86 @@ testModule <- function(module, expr, ...) {
...
-  # Add `session` argument if not present
-  fn_formals <- formals(server)
-  if (! "session" %in% names(fn_formals)) {
-    fn_formals$session <- bquote()
-    formals(server) <- fn_formals
+  if (is.character(app) && length(app) == 1) {
+    server <- shinyAppDir(app)$serverFuncSource()
+    message("Testing application found in: ", app)
+    fn_formals <- formals(server)
+    if (! "session" %in% names(fn_formals)) {
```
wch: Instead of injecting `session`, we should just error and tell people that if they want to test, they need to have `session`. Also include that in docs.

R/test-server.R
```diff
@@ -0,0 +1,138 @@
...
+}
+
+#' @noRd
+isModuleServer <- function(x) {
+  is.function(x) && names(formals(x))[1] == "id"
+}
+
+#' @noRd
+coercableToAppObj <- function(x) {
+  !is.null(getS3method("as.shiny.appobj", class(x), optional = TRUE))
```
alandipert, line 36: 🤔 

There's probably a better way to do this

R/test-server.R
```diff
@@ -0,0 +1,138 @@
...
+#'   # !!/:= and !!! from rlang are used below to splice computed arguments
+#'   # into the testModule() argument list.
+#' }, !!multiplier_arg_name := 2, !!!more_args)
+#' @export
+testServer <- function(app, expr, ...) {
+
+  session <- MockShinySession$new()
+  on.exit(if (!session$isClosed()) session$close())
+
+  if (coercableToAppObj(app)) {
```
alandipert, line 103: Consider inverting this so most specific check is first. Then appObj coercion fails naturally if app isn't coercable.

tests/testthat/test-test-server-app.R
```diff
@@ -0,0 +1,31 @@
...
+  testServer(test_path("..", "test-modules", "server_r"), {
+    session$setInputs(dist="norm", n=5)
+    expect_length(d(), 5)
+
+    session$setInputs(dist="unif", n=6)
+    expect_length(d(), 6)
+  })
+})
+
+test_that("testServer works when referencing external globals", {
```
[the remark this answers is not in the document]
alandipert: Yes, although if `testthat` is involved, care must be taken to override the `env` parameter of e.g. `test_dir()` with `environment()`.

PR #2812, opened 2020-04-08
Test server small updates

alandipert:
This addresses some feedback left unaddressed in a previous `testServer()` PR.


R/test-server.R
```diff
@@ -1,13 +1,31 @@
...
+# resolved in the environment in which testServer() is called.
+#
+# env is cloned because rlang::new_data_mask() mutates the parent of its `top`
+# argument.
+#' @importFrom rlang env_clone
+buildMask <- function(env) {
+  if (identical(parent.env(env), emptyenv()))
+    stop("env must have a non-empty parent")
+  clone <- env_clone(env, env_clone(parent.env(env), emptyenv()))
+  rlang::new_data_mask(clone, parent.env(clone))
```
wch, line 28: Can we skip the creation of `clone` and just use `new_data_mask(env, parent.env(env))`?

alandipert: No, it's important we pass clones to `new_data_mask` because it reserves the right to modify its arguments.

PR #2813, opened 2020-04-08
Test server assign env once

alandipert:
This fixes a bug demonstrated by https://github.com/rstudio/shiny/tree/jeff-skeleton/inst/examples/12_counter.

The bug arises when modules are used within a server function under test. Each module invocation overwrites `MockShinySession$env`, and so by the time the test expression runs, `session$env` doesn't correspond to the environment captured in the server function. It instead corresponds to the environment captured by the last module function that was invoked.

This PR ensures that only the first write to `session$env` via `MockShinySession$setEnv()` has an effect and documents why.

It also documents why last-one-wins is the desired behavior of `MockShinySession$setReturned()`, which was not modified.

### DIFF
--- a/R/mock-session.R
+++ b/R/mock-session.R
@@ -402,12 +402,18 @@ MockShinySession <- R6Class(
         setInputs = function(...) do.call(self$setInputs, mapNames(ns, ...))
       )
     },
-    #' @description Set the environment associated with a testServer() call.
+    #' @description Set the environment associated with a testServer() call, but
+    #'   only if it has not previously been set. This ensures that only the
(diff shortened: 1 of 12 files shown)


wch:
It would be good to have another test that uses an inline shiny app object with modules, in addition to the ones we have here that use an app on disk. 

```R
mymoduleServer <- ....

app <- shinyApp(
  ui = NULL,
  server = function() {
    mymoduleServer("foo")
  }
)

testServer(app, {
  ...
})
```

This would provide additional coverage for modules in an app, and it will also give us a test for app objects (as opposed to apps on disk) -- I don't think we have any of those at this point.

Issue #2680, opened 2019-10-23
tabsetPanel(type = "hidden")


hadley:
It would be useful to have a tabset variant where the user level controls for switching between tabs are hidden, and they can only be changed via `updateTabsetPanel()`. I think that this would facilitate a useful pattern for dynamic UI, where instead of having to use `renderUI()` or `insertUI()`/`removeUI()` you'd instead reveal different UI by changing the active tab.

For example, you could imagine implementing a wizard by having next and previous action buttons that adjusted the active tab.

cpsievert:
FWIW, this is already possible with a bit of css:

```css
#inputId { display: none; }
```

For example,

```r
ui <- fluidPage(
  tags$style("#inTabset { display:none; }"), 
  sidebarLayout(
    sidebarPanel(
      sliderInput("controller", "Controller", 1, 3, 1)
    ),
    mainPanel(
      tabsetPanel(id = "inTabset",
                  tabPanel(title = "Panel 1", value = "panel1", "Panel 1 content"),
                  tabPanel(title = "Panel 2", value = "panel2", "Panel 2 content"),
                  tabPanel(title = "Panel 3", value = "panel3", "Panel 3 content")
      )
    )
  )
)

server <- function(input, output, session) {
  observeEvent(input$controller, {
     updateTabsetPanel(session, "inTabset", selected = paste0("panel", input$controller))
  })
}

shinyApp(ui, server)
```

hadley:
Yeah, I think it's worth baking in though.

hadley:
Thanks all!

PR #2814, opened 2020-04-09
Add `type = "hidden"` to `tabsetPanel`

schloerke:
Fixes https://github.com/rstudio/shiny/issues/2680

Removes the titles at the top of a `tabsetPanel`.  To change the tab, `updateTabsetPanel()` **must** be called.

This allows for quick switching between different UI sets.

* Added function `tabPanelBody()` which wraps `tabPanel(title = NULL)`
* Added option `tabsetPanel(type = "hidden")`
* Added example for `tabsetPanel(type = "hidden")`
* Added `!important` on `display: none` to override `bs4`.
* [ ] Is `tabPanelBody()` a good name?

cc @hadley @cpsievert


R/bootstrap.R
```diff
@@ -510,8 +517,10 @@ tabPanel <- function(title, ..., value = title, icon = NULL) {
 #' @param selected The `value` (or, if none was supplied, the `title`)
 #'   of the tab that should be selected by default. If `NULL`, the first
 #'   tab will be selected.
-#' @param type Use "tabs" for the standard look; Use "pills" for a more plain
-#'   look where tabs are selected using a background fill color.
+#' @param type Use `"tabs"` for the standard look; Use `"pills"` for a more plain
```
hadley, line 520: I think this would be slightly better as a bulleted list.

wch:
@cpsievert Does this approach with the CSS classes work with bootstrap 4?

R/bootstrap.R
```diff
@@ -510,8 +517,10 @@ tabPanel <- function(title, ..., value = title, icon = NULL) {
...
+#'   look where tabs are selected using a background fill color; Use `"hidden"`
+#'   to hide the tab titles.  [updateTabsetPanel()] must be used to switch tabs.
+#'   Using `type = "hidden"` allows for quick switching between different UI sets.
```
cpsievert, line 523:
```suggestion
#'   look where tabs are selected using a background fill color; Use `"hidden"`
#'   in conjunction with [tabPanelBody()] and [updateTabsetPanel()] to control the
#'   active tab via other input controls (see examples section below).
```

cpsievert:
> Does this approach with the CSS classes work with bootstrap 4?

Mostly yes, but `.nav`'s (new) `display: flex` takes precedence over `.nav-hidden`'s `display: none`, so maybe we should make that `display: none !important` (or change the class ordering). 

Also, I'm not sure I see the reason to introduce `tabPanelBody()`? The title will be hidden regardless (at least it will in BS3...with this current approach, it wouldn't be in BS4).

schloerke:
@cpsievert 

> I'm not sure I see the reason to introduce `tabPanelBody()`?

I didn't like the api of `tabPanel(NULL, contentObj1, contentObj2, id = "panelID")`. 

As you said, the title will always be hidden.  But it is the first argument and it has to be set to supply content values.

NEWS.md
```diff
@@ -27,6 +29,8 @@ shiny 1.4.0.9001
 
 * Added a `'function'` class to `reactive()` and `reactiveVal()` objects. ([#2793](https://github.com/rstudio/shiny/pull/2793))
 
+* Added function `tabPanelBody()` which wraps `tabPanel()` without the title argument.  This function should be used within `tabsetPanel(type = "hidden")` ([#2814](https://github.com/rstudio/shiny/pull/2814))
+
```
cpsievert, line 33: I would collapse these two bullets into a single bullet that gets the idea across more quickly, something like

* Added a new option (`type = "hidden"`) to `tabsetPanel()`, making it easier to set the active tab via other input controls (e.g., `radioButtons()`) rather than tabs or pills. Use this option in conjunction with `updateTabsetPanel()` and the new `tabsetPanelBody()` function (see `help(tabsetPanel)` for an example and more details).

PR #2704, opened 2019-11-15
[WIP] Add skeleton function and example

trestletech:
Create a silly example app that has all the right structure.

TODO: 
 - come up with a more compelling Shiny app to test -- but hopefully one that doesn't have additional dependencies
 - Add real testthat unit tests. Right now these are just placeholders.
 - unit tests for the new skeleton function
 - roxygen tests for the new skeleton function
 - more in-line comments sprinkled throughout?
 - a README or pointers to relevant shiny dev center articles?
 - Other skeleton functions offer a way to seed your app with some files. Here it's unclear how we'd direct certain files to app/ui/server and others to R/ helpers. So I just punted on that for now
 - Make return value of this function useful or invisible

PR #2815, opened 2020-04-10
Consolidate testServer() fixes and enhancements

alandipert:
Closes #2812 
Closes #2813 
Merged #2704 
Spurred #2816 

This consolidates and fixes all existing outstanding branches related to `testServer()`.

In particular, it resolves a problem in the branch of `testServer()` in the case that `app` is a Shiny app obj or coercable to one. Some Shiny app objs may have a non-null `$onStart()` method.

In the case of a directory-backed app obj, this method is responsible for changing to the app directory before the server function is invoked. A corresponding `$onStop()` method is responsible for switching to the previous working directory.

In `testServer()`, we weren't calling either of these functions, under any circumstances. This was problematic, because in the case of an app with supporting files, the relevant `loadSupport()` call was inside `$onStart()`.

Now, we check for the presence of `$onStart()` and `$onStop()` and call them if they exist. If `$onStop()` is defined, it's called regardless of whether or not the server function signals an error. 

This ensures that the working directory in effect when the server function is invoked is that of the app. It also ensures that the working directory in effect when the test expression is evaluated is that of the enclosing test file.


R/test-server.R
```diff
@@ -1,13 +1,31 @@
...
+# }
+#
+# The data mask returned by this function should include z, session,
+# output, input, y, and id, but *not* x. Definitions not masked are
+# resolved in the environment in which testServer() is called.
+#
+# env is cloned because rlang::new_data_mask() mutates the parent of its `top`
+# argument.
+#' @importFrom rlang env_clone
+buildMask <- function(env) {
```
alandipert, line 24: buildMask should be called differently, and the number of environments cloned.

schloerke:
Deleting branch to avoid confusion

PR #2823, opened 2020-04-15
testServer() and loadSupport(): if app is a path, and not an app, walk up dirs

alandipert:


R/test-server.R
```diff
@@ -94,6 +115,12 @@ testServer <- function(app = ".", expr, ...) {
       )
     )
   } else {
+    # If app is a character vector it is assumed to be a path. If the path does
+    # not constitute a Shiny app, the path is traversed upward until one is
+    # found. If one is not found, an error is signaled.
+    if (is.character(app)) {
+      app <- findEnclosingApp(app)
+    }
```
schloerke, line 123: Can we move this logic up so that it is 

```
testServer <- function(app = findEnclosingApp("."), ....
```

R/test-server.R
```diff
@@ -3,6 +3,27 @@ isModuleServer <- function(x) {
...
+  FALSE
+}
+
+#' @noRd
+findEnclosingApp <- function(path = ".") {
+  rprojroot::find_root(
+    rprojroot::root_criterion(isAppDir, "is a Shiny app"),
+    path
+  )
+}
```
schloerke, line 25: Can this function also be set for the default of loadSupport?

```r
loadSupport <- function(appDir = findEnclosingApp("."), ....
```

schloerke: I don't know if it would be needed to be exported, but it wouldn't hurt if it is set as a default value.

wch:
The target branch for this PR, `alan-testServer-wip`, was squash-merged into master in #2815, so this should be re-targeted to master.

R/test-server.R
```diff
@@ -3,6 +3,27 @@ isModuleServer <- function(x) {
...
+  if (file.exists(file.path(path, "server.R"))
+      && file.exists(file.path(path, "ui.R")))
+    return(TRUE)
+
+  FALSE
+}
+
+#' @noRd
+findEnclosingApp <- function(path = ".") {
+  rprojroot::find_root(
```
wch, line 21: I think we should do this without taking a dependency on rprojroot, since the task is relatively small.

alandipert: OK, shall I copy the relevant code from rprojroot and include it in Shiny, or write new, rprojroot-inspired code?

I decided to depend on rprojroot mostly because it encapsulates subtleties related to functioning properly on Windows.

wch: What kind of issues are there with Windows?

R/test-server.R
```diff
@@ -50,7 +71,9 @@ isModuleServer <- function(x) {
 #'   # Any additional arguments, below, are passed along to the module.
 #' }, multiplier = 2)
 #' @export
-testServer <- function(app = ".", expr, ...) {
```
wch, line 53: I think we can keep `app = "."` and automatically search up from there, similar to how the functions in devtools and usethis work.

R/test-server.R
alandipert: There are no issues per se; there are only decisions about which conventions to use, such as which value to use consistently for the `winslash` argument of `normalizePath()` or how many parents to traverse before erroring. The benefit I saw to depending on rprojroot was less that it provided some algorithm that was difficult to reproduce, and more that it encapsulates a standard approach to this mundane problem, with limitations and capabilities that are already universally understood. One potential end-user benefit of this kind of reuse is that related error messages might be easier to Google.

That said, I understand that adding dependencies carries its own tradeoffs, and an argument for reuse based on convention adoption is relatively weak. It's straightforward for me to write new code in Shiny that follows the `winslash` and `.MAX_DEPTH` conventions established by rprojroot but is otherwise more focused on our particular use-case. Shall I do that?

R/test-server.R
alandipert: While `app` now defaults to `findEnclosingApp(".")` it sounds to me like the behavior is that which you describe, i.e. similar to devtools etc.

Are you suggesting an arrangement more like the following?

```
testServer <- function(app = ".", expr, ...) {
  app <- if (is.character(app)) findEnclosingApp(app) else app
}
```
The complication relative to devtools here is that `app` might be an appobj.

alandipert: Oops, I get it now, if we default with `findEnclosingApp()` than user-supplied dirs aren't traversed. Will compute.

R/test-server.R
wch: Since we're just using this internally, can we just do something simple like use `normalizePath()` and then call `dirname()` until we hit the top level?

alandipert: I'll try it out, and I think for cross-platform root detection we can borrow this from rprojroot: https://github.com/r-lib/rprojroot/blob/master/R/root.R#L67-L73

alandipert:
Update: latest thinking on `app`/`appDir` defaulting to `NULL` in `testServer()` and `loadSupport()` has been implemented.

R/utils.R
```diff
@@ -1838,3 +1838,36 @@ select_menu <- function(choices, title = NULL, msg = "Enter one or more numbers
...
+
+#' @noRd
+findEnclosingApp <- function(path = ".") {
+  path <- normalizePath(path, winslash = "/", mustWork = TRUE)
+  while (!is_root(path)) {
+    if (isAppDir(path))
+      return(path)
+    path <- dirname(path)
+  }
+  stop("Shiny app not found")
```
wch, line 1872: It would be helpful to tell the user where it's looking:

```R
stop("Shiny app not found at ", orig_path, " or an ancestor directory.")
```

Where `orig_path` is the value passed to the function.

R/app.R
```diff
@@ -325,20 +325,26 @@ initAutoReloadMonitor <- function(dir) {
 #' @details The files are sourced in alphabetical order (as determined by
 #'   [list.files]). `global.R` is evaluated before the supporting R files in the
 #'   `R/` directory.
-#' @param appDir The application directory
+#' @param appDir The application directory. If `appDir` is `NULL` or
+#'   not supplied, the nearest enclosing directory that is a Shiny app, starting
+#'   with the current directory, is used.
 #' @param renv The environmeny in which the files in the `R/` directory should
```
jcheng5, line 331: environmeny => environment

PR #2828, opened 2020-04-17
shinyAppTemplate tweaks

wch:
This PR does the following:
* adds the `dryrun` option
* prints out which files are being created
* Errors if a file already exists

### DIFF
--- a/R/app_template.R
+++ b/R/app_template.R
@@ -138,9 +138,14 @@ shinyAppTemplate <- function(path = NULL, examples = "default", dryrun = FALSE)
     return(length(files) != 0)
   }
 
-  # Helper to resolve paths relative to our example
-  example_path <- function(path) {
-    system.file("app_template", path, package = "shiny")
+  # Helper to resolve paths relative to our template
+  template_path <- function(...) {
+    system.file("app_template", ..., package = "shiny")
+  }
+
+  # Resolve path relative to destination
+  dest_path <- function(...) {
+    file.path(path, ...)
   }
 
   mkdir <- function(path) {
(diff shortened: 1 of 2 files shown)


R/app_template.R
```diff
@@ -122,58 +132,79 @@ shinyAppTemplate <- function(path = NULL, examples = "default")
...
+  mkdir <- function(path) {
+    if (!dirExists(path)) {
+      message("Creating ", ensure_trailing_slash(path))
+      if (!dryrun) {
+        dir.create(path, recursive = TRUE)
+      }
+    }
+  }
+
+  copy_file <- function(from, to, is_template = FALSE) {
```
schloerke, line 144: Can maybe move the `is_template` logic into the function as it's consistent across the board

schloerke: or make the `is_template` logic a helper function

R/app_template.R
```diff
@@ -109,6 +115,10 @@ shinyAppTemplate <- function(path = NULL, examples = "default")
     return(invisible())
   }
 
+  # =======================================================
```
schloerke, line 118: Before this line, can you check for `shinytest` version?

R/app_template.R
```diff
@@ -115,6 +115,17 @@ shinyAppTemplate <- function(path = NULL, examples = "default", dryrun = FALSE)
...
   }
 
+  if ("shinytest" %in% examples) {
+    if (system.file(package = "shinytest") != "" &&
+        packageVersion("shinytest") <= "1.4.0")
+    {
+      message(
+        "The tests/shinytest directory needs shinytest 1.4.0 or later to work properly.\n",
+        "You currently have shinytest ", packageVersion("shinytest"), " installed."
+      )
```
schloerke, line 125: Should this be a `stop()`?

schloerke: Wondering if the second line should be displayed if the package is not installed.

wch: I don't think it should be a stop -- we shouldn't force people to change their system for this. But they should know about it.

I agree the second line shouldn't be displayed if shinytest is not installed.

PR #2831, opened 2020-04-21
loadSupport(): fix global.R support, run global.R in appropriate dir

alandipert:
* `global.R` support was broken as a result our changes to the `appDir` argument; this fixes it.
* I added ye olde `require(shiny)` so `shiny::loadSupport()` works.
* Ensures `global.R` is evaluated in the app directory instead of whichever directory `loadSupport()` was called in.
* Replace use of `on.exit()/setwd()` with a `withr::with_dir()`
* The order of files in `R/` is sourced is now given by `sort(..., method="radix")` which aligns with R conventions
* `findEnclosingApp()` now uses `file.path.ci()` internally so it now works on case-insensitive file systems (tested by mounting fat32 loopback disk on Linux)

These items came up preparing https://github.com/CRI-iAtlas/shiny-iatlas for module testing.

### DIFF
--- a/R/app.R
+++ b/R/app.R
@@ -359,6 +359,10 @@ loadSupport <- function(appDir=NULL, renv=new.env(parent=globalenv()), globalren
   }
 
   helpers <- list.files(helpersDir, pattern="\\.[rR]$", recursive=FALSE, full.names=TRUE)
+  # Ensure files in R/ are sorted according to the 'C' locale before sourcing.
+  # This convention is based on the default for packages. For details, see:
+  # https://cran.r-project.org/doc/manuals/r-release/R-exts.html#The-DESCRIPTION-file
+  helpers <- sort(helpers, method = "radix")
   helpers <- normalizePath(helpers)
 
   withr::with_dir(appDir, {
--- a/R/app.R
+++ b/R/app.R
@@ -342,9 +342,10 @@ loadSupport <- function(appDir=NULL, renv=new.env(parent=globalenv()), globalren
 
   if (!is.null(globalrenv)){
     # Evaluate global.R, if it exists.
-    if (file.exists(file.path.ci(appDir, "global.R"))){
+    globalPath <- file.path.ci(appDir, "global.R")
+    if (file.exists(globalPath)){
       withr::with_dir(appDir, {
-        sourceUTF8("global.R", envir=globalrenv)
+        sourceUTF8(basename(globalPath), envir=globalrenv)
       })
     }
   }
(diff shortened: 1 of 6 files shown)


R/app.R
```diff
@@ -334,16 +334,22 @@ initAutoReloadMonitor <- function(dir) {
...
+
+  if (is.null(appDir)) {
+    appDir <- findEnclosingApp(".")
+  }
+
   if (!is.null(globalrenv)){
     # Evaluate global.R, if it exists.
     if (file.exists(file.path.ci(appDir, "global.R"))){
-      sourceUTF8(file.path.ci(appDir, "global.R"), envir=globalrenv)
+      with_save_wd({
```
schloerke, line 346: I believe we have `withr` as a dep.

Can call 
```r
withr::with_dir(appDir, { 
  sourceUTF8.... 
})
```

and drop `with_save_wd`

wch:
I think the other files, in the R/ directory, need to be sourced in the app directory as well, though this should be checked first to make sure it's the correct behavior.

alandipert:
@wch I think you're right, I'll fix it up.

alandipert:
I did some experimentation with `devtools::load_all()`, which seems to me like the most analogous existing behavior. I found it to be true that sources in `R/` are evaluated with the working dir initially set to the package dir. I also found that the directory is not reset *between* scripts. That is, one script may call `setwd()` and influence the working directory of subsequent scripts.

R/app.R
```diff
@@ -334,16 +334,21 @@ initAutoReloadMonitor <- function(dir) {
...
+  if (is.null(appDir)) {
+    appDir <- findEnclosingApp(".")
+  }
+
   if (!is.null(globalrenv)){
     # Evaluate global.R, if it exists.
     if (file.exists(file.path.ci(appDir, "global.R"))){
-      sourceUTF8(file.path.ci(appDir, "global.R"), envir=globalrenv)
+      withr::with_dir(appDir, {
+        sourceUTF8("global.R", envir=globalrenv)
```
wch, line 347: I think this needs `file.path.ci` in order to work on systems with a case-sensitive filesystem where global.R has a different capitalization, like Global.r. Should be able to test this out on Linux.

alandipert: Thanks, got it. Also applied to `findEnclosingApp()`

alandipert:
@wch apprised me of https://github.com/r-lib/roxygen2/issues/1077 and we now sort files from R/ using `sort(..., method = "radix")` which is documented to use 'C' collation order.

PR #2835, opened 2020-04-22
MockShinySession$click()

alandipert:
This adds a convenience function to `MockShinySession` to simulate clicking an `actionButton`. Motivating usage: https://github.com/alandipert/memory-hex/blob/1f78982f0d1f5b0ff87aabe9757b2395b45ef042/tests/testthat/test-hex.R

- [x] Test

### DIFF
--- a/R/mock-session.R
+++ b/R/mock-session.R
@@ -206,6 +206,9 @@ MockShinySession <- R6Class(
 
     #' @description Simulates clicking an action button.
     #' @param id The id of the button to click.
+    #' @examples
+    #' session <- MockShinySession$new()
+    #' session$click("button1")
     click = function(id) {
       val <- (private$.input$get(id) %OR% 0) + 1
       private$.input$set(id, val)
--- a/R/mock-session.R
+++ b/R/mock-session.R
@@ -194,8 +194,9 @@ MockShinySession <- R6Class(
     #'   and flushes the reactives.
     #' @param ... The inputs to set.
     #' @examples
-    #' session <- MockShinySession$new()
+    #' \dontrun{
     #' session$setInputs(x=1, y=2)
+    #' }
     setInputs = function(...) {
       vals <- list(...)
       mapply(names(vals), vals, FUN = function(name, value) {
@@ -207,8 +208,9 @@ MockShinySession <- R6Class(
     #' @description Simulates clicking an action button.
     #' @param id The id of the button to click.
     #' @examples
-    #' session <- MockShinySession$new()
+    #' \dontrun{
     #' session$click("button1")
+    #' }
     click = function(id) {
       val <- (private$.input$get(id) %OR% 0) + 1
       private$.input$set(id, val)
--- a/R/mock-session.R
+++ b/R/mock-session.R
@@ -194,14 +194,23 @@ MockShinySession <- R6Class(
     #'   and flushes the reactives.
     #' @param ... The inputs to set.
     #' @examples
-    #' s <- MockShinySession$new()
-    #' s$setInputs(x=1, y=2)
+    #' session <- MockShinySession$new()
+    #' session$setInputs(x=1, y=2)
     setInputs = function(...) {
       vals <- list(...)
       mapply(names(vals), vals, FUN = function(name, value) {
         private$.input$set(name, value)
       })
       private$flush()
+    },
+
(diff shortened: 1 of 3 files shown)


schloerke:
I really like the idea. This starts to venture into having a testing mock session vs a true mocking session object.

Is there any way we can distinguish between a true mock session and one with extra mock helpers?  Ex: `$_click()` . 

Or should we not worry about it?

alandipert:
@schloerke I'm not sure what you mean since there's currently only one kind of mock session, `MockShinySession`. Would you care to elaborate?

schloerke:
For regular session objects, there will probably never be a `$click()` method implemented.  My understanding of "mock" is to imitate all implemented methods.  Having the `MockShinySession` implement `$click()` breaks this concept.

alandipert:
Ah, yes. `MockShinySession` has things a regular session doesn't have. `$setInputs()` was the original "breaking" concept; `$click()` is a new specialization.

schloerke:
Ah. Great! Ok.

To round out the `$setInputs(...)`, can we add two public methods `$getInput(id)` and `$getInputs()`?  ... Then implement `$click(id)` using those methods?

alandipert:
To "get" inputs by id in a testing context, one would normally use `input$<id>`, and `input$names` to get all of the ids. `input` here is a `ReactiveValues` stored in a `MockShinySession`, with access possibly mediated by "proxies" that impose a namespace.

These accessors imply using `$<-` for assignment, but in a normal app, `$<-` for `input` is disallowed, because it would short-circuit the bidirectional data flow between the client (canonical source of input values) and the server.

So, we have `session$setInputs()`, which effectively performs assignment to inputs, and suggests explicitly that the "special" thing in a testing expression is the `session`.

One thing we might consider is implementing `$click()` in terms of `$setInputs()`. I decided against this because setting up the arguments for `$setInputs()` was much more involved than just manipulating `private$.input`, but I'm open to other ways.

schloerke:
I'm trying to protect against the manual call to flush.  

Maybe we can use `do.call(self$setInputs, list2(!!id := val))` since we have `rlang` imported?

alandipert:
Thanks, excellent idea, I was able to achieve something along those lines by making `setInputs` take dynamic dots, which is a cool bonus feature.

wch:
I just had a discussion with @schloerke about this, and after some back and forth, I think we actually should not have a click method or shortcut at all.

Users can just do something like this (we'll need to document this):

```R
testServer(s, {
  session$setInputs(x = 0)
  session$setInputs(x = input$x + 1)
})
```

The drawbacks of this are:
* It is a little more code than `session$click()`
* Users have to know a bit about how actionButtons work

The benefits are:
* We don't have to add a method for actionButtons that's inconsistent with all other inputs
* It more accurately reflects how actionButtons work


Regarding the last point, about being accurate: The initial value of an `actionButton` is 0. When it's clicked, it's incremented by 1. But if we have an API like this:


```R
testServer(s, {
  session$click("x")
})
```

Then the value of `input$x` goes from `NULL` to `1`. But when running an app for real, it starts at 0 and goes to 1. If you're using an `observeEvent(input$x, { ... })`, then the behavior would be the same, because both `NULL` and `0` are considered falsy values, and the expression will therefore not be run. But if you're using a regular `observe()` that uses `input$x`, then the behavior would be different.

For example, if you run this app, it'll print `0` at the console, then when you click the action button, it'll print `1`.

```R
s <- shinyApp(
  fluidPage(
    actionButton("x", "Click me")
  ),
    observe({
      message(input$x)
    })
  }
)

s
```


If you use `$click("x")`, it'll just print `1`:

```R
testServer(s, {
  session$click("x")
})
```

But if you do it the explicit way, it will print `0` and then `1`:

```R
testServer(s, {
  session$setInputs(x = 0)
  session$setInputs(x = input$x + 1)
})
```


The `rlang::dots_list()` stuff is good, though. Maybe make a separate PR with that stuff?

wch:
One more quick note: we also considered interfaces like these, but they have the same problems with not having the right initial value.

```R
session$setInput(x = inc(input$x))

session$setInput(x = session$inc(input$x))

session$setInput(x = session$trigger())
```

alandipert:
Thanks for thinking about this, I'll close here and open up a separate PR for the dots.

PR #2836, opened 2020-04-23
Combine docs for nearPoints() and brushedPoints()

hadley:
* Mouse -> pointer
* Simplify panelvar docs
* Add new ggplot2 and brushing sections


wch:
I think you need to rebuild the .Rd files -- they're not consistent with the source.

hadley:
Oops, sorry, fixed now.

PR #2837, opened 2020-04-23
Add parameter `args` to `testServer()`. Remove `../` from `loadSupport('../')` calls

schloerke:
* Removes the parameter of `../` from `loadSupport()` calls as the app will be found automatically now
* Moves `...` parameters in `testServer()` to `args` to avoid name clashing.  
  * Removed dynamic arg test as that is not needed anymore
  * Users can still create "dynamic args" using `rlang::list2()` mixed with `!!` or `!!!` when creating `args`.  This is why the default value for `args = rlang::list2()`

### DIFF
--- a/R/test-server.R
+++ b/R/test-server.R
@@ -18,10 +18,9 @@ isModuleServer <- function(x) {
 #'   in the server function environment, meaning that the parameters of the
 #'   server function (e.g. `input`, `output`, and `session`) will be available
 #'   along with any other values created inside of the server function.
-#' @param ... Additional arguments to pass to the module function. These
-#'   arguments are processed with [rlang::list2()] and so are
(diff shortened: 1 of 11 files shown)


R/test-server.R
```diff
@@ -49,14 +48,13 @@ isModuleServer <- function(x) {
 #'   stopifnot(myreactive() == 4)
 #'   stopifnot(output$txt == "I am 4")
 #'   # Any additional arguments, below, are passed along to the module.
-#' }, multiplier = 2)
+#' })
 #' @export
-testServer <- function(app = NULL, expr, ...) {
+testServer <- function(app = NULL, expr, args = rlang::list2()) {
```
wch, line 53: Why not just `list()` here? I think `rlang::list2()` will confuse people.

schloerke: Sounds good. Changing to `list()`

PR #2842, opened 2020-04-24
Safeguard monitorHandle function call

schloerke:
If an app source code fails to source, `monitorHandle` is not set as a function.  In the app cleanup, `monitorHandle()` is called.  

Now we check if it is a function and only call it if it is a function.

Before:
```r
tryCatch({
  # app 
  shiny::runApp("missing_highcharter_pkg")
}, error = function(e) { message("", e)})
Error in monitorHandle(): could not find function "monitorHandle"
```

With PR:
```r
tryCatch({shiny::runApp("missing_highcharter_pkg")}, error = function(e) { message("", e)})
Error in library(highcharter): there is no package called ‘highcharter’
```

(Used a `tryCatch` to capture the last error produced.  The `monitorHandle` error was stomping the true error of `highcharter` is missing.)

### DIFF
--- a/R/app.R
+++ b/R/app.R
@@ -238,8 +238,13 @@ shinyAppDir_serverR <- function(appDir, options=list()) {
   }
   onStop <- function() {
     setwd(oldwd)
-    monitorHandle()
-    monitorHandle <<- NULL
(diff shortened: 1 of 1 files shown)


schloerke:
@wch No changes are needed to shinytest for clearer error reporting

PR #2848, opened 2020-04-24
Add miscellaneous MockShinySession stubs

alandipert:
I was working on testing the `server` function of https://github.com/alandipert/memory-hex today and it didn't work, and I boiled it down to the fact that `MockShinySession` was missing a `session$sendModal()` which is called in `shiny::showModal()`.

I looked around and spotted a few others that we should probably add. I'll do a more thorough audit ASAP.

- [x] `sendModal`
- [ ] `showProgress`
- [ ] `sendProgress`


alandipert:
Defunct

PR #2851, opened 2020-04-27
Make `value` required for `tabPanelBody`

schloerke:
Addressing feedback on `tabPanelBody`. Currently not pursuing a new family of functions that do not involve the word "tab" (as there is no UI tab involved anymore). 

* `tabPanel()`
  * Change argument `value` to be arg `id`. 
  * This has the added side effect of setting the `id` of the `<div />` container using the `id` value. Since it was setting `data-value` attribute beforehand, this should not cause any new errors.
* `tabPanelBody()`
  * Move `id` to first arg position
  * `id` is now a required argument
* Added tests for both methods.

Followup to https://github.com/rstudio/shiny/pull/2814

cc @hadley


R/bootstrap.R
```diff
@@ -500,7 +500,18 @@ tabPanel <- function(title, ..., value = title, icon = NULL) {
 #' @export
 #' @describeIn tabPanel Create a tab panel that drops the title argument.
 #'   This function should be used within `tabsetPanel(type = "hidden")`. See [tabsetPanel()] for example usage.
-tabPanelBody <- function(..., value = NULL, icon = NULL) {
+#'   `value` must be set when using `tabPanelBody`.
+tabPanelBody <- function(value, ..., icon = NULL) {
+  if (
+    missing(value) ||
```
hadley, line 506: I don't think you need to explicitly test for absence

R/bootstrap.R
```diff
@@ -500,7 +500,18 @@ tabPanel <- function(title, ..., value = title, icon = NULL) {
 #' @export
 #' @describeIn tabPanel Create a tab panel that drops the title argument.
 #'   This function should be used within `tabsetPanel(type = "hidden")`. See [tabsetPanel()] for example usage.
-tabPanelBody <- function(..., value = NULL, icon = NULL) {
+#'   `value` must be set when using `tabPanelBody`.
+tabPanelBody <- function(value, ..., icon = NULL) {
```
hadley, line 504: I wish this could be called `id` instead of `value`.

tests/testthat/test-tabPanel.R
```diff
@@ -0,0 +1,14 @@
+
+context("tabPanel")
+
+test_that("tabPanelBody", {
```
hadley, line 4: I think you should include a test that verifies the previous syntax continues to work too.

tests/testthat/test-tabPanel.R
```diff
@@ -0,0 +1,14 @@
+
+context("tabPanel")
```
hadley, line 2: Modern practice is to omit `context()`

R/bootstrap.R
```diff
@@ -500,7 +500,18 @@ tabPanel <- function(title, ..., value = title, icon = NULL) {
...
+  if (
+    missing(value) ||
+    nullOrEmpty(value) ||
```
hadley, line 507: I think this is redundant as you're already checking that it's a character vector of length 1.

tests/testthat/test-tabPanel.R
schloerke: I ran into a situation with `shiny::shinyAppTemplate()`... if no file had a context, then the errors were not printed.  By adding the context, then errors were reported correctly.

tests/testthat/test-tabPanel.R
schloerke: Which syntax? `tabPanelBody(someHTML, value = X)`?

hadley: Yeah

tests/testthat/test-tabPanel.R
```diff
@@ -0,0 +1,28 @@
+
+
+test_that("tabPanelBody", {
```
hadley, line 3:
```suggestion
test_that("tabPanelBody validates it's input", {
```

tests/testthat/test-tabPanel.R
```diff
@@ -0,0 +1,28 @@
...
+  expect_error(tabPanelBody(NULL), "single, non-empty string")
+  expect_error(tabPanelBody(1), "single, non-empty string")
+  expect_error(tabPanelBody(TRUE), "single, non-empty string")
+  expect_error(tabPanelBody(NA), "single, non-empty string")
+  expect_error(tabPanelBody(NA_character_), "single, non-empty string")
+  expect_error(tabPanelBody(""), "single, non-empty string")
+  expect_error(tabPanelBody(letters[1:2]), "single, non-empty string")
+})
+
+test_that("tabPanel", {
```
hadley, line 16:
```suggestion
test_that("tabPanel works with deprecated interface", {
```

R/bootstrap.R
```diff
@@ -488,20 +489,33 @@ helpText <- function(...) {
 #' )
 #' @export
 #' @describeIn tabPanel Create a tab panel that can be included within a [tabsetPanel()] or a [navbarPage()].
-tabPanel <- function(title, ..., value = title, icon = NULL) {
+tabPanel <- function(title, ..., id = title, icon = NULL, value = NULL) {
```
wch, line 492: I think using `id` is too drastic. The change here will cause lots of messages to be printed for some users, and it will also add a HTML id to the DOM element, which will cause problems if the it happens to be the same ID as some other DOM element. It's not hard to imagine a case where the `value` is the same is as the `tabsetPanel`'s `id`.

The name `id` is also a bit misleading, since everywhere else that we use it, it's meant to set the DOM ID, but in this case, that's sort of a side effect of what we really want to do, which is provide a value when this tab is selected, if the `tabsetPanel` is an input.

R/bootstrap.R
```diff
@@ -500,7 +500,15 @@ tabPanel <- function(title, ..., value = title, icon = NULL) {
 #' @export
 #' @describeIn tabPanel Create a tab panel that drops the title argument.
 #'   This function should be used within `tabsetPanel(type = "hidden")`. See [tabsetPanel()] for example usage.
-tabPanelBody <- function(..., value = NULL, icon = NULL) {
+tabPanelBody <- function(value, ..., icon = NULL) {
+  if (
```
wch, line 504: Did you decide against moving this check into `tabPanel`?

R/update-input.R
```diff
@@ -305,9 +305,9 @@ updateDateRangeInput <- function(session, inputId, label = NULL,
 #'   ),
 #'   mainPanel(
 #'     tabsetPanel(id = "inTabset",
-#'       tabPanel(title = "Panel 1", value = "panel1", "Panel 1 content"),
-#'       tabPanel(title = "Panel 2", value = "panel2", "Panel 2 content"),
-#'       tabPanel(title = "Panel 3", value = "panel3", "Panel 3 content")
+#'       tabPanel(title = "Panel 1", id = "panel1", "Panel 1 content"),
```
wch, line 308: These should be `value`, not `id`.

R/bootstrap.R
schloerke: Yes.  It could break others since it is a `stop`. Having an `title = NA` with no value shouldn't prevent users from having the `tabPanel` work.

PR #2854, opened 2020-04-29
testServer(): Properly capture module return values

alandipert:
Previously, the return value of a module under test via `testServer()` was captured by modifying the module function body. The expression value of the function body was retained. However, this doesn't work when module function body includes early `return()` calls. `session$returned` in these cases is `NULL`.

In this PR, we introduce a different technique for retaining module function return values. We wrap the call to the module function in a call to `session$setReturned()`.

Note that the value of app server functions is not retained, because we [document](https://shiny.rstudio.com/reference/shiny/0.12.2/shinyServer.html) that server function return values are discarded. We may consider producing an error if `session$returned` is referenced in the test expression associated with an app server function.


alandipert:
Gonna roll this into a new PR shortly.

PR #2864, opened 2020-05-04
Add 'auto' brush fill and stroke

cpsievert:
Allows `fill`/`stroke` in `brushOpts()` to be `'auto'`; in which case, `fill` is based on `accent` (i.e., link color) and `stroke` is based on `fg` (i.e., foreground color). This works regardless of how the container is styled, but just to demonstrate how this can be used with **thematic** and **bslib**:

```r
library(shiny)
library(ggplot2)
thematic::thematic_shiny()
ui <- fluidPage(
  theme = bslib::bs_theme(bootswatch = "darkly"),
  plotOutput("p", brush = brushOpts("brush", fill = "auto", stroke = "auto"))
)
  output$p <- renderPlot({
    qplot(rnorm(1000), rnorm(1000))
  })
}
shinyApp(ui, server)
```
Screen Shot 2020-05-04 at 1 10 07 PM


R/image-interact-opts.R
```diff
@@ -137,3 +148,8 @@ brushOpts <- function(id, fill = "#9cf", stroke = "#036",
     resetOnNew = resetOnNew
   )
 }
+
+getThematicOption <- function(...) {
+  if (system.file(package = "thematic") == "") return(NULL)
```
wch, line 153: We should make this `system.file` pattern a separate function, like `is_installed`.

wch (approved):

PR #2867, opened 2020-05-06
Add missing methods to MockShinySession

alandipert:
- [x] Implement every public method, field, active binding except those related to `downloadHandler()`
- [x] `downloadHandler()` and corresponding `getOutput()` support
- [x] `getCurrentOutputInfo()` support - only ever returns `NULL` or `list(name = <output-id>)`
- [x] Document remaining undocumented `MockShinySession` fields and methods
- [x] Ensure `renderCachedPlot` works
- [x] Test apps with bookmarking
- [x] Add reactivity + promises test
- [x] Address outstanding review comments


R/mock-session.R
```diff
@@ -75,41 +75,149 @@ mapNames <- function(func, vals) {
...
+#' @noRd
+addGeneratedInstanceMethods <- function(instance, methods = makeExtraMethods()) {
+  mapply(function(name, impl) {
+    environment(impl) <- instance$.__enclos_env__
+    instance[[name]] <- impl
+  }, names(methods), methods)
+}
+
+#' @export
+makeMockSession <- function() {
```
wch, line 186: Can we get rid of this now?

R/mock-session.R
```diff
@@ -75,41 +75,149 @@ mapNames <- function(func, vals) {
...
+
 #' Mock Shiny Session
 #'
 #' @description
 #' An R6 class suitable for testing that simulates the `session` parameter
 #' provided to Shiny server functions or modules.
 #'
-#' @include timer.R
-#' @export
+#' @noRd
```
wch, line 197: Can we restore the previous state of the roxygen comments (that is, export it, and have the methods documented)?

R/mock-session.R
```diff
@@ -75,41 +75,149 @@ mapNames <- function(func, vals) {
...
+makeNoop <- function(name, msg = paste0(name, " is a noop.")) {
+  if (!(name %in% names(ShinySession$public_methods)))
+    stop(name, " is not public method of ShinySession.")
+  impl <- ShinySession$public_methods[[name]]
+  body(impl) <- rlang::expr({
+    # Force arguments
+    !!lapply(formalArgs(impl), rlang::sym)
+    private$noopWarn(!!name, !!msg)
+    NULL
```
wch, line 89: Returning `invisible()` would make the UI a little cleaner.

R/mock-session.R
```diff
@@ -75,41 +75,149 @@ mapNames <- function(func, vals) {
...
+  })
+  impl
+}
+
+#' Accepts a series of symbols as arguments and generates corresponding noop
+#' implementations.
+#' @noRd
+makeWarnNoops <- function(...) {
+  methods <- as.character(rlang::ensyms(...))
+  sapply(methods, makeNoop, USE.NAMES = TRUE, simplify = FALSE)
```
wch, line 99: Please use `lapply` -- in general, avoid `sapply` because it's not clear from looking at the code what type of object it returns (and also, the type can change depending on the input).

R/mock-session.R
```diff
@@ -75,41 +75,149 @@ mapNames <- function(func, vals) {
...
+    NULL
+  })
+  impl
+}
+
+#' Accepts a series of symbols as arguments and generates corresponding noop
+#' implementations.
+#' @noRd
+makeWarnNoops <- function(...) {
+  methods <- as.character(rlang::ensyms(...))
```
wch, line 98: Simpler to use `as.character(list(...))`

tests/testthat/test-test-server.R
```diff
@@ -693,12 +693,37 @@ test_that("It's an error to pass arguments to a server", {
   expect_error(testServer(test_path("..", "test-modules", "06_tabsets"), {}, args = list(an_arg = 123)))
 })
 
+# Provided an instance of an R6 object, returns a list with `methods` and
+# `fields`. `methods` contains a character vector of names of public methods.
+# `fields` is a character vector of public fields. Any active bindings are
+# considered `fields`.
+get_mocked_publics <- function(instance) {
+  generator <- get(class(instance)[[1]])
+  publics <- ls(mock_session)
```
wch, line 702: Should this be `instance` instead of `mock_session`?

tests/testthat/test-test-server.R
```diff
@@ -693,12 +693,37 @@ test_that("It's an error to pass arguments to a server", {
...
+  actives <- names(generator$active)
+  # Active bindings are considered fields.
+  methods_or_fields <- publics[!(publics %in% actives)]
+  methods <- character(0)
+  fields <- c(character(0), actives)
```
wch, line 707: Any particular reason to not do:

```
fields <- actives
```

R/test-server.R
```diff
@@ -50,20 +50,21 @@ isModuleServer <- function(x) {
...
 
   quosure <- rlang::enquo(expr)
-  session <- getDefaultReactiveDomain()
 
-  if (inherits(session, "MockShinySession"))
-    stop("Test expressions may not call testServer()")
-  if (inherits(session, "session_proxy")
-      && inherits(get("parent", envir = session), "MockShinySession"))
-    stop("Modules may not call testServer()")
+  local({
```
wch, line 59: If `session` now defaults to `makeMockSession()`, what is the purpose of moving this code into local and checking the `getDefaultReactiveDomain()`?

tests/testthat/test-test-server.R
```diff
@@ -651,3 +651,79 @@ test_that("session flush handlers work", {
...
+      123
+    })
+  }
+
+  testServer(module_implicit_return, {
+    expect_equal(session$returned, 123)
+  })
+
+  module_early_returns <- function(id, n) {
+    retval <<- NULL
```
wch, line 667: This should be `<-`; otherwise it will assign `retval` in the global environment.

R/mock-session.R
alandipert: Since lapply doesn't have a USE.NAMES option I switched to lapply and an assignment to names(), please let me know if you were suggesting something different

tests/testthat/test-test-server.R
alandipert: `fields` might then be `NULL` instead of an empty character vector and I thought there might be some value in the field being type stable

R/test-server.R
alandipert: The idea for exposing the `session` argument was that expert users might want to subclass/override parts of a MockShinySession to do stuff like assign to `user` or `groups` to test auth, or to provide bogus clientData.

The check endures because the average user should still be aware that nested testServer is bad.

R/mock-session.R
```diff
@@ -317,73 +471,17 @@ MockShinySession <- R6Class(
       v <- extract(private$outs[[name]]$promise)
       if (!is.null(v$err)){
         stop(v$err)
+      } else if (private$file_generators$has(self$ns(name))) {
+        download <- private$file_generators$get(self$ns(name))
+        private$renderFile(name, download)
```
wch, line 476: Might be helpful to use `ns` for this as well.

R/mock-session.R
```diff
@@ -407,53 +506,187 @@ MockShinySession <- R6Class(
...
     outs = list(),
+    # @field nsPrefix Prefix with which to namespace inputs and outputs.
     nsPrefix = "mock-session",
+    # @field idCounter Incremented every time `$genId()` is called.
     idCounter = 0,
+    # @field file_generators Map of namespaced output names to lists with
+    #   `filename` and `output` elements, each a function. Updated by
+    #   `$registerDownload()` and read by `$getOutput()`. Files are generated
+    #   on demand when the output is accessed.
+    file_generators = fastmap(),
```
wch, line 620: This needs to be done in the constructor.

R/mock-session.R
```diff
@@ -407,53 +506,187 @@ MockShinySession <- R6Class(
...
+    #   function associated with a download handler does not write a file, an
+    #   error is signaled.
+    # @param name The un-namespaced output name associated with the
+    #   downloadable file.
+    # @param download List with two names, `filename` and `content`. Both
+    #   should be functions. `filename` should take no arguments and return a
+    #   string. `content` should accept a path argument and create a file at
+    #   that path.
+    # @return A path to a temp file.
+    renderFile = function(name, download) {
```
wch, line 635: Should delete all download files when session is closed.

R/test-server.R
```diff
@@ -50,20 +50,21 @@ isModuleServer <- function(x) {
...
```
wch, line 59: Can remove `local` because it will make the stack trace more confusing.

Also, a comment explaining why this is here would be helpful.

R/test-server.R
```diff
@@ -50,20 +50,21 @@ isModuleServer <- function(x) {
...
+    active_session <- getDefaultReactiveDomain()
```
wch, line 60: We can just check for `!is.null(getDefaultReactiveDomain())` and throw an error in that case. Maybe a message like "Can't call `testServer()` with an active reactive domain."

tests/testthat/test-test-server.R
```diff
@@ -651,3 +651,123 @@ test_that("session flush handlers work", {
...
+})
+
+# Provided an instance of an R6 object, returns a list with `methods` and
+# `fields`. `methods` contains a character vector of names of public methods.
+# `fields` is a character vector of public fields. Any active bindings are
+# considered `fields`.
+get_mocked_publics <- function(instance) {
+  generator <- get(class(instance)[[1]])
+  publics <- ls(instance, all.names = TRUE)
+  actives <- names(generator$active)
```
wch, line 703:
```
names(generator$active) %OR% character(0)
```

And then the `fields <-` below doesn't need the `c()`.

tools/documentation/pkgdown.yml
```diff
@@ -220,4 +220,3 @@ reference:
     contents:
       - runTests
       - testServer
```
wch, line 222: `MockShinySession` should go back in here.

R/mock-session.R
alandipert: Turns out session close callbacks weren't actually implemented and the test for them was broken, so added that, and then this

R/mock-session.R
```diff
@@ -85,7 +85,8 @@ makeNoop <- function(name, msg = paste0(name, " is a noop.")) {
   body(impl) <- rlang::expr({
     # Force arguments
     !!lapply(formalArgs(impl), rlang::sym)
-    private$noopWarn(!!name, !!msg)
+    # Evade "no visible binding" note for reference to `private`
+    get("private")$noopWarn(!!name, !!msg)
```
wch, line 89: Instead of putting a `get()` in each function, we can put the symbol in there directly with something like:

```R
private_sym <- as.symbol("private")
rlang::expr({ !!private_sym })
```

An alternate way of doing this is by calling `globalVariables()` (though we use it elsewhere, so you'd need to use `add=TRUE`).

alandipert: I thought about the globalVariables() way, but I thought it might be more confusing than helpful... since `private` ever isn't actually global

PR #2873, opened 2020-05-07
Document the return value of getCurrentOutputInfo()

cpsievert:
Including the new info added in #2740.

### DIFF
--- a/R/shiny.R
+++ b/R/shiny.R
@@ -2144,9 +2144,8 @@ outputOptions <- function(x, name, ...) {
 
 #' Get output information
 #'
-#' Calling this function within an output context returns information about
-#' the output, including its `name` (i.e., `outputId`); and in some cases,
-#' sizing and styling information.
+#' Returns information about the currently executing output, including its `name` (i.e., `outputId`);
+#' and in some cases, relevant sizing and styling information.
 #'
 #' @param session The current Shiny session.
 #'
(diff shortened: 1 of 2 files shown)


R/shiny.R
```diff
@@ -2142,15 +2142,71 @@ outputOptions <- function(x, name, ...) {
 }
 
 
-#' Get information about the output that is currently being executed.
+#' Get output information
 #'
-#' @return A list with information about the current output, including the
-#'   `name` of the output. If no output is currently being executed, this will
-#'   return `NULL`.
+#' Calling this function within an output context returns information about
```
wch, line 2147: Oh, actually, this is not quite right, or at least a little confusing. You can of course call `getCurrentOutputInfo()` in a `renderPlot()`, but you can also call it in any reactive called by the `renderPlot()`.

cpsievert: Sure, but in that case you're still calling the function within the context, right? How about:

> When this function is called inside a reactive output context

Instead of 

> Calling this function within an output context

wch: I just mean that "output context" isn't a well-defined concept, whereas "currently executing output" is slightly less confusing (at least to me). I'm open to other phrasing as well.

We actually do use "output context" here, but the meaning is explicitly different from what you're saying: https://github.com/rstudio/shiny/blob/master/R/utils.R#L1318-L1320

cpsievert: Ah, I see, thanks for clarifying

PR #2874, opened 2020-05-07
shinyAppTemplate tweaks

wch:
This does the following:

- Rename R/sort.R to R/example.R
- Rename R/mymodule.R to R/example-module.R
- Changes the appearance of the app so that it's clearer that the modules are separate from the slider.

The full app now looks like this:
image

@mine-cetinkaya-rundel


R/app_template.R
```diff
@@ -38,7 +38,7 @@
 #' * `app.R` is the main application file.
 #' * All files in the `R/` subdirectory are automatically sourced when the
 #'   application is run.
-#' * `R/sort.R` and `R/my-module.R` are automatically sourced when
+#' * `R/example.R` and `R/example-module.R` are automatically sourced when
 #'   the application is run. The first contains a function `lexical_sort()`,
 #'   and the second contains code for a [Shiny module](moduleServer()) which
```
mine-cetinkaya-rundel, line 43: `[Shiny module](moduleServer())` doesn't actually take me to the docs for the `moduleServer()` from the help file, when viewing it in the IDE. Could either link to the article https://shiny.rstudio.com/articles/modules.html or rephrase as "contains code for a Shiny module created with the [moduleServer()]" function which is ..."

wch:
Thanks!

PR #2881, opened 2020-05-08
Safer default file deletion behavior for renderImage

jcheng5:
## Testing notes

```r
library(shiny)

ui <- fluidPage(
  imageOutput("img")
)

  output$img <- renderImage({
    list(src = "~/Desktop/image.png")
  })
}

shinyApp(ui, server)
```

In previous versions of Shiny, running this app would cause `~/Desktop/image.png` to be deleted (!!).

With this change, instead, the file will not be deleted; but you will get a warning asking that an explicit `deleteFile` argument be passed.

If however, you replace `src` with a file path that's under `tempdir()`, then the file WILL be deleted (you'll still get a warning).

If you pass `deleteFile=TRUE`/`FALSE` then it should do that, regardless of where the file lives, and no warning should be emitted.

### DIFF
--- a/R/shinywrappers.R
+++ b/R/shinywrappers.R
@@ -293,7 +293,7 @@ renderImage <- function(expr, env=parent.frame(), quoted=FALSE,
 
       # jcheng 2020-05-08
       #
-      # Until Shiny 1.4.1, the default for deleteFile was, incredibly, TRUE.
+      # Until Shiny 1.5.0, the default for deleteFile was, incredibly, TRUE.
       # Changing it to default to FALSE might cause existing Shiny apps to pile
       # up images in their temp directory (for long lived R processes). Not
       # having a default (requiring explicit value) is the right long-term move,
@@ -308,9 +308,9 @@ renderImage <- function(expr, env=parent.frame(), quoted=FALSE,
           warned <<- TRUE
           warning("The renderImage output named '",
             getCurrentOutputInfo()$name,
-            "' is missing the deleteFile argument; please provide TRUE or ",
-            "FALSE. (This warning will become an error in a future version of ",
-            "Shiny.)",
+            "' is missing the deleteFile argument; as of Shiny 1.5.0, you must ",
+            "use deleteFile=TRUE or deleteFile=FALSE. (This warning will ",
+            "become an error in a future version of Shiny.)",
             call. = FALSE
           )
         }
@@ -339,9 +339,7 @@ isTempFile <- function(path, tempDir = tempdir(), default = FALSE) {
     return(default)
   }
   tempDir <- normalizePath(tempDir, winslash = "/", mustWork = FALSE)
-  if (substr(tempDir, nchar(tempDir), nchar(tempDir)) != .Platform$file.sep) {
-    tempDir <- paste0(tempDir, .Platform$file.sep)
-  }
+  tempDir <- ensure_trailing_slash(tempDir)
 
   path <- normalizePath(path, winslash = "/", mustWork = FALSE)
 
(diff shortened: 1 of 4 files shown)


jcheng5:
Oops forgot a unit test

R/shinywrappers.R
```diff
@@ -295,6 +333,20 @@ renderImage <- function(expr, env=parent.frame(), quoted=FALSE,
     imageOutput, outputArgs)
 }
 
+isTempFile <- function(path, tempDir = tempdir(), default = FALSE) {
+  if (nchar(tempDir) == 0) {
+    # This should never happen, but just to be super paranoid...
+    return(default)
+  }
+  tempDir <- normalizePath(tempDir, winslash = "/", mustWork = FALSE)
+  if (substr(tempDir, nchar(tempDir), nchar(tempDir)) != .Platform$file.sep) {
```
wch, line 342: Not super important, but there's a function `ensure_trailing_slash` that does the same thing and conveys intent better. (Also `.Platform$file.sep` is apparently `/` on all platforms.)

R/shinywrappers.R
```diff
@@ -271,15 +274,50 @@ markOutputAttrs <- function(renderFunc, snapshotExclude = NULL,
...
     transform = function(imageinfo, session, name, ...) {
-      # Should the file be deleted after being sent? If .deleteFile not set or if
-      # TRUE, then delete; otherwise don't delete.
-      if (deleteFile) {
-        on.exit(unlink(imageinfo$src))
+      shouldDelete <- deleteFile
+
+      # jcheng 2020-05-08
+      #
+      # Until Shiny 1.4.1, the default for deleteFile was, incredibly, TRUE.
```
wch, line 296: 1.5.0

R/shinywrappers.R
```diff
@@ -271,15 +274,50 @@ markOutputAttrs <- function(renderFunc, snapshotExclude = NULL,
...
+      # Compromise we decided on was to eventually require TRUE/FALSE, but for
+      # now, change the default behavior to only delete temp files; and emit a
+      # warning encouraging people to not rely on the default.
+      if (is.null(shouldDelete)) {
+        shouldDelete <- isTempFile(imageinfo$src)
+        if (!warned) {
+          warned <<- TRUE
+          warning("The renderImage output named '",
+            getCurrentOutputInfo()$name,
+            "' is missing the deleteFile argument; please provide TRUE or ",
```
wch, line 311: Can you say, "As of Shiny 1.5.0, the deleteFile argument must be either TRUE or FALSE..."

R/shinywrappers.R
```diff
@@ -271,15 +274,50 @@ markOutputAttrs <- function(renderFunc, snapshotExclude = NULL,
 #' shinyApp(ui, server)
 #' }
 renderImage <- function(expr, env=parent.frame(), quoted=FALSE,
-                        deleteFile=TRUE, outputArgs=list()) {
+                        deleteFile, outputArgs=list()) {
   installExprFunction(expr, "func", env, quoted)
 
+  # missing() must be used directly within the function with the given arg
+  if (missing(deleteFile)) {
```
hadley, line 281: Why not set `deleteFile = NULL` in the function def?

jcheng5: So you could look at the signature and know that `deleteFile` is "required". (Even though it's not truly required right now, because we're being nice and giving you a warning instead of an error.)

hadley: Oh yeah, makes sense.

wch: Since our end goal is for people to always use either `TRUE` or `FALSE`, providing `NULL` as a default might make give people the idea that it's an OK value to use (even if only temporarily).

PR #2886, opened 2020-05-11
isTempPath is now isTemp, simplified semantics

jcheng5:
Tests were failing on Mac due to `tempdir()` returning a non-normalized path, and `tempfile()` returning paths that could not be normalized (because the files didn't exist).

- Now requires both the path and temp dir to exist
- isTemp(tempdir()) is now false
- Doesn't matter if it's files or directories

## Testing notes

See https://github.com/rstudio/shiny/pull/2881

### DIFF
--- a/R/shinywrappers.R
+++ b/R/shinywrappers.R
@@ -303,7 +303,10 @@ renderImage <- function(expr, env=parent.frame(), quoted=FALSE,
       # now, change the default behavior to only delete temp files; and emit a
       # warning encouraging people to not rely on the default.
       if (is.null(shouldDelete)) {
-        shouldDelete <- isTempFile(imageinfo$src)
+        shouldDelete <- isTRUE(try(silent = TRUE,
+          file.exists(imageinfo$src) && isTemp(imageinfo$src, mustExist = TRUE)
+        ))
+
         if (!warned) {
           warned <<- TRUE
           warning("The renderImage output named '",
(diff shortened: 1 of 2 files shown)


tests/testthat/test-shinywrappers.R
```diff
@@ -1,27 +1,50 @@
 context("shinywrappers")
 
-test_that("isTempFile passes sanity checks", {
-  expect_true(isTempFile(tempfile()))
-  expect_false(isTempFile(path.expand("~")))
-  expect_false(isTempFile("."))
+test_that("isTemp passes sanity checks", {
+  t <- tempfile(fileext = ".txt")
+  writeLines("hello", t)
+  on.exit(unlink(t), add = TRUE, after = FALSE)
```
schloerke, line 6: This fails on the ubuntu tests

R/shinywrappers.R
```diff
@@ -333,15 +336,38 @@ renderImage <- function(expr, env=parent.frame(), quoted=FALSE,
     imageOutput, outputArgs)
 }
 
-isTempFile <- function(path, tempDir = tempdir(), default = FALSE) {
-  if (nchar(tempDir) == 0) {
+# TODO: If we ever take a dependency on fs, it'd be great to replace this with
+# fs::path_as_parent().
```
cpsievert, line 340:
```suggestion
# fs::path_has_parent().
```

PR #2889, opened 2020-05-12
Move callModule() to separate .Rd; add various links to docs

alandipert:


R/modules.R
```diff
@@ -143,7 +142,19 @@ moduleServer <- function(id, module, session = getDefaultReactiveDomain()) {
 }
 
 
-#' @rdname moduleServer
+#' Invoke a Shiny module
+#'
+#' **Deprecated** `callModule` has been deprecated in favor of
+#' [`moduleServer()`] starting in Shiny 1.5.0.
```
wch, line 148:
```suggestion
#' Note: As of Shiny 1.5.0, we recommend using [`moduleServer()`] instead of
#' [`callModule()`], because the syntax is a little easier
#' to understand, and modules created with `moduleServer` can be tested with
#' [`testServer()`].
```

PR #2897, opened 2020-05-15
Fix collapsible navbar

cpsievert:
Fixes #2896

Issue #2896, opened 2020-05-15
navbarPage(collapsible = TRUE) doesn't work


cpsievert:
Minimal example (make window narrow, then click on the collapsible menu, it doesn't expand)

```r
library(shiny)

ui <- navbarPage(
  "Navbar!", collapsible = TRUE,
  tabPanel(
    "Plot",
    sidebarLayout(
      sidebarPanel(
        radioButtons("plotType", "Plot type", c("Scatter"="p", "Line"="l"))
      ),
      mainPanel(plotOutput("plot"))
    )
  ),
  tabPanel(
    "Summary",
    verbatimTextOutput("summary")
  )
)

server <- function(input, output, session) {
  output$plot <- renderPlot({
    plot(cars, type=input$plotType)
  })
  
  output$summary <- renderPrint({
    summary(cars)
  })
}

shinyApp(ui, server)
```

cpsievert:
For posterity, this issue was surfaced by bumping jQuery from 3.4.1 to 3.5.0 (for #2857), but we've added a workaround by patching Bootstrap 3

PR #2900, opened 2020-05-18
Bump jQuery 3.5.0 to 3.5.1

cpsievert:
This bumps jQuery 3.5.0 -> 3.5.1 and also rolls back the BS3 patch done in #2897 (since the problem (#2896) can be fixed by just bumping jQuery).


tools/updateAndPatchBootstrap.R
```diff
@@ -1,38 +0,0 @@
-version <- "3.4.1"
```
wch, line 1: Would it make sense to keep this script, just without the patch stuff?

cpsievert: I guess, but hopefully in the near future we'll be using `bootstraplib::bootstrap()` ;)

wch:
Can you also add a script for updating jQuery? We'll be doing this again in the future and it would be nice to have a script for it.

tools/updateAndPatchBootstrap.R
cpsievert: (Also, even if we don't, I doubt it'll be that useful as BS3 is end-of-life)

PR #2911, opened 2020-05-29
Added Bootstrap accessibility plugin

jooyoungseo:
* Added [bootstrap accessibility plugin](https://github.com/paypal/bootstrap-accessibility-plugin) under the hood to improve accessibility of shiny apps for screen-reader and keyboard users: the enhancements include better navigations for alert, tooltip, popover, modal dialog, dropdown, tab Panel, collapse, and carousel elements.
* Added the plugin authors and corresponding copyright info to `DESCRIPTION` file.
* I have confirmed that screen reader navigation has improved after this update; however, this is a global fix. I have to address shiny-specific UI and widgets one by one moving forward.
* Made `accessibility` section with the updates in `NEWS.md` file.


wch:
Thanks! Can you also add a script for updating the library. Something similar to this:
https://github.com/rstudio/shiny/blob/68f0c12c/tools/updateBootstrapDatepicker.R

And also update src/README.md with appropriate instructions on how to update, under "Updating web libraries".

We are about to release Shiny 1.5.0, so the 1.4.0.9001 section in the NEWS isn't the right place to put the news item (we would merge this after the release). No matter where you put the news, there will be a merge conflict on master, but I think it would make sense for now to put it at the top, above the `Shiny 1.4.0.9001`, and then we can resolve the merge conflict later.

jooyoungseo:
@wch, please review again at your best convenience. :) I have intentionally placed the plugin under the bootstrap directory without creating a separate folder for a better management. And, I don't think we need a patch script for this plugin beyond the updater. But, just let me know.

tools/updateBootstrapAccessibilityPlugin.R
```diff
@@ -0,0 +1,34 @@
...
+sha   <- "fbbf870eafc1ee5d4547fabbeea71778ddbe2166"
+dest_dir  <- rprojroot::find_package_root_file("inst/www/shared/bootstrap")
+
+js_url       <- sprintf("https://raw.githubusercontent.com/paypal/bootstrap-accessibility-plugin/%s/plugins/js/bootstrap-accessibility.js", sha)
+min_js_url       <- sprintf("https://raw.githubusercontent.com/paypal/bootstrap-accessibility-plugin/%s/plugins/js/bootstrap-accessibility.min.js", sha)
+css_url       <- sprintf("https://raw.githubusercontent.com/paypal/bootstrap-accessibility-plugin/%s/plugins/css/bootstrap-accessibility.css", sha)
+url <- c(js_url, min_js_url, css_url)
+
+# Downloading each required file:
+invisible(sapply(url, function(x) {download.file(x, file.path(tempdir(), basename(x)))}))
```
wch, line 18: Minor stylistic things that don't really matter in this case, but are good practice:

* No need for `invisible()` here, since this in the middle of a script.
* In general, please use `lapply` instead of `sapply`, since the type of object returned by `sapply` isn't clear from reading the code, and can change depending on what the applied function returns.

tools/updateBootstrapAccessibilityPlugin.R
```diff
@@ -0,0 +1,34 @@
...
+#  https://github.com/paypal/bootstrap-accessibility-plugin
+# Since the last version release has been too behind the latest commit, we will use a specific sha instead.
+
+
+library(rprojroot)
+
+sha   <- "fbbf870eafc1ee5d4547fabbeea71778ddbe2166"
+dest_dir  <- rprojroot::find_package_root_file("inst/www/shared/bootstrap")
+
+js_url       <- sprintf("https://raw.githubusercontent.com/paypal/bootstrap-accessibility-plugin/%s/plugins/js/bootstrap-accessibility.js", sha)
```
wch, line 12: We don't need to download the un-minified JS, since we never use it.

tools/updateBootstrapAccessibilityPlugin.R
```diff
@@ -0,0 +1,34 @@
...
+min_js_url       <- sprintf("https://raw.githubusercontent.com/paypal/bootstrap-accessibility-plugin/%s/plugins/js/bootstrap-accessibility.min.js", sha)
+css_url       <- sprintf("https://raw.githubusercontent.com/paypal/bootstrap-accessibility-plugin/%s/plugins/css/bootstrap-accessibility.css", sha)
+url <- c(js_url, min_js_url, css_url)
+
+# Downloading each required file:
+lapply(url, function(x) {download.file(x, file.path(tempdir(), basename(x)))})
+
+# Copying js files:
+dir.create(file.path(dest_dir, "js"), recursive = TRUE)
```
wch, line 21: Can you put the files into `inst/www/shared/bootstrap/accessibility`? And no need for separate js and css subdirectories.

PR #2913, opened 2020-06-02
Trying to reconnect when $socket is null in ShinyApp.$sendMsg

apzentral:
I ran into this error below from time to time.

Uncaught TypeError: Cannot read property 'readyState' of null
    at ShinyApp.$sendMsg (shinyapp.js:288)
    at ShinyApp.sendInput (shinyapp.js:140)
    at InputBatchSender.$sendNow (input_rate.js:220)

Maybe in this method, we could check for $socket first, if it null then we try to reconnect.

2020-06-01 23_50_38-Window
This error cause the shiny application to have a grey overlay and no error message show. This cause some user doesn't know what to do with that error.

Thank you.

### DIFF
--- a/srcjs/shinyapp.js
+++ b/srcjs/shinyapp.js
@@ -282,6 +282,9 @@ var ShinyApp = function() {
   };
 
   this.$sendMsg = function(msg) {
+    if (!this.$socket) {
+      this.reconnect();
+    }
     if (!this.$socket.readyState) {
       this.$pendingMessages.push(msg);
     }


wch:
When do you encounter this error? Is it when the application starts, or after it's been running a while? Also, what serving platform do you use (like Shiny Server, Shiny Server Pro, Connect, etc)?

In general, I don't think it's wise for it to try to reconnect like this. We already have logic to handle reconnecting (https://shiny.rstudio.com/articles/reconnecting.html), so I would like to understand the conditions in which the disconnection is happening, and why the existing code is not working.

sada1993:
Experiencing the same issue on chrome, not experiencing the problem on Safari. Recreated the issue with any app from the Rstudio gallary, https://youtu.be/29ZsZ8VoF8I. 

Also noticed this strange behaviour with network activity:

network

JS-console

PR #2917, opened 2020-06-04
Improved accessibility for icon()

jooyoungseo:
* Added `role = "presentation"` to `icon()` to remove redundant semantic info for screen readers.
* Added appropriate labels to `icon()` element to provide screen-reader users with alternative descriptions of the `fontawesome`: `aria-label` is automatically applied based on the fontawesome name. For example, `icon("calendar")` will be announced as "calendar icon" to screen readers now.


R/bootstrap.R
```diff
@@ -1422,7 +1422,7 @@ icon <- function(name, class = NULL, lib = "font-awesome") {
   if (!is.null(class))
     iconClass <- paste(iconClass, class)
 
-  iconTag <- tags$i(class = iconClass)
+  iconTag <- tags$i(class = iconClass, role = "presentation", `aria-label` = paste(c(name, "icon"), collapse = " "))
```
wch, line 1425: A bit simpler to do:

```
paste(name, "icon")
```

Unless `name` can have more than one element, but I don't think that's the case here.

NEWS.md
```diff
@@ -1,3 +1,8 @@
+### Accessibility
+
+* Added appropriate labels to `icon()` element to provide screen-reader users with alternative descriptions of the `fontawesome`: `aria-label` is automatically applied based on the fontawesome name. For example, `icon("calendar")` will be announced as "calendar icon" to screen readers now.
```
wch, line 3: The label also applies to glyphicon, right?

R/bootstrap.R
jooyoungseo: Thanks! The reason why I did like this was because I wanted to address NULL name so that it can be just "icon" when no name is provided. However, that does not happen, right?

NEWS.md
jooyoungseo: Yes, correct! I have just pushed the updated `NEWS.md`.

R/bootstrap.R
jooyoungseo: Fixed!

PR #2920, opened 2020-06-09
Added lang attribute for accessibility and search-engine parsers

### LINKED ISSUE #2844: Shiny HTML element has no lang attribute
### System details
- Shiny 1.4.0.9002
- R 3.6.3

### Repro Steps
- Create a simple Shiny App following tutorials or use the default template created in the RStudio IDE (New Project / Shiny Web Application)
- Run the app and load into browser (I'm using Chrome on macOS)
- Examine the HTML or run an accessibility validation tool such as Axe (https://www.deque.com/axe/axe-for-web/)

### Problem
The HTML element has no `lang` attribute. This is a WCAG 2.1 A violation: https://www.w3.org/TR/WCAG21/#language-of-page

### Expected
Pages should be accessible by default. Thus, there should be a reasonable default, and the ability to specify a different language. For English:

`<html lang="en">`

This is critical for screen readers, otherwise you can have (for example) a Spanish user's screen reader trying to pronounce English words using Spanish language rules.

There was a community question on this:
https://community.rstudio.com/t/how-to-specify-the-language-to-a-shiny-app/26472

jooyoungseo:
* This PR closes #2844.
* Added the `lang` argument to `shinyApp()` that specifies document-level language within the app for the accessibility of screen readers and search-engine parsers.
* Users can pass [ISO 639-1 language code](https://www.w3schools.com/tags/ref_language_codes.asp) for this argument.
* Defaulting to `lang = ""` when no input is given by users. I was pondering the default  lang value set to either "en" or auto-locale-detect method; however, that can lead to unexpected issues. For example, many people use English R and locale while creating shiny apps in their own languages. We cannot assume only based upon users' locale which can lead to other accessibility side effects for screen readers. Thus, defaulting to empty string is a safer way.


Biggsen:
Would be lovely to get this in. Please fix the News conflict and merge! :)

R/app.R
```diff
@@ -96,6 +97,11 @@ shinyApp <- function(ui, server, onStart=NULL, options=list(),
   shinyOptions(appDir = getwd())
   appOptions <- consumeAppOptions()
 
+  # Also store `lang` option to be passed for rendering processors.
+  if(!is.null(lang) && is.character(lang)) {
```
cpsievert, line 101: Can `lang` be anything other than a string? If not, then when `is.character(lang)`, we should make sure it's length 1.

cpsievert:
@wch do you think it'd make more sense for lang to come in via `options` (so you can specify with the `ui.R`/`server.R` setup)?

wch:
Adding it via a parameter to `shinyApp()` doesn't seem right to me. Adding it to the UI object makes more sense to me. I think we could add a `lang` param to `bootstrapPage()` and maybe add it to the returned object as an attribute. Then, in the `renderPage()` function, it could try to extract the attribute's value.

One limitation of doing that, though, is that `lang` wouldn't get passed to the error page. How important is that? And does it even make sense, given that the text of the error page is in English?

inst/template/default.html
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html>
+<html lang = {{ lang }}>
```
wch, line 2: What happens if `lang` is `NULL`?

jooyoungseo: When it's NULL, lang is set to "". This means assisitve technologies and search parsers will use browsers' default language setting.

jooyoungseo:
Thanks very much for the insightful review, @wch! That  makes sense. We can fix lang to "en" for error page given that it is in English.

inst/template/default.html
wch: I think this needs quotes, as in:

```
<html lang = "{{ lang }}">
```

R/shinyui.R
```diff
@@ -25,6 +25,13 @@ withMathJax <- function(...) {
 }
 
 renderPage <- function(ui, showcase=0, testMode=FALSE) {
+  # Check if ui has lang attribute; otherwise, NULL
+  lang <- if ("lang" %in% names(attributes(ui))) {
+    attributes(ui)$lang
```
cpsievert, line 30:
```suggestion
    attr(ui, "lang")
```

cpsievert:
@jooyoungseo is there any added benefit to attaching `lang` to the root `<html>` elements as opposed to, say, `<head>` or `<body>`?

R/bootstrap.R
```diff
@@ -44,6 +45,19 @@ bootstrapPage <- function(..., title = NULL, responsive = NULL, theme = NULL) {
     ),
     bootstrapLib()
   )
+
+  # Also store `lang` to be passed for rendering processor.
+  if (!is.null(lang)) {
+    if (is.character(lang) && length(lang) == 1) {
+      # Append lang attribute to be passed to renderPage function
+      attributes(ui) <- c(attributes(ui), list(lang = lang))
```
wch, line 53: This can be:

```R
attr(ui, "lang") <- lang
```

R/bootstrap.R
```diff
@@ -44,6 +45,19 @@ bootstrapPage <- function(..., title = NULL, responsive = NULL, theme = NULL) {
...
+    } else {
+      warning("Invalid lang value: defaulting to empty string.")
+      attributes(ui) <- c(attributes(ui), list(lang = NULL))
```
wch, line 56: Can we remove this line? Then the behavior would be consistent with the case where `lang` actually is `NULL`.

R/bootstrap.R
```diff
@@ -159,22 +174,37 @@ basicPage <- function(...) {
...
+  theme = NULL, lang = NULL) {
 
   fillCSS <- tags$head(tags$style(type = "text/css",
     "html, body { width: 100%; height: 100%; overflow: hidden; }",
     sprintf("body { padding: %s; margin: 0; }", collapseSizes(padding))
   ))
 
-  if (isTRUE(bootstrap)) {
-    bootstrapPage(title = title, theme = theme, fillCSS, ...)
+  ui <- if (isTRUE(bootstrap)) {
```
wch, line 184: The overall logic would be clearer if this were structured like so:

```R

if (isTRUE(bootstrap)) {
  ui <- ....
} else {
  ui <- ....

  # Add attribute stuff here
}

File: R/bootstrap-layout.R [hunk 7]
```diff
@@ -118,6 +120,7 @@ fluidRow <- function(...) {
 #' @param theme Alternative Bootstrap stylesheet (normally a css file within the
 #'   www directory). For example, to use the theme located at
 #'   `www/bootstrap.css` you would use `theme = "bootstrap.css"`.
+#' @param lang ISO 639-1 language code used within the shiny app. Default is set to empty value.
```
wch, line 123: It would help to explain with examples:

> ISO 639-1 language code for the HTML page, such as "en" or "ko". This will be used as the `lang` in the `<html>` tag, as in `<html lang="en">`. The default (NULL) results in an empty string.

R/shinyui.R
```diff
@@ -25,6 +25,13 @@ withMathJax <- function(...) {
 }
 
 renderPage <- function(ui, showcase=0, testMode=FALSE) {
+  # Check if ui has lang attribute; otherwise, NULL
+  lang <- if ("lang" %in% names(attributes(ui))) {
```
wch, line 29: Simpler to do:

```R
lang <- attr(ui, "lang", exact = TRUE)
```

(This will default to `NULL` if there's no such attribute)

wch:
`basicPage` also needs the `lang` parameter.

jooyoungseo:
> @jooyoungseo is there any added benefit to attaching `lang` to the root `<html>` element instead of `<head>`? I mainly ask because it seems the implementation would be a lot easier we could just attach `lang` to `<head>`.

@cpsievert, thanks for asking!

Declaring lang globally in the `<html>` root is sort of conventional standar, which allows screen readers to recognize document-level/primary language.

[W3C spec](https://www.w3.org/International/articles/language-tags/) also recommends this:

> Use the lang attribute to specify language tags in HTML, and the xml ...

jooyoungseo:
I think [this failure](https://github.com/rstudio/shiny/runs/1053571269#step:14:95) is a false-positive error. If not, I will address this.

R/bootstrap.R
```diff
@@ -132,6 +145,9 @@ basicPage <- function(...) {
 #'   shown in the document).
 #' @param bootstrap If `TRUE`, load the Bootstrap CSS library.
 #' @param theme URL to alternative Bootstrap stylesheet.
+#' @param lang ISO 639-1 language code for the HTML page, such as "en" or "ko".
```
wch, line 148: We can use `@inheritParams` here

R/bootstrap.R
```diff
@@ -81,8 +94,8 @@ bootstrapLib <- function(theme = NULL) {
 
 #' @rdname bootstrapPage
 #' @export
-basicPage <- function(...) {
-  bootstrapPage(div(class="container-fluid", list(...)))
+basicPage <- function(..., lang = NULL) {
```
wch, line 97: I think we can skip this one, since `basicPage` is deprecated

R/bootstrap.R
```diff
@@ -44,6 +47,16 @@ bootstrapPage <- function(..., title = NULL, responsive = NULL, theme = NULL) {
     ),
     bootstrapLib()
   )
+
+  # Also store `lang` to be passed for rendering processor.
+  if (!is.null(lang)) {
```
wch, line 52: We can get rid of all these `if` statements and assign the attribute unconditionally

inst/template/default.html
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html>
+<html lang = "{{ lang }}">
```
wch, line 2: It would be a bit nicer if the whole `lang = "foo"` was conditional. I know that `lang=""` is valid, but it feels better to not have it at all in that case.

wch (approved):
Just a few small comments. @schloerke will take care of them.

schloerke:
Closing in favor of PR: https://github.com/rstudio/shiny/pull/3087

PR #2934, opened 2020-06-23
Combine documentation of renderPrint() and renderText()

hadley:
Since they're so closely related, and it makes it easier to see how they differ.


wch:
Looks good - the only issue is that the _pkgdown.yml file needs an update (see the failing check).

PR #2943, opened 2020-06-26
shinyAppTemplate()'s shinytest tests should be calling expect_pass()

cpsievert:


inst/app_template/tests/shinytest.R
```diff
@@ -1,3 +1,3 @@
 library(shinytest)
-shinytest::testApp("../")
+expect_pass(testApp("../", compareImages = FALSE))
```
cpsievert, line 2:
```suggestion
expect_pass(testApp("../", suffix = strsplit(utils::osVersion, " ")[[1]][1]))
```

wch: (Repeating discussion from elsewhere)
The overall idea is good, but I think that we should have a utility function in shinytest that provides the platform suffix. We'll also need to update the help for `shinyAppTemplate` to explain why the `suffix` is used in the template.

wch: As @schloerke mentioned elsewhere, `osVersion` was added only recently, in R 3.6. So we'll need something different here and/or in shinytest.

schloerke: Am now using `utils::sessionInfo()$running` in shinycoreci https://github.com/rstudio/shinycoreci/commit/43ac023e87f864469e9d9e22ed995ed51e823ec4

cpsievert: Maybe we should just copy the implementation of `rappdirs::get_os`?

```r
function () {
    if (.Platform$OS.type == "windows") {
        "win"
    }
    else if (Sys.info()["sysname"] == "Darwin") {
        "mac"
    }
    else if (.Platform$OS.type == "unix") {
        "unix"
    }
    else {
        stop("Unknown OS")
    }
}
```

wch: Holding off on this change until shinytest has a function for it.

wch (approved):

PR #2954, opened 2020-06-30
Remove context() calls from example app

wch:
`context()` is no longer recommended by the testthat package, so this PR removes it from the app template.

### DIFF
--- a/inst/app_template/tests/testthat/test-server.R
+++ b/inst/app_template/tests/testthat/test-server.R
@@ -1,5 +1,3 @@
-context("app")
-
 testServer(expr = {
   # Set the `size` slider and check the output
   session$setInputs(size = 6)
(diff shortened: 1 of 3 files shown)


inst/app_template/tests/testthat/test-sort.R
```diff
@@ -1,6 +1,4 @@
 # Test the lexical_sort function from R/example.R
```
hadley, line 1: We also recommend naming the test file to match the R file so this should be `test-example.R`

inst/app_template/tests/testthat/test-server.R
```diff
@@ -1,5 +1,3 @@
-context("app")
-
 testServer(expr = {
```
hadley, line 1: Why doesn't this have a first argument?

hadley: Why not make it `testServer(".", {`?

PR #2950, opened 2020-06-29
More themable selectInput()

cpsievert:
This PR allows `selectInput()` to be styled via {bootstraplib}  (see also https://github.com/rstudio/shiny/pull/2964 and https://github.com/rstudio/shiny/pull/2958)

## Testing notes

Install `remotes::install_github("rstudio/shiny")`. Now if you run the following app you should see a dropdown with the same colors/fonts as the screenshot. Now change `bs_theme_new()` to `bs_theme_new(3)` and re-run the app. You should see a similar result.

```r
library(shiny)
library(bootstraplib)
shinyOptions(bootstraplib = TRUE)

bs_theme_new()
bs_theme_base_colors(bg = "#002B36", fg = "#EEE8D5")
bs_theme_accent_colors(primary = "#2AA198")

ui <- fluidPage(
  tags$h4("With selectize"),
  selectInput("single", "Choose single", choices = state.abb),
  selectInput("multiple", "Choose multiple", choices = state.abb, multiple = TRUE),
  tags$h4("Without selectize"),
  selectInput("single2", "Choose single", choices = state.abb, selectize = FALSE),
  selectInput("multiple2", "Choose multiple", choices = state.abb, multiple = TRUE, selectize = FALSE),
)

shinyApp(ui, function(input, output) {})
```

Screen Shot 2020-08-17 at 1 52 29 PM

PR #2958, opened 2020-07-02
More themable sliderInput()

cpsievert:
This PR:

* Upgrades ion.rangeSlider library from 2.1.6 to 2.3.1.
* Allows `sliderInput()` to be styled via `{bootstraplib}`  (see also https://github.com/rstudio/shiny/pull/2964 and https://github.com/rstudio/shiny/pull/2950)

## Testing notes

Install `remotes::install_github("rstudio/shiny")`. Now if you run the following app you should see a slider with the same colors/fonts as the screenshot. Now change `bs_theme_new()` to `bs_theme_new(3)` and re-run the app. You should see a similar result.

```r
library(shiny)
library(bootstraplib)

bs_theme_new()
bs_theme_base_colors(bg = "#002B36", fg = "#EEE8D5")
bs_theme_accent_colors(primary = "#2AA198")
bs_theme_fonts("Pacifico")

ui <- fluidPage(
  tags$head(tags$link(rel="stylesheet", href="https://fonts.googleapis.com/css?family=Pacifico")),
  sliderInput("demo", "Demo", min = 0, max = 100, value = c(30, 70))
)

```

Screen Shot 2020-08-26 at 4 11 24 PM


R/input-slider.R
```diff
@@ -79,8 +88,9 @@ sliderInput <- function(inputId, label, min, max, value, step = NULL,
                         round = FALSE, format = NULL, locale = NULL,
                         ticks = TRUE, animate = FALSE, width = NULL, sep = ",",
                         pre = NULL, post = NULL, timeFormat = NULL,
-                        timezone = NULL, dragRange = TRUE)
-{
+                        timezone = NULL, dragRange = TRUE,
+                        skin = c("shiny", "big", "flat", "modern", "round", "sharp", "square"),
```
cpsievert, line 92: Perhaps the ordering here should reflect our reflect our preferences (i.e., flat and round are pretty nice)

R/input-slider.R
```diff
@@ -227,6 +282,164 @@ hasDecimals <- function(value) {
...
+      "line_color_1"      = "$white",
+      "line_color_2"      = "gray('200')",
+      "line_color_3"      = "gray('100')",
+      "handle_color_1"    = "gray('200')",
+      "handle_color_2"    = "gray('300')",
+      "handle_color_3"    = "$white",
+      "handle_color_4"    = "gray('400')",
+      "minmax_text_color" = "$white",
+      "minmax_bg_color"   = "gray('500')",
+      "label_color_2"     = "$white"
```
cpsievert, line 354: Turns out that Bootswatch doesn't set `$white`/`$black` to true "bg"/"fg" colors, so I think we should replace `$white` with `$body-bg` and `$black` with `$body-color`

R/input-slider.R
```diff
@@ -227,6 +282,164 @@ hasDecimals <- function(value) {
...
+      "handle_color_3" = accentColorLight,
+      "label_color_1"  = accentColor
+    ),
+    stop("Couldn't find skin named '", skin, "'", call. = FALSE)
+  )
+
+  lapply(vars, function(x) paste(x, "!default"))
+}
+
+# TODO: BS3 version!
```
cpsievert, line 337: Implement this

DESCRIPTION
```diff
@@ -103,7 +103,10 @@ Suggests:
     future,
     dygraphs,
     ragg,
-    showtext
+    showtext,
+    sass,
+    bootstraplib
+Remotes: rstudio/bootstraplib
```
wch, line 109: Also need `rstudio/sass`

DESCRIPTION
```diff
@@ -103,7 +103,10 @@ Suggests:
     future,
     dygraphs,
     ragg,
-    showtext
+    showtext,
+    sass,
+    bootstraplib
```
wch, line 108: Need version requirement

R/bootstrap.R
```diff
@@ -42,10 +42,37 @@ bootstrapPage <- function(..., title = NULL, responsive = NULL, theme = NULL) {
       # remainder of tags passed to the function
       list(...)
     ),
-    bootstrapLib()
+    if (useBsTheme()) bootstraplib::bootstrap() else bootstrapLib()
```
wch, line 45: Move this logic into `bootstrapLib()`

cpsievert: Done in 61509ccbc55900bad795eb3d320f0adddb2dc298

R/bootstrap.R
```diff
@@ -42,16 +42,60 @@ bootstrapPage <- function(..., title = NULL, responsive = NULL, theme = NULL) {
...
+#' [basicPage()], [navbarPage()], etc) implicitly depend on Bootstrap.
+#'
+#' @param theme deprecated (it never worked).
+#' @export
+#' @keywords internal
+bootstrapLib <- function(theme = NULL) {
+  if (!is.null(theme)) {
+    stop("shiny::bootstrapLib()'s theme argument never worked as intended, don't use it.")
+  }
+  shinyDeprecated("bootstraplib::bootstrap()", old = "shiny::bootstrapLib()")
```
wch, line 64: I think using `shinyDeprecated` is too aggressive -- people who are using HTML templates will upgrade their packages or install them on a new system, and all of a sudden they'll start getting the deprecation message and be alarmed or confused by it, and they'll feel like it's kind of an emergency and they have to fix their apps right away. (I've had that happen with other packages and it's always a disconcerting experience.)  I think it's better to just allow it to work without a message, and update our docs to suggest using `bootstraplib::bootstrap()`.

We could leave this in, but comment it out, and add a comment with the current date and a reminder that we can un-comment the line in, say a year or two.

R/bootstrap.R
```diff
@@ -42,16 +42,60 @@ bootstrapPage <- function(..., title = NULL, responsive = NULL, theme = NULL) {
...
+#' CSS. Note that most users won't need to call this function directly since
+#' [bootstrapPage()] and it's special cases ( e.g., [fluidPage()],
+#' [basicPage()], [navbarPage()], etc) implicitly depend on Bootstrap.
+#'
+#' @param theme deprecated (it never worked).
+#' @export
+#' @keywords internal
+bootstrapLib <- function(theme = NULL) {
+  if (!is.null(theme)) {
+    stop("shiny::bootstrapLib()'s theme argument never worked as intended, don't use it.")
```
wch, line 62: I think this should be a message -- if people are currently using it, no need to make their app stop working.

R/bootstrap.R
cpsievert: This deprecation message will only happen if users are calling `bootstrapLib()` directly (internally we use `bootstrapLib_()`  which won't throw the message -- I don't think that's too aggressive?)

R/bootstrap.R
cpsievert: Ok, yea.

R/bootstrap.R
wch: In our docs for HTML templates, we suggest that they call `bootstrapLib()`, and there are probably a lot of apps out there that do this.
https://shiny.rstudio.com/articles/templates.html

wch (approved):

PR #2964, opened 2020-07-08
More themable dateInput()/dateRangeInput()

cpsievert:
This PR:

* Upgrades bootstrap-datepicker from 1.6.4 to 1.9.0
* Allows `dateInput()` and `dateRangeInput()` to be styled via `{bootstraplib}` (see also https://github.com/rstudio/shiny/pull/2958 and https://github.com/rstudio/shiny/pull/2950)

## Testing notes

Install `remotes::install_github("rstudio/shiny")`. Now if you run the following app you should see date input(s) with the same colors/fonts as the screenshot. Now change `bs_theme_new()` to `bs_theme_new(3)` and re-run the app. You should see a similar result.

```r
library(shiny)
library(bootstraplib)

bs_theme_new()
bs_theme_base_colors("black", "white")
bs_theme_accent_colors("violet")

ui <- fluidPage(
  dateInput("date", "Date", value = Sys.Date()),
  dateRangeInput("date", "Date", start = Sys.Date(), end = Sys.Date() + 7),
)
```

Screen Shot 2020-07-08 at 4 06 13 PM


R/input-date.R
```diff
@@ -124,19 +125,103 @@ dateInput <- function(inputId, label, value = NULL, min = NULL, max = NULL,
...
+datePickerDependencies <- function(sassVariables = list()) {
+
+  if (useBsTheme() || length(sassVariables) == 0) {
+    scssDir <- system.file(package = "shiny", "www", "shared", "datepicker", "scss")
+    tmpDir <- tempfile("datepicker-scss")
+    dir.create(tmpDir)
+    outFile <- file.path(tmpDir, "custom.css")
+    sassFunc <- if (useBsTheme()) bootstraplib::bootstrap_sass else sass::sass
+    defaults <- list(
+      sassVariables %OR% "",
```
wch, line 141: Is the `%OR%` ever needed here?

R/input-date.R
```diff
@@ -124,19 +125,103 @@ dateInput <- function(inputId, label, value = NULL, min = NULL, max = NULL,
...
+      bsThemeDateDefaults()
+    )
+    sassFunc(
```
wch, line 144: I want to make sure I understand here. The documentation for how `sass()` deals with nested lists isn't totally clear to me, so I tried this out:

```R
sass::sass(
  list(
    list(list(a=1), list(a=2)),
    "x {a: $a; }"
  )
)
#> /* CSS */
#> x {
#>   a: 2;
#> }
```

Does that mean that the `bsThemeDateDefaults()` will override `sassVariables`?

R/input-date.R
```diff
@@ -124,19 +125,103 @@ dateInput <- function(inputId, label, value = NULL, min = NULL, max = NULL,
...
+      list(
+        defaults,
+        sass::sass_file(file.path(scssDir, "build3.scss"))
+      ),
+      output = outFile
+    )
+    # Hash the skin and defaults so that the html dependency from one date-picker won't stomp another
+    cssDependency <- htmlDependency(
+      paste0("datepicker-datepicker-css-", digest::digest(defaults, "xxhash64")),
```
wch, line 153: Maybe this `digest()` strategy can be used for the temp dir creation. One quick thought: maybe a directory for run-time-compiled CSS resources can be created for an app, and then when the app stops, the directory could be removed.

R/input-date.R
```diff
@@ -124,19 +125,103 @@ dateInput <- function(inputId, label, value = NULL, min = NULL, max = NULL,
...
+      version = "1.9.0",
```
wch, line 154: Now that I see all these copies of the version number, it makes me think that we maybe should have the update script automatically write the version number in a file named `aaa_lib_versions.R`, with something like:

```R
bootstrap_datepicker_version <- "1.9.0"
```

And then use that variable here. We've certainly forgotten to update JS/CSS library version numbers in the past.

R/input-date.R
```diff
@@ -124,19 +125,103 @@ dateInput <- function(inputId, label, value = NULL, min = NULL, max = NULL,
...
+     </script>"
+    )
+  )
+}
+
+
+
+bsThemeDateDefaults <- function() {
+  if (!useBsTheme()) {
+    return("")
```
wch, line 188: Would a `list()` make more sense, to be consistent with the other possible return values?

R/input-daterange.R
```diff
@@ -78,7 +78,7 @@
 dateRangeInput <- function(inputId, label, start = NULL, end = NULL,
     min = NULL, max = NULL, format = "yyyy-mm-dd", startview = "month",
     weekstart = 0, language = "en", separator = " to ", width = NULL,
-    autoclose = TRUE) {
+    autoclose = TRUE, sassVariables = list()) {
```
wch, line 81: Same as comment above, about `sassVars` and possibly adding `...`.

R/input-date.R
```diff
@@ -94,7 +94,8 @@
 dateInput <- function(inputId, label, value = NULL, min = NULL, max = NULL,
   format = "yyyy-mm-dd", startview = "month", weekstart = 0,
   language = "en", width = NULL, autoclose = TRUE,
-  datesdisabled = NULL, daysofweekdisabled = NULL) {
+  datesdisabled = NULL, daysofweekdisabled = NULL,
+  sassVariables = list()) {
```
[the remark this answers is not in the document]
cpsievert: Yea, I'm not in love with the argument name. I was also thinking maybe `defaults` since it's intended for variable defaults and matches naming used in `sass::sass_layer()`.

I also really like the idea of adding `...`. Also, FWIW, I could maybe be swayed into not including the argument at all for this input, but it seems like it could be quite useful for `sliderInput()` and `selectInput()`,  so I think for consistency sake we should take similar approaches in all of them

R/input-date.R
```diff
@@ -124,19 +125,103 @@ dateInput <- function(inputId, label, value = NULL, min = NULL, max = NULL,
...
-  var datepicker = $.fn.datepicker.noConflict();
-  $.fn.bsDatepicker = datepicker;
-})();
-</script>"
-)
+datePickerDependencies <- function(sassVariables = list()) {
+
+  if (useBsTheme() || length(sassVariables) == 0) {
+    scssDir <- system.file(package = "shiny", "www", "shared", "datepicker", "scss")
+    tmpDir <- tempfile("datepicker-scss")
```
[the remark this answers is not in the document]
cpsievert: Yea, and I think that's necessary if we want to use `htmlDependency()`s? Just FYI, this is also how  `bootstraplib::bootstrap()` works

R/input-date.R
cpsievert: Oh, I guess not

R/input-date.R
cpsievert: Ahhh, yea, good catch...I was originally thinking that `bsThemeDateDefaults()` would include `!default` flags (that's how `sliderInput()` and `selectInput()` will work), but that doesn't actually work for bootstrap-datepicker since its Sass forces you to override already existing Bootstrap Sass variables. Thus, I think we should switch the ordering and explain in the documentation that `sassVariables` shouldn't include `!default` flags

R/input-date.R
```diff
@@ -94,8 +100,7 @@
 dateInput <- function(inputId, label, value = NULL, min = NULL, max = NULL,
   format = "yyyy-mm-dd", startview = "month", weekstart = 0,
   language = "en", width = NULL, autoclose = TRUE,
-  datesdisabled = NULL, daysofweekdisabled = NULL,
-  sassVariables = list()) {
+  datesdisabled = NULL, daysofweekdisabled = NULL, ..., sassVars = list()) {
```
wch, line 103: Should also call `ellipsis::check_dots_empty()` in this function to make sure there are no `...` args.

R/aaa_lib_versions.R
```diff
@@ -0,0 +1 @@
+datePickerVersion <- "1.9.0"
```
wch, line 1: Can you update this from the `updateBootstrapDatepicker.R` script? Or at least modify that script to print out a message saying to update this value. (But just updating this file automatically doesn't seem like much more work.)

R/input-date.R
cpsievert: Do we already depend on that package? If not, we could also implement our own?

wch: We don't import it yet, but the package is lightweight and maintained by Hadley and Lionel, and it's used in tidyverse packages, so I think it's OK to bring it in.

R/input-date.R
```diff
@@ -124,19 +130,98 @@ dateInput <- function(inputId, label, value = NULL, min = NULL, max = NULL,
...
+    "btn-primary-color" = "color-yiq($brand-primary)",
+    "btn-primary-color-hover" = "color-yiq($gray-lighter)",
+    "btn-primary-bg" = "$brand-primary",
+    "btn-primary-border" = "darken($btn-primary-bg, 5%)",
+    "btn-link-disabled-color" = "$gray-light",
+    "state-info-bg" = "mix($white, $brand-primary, 20%)",
+    "line-height-base" = "20/14",
+    "border-radius-base" = "4px",
+    "dropdown-bg" = "$white",
+    "dropdown-border" = "mix($black, $white, 15%)"
```
cpsievert, line 225: Turns out that Bootswatch doesn't set `$white`/`$black` to true bg/fg colors, so I think we should replace `$white` with `$body-bg` and `$black` with `$body-color`

R/input-date.R
cpsievert: No longer relevant now that we're no longer doing `sassVars`

R/input-date.R
cpsievert: (The other theming PRs suffer from the same issue....maybe we should "shim" bootswatch sass variables in bootstraplib so that they're structured more like `bs_theme_base_colors()`'s model

R/bootstrap.R
```diff
@@ -66,8 +66,6 @@ bootstrapLib <- function(theme = NULL) {
   if (!is.null(theme)) {
     message("shiny::bootstrapLib()'s theme argument never worked as intended, don't use it.")
   }
-  # Eventually this should be deprecated in favor of bootstraplib::bootstrap()
```
wch, line 69: I don't think we want to remove these comments.

wch (approved):

PR #2965, opened 2020-07-09
Allow passing server function directory to testServer()

hadley:
Includes:

* Pulling out withMockContext() to make flow of function easier to understand

* Eliminated tryCatch() in favour of on.exit()

* Clarified expected inputs in documentation

* Removed default for `app` argument to make it more clear that this should usually be provided

* Added tests for the error messages

If the basic approach looks good to you, I think it also needs:

* Deprecate the existing module interface; now it's just easier to call the module yourself to generate a server function.
* Update the example
* Update the test


R/test-server.R
```diff
@@ -55,75 +58,78 @@ isModuleServer <- function(x) {
 #'   # Any additional arguments, below, are passed along to the module.
 #' })
 #' @export
-testServer <- function(app = NULL, expr, args = list(), session = MockShinySession$new()) {
+testServer <- function(app, expr, args = list(), session = MockShinySession$new()) {
```
wch, line 61: We had it default to `NULL` so that in the case where you're testing an application on disk with `runTests()`, it just tests the server function from the current app. This is done in the app template (ignore the stuff with `{{`, since that's templating code): https://github.com/rstudio/shiny/blob/master/inst/app_template/tests/testthat/test-server.R#L3 

Does this now require users to explicitly use `app=NULL` if they want to test the current app?

hadley: Yes, it would. I'm happy to change the default back, but I personally think the most common use of this function will be in regular unit test files 😄

hadley:
This is ready for another review:

* I've reverted to the previous default for `app`, and fixed a silly bug so it now works again
* Updated all the tests to avoid now unnecessary creation of a module
* Added a simple example to the `testServer()` docs
* Backed off from my idea that passing a module function could be deprecated; for some reason I was thinking that calling a module server function returned a server function, but that's clearly not true.

R/test-server.R
```diff
@@ -1,19 +1,22 @@
-#' @noRd
 isModuleServer <- function(x) {
   is.function(x) && names(formals(x))[1] == "id"
 }
 
+isServer <- function(x) {
+  is.function(x) && all(c("input", "output", "session") %in% names(formals(x)))
```
wch, line 6: Might be good to check that input, output, and session are also the first three arguments.

R/test-server.R
```diff
@@ -1,19 +1,22 @@
...
 #' interactions are defined in the server function of applications and in
 #' modules.
-#' @param app The path to an application or module to test. In addition to
-#'   paths, applications may be represented by any object suitable for coercion
-#'   to an `appObj` by [`as.shiny.appobj`]. Application server functions must
-#'   include a `session` argument in order to be tested. If `app` is `NULL` or
-#'   not supplied, the nearest enclosing directory that is a Shiny app, starting
-#'   with the current directory, is used.
+#' @param app A server function (i.e. a function with `input`, `output`,
+#'   and `session()`), or a module function (i.e. a function with first
```
wch, line 15: No `()`

R/test-server.R
hadley: Done

R/test-server.R
hadley: Ooops, fixed.

PR #2968, opened 2020-07-14
Allow function-based UI to return a complete HTTP response

jcheng5:
This allows the UI handler to have total control over the response,
including status code (for redirects) or content type (for serving
up files).

## Testing

This app should show the R logo. Without this PR, you get a broken image and an error at the console.

```r
library(shiny)

ui <- function(req) {
  if (req[["PATH_INFO"]] == "/") {
    tags$img(src = "image.jpg")
  } else if (req[["PATH_INFO"]] == "/image.jpg") {
    path <- file.path(R.home("doc"), "html", "logo.jpg")
    httpResponse(
      status = 200L,
      content_type = "image/jpg",
      content = readBin(path, what = raw(0), n = file.info(path)$size)
    )
  }
}

}

shinyApp(ui, server, uiPattern = function(req) {
  req[["PATH_INFO"]] %in% c("/", "/image.jpg")
})
```


jcheng5:
Might be able to get away with `uiPattern = ""` and return `NULL` from the ui function.

jcheng5:
Replaced by #2970

PR #2969, opened 2020-07-15
Calling `runApp("app.R")` ignored options passed into `shinyApp()`

### LINKED ISSUE #1942: Calling `runApp("app.R")` ignores options passed into `shinyApp()`
When developing locally, I sometimes specify a port in the `options` argument of `shinyApp()` so that I can make sure I get a consistent URL every time I restart my session.

```
library(shiny)

ui <- fluidPage(
  "hello"
)

}

shinyApp(ui, server, options = list(port = 5900))
```

Running this code manually works fine. But clicking on the green "Run App" button in RStudio seems to ignore the options list, and it runs on a different port each time. The same is true for using other options: for example, setting the display.mode will also be ignored with the "Run App" button.

I'm not sure if this is a feature or a bug, and it's very possible that it's been reported before but I couldn't find any traces of it

jcheng5:
Fixes #1942.

This was caused by shinyAppDir_appR (which was called by shinyAppDir
and shinyAppFile, which were called when runApp() was used with either
no args or a path) dropping shinyApp() options on the floor.

Fixing this was not as simple as not dropping those options. Before
this commit, the app.R file was not invoked until after the host and
port options (possibly others) were already finalized. This commit
changes the effective sequence of events during startup.

- Calling shinyAppFile() or shinyAppDir() that points to a single-file
  app, now causes the app.R file to be sourced immediately; previously,
  sourcing would happen only at onStart(). (Honestly, the new behavior
  seems less surprising anyway.)

- The support files (R/*.R) for app.R were sourced during onStart. I've
  moved this so that the support files are sourced right before app.R
  is, and re-sourced every time app.R reloads.

## Testing notes

Save the following as app.R somewhere (it is important that you actually save it). Then hit Run App in RStudio.

```r
library(shiny)

ui <- fluidPage(
  
)

}

shinyApp(ui, server, options = list(port = 8100))
```

Without this commit, a random port number is picked (`Listening on http://127.0.0.1:4892` or whatever). With this commit, port 8100 is used.


tests/testthat/test-app.R
```diff
@@ -210,3 +210,46 @@ test_that("global.R and sources in R/ are sourced in the app directory", {
   # Set by ../test-helpers/app1-standard/R/helperCap.R
   expect_equal(normalizePath(appEnv$source_wd), normalizePath(appDir))
 })
+
+test_that("Setting options in various places works", {
```
jcheng5, line 214: Guarantee that launch.browser is always FALSE (for CRAN and CI)

tests/test-helpers/app7-port/option.R
```diff
@@ -0,0 +1,13 @@
+library(shiny)
+
+options(shiny.port = 7777)
```
wch, line 3: Need to make sure this gets unset. In `onStop()`?

R/app.R
```diff
@@ -480,7 +481,7 @@ shinyAppDir_appR <- function(fileName, appDir, options=list())
       serverFuncSource = dynServerFuncSource,
       onStart = onStart,
       onStop = onStop,
-      options = options
+      options = joinOptions(appObjOptions, options)
```
wch, line 484: You should be able to use `mergeVectors`.

tests/testthat/test-app.R
```diff
@@ -249,7 +251,13 @@ test_that("Setting options in various places works", {
 
   # Calls to options(shiny.port = xxx) within app.R should also work reliably
   expect_port(runApp(file.path(appDir, "option.R")), 7777)
-  # But still overrideable
+  # Ensure that option was unset
+  expect_null(getOption("shiny.port"))
```
wch, line 255: This test makes the assumption that the starting value for `shiny.port` was NULL, but the app itself restores whatever the starting value was, even if it was something other than NULL.

For these tests, maybe set `shiny.port` to NULL before doing them, or save the original value for `shiny.port`, and compare to that in the tests?

PR #2970, opened 2020-07-15
Allow function-based UI to return a complete HTTP response (2nd try)

jcheng5:
Replaces #2968; this PR has a simpler API with a less invasive implementation.

This allows the UI handler to have total control over the response,
including status code (for redirects) or content type (for serving
up files).

## Testing

This app should show the R logo. Without this PR, you get a broken image and an error at the console.

```r
library(shiny)

ui <- function(req) {
  if (req[["PATH_INFO"]] == "/") {
    tags$img(src = "image.jpg")
  } else if (req[["PATH_INFO"]] == "/image.jpg") {
    path <- file.path(R.home("doc"), "html", "logo.jpg")
    httpResponse(
      status = 200L,
      content_type = "image/jpg",
      content = readBin(path, what = raw(0), n = file.info(path)$size)
    )
  } else {
    NULL
  }
}

}

shinyApp(ui, server, uiPattern = ".*")
```


NEWS.md
```diff
@@ -6,6 +6,12 @@ shiny 1.5.0.9000
 
 * Added appropriate labels to `icon()` element to provide screen-reader users with alternative descriptions for the `fontawesome` and `glyphicon`: `aria-label` is automatically applied based on the fontawesome name. For example, `icon("calendar")` will be announced as "calendar icon" to screen readers. "presentation" aria role has also been attached to `icon()` to remove redundant semantic info for screen readers.
 
+### Minor new features and improvements
+
+* When UI is specified as a function (e.g. `ui <- function(req) { ... }`), the
```
wch, line 11: This is a tiny thing, but can you not use hard line breaks here?

R/shinyui.R
```diff
@@ -150,8 +152,15 @@ uiHttpHandler <- function(ui, uiPattern = "^/$") {
     if (is.null(uiValue))
       return(NULL)
 
-    renderPage(uiValue, textConn, showcaseMode, testMode)
-    html <- paste(readLines(textConn, encoding = 'UTF-8'), collapse='\n')
-    return(httpResponse(200, content=enc2utf8(html)))
+    if (inherits(uiValue, "httpResponse")) {
+      return(uiValue)
+    } else {
+      textConn <- file(open = "w+")
```
wch, line 158: Looking at this code, I realized how awkward and unnecessary all this `textConn` stuff is. This function calls `renderPage()`, which calls `renderDocument()`. That function actually returns a string, but then then `renderPage` takes that string and writes it to a connection, which we read from here and then paste together into a string. We might as well take out the text connection entirely. But that can be another PR if you don't want to do it here.

wch:
Do you mind adding a simple test app to https://github.com/rstudio/shinycoreci-apps/tree/master/apps ?

R/shinyui.R
```diff
@@ -150,8 +152,15 @@ uiHttpHandler <- function(ui, uiPattern = "^/$") {
...
+      on.exit(close(textConn))
```
cpsievert, line 159:
```suggestion
      on.exit(close(textConn), add = TRUE)
```

R/shinyui.R
```diff
@@ -77,7 +77,7 @@ renderPage <- function(ui, connection, showcase=0, testMode=FALSE) {
   }
 
   html <- renderDocument(ui, shiny_deps, processDep = createWebDependency)
-  writeUTF8(html, con = connection)
+  enc2utf8(paste(collapse = "\n", html))
```
wch, line 80: I'm pretty sure `renderDocument` returns a UTF-8 encoded string vector with one element (and class `"html"`), so you should be able to just call `as.character` (to remove class and other attributes) on the value of `renderDocument()`.

https://github.com/rstudio/htmltools/blob/a8025f360/R/template.R#L134-L136

Example:

```R
library(htmltools)
library(shiny)

ui <- htmlTemplate(
  system.file("template", "default.html", package = "shiny"),
  body = fluidPage(div("hello world"))
)

shiny_deps <- list(
  htmlDependency("json2", "2014.02.04", c(href="shared"), script = "json2-min.js"),
  htmlDependency("shiny", utils::packageVersion("shiny"), c(href="shared"),
    script = if (getOption("shiny.minified", TRUE)) "shiny.min.js" else "shiny.js",
    stylesheet = "shiny.css")
)

res <- renderDocument(ui, shiny_deps)
length(res)
#> [1] 1
str(res)
#>  'html' chr "<!DOCTYPE html>\n<html>\n<head>\n  <meta http-equiv=\"Content-Type\" content=\"text/html; charset=utf-8\"/>\n  "| __truncated__
#>  - attr(*, "html")= logi TRUE
```

jcheng5: TBH I'd rather keep it the way I have it just to be safe, enc2utf8 and paste should have no perf impact if it's already a UTF-8 encoded single element character.

jcheng5: (@wch told me offline it's ok to resolve)

jcheng5:
Added test https://github.com/rstudio/shinycoreci-apps/pull/69

Issue #1791, opened 2017-07-13
Investigate freeze/thaw behavior inconsistencies


jcheng5:
1. reactiveVal thaws on onFlushed, reactiveValues... doesn't?
2. reactiveVal doesn't thaw on set, seems like both should (don't know if reactiveValues does)
3. Should we auto-unset all input$setMeta for a given slot when that slot is written to?

jcheng5:
Here's an example of `freezeReactiveValue` not behaving correctly with a `textInput` that is re-rendered with its own value:

```r
library(shiny)

ui <- fluidPage(
  uiOutput("ui"),
  actionButton("redraw", "Redraw"),
  p("Status:", textOutput("status", inline = TRUE))
)

server <- function(input, output, session) {
  output$ui <- renderUI({
    input$redraw
    freezeReactiveValue(input, "outval")
    textInput("outval", "Dynamic input", value = "hello")
  })
  
  output$status <- renderText({
    input$redraw
    tryCatch({
      input$outval
      "Thawed"
    }, error = function(err) {
      "Frozen"
    })
  })
}

shinyApp(ui, server)
```

If you click "Redraw" without changing the value from "hello", then the output only sees the input as frozen. The correct behavior would be for the input to only be frozen for a brief moment (the time between `output$ui` being refreshed and the text input being re-bound). The reason for this is because when the text input is re-bound, its value is the same as the last known value for the `"outval"` input, so nothing is sent to the server.

One solution for this would be for every `bindInput` to send values to the server regardless of whether it's a duplicate of the last value or not.

jcheng5:
Similar problem as my previous comment, but this time with `updateXXXInput`:

```r
library(shiny)

ui <- fluidPage(
  textInput("outval", "Dynamic input", value = "hello"),
  actionButton("update", "Update"),
  p("Status:", textOutput("status", inline = TRUE))
)

server <- function(input, output, session) {
  observeEvent(input$update, {
    freezeReactiveValue(input, "outval")
    updateTextInput(session, "outval", value = "hello")
  })

  output$status <- renderText({
    input$update
    tryCatch({
      input$outval
      "Thawed"
    }, error = function(err) {
      "Frozen"
    })
  })
}

shinyApp(ui, server)
```

The previous solution I proposed would not help for this case.

fweber144:
Isn't this fixed and could be closed?

PR #2975, opened 2020-07-17
When initializing inputs, make sure their values actually get to server

jcheng5:
Partial fix for #1791

Also necessary for some client-side validation logic I'm developing, details to come


jcheng5:
After discussing with @wch, this seems overbroad. In cases where `freezeReactiveValue` wasn't used, this would cause unnecessary invalidations.

I'll create a new PR with a more surgical fix: when `freezeReactiveValue` is applied to an input, the client is notified and flags that input. The next time the input is set, deduplication will not apply. This should fix both `renderUI` and `updateXXXInput` cases at once.

jcheng5:
Superceded by https://github.com/rstudio/shiny/pull/3055

PR #2978, opened 2020-07-22
* Fixed #2951: Added labels and title for dateInput and dateRangeInput

### LINKED ISSUE #2951: dateInput lacks semantic accessibility
### System details

* Browser Version: any

* Session info:

``` r
xfun::session_info("shiny")
#> R version 4.0.2 (2020-06-22)
#> Platform: x86_64-w64-mingw32/x64 (64-bit)
#> Running under: Windows 10 x64 (build 19041)
#> 
#> Locale:
#>   LC_COLLATE=English_United States.1252 
#>   LC_CTYPE=English_United States.1252   
#>   LC_MONETARY=English_United States.1252
#>   LC_NUMERIC=C                          
#>   LC_TIME=English_United States.1252    
#> system code page: 949
#> 
#> Package version:
#>   base64enc_0.1.3   BH_1.72.0.3       commonmark_1.7    crayon_1.3.4     
#>   digest_0.6.25     fastmap_1.0.1     glue_1.4.1        graphics_4.0.2   
#>   grDevices_4.0.2   htmltools_0.5.0   httpuv_1.5.4      jsonlite_1.7.0   
#>   later_1.1.0.1     magrittr_1.5      methods_4.0.2     mime_0.9         
#>   promises_1.1.1    R6_2.4.1          Rcpp_1.0.4.6      rlang_0.4.6      
#>   shiny_1.5.0       sourcetools_0.1.7 stats_4.0.2       tools_4.0.2      
#>   utils_4.0.2       withr_2.2.0       xtable_1.8.4
```

<sup>Created on 2020-06-30 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

### Example application

``` r
library(shiny)

example(dateInput)
```

1. Each input label element is annotated with `for=` attribute but refers to the `<div>` that contains both the label and the following textbox, which means the textbox itself has no accessible label. So that's messed up for assistive technology users.
2. In the pop-up calendar grid, the three buttons (previous, next, and datepicker switch) are completely unannotated and inaccessible for screen readers.
3. No semantic hint is presented to screen readers when the calendar popup appears and disappears.

Greatly assisted by @@gtritchie in identifying those issues.

jooyoungseo:
* Fixed #2951: screen readers correctly announce labels and date formats for `dateInput()` and `dateRangeInput()` widgets.
* Extended `shinyInputLabel()` function to produce `id` attribute to be matched with `aria-labelledby` for accessibility: This will be playing a great role in triggering appropriate label info for other  accessibility fixes moving forward.
* Added a safe workaround to avoid doubled label issues for dateInput and dateRangeInput using `aria-labelledby`.
* Tested with NVDA and JAWS screen readers for this fix, and confirmed that this works perfectly.

P.S.: accessibility fixes for the datepicker-dropdown calendar will have to be done via a separate PR later on.


R/input-daterange.R
```diff
@@ -118,6 +122,10 @@ dateRangeInput <- function(inputId, label, start = NULL, end = NULL,
         tags$input(
           class = "form-control",
           type = "text",
+          # `aria-labelledby` attribute is required for accessibility to avoid doubled labels (#2951).
+          `aria-labelledby` = paste0(inputId, "-label"),
+          # title attribute is announced for screen readers for date format.
+          title = paste("format:", format),
```
cpsievert, line 128: Would it be helpful for this to be `Date format:` instead of `format:`? That is, is it already obvious that this is a _date_ input?

R/input-utils.R
```diff
@@ -3,6 +3,8 @@ shinyInputLabel <- function(inputId, label = NULL) {
     label,
     class = "control-label",
     class = if (is.null(label)) "shiny-label-null",
+    # `id` attribute is required for `aria-labelledby` used by screen readers:
+    id = paste0(inputId, "-label"),
```
cpsievert, line 7: By any chance, do you have any reference that would explain why the `id` is needed (in addition to the `for` below)?

jooyoungseo:
That's a good point, @cpsievert! Typically, we don't need `aria-labelledby` attribute when a `<label>` is attached to an element using `for` attribute. However, as explained #2951, dateInput and dateRangeInput has the following issue:

> Each input label element is annotated with for= attribute but refers to the <div> that contains both the label and the following textbox, which means the textbox itself has no accessible label. So that's messed up for assistive technology users.

In other words, the `for` value of the `<label>` tag has been double-matched with both the dateInput `<input>` tag and its outer (`<div>`) tag as they both share the same inputId.

Since we only want to pass label text to each input tag so that screen readers do not get confused with its `<div>` parent, I explicitly made a connection between the input tags and label via `id` attribute.

According to this [accessibility MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute) ...:

> aria-labelledby attribute establishes relationships between objects and their label(s), and its value should be one or more element IDs, which refer to elements that have the text needed for labeling.

Please kindly let me know if there would be anything I need to tidy up.

jooyoungseo:
> Looks good, other than the test that needs to be fixed.

`devtools::test()` does not give me any errors on my local machine after my latest commit (e8eadc1); I have no idea why GitHub throws some failing results.

cpsievert:
I'm not exactly sure why this fixes the test failures, but https://github.com/rstudio/shiny/pull/2989/commits/8693eed3ec5104cbfb2c44bf214c86658734ce86 seems to do the trick

wch:
Any idea why `expect_equal` works but `grepl(fixed=T)` doesn't?

PR #2981, opened 2020-07-24
Allow to pass resource path that starts with `.`

### LINKED ISSUE #2980: addResourcePath should allow resources that starts with a dot
It's pretty common to have on a website a folder called `/.well-known`, notably for adding a `security.txt`: https://securitytxt.org/

It would be doable by doing `addResourcePath(".well-known", "inst/.well-known")`, but `addResourcePath()` regex for prefix name excludes prefixes that starts with dot. 

https://github.com/rstudio/shiny/blob/6ad2125ee74a0a6cab723fefc25449c50a4568fa/R/server.R#L74

Could this regex be turned to `"^[a-z0-9\\-_\\.][a-z0-9\\-_\\.]*$"` instead?

Happy to make a PR in that sense. 

Cheers

ColinFay:
This will close #2980

### DIFF
--- a/R/server.R
+++ b/R/server.R
@@ -72,7 +72,7 @@ registerClient <- function(client) {
 addResourcePath <- function(prefix, directoryPath) {
   if (length(prefix) != 1) stop("prefix must be of length 1")
   if (grepl("^\\.+$", prefix)) stop("prefix can't be composed of dots only")
-  if (!grepl('^[a-z0-9\\-_\\.][a-z0-9\\-_.]*$', prefix, ignore.case = TRUE, perl = TRUE)) {
+  if (!grepl('[a-z0-9\\-_.]*$', prefix, ignore.case = TRUE, perl = TRUE)) {
     stop("addResourcePath called with invalid prefix; please see documentation")
   }
   if (prefix %in% c('shared')) {
--- a/R/server.R
+++ b/R/server.R
@@ -71,6 +71,7 @@ registerClient <- function(client) {
 #' @export
 addResourcePath <- function(prefix, directoryPath) {
   if (length(prefix) != 1) stop("prefix must be of length 1")
+  if (grepl("^\\.+$", prefix)) stop("prefix can't be composed of dots only")
   if (!grepl('^[a-z0-9\\-_\\.][a-z0-9\\-_.]*$', prefix, ignore.case = TRUE, perl = TRUE)) {
     stop("addResourcePath called with invalid prefix; please see documentation")
   }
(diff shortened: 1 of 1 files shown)


ColinFay:
Not sure why this is failing here: https://github.com/rstudio/shiny/pull/2981/checks?check_run_id=905762561

R/server.R
```diff
@@ -71,7 +71,7 @@ registerClient <- function(client) {
 #' @export
 addResourcePath <- function(prefix, directoryPath) {
   if (length(prefix) != 1) stop("prefix must be of length 1")
-  if (!grepl('^[a-z0-9\\-_][a-z0-9\\-_.]*$', prefix, ignore.case = TRUE, perl = TRUE)) {
+  if (!grepl('^[a-z0-9\\-_\\.][a-z0-9\\-_.]*$', prefix, ignore.case = TRUE, perl = TRUE)) {
```
wch, line 74: I think this can be simplified to just:
```
'^[a-z0-9\\-_.]+$'
```

However, we should also disallow prefixes that consist only of `.` characters - that could lead to unnecessary confusion.

ColinFay:
Ah, I hadn't think about the use of dot only. 

I've pushed a new test to check if the prefix is made only of dots (that would also cacth `".."`).

``` r
fn <- function(prefix) {
  if (grepl("^\\.+$", prefix)) stop("prefix can't be composed of dots only") else "ok"
}

fn(".")
#> Error in fn("."): prefix can't be composed of dots only
fn("..")
#> Error in fn(".."): prefix can't be composed of dots only
fn(".well-known")
#> [1] "ok"
```

<sup>Created on 2020-07-25 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

R/server.R
```diff
@@ -72,7 +72,7 @@ registerClient <- function(client) {
 addResourcePath <- function(prefix, directoryPath) {
   if (length(prefix) != 1) stop("prefix must be of length 1")
   if (grepl("^\\.+$", prefix)) stop("prefix can't be composed of dots only")
-  if (!grepl('^[a-z0-9\\-_\\.][a-z0-9\\-_.]*$', prefix, ignore.case = TRUE, perl = TRUE)) {
+  if (!grepl('[a-z0-9\\-_.]*$', prefix, ignore.case = TRUE, perl = TRUE)) {
```
wch, line 75: This needs a `+` instead of `*` -- otherwise it'll allow empty strings.

ColinFay:
Argh, sure. 

Done in the last commit ✅

Issue #2998, opened 2020-08-11
getCurrentOutputInfo() doesn't necessarily respond to dynamic CSS changes 


cpsievert:
If relevant CSS changes occur the tick after image information is sent, it's currently too late for `getCurrentOutputInfo()` to respond to the change. For example, when the foreground changes from red to blue here, `getCurrentOutputInfo()` isn't updated:

```r
library(shiny)
ui <- fluidPage(
  h3("This is the foreground color"),
  tags$head(
    tags$style(HTML("body {color: red}")),
    tags$script(HTML(
      "setTimeout(function() { document.body.style.setProperty('color', 'blue'); }, 1000);"
    ))
  ),
  plotOutput("p")
)
server <- function(input, output) {
  output$p <- renderPlot({
    plot(1:10, pch = 19, col = getCurrentOutputInfo()$fg())
  })
}
shinyApp(ui, server)
```

It appears this problem can manifest itself in other interesting ways. Take this `runtime: shiny` example. After installing `remotes::install_github("rstudio/rmarkdown#1706")`, and switching to a slow 3G connection on Chrome, `getCurrentOutputInfo()` returns a `bg` of `NA`, even though it eventually gets painted as `#222`

````
---
runtime: shiny
output: 
  html_document:
    theme: darkly
---

<style>
  body{color: red}
</style>

```{r}
renderPlot({
  info <- shiny::getCurrentOutputInfo()
  plot.new()
  text(0.5, 0.5, sprintf("bg: %s; fg: %s", info$bg(), info$fg()))
})
```
````

<img width="1226" alt="Screen Shot 2020-08-11 at 5 07 19 PM" src="https://user-images.githubusercontent.com/1365941/89959832-f237a600-dc02-11ea-8815-92befc7b3772.png">

<img width="369" alt="Screen Shot 2020-08-11 at 5 06 29 PM" src="https://user-images.githubusercontent.com/1365941/89959954-2c08ac80-dc03-11ea-8cc2-4d462ca55d1f.png">

PR #2999, opened 2020-08-11
Make getCurrentOutputInfo() reactive to dynamic CSS modifications

cpsievert:
To start a discussion on how we might address #2998


cpsievert:
See https://github.com/rstudio/shiny/issues/3196#issuecomment-742138836 a discussion of why this is still somewhat of a problem, but I don't think the solution will ever be polling (so closing this)

PR #3001, opened 2020-08-13
Unified eol in Gruntfile.js for all the platforms

jooyoungseo:
This PR attempts to address nodejs failing issues on GitHub against Windows just like [this commit](https://github.com/rstudio/shiny/runs/977665311).

We have followed [this StackOverflow comment](https://stackoverflow.com/questions/29817511/grunt-issue-with-line-endings).

Let's see this tweak would work.


jooyoungseo:
@wch, I am sorry... I didn't know that @schloerke had requested additional review. Please feel free to revert if necessary.

PR #3002, opened 2020-08-14
Added `aria-live` for *Output functions for accessibility of dynamic content

### LINKED ISSUE #2987: Set `aria-live` and `aria-busy` on outputs
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions
https://www.w3.org/TR/wai-aria-1.1/#aria-busy

jooyoungseo:
* Closed #2987: Dynamic content within `*Output()` functions is now automatically announced to assistive technologies via `aria-live = "polite"` by default. All the `*Output()` functions now have `aria_live` param to control how their dynamic changes are announced to assistive technologies (see [ARIA live regions - Accessibility MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) for technical details).

* Screen reader testing has been carried out for various `*Output()` functions, and dynamic changes are auto-announced now.

* For example, you can test with the following app using assistive technologies (e.g., screen readers):

``` r
library(shiny)
runExample("11_timer")

# Output is auto-announced every second.
```

* @jcheng5, I don't think we need `aria-busy` attribute because `aria-live` autofetches only when new events occur via `render*()` functions. Please advise me if there would be anything I could improve further.


jcheng5:
Do you happen to know if the `aria-live` attribute needs to be included in the initial, server-rendered HTML in order to be respected by screen readers, or whether it's OK to have `aria-live` added in JavaScript after the page has loaded? If the latter, then we could add the attribute from shiny.js at the time that we're discovering and binding output elements; this would mean that we could make the default of ALL outputs, even custom ones like HTML widgets, become `aria-live=polite`.

jooyoungseo:
Thanks very much for your great comment, @jcheng5.

1. As far as I tested using screen readers, `aria-live` can be added to client side after the page has loaded.
2. Would there be any special benefit to add `aria-live` to `shiny.js`? As we may want to allow users to set `aria-live` as "off" for some informative and redundant dynamics, I intentionally put the param to `*Output()` functions. Please advise me.

wch:
Here another possibility:

* Add `...` to the *Output functions. Check to make sure they're all named args, and then pass the `...` to the `div()`. Then users can do something like `` textOutput("x", "x", `aria-live` = "off") ``.
* On the JS side, any output div/span that doesn't already have an `aria-live` attribute will have `aria-live="polite"` added automatically to the tag.

The only thing that I'm a little wary of is opening up the wrapper div to take any arbitrary attributes.

@jcheng5 Thoughts?

cpsievert:
Closing in favor of https://github.com/rstudio/shiny/pull/3042

Issue #2845, opened 2020-04-24
sliderInput control inaccessible for keyboard or screen reader users


gtritchie:
### System details

Output of `sessionInfo()`:

```
R version 3.6.3 (2020-02-29)
Platform: x86_64-apple-darwin15.6.0 (64-bit)
Running under: macOS Catalina 10.15.4

Matrix products: default
BLAS:   /System/Library/Frameworks/Accelerate.framework/Versions/A/Frameworks/vecLib.framework/Versions/A/libBLAS.dylib
LAPACK: /Library/Frameworks/R.framework/Versions/3.6/Resources/lib/libRlapack.dylib

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] shiny_1.4.0.9002

loaded via a namespace (and not attached):
 [1] Rcpp_1.0.4.6         withr_2.1.2          packrat_0.5.0        digest_0.6.25        later_1.0.0         
 [6] mime_0.9             R6_2.4.1             jsonlite_1.6.1       xtable_1.8-4         magrittr_1.5        
[11] rlang_0.4.5          promises_1.1.0       tools_3.6.3          tinytex_0.22         rsconnect_0.8.16    
[16] httpuv_1.5.2         xfun_0.13            fastmap_1.0.1        compiler_3.6.3       htmltools_0.4.0.9003
```

### Example application *or* steps to reproduce the problem

Using the default app created by RStudio IDE 1.3.947, but specifically looking at:

```R
            sliderInput("bins",
                        "Number of bins:",
                        min = 1,
                        max = 50,
                        value = 30)
```

<img width="1131" alt="2020-04-24_14-58-13" src="https://user-images.githubusercontent.com/10569626/80260128-0b105180-863c-11ea-9e2f-73e85d1eb0f6.png">

### Describe the problem in detail
Load the page into browser, and try to use the slider via keyboard-only.

The slider control is completely inaccessible. You cannot get focus to it via keyboard (e.g. by Tab key), thus you cannot manipulate it via keyboard. Similarly, it is not annotated in a way that permits screen readers to recognize it as a slider, nor to manipulate it.

This violates numerous WCAG 2.1 level A requirements.

### Expected

- slider should be reachable via Tab sequence (i.e. tabindex=0)
[...]

PR #3003, opened 2020-08-14
Improving sliderInput accessibility

jooyoungseo:
This draft attempts to address #2845 (sliderInput accessibility) and it is still under development.

## To-Do List

- [ ] Patch `IonRangeSlider.js` to not set `tabindex` to "-1": We need to fix this to `tabindex` = "0" to allow keyboard tabbing focus.
- [ ] Write a patch js to slider value changes to be announced to screen readers: see [Deque University](https://dequeuniversity.com/library/aria/slider) and [aria practice](https://www.w3.org/TR/wai-aria-practices-1.1/examples/slider/slider-1.html).


itsozz:
When testing this widget using [WAVE](https://wave.webaim.org) "_very low contrast_" errors and "_small text_" alerts are reported. The small text alerts are reported for all elements except the current slider value(s) displayed above the 'thumb' control(s) and the very low contrast errors are for all. The number of errors and alerts reported depends on whether you have the grid number markings displayed or not, and if so - how many there are.

I've used the following CSS to overcome these issues. (NOTE: although the slider bar colour doesn't cause an error, I've changed it to maintain consistency with the current slider value(s) background which does require modification.)

```css
/* slider scale, min and max numbers */
.irs-grid-text, .irs-min, .irs-max {
    color: #333;
    font-size: 0.8em;
}   
                               
/* value(s) of slider appearing above the 'thumb' control(s) */
.irs-from, .irs-to, .irs-single {
    background-color: #5d77a3;
    color: #fff;
}
                               
/* slider highlighted/filled range */
.irs-bar, .irs-bar-edge {
    background-color: #5d77a3;
}
```

sasscal-dwd-apps:
Dear Developers,
I am also very interested on this topic, since I am developing different Apps with Shiny, and they all should be "accessible". I was also trying to use "noUISliderInput" instead of "sliderInput", but it is also inaccessible with the keyboard (at least using "tab")
Looking forward to seeing the solution on this issue.

cpsievert:
I'll take this over the finish line in #3011, thanks @jooyoungseo!

tpetzoldt:
I came across the same problem (keyboard accessibility of sliderInput) and tested it again right now, so I want to ask if there was any progress in the meantime. Thanks!

PR #3006, opened 2020-08-18
Added alt param

jooyoungseo:
This ready-for-review PR attempts to follow up with the PR #2494 to address the alt text issue #612.

We have to make sure `alt` attribute to be dynamic (beyond static).

Adding `alt` param to `render*()` instead of `*Output()` makes more sense to me (the actual screen reader user) given that we may want to have dynamic `alt` text depending on graph changes.

## Reprex

* Confirmed the following dynamic alt text works very well for screen readers

``` r
library(shiny)

shinyApp(
  ui = fluidPage(
    actionButton("btn1", "Update plots"),
    plotOutput("plot1", width = "400px")
  ),
    dt <- reactive({
      invalidateLater(3 * 1000)
      rnorm(30)
    })
    output$plot1 <- renderPlot(
      {
        hist(dt())
      },
      alt = reactive({
        paste("Dynamic alt text. Mean(x):", round(mean(dt()), 3))
      })
    )
  }
)
```

### DIFF
--- a/R/render-cached-plot.R
+++ b/R/render-cached-plot.R
@@ -446,6 +446,7 @@ renderCachedPlot <- function(expr,
       function(userCacheKeyResult) {
         width  <- fitDims$width
         height <- fitDims$height
+        alt <- altWrapper()
         pixelratio <- session$clientData$pixelratio %OR% 1
 
         key <- digest::digest(list(outputName, userCacheKeyResult, width, height, res, pixelratio), "xxhash64")
@@ -461,6 +462,7 @@ renderCachedPlot <- function(expr,
             plotObj = plotObj,
             width = width,
             height = height,
+            alt = alt,
             pixelratio = pixelratio
           ))
         }
--- a/R/render-cached-plot.R
+++ b/R/render-cached-plot.R
@@ -150,7 +150,9 @@
 #'   `"app"` (the default), `"session"`, or a cache object like
 #'   a [diskCache()]. See the Cache Scoping section for more
 #'   information.
-#' @param alt Optional parameter to pass `alt` to the img-tag.
+#' @param alt Descriptive text for a plot object. The default is "Plot object".
+#'   Dynamic alt text can be made when reactive function is passed.
+#'   (NULL or "" is not recommended because those should be limited to decorative images)
 #'
 #' @seealso See [renderPlot()] for the regular, non-cached version of
 #'   this function. For more about configuring caches, see
--- a/R/render-cached-plot.R
+++ b/R/render-cached-plot.R
@@ -343,6 +343,14 @@ renderCachedPlot <- function(expr,
   # values get filled by an observer below.
   fitDims <- reactiveValues(width = NULL, height = NULL)
 
+  # Make sure alt param to be reactive function
+  if (is.reactive(alt))
+    altWrapper <- alt
+  else if (is.function(alt))
+    altWrapper <- reactive({ alt() })
+  else
+    altWrapper <- function() { alt }
+
   resizeObserver <- NULL
   ensureResizeObserver <- function() {
     if (!is.null(resizeObserver))
@@ -389,6 +397,8 @@ renderCachedPlot <- function(expr,
         isolate({
           width  <- fitDims$width
           height <- fitDims$height
+          # Make sure alt text to be reactive function
+          alt <- altWrapper() 
         })
 
         pixelratio <- session$clientData$pixelratio %OR% 1
@@ -400,7 +410,6 @@ renderCachedPlot <- function(expr,
             func = isolatedFunc,
             width = width,
             height = height,
-#jy: default to title when NULL
             alt = alt,
             pixelratio = pixelratio,
             res = res
@@ -475,7 +484,6 @@ renderCachedPlot <- function(expr,
               plotObj = drawReactiveResult,
               width = width,
               height = height,
-#jy: default to title when NULL
               alt = alt,
               pixelratio = pixelratio
             )
@@ -486,6 +494,7 @@ renderCachedPlot <- function(expr,
         hybrid_chain(possiblyAsyncResult, function(result) {
           width      <- result$width
           height     <- result$height
+          alt     <- result$alt
           pixelratio <- result$pixelratio
 
           # Three possibilities when we get here:
--- a/R/render-cached-plot.R
+++ b/R/render-cached-plot.R
@@ -476,7 +476,7 @@ renderCachedPlot <- function(expr,
               width = width,
               height = height,
 #jy: default to title when NULL
-            alt = alt,
+              alt = alt,
               pixelratio = pixelratio
             )
           }
--- a/R/render-cached-plot.R
+++ b/R/render-cached-plot.R
@@ -150,9 +150,6 @@
 #'   `"app"` (the default), `"session"`, or a cache object like
 #'   a [diskCache()]. See the Cache Scoping section for more
 #'   information.
-#' @param alt Descriptive text for a plot object. The default is "Plot object".
-#'   Dynamic alt text can be made when reactive function is passed.
-#'   (NULL or "" is not recommended because those should be limited to decorative images)
 #'
 #' @seealso See [renderPlot()] for the regular, non-cached version of
 #'   this function. For more about configuring caches, see
--- a/R/render-cached-plot.R
+++ b/R/render-cached-plot.R
@@ -493,7 +493,7 @@ renderCachedPlot <- function(expr,
         hybrid_chain(possiblyAsyncResult, function(result) {
           width      <- result$width
           height     <- result$height
-          alt     <- result$alt
+          alt        <- result$alt
           pixelratio <- result$pixelratio
 
           # Three possibilities when we get here:
--- a/R/render-cached-plot.R
+++ b/R/render-cached-plot.R
@@ -150,6 +150,7 @@
 #'   `"app"` (the default), `"session"`, or a cache object like
 #'   a [diskCache()]. See the Cache Scoping section for more
 #'   information.
+#' @param alt Optional parameter to pass `alt` to the img-tag.
 #'
 #' @seealso See [renderPlot()] for the regular, non-cached version of
 #'   this function. For more about configuring caches, see
@@ -293,6 +294,7 @@ renderCachedPlot <- function(expr,
   cacheKeyExpr,
   sizePolicy = sizeGrowthRatio(width = 400, height = 400, growthRate = 1.2),
   res = 72,
+  alt = "Plot object",
   cache = "app",
   ...,
   outputArgs = list()
@@ -398,6 +400,8 @@ renderCachedPlot <- function(expr,
             func = isolatedFunc,
             width = width,
             height = height,
+#jy: default to title when NULL
+            alt = alt,
             pixelratio = pixelratio,
             res = res
           ),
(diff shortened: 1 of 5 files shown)


R/render-cached-plot.R
```diff
@@ -150,6 +150,7 @@
 #'   `"app"` (the default), `"session"`, or a cache object like
 #'   a [diskCache()]. See the Cache Scoping section for more
 #'   information.
+#' @param alt Optional parameter to pass `alt` to the img-tag.
```
schloerke, line 153: Let's see if we can remove this and retrieve it from the rendered plot

R/render-cached-plot.R
```diff
@@ -293,6 +294,7 @@ renderCachedPlot <- function(expr,
   cacheKeyExpr,
   sizePolicy = sizeGrowthRatio(width = 400, height = 400, growthRate = 1.2),
   res = 72,
+  alt = "Plot object",
```
schloerke, line 297: (to remove)

R/render-cached-plot.R
```diff
@@ -398,6 +400,8 @@ renderCachedPlot <- function(expr,
             func = isolatedFunc,
             width = width,
             height = height,
+#jy: default to title when NULL
+            alt = alt,
```
schloerke, line 404: (remove this. Retrieve later)

R/render-cached-plot.R
```diff
@@ -471,6 +475,8 @@ renderCachedPlot <- function(expr,
               plotObj = drawReactiveResult,
               width = width,
               height = height,
+#jy: default to title when NULL
+            alt = alt,
```
schloerke, line 479: (to remove)

R/render-plot.R
```diff
@@ -36,6 +36,7 @@
 #' @param res Resolution of resulting plot, in pixels per inch. This value is
 #'   passed to [grDevices::png()]. Note that this affects the resolution of PNG
 #'   rendering in R; it won't change the actual ppi of the browser.
+#' @param alt Optional parameter to pass `alt` to the img-tag.
```
schloerke, line 39: Same as `renderCachedPlot`, let's retrieve `alt` from the plot output

R/render-plot.R
```diff
@@ -51,7 +52,7 @@
 #'   call to [plotOutput()] when `renderPlot` is used in an
 #'   interactive R Markdown document.
 #' @export
-renderPlot <- function(expr, width='auto', height='auto', res=72, ...,
+renderPlot <- function(expr, width='auto', height='auto', res=72, alt="Plot object", ...,
```
schloerke, line 55: (to remove)

R/render-plot.R
```diff
@@ -112,6 +113,7 @@ renderPlot <- function(expr, width='auto', height='auto', res=72, ...,
             func = func,
             width = dims$width,
             height = dims$height,
+            alt = alt,
```
schloerke, line 116: (to remove)

R/render-plot.R
```diff
@@ -140,7 +142,7 @@ renderPlot <- function(expr, width='auto', height='auto', res=72, ...,
         dims <- getDims()
         pixelratio <- session$clientData$pixelratio %OR% 1
         result <- do.call("resizeSavedPlot", c(
-          list(name, shinysession, result, dims$width, dims$height, pixelratio, res),
+          list(name, shinysession, result, dims$width, dims$height, alt, pixelratio, res),
```
schloerke, line 145: (to remove)

R/render-cached-plot.R
```diff
@@ -500,6 +506,7 @@ renderCachedPlot <- function(expr,
                 result$plotObj,
                 width,
                 height,
+                alt,
```
schloerke, line 509: (to remove)

R/render-plot.R
```diff
@@ -181,14 +183,15 @@ resizeSavedPlot <- function(name, session, result, width, height, pixelratio, re
     src = session$fileUrl(name, outfile, contentType = "image/png"),
     width = width,
     height = height,
+    alt = alt,
```
schloerke, line 186:
```suggestion
    alt = result$alt,
```

R/render-plot.R
```diff
@@ -159,7 +161,7 @@ renderPlot <- function(expr, width='auto', height='auto', res=72, ...,
   markRenderFunction(outputFunc, renderFunc, outputArgs = outputArgs)
 }
 
-resizeSavedPlot <- function(name, session, result, width, height, pixelratio, res, ...) {
+resizeSavedPlot <- function(name, session, result, width, height, alt, pixelratio, res, ...) {
```
schloerke, line 164: (to remove)

R/render-plot.R
```diff
@@ -181,14 +183,15 @@ resizeSavedPlot <- function(name, session, result, width, height, pixelratio, re
...
+    alt = alt,
     coordmap = coordmap,
     error = attr(coordmap, "error", exact = TRUE)
   )
 
   result
 }
 
-drawPlot <- function(name, session, func, width, height, pixelratio, res, ...) {
+drawPlot <- function(name, session, func, width, height, alt, pixelratio, res, ...) {
```
schloerke, line 194: (to remove)

R/render-plot.R
```diff
@@ -272,6 +275,7 @@ drawPlot <- function(name, session, func, width, height, pixelratio, res, ...) {
         src = session$fileUrl(name, outfile, contentType='image/png'),
         width = width,
         height = height,
+        alt = alt,
```
schloerke, line 278: Similar to `result$coordmap`, we need to store the posibility of `alt` being returned from the function above (where `coordmap` is created).

In that return value, we would add `alt` if `value` inherits from `shiny_rendered_plot`. This will be attached to the result object which will be passed around, making the extra `alt` arguments unnecessary.

schloerke: This would be the location to default `alt` to `"Plot object"`

schloerke: This location may be too late to capture the `alt` value.  At this point, the print method has been called on the plot object.  Instead, we'd like to capture the alt before it's printed.

May have to add a check in-between :

```r
          function(value, .visible) {
            if (.visible) {
```

... in the `hybrid_chain` above.

This way, we can always check for the `alt` information even if nothing is visible (such as with `base` graphics).

R/render-plot.R
schloerke: (not true anymore)

R/render-plot.R
schloerke: (not true anymore)

R/render-plot.R
schloerke: (not true anymore)

R/render-plot.R
schloerke: (not true anymore)

R/render-plot.R
schloerke: (not true anymore)

R/render-plot.R
schloerke: (not true anymore. Resolving)

R/render-cached-plot.R
schloerke: (not true anymore)

R/render-cached-plot.R
schloerke: (not true anymore)

R/render-plot.R
schloerke, line 116: Also, above (after `heightWrapper`), define `altWrapper` to ingest `alt` similar to how `height` and `width` are ingested.

```suggestion
            alt = altWrapper(),
```

R/render-plot.R
schloerke, line 145:
```suggestion
          list(name, shinysession, result, dims$width, dims$height, altWrapper(), pixelratio, res),
```

R/render-cached-plot.R
schloerke, line 509: retrieve `alt` from `result$alt`, just like `width`/`height`/`pixelratio` 20 lines up

R/render-cached-plot.R
schloerke, line 479: Also set `alt <- altWrapper()` about 40 lines up where `width` / `height` / `pixelratio` are set.

```suggestion
              alt = alt,
```

R/render-cached-plot.R
schloerke, line 404: Set `alt <- altWrapper()` about 15 lines up where `height` and `width` are set inside the `isolate()` call.

schloerke, line 404: Similar to `renderPlot()`, define `altWrapper` right after `fitDims` is defined on line 344.

schloerke:
reprex goal:

```r
shinyApp(
  ui = fluidPage(
    actionButton("btn1", "Update plots"),
    plotOutput("plot1", width = "400px")
  ),
    dt <- reactive({
      invalidateLater(3 * 1000)
      rnorm(30)
    })
    output$plot1 <- renderPlot(
      {
        hist(dt())
      },
      alt = reactive({
        paste("Dynamic alt text. Mean(x):", round(mean(dt()), 3))
      })
    )
  }
)
```

jooyoungseo:
@schloerke, we've made it!

I have confirmed the reactive expr is dynamically passed to `alt` param. When not provided, defaulting to "plot object" announced to screen readers :)

R/render-cached-plot.R
cpsievert, line 153: I believe this can be removed since we have the `@inheritParams renderPlot` above

```suggestion
```

R/render-plot.R
cpsievert, line 39:
```suggestion
#' @param alt Alternate text for the HTML `<img>` tag if it cannot be displayed or viewed (i.e., the user uses a screen reader). In addition to a character string, the value may be a reactive expression (or a function referencing reactive values) that returns a character string.
```

R/render-cached-plot.R
```diff
@@ -293,6 +293,7 @@ renderCachedPlot <- function(expr,
   cacheKeyExpr,
   sizePolicy = sizeGrowthRatio(width = 400, height = 400, growthRate = 1.2),
   res = 72,
+  alt = "Plot object",
```
cpsievert, line 296: Maybe we could be a bit more descriptive here?

```suggestion
  alt = "A plot image generated by R Shiny",
```

jooyoungseo: Thanks very much for the great suggestion, Carson! The reason why I have simplified it as "Plot object" is because I thought that was the minimal nitty-gritty. Users would know the plot has been generated by Shiny if they came  across the plot object within Shiny apps. Even it is not the case, I think shorter alt text would be preferred; otherwise, they would need to consume extra seconds to listen to decorative info. :)

cpsievert: Ok, sounds good to me!

R/render-cached-plot.R
```diff
@@ -480,6 +493,7 @@ renderCachedPlot <- function(expr,
         hybrid_chain(possiblyAsyncResult, function(result) {
           width      <- result$width
           height     <- result$height
+          alt     <- result$alt
```
schloerke, line 496: (White space formatting to match the lines together)

```suggestion
          height     <- result$height
          alt        <- result$alt
```

R/render-cached-plot.R
```diff
@@ -471,6 +483,7 @@ renderCachedPlot <- function(expr,
               plotObj = drawReactiveResult,
               width = width,
               height = height,
+              alt = alt,
```
schloerke, line 486: `alt` must be retrieved from `altWrapper()` near line 448, similar to `height` and `width` are retrieved from `fitDims`.

jooyoungseo: Thank you! Nice catch! Resolved via c844ea6

PR #3007, opened 2020-08-18
Experiment with error message

hadley:
Created for the purposes of discussion.

Before:

```
Error in .getReactiveEnvironment()$currentContext() : 
  Operation not allowed without an active reactive context. (You tried to do something that can only be done from inside a reactive expression or observer.)
```

After:

```
Error: Operation not allowed without an active reactive context.
* You tried to do something that can only be done from inside a reactive consumer.
```

Also worth consider if it would be useful to include a hint here, or if that's ultimately too misleading:

```
Error: Operation not allowed without an active reactive context.
* You tried to do something that can only be done from inside a reactive consumer.
i Do you need to wrap the expression in `isolate()`?
```


wch:
Looks reasonable to me. I think suggesting `isolate()` could point people to write bad code and not understand what it's doing.

hadley:
You can merge this if you want, or leave it open and I'll collect a few more error messages.

wch:
I'll leave it open for now - let me know when you feel like it's done.

hadley:
This is worth another review — I've updated all errors in `react.R` and `reactive.R` following http://style.tidyverse.org/error-messages.html

hadley:
Discovered a few older issues that I filed with suggestions for improved error mesages, so I tweaked those too.

daattali:
Just noticed the new error messages! It's nice that very old messages from version 0.0.1 are being reexamined :)

Regarding the change of "reactive expression or observer" -> "reactive consumer", was that motivated by people being confused by the original message, or is it an attempt to be more "correct"? From talking to many intermediate shiny developers, most don't seem to know what "reactive consumer" means and I think this message might cause more confusion than it fixes

PR #3010, opened 2020-08-20
`downloadButton()` icon can now be changed via the `icon` parameter

### LINKED ISSUE #2692: Choice of the icon in downloadButton
Hello,

 It would be nice if we could choose the icon of the download button (for example I like `icon("save")`).

Currently I do it by changing the code of `downloadButton` or by using JavaScript.

ColinFay:
Will close #2692


R/bootstrap.R
```diff
@@ -1342,13 +1343,15 @@ uiOutput <- htmlOutput
 #' @export
 downloadButton <- function(outputId,
                            label="Download",
-                           class=NULL, ...) {
+                           class=NULL,
+                           icon=icon("download"),
```
wch, line 1347: Can you put `icon` after the `...`? That would reduce the risk of breaking existing code, and it also is a pattern we are using more and more in general (forcing users to name the arguments).

ColinFay: Done ✅

R/bootstrap.R
```diff
@@ -1342,13 +1343,15 @@ uiOutput <- htmlOutput
 #' @export
 downloadButton <- function(outputId,
                            label="Download",
-                           class=NULL, ...) {
+                           class=NULL,
+                           ..., 
+                          icon=icon("download")) {
```
wch, line 1348: Small nit: can you add one more space, so that this aligns this with the previous lines?

Also, the .Rd file needs to be re-generated. Otherwise, looks good!

ColinFay: Done via last commit ✅

wch:
Thanks!

PR #3016, opened 2020-08-24
Issue 2984

aalucaci:


wch:
Instead of adding and exporting a new function, I think it makes more sense for `renderCachedPlot` to accept `width` and `height` arguments, like this:

```
renderCachedPlot <- function(expr,
  cacheKeyExpr,
  res = 72,
  cache = "app",
  ...,
  outputArgs = list(),
  width,
  height
) {
```

Then, if `!missing(width)`, then raise a warning. (And same for `height`.)

The help for `width` and `height` should also say that they aren't used, and that `sizePolicy` is used instead.

aalucaci:
Thank you for your suggestion. I appreciate your time (I guess it took you longer to write the suggestion, than it would have taken to implement it yourself...). I gave it another try.

Kind regards,
Angela

wch:
Thanks!

PR #3017, opened 2020-08-25
Fix reactiveValuesToList reactlog label

schloerke:
Was using the old method label before. Now using the valid method as the label


NEWS.md
```diff
@@ -22,6 +24,8 @@ shiny 1.5.0.9000
 
 * Closed #2972: `runExample()` now supports the `shiny.port` option (thanks to @ColinFay). (#2982)
 
+* `reactiveValuesToList()` will save it's `reactlog` label as `reactiveValuesToList(<ID>)` vs `as.list(<ID>)` (#3017)
```
wch, line 27:
```suggestion
* `reactiveValuesToList()` will save its `reactlog` label as `reactiveValuesToList(<ID>)` vs `as.list(<ID>)` (#3017)
```

schloerke: I swear my IDE is changing my text to bad grammar

PR #3021, opened 2020-08-25
Issue 629

ColinFay:


jcheng5:
Thanks, @ColinFay! A couple of comments:

1. Rather than extracting the `stopText` for reuse, I'd create a helper function there (i.e. `validateSessionForUpdate <- function(session) { ... }`).
2. We usually use `inherits(x, "ShinySession")` instead of `"ShinySession" %in% class(x)`.
3. We should also allow `inherits(x, "MockShinySession")`.
4. Given 2) and 3), perhaps it's best to introduce an `is.session()` function as well.

ColinFay:
Thanks @jcheng5!

Do you think this validation function should become an exported function? 
I think that could be useful for other packages with updateinput functions.

wch:
This is superseded by #3195 since I'm not able to push to this branch.

PR #3034, opened 2020-09-02
Fix DiskCache pruning logic when max_n and max_size are both used.

### LINKED ISSUE #3033: DiskCache may not prune enough items if max_n and max_size are both used
In the example below, just the last item, `"f"`, should not be pruned; the rest of the objects exceed the max size of 200 bytes when serialized.

```R
d <- diskCache(max_n = 3, max_size = 200)
# Set some values. Use rnorm so that object size is large; a simple vector
# like 1:100 will be stored very efficiently by R's ALTREP, and won't exceed
# the max_size. We want each of these objects to exceed max_size so that they'll be
# pruned.
d$set("a", rnorm(100))
d$set("b", rnorm(100))
d$set("c", rnorm(100))
d$set("d", rnorm(100))
d$set("e", rnorm(100))
Sys.sleep(0.1)  # For systems that have low mtime resolution.
d$set("f", 1)   # This object is small and shouldn't be pruned.
d$prune()
d$keys()
#> [1] "d" "e" "f"
```

wch:
This closes #3033.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -40,6 +40,8 @@ shiny 1.5.0.9000
 
 * Fixed #1942: Calling `runApp("app.R")` no longer ignores options passed into `shinyApp()`. This makes it possible for Shiny apps to specify what port/host should be used by default. (#2969)
 
+* Fixed #3033: When a `DiskCache` was created with both `max_n` and `max_size`, too many items could get pruned when `prune()` was called. (#3034)
+
 ### Library updates
 
 * Removed html5shiv and respond.js, which were used for IE 8 and IE 9 compatibility. (#2973)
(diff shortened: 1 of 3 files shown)


NEWS.md
```diff
@@ -40,6 +40,8 @@ shiny 1.5.0.9000
 
 * Fixed #1942: Calling `runApp("app.R")` no longer ignores options passed into `shinyApp()`. This makes it possible for Shiny apps to specify what port/host should be used by default. (#2969)
 
+* Fixed #3033: When a `DiskCache` was created with both `max_n` and `max_size`, too many items could get pruned when `prune()` was called. (#3034)
```
cpsievert, line 43:
```suggestion
* Fixed #3033: When a `DiskCache` was created with both `max_n` and `max_size`, `prune()` didn't always prune objects that it should've. (#3034)
```

wch: Oh, actually, now that I think about it again, the original news item was correct. If only `max_n` or `max_size` was used, it would remove the correct items. It's only after the removal that the `info` data frame was modified incorrectly, and that's a problem only if `info` is used again. That would only happen if `max_n` and `max_size` were both used.

(Note that the first step of pruning, for `max_age`, did not record the sucessfully removed files -- it just assumed that the removal worked -- so it wasn't a problem there. But I have added the check for successful removal in that stage in this PR.)

I'm going to change the news back to reflect this.

PR #3035, opened 2020-09-02
Make sure cache logging appends to file

wch:
Before this change, if a DiskCache or MemoryCache was logging to a file, then the file would get overwritten each text was logged. This change makes it append to the file.

(Note that if logging to `stdout()`, this wasn't an issue.)


wch:
Due to a potential merge conflict, I'll add a NEWS item after merging. I don't think this merits a unit test, given the use cases for cache logging (basically just debugging) and the very low likelihood of accidental breakage in the future.

PR #3039, opened 2020-09-03
UI errors with Unicode characters rendered poorly

jcheng5:
Repro:

```r
library(shiny)

ui <- function(req) {
  # Throws an error
  match.arg("one", c("two", "three"))
}

}

shinyApp(ui, server)
```

Without this fix, you get incorrectly encoded characters.

image

Fix was tested on Ubuntu and Windows.

### DIFF
--- a/inst/template/error.html
+++ b/inst/template/error.html
@@ -2,6 +2,7 @@
 
 <head>
   <title>An error has occurred</title>
+  <meta charset="UTF-8">
 </head>
 
 <body>


cpsievert (approved):

wch:
@jcheng5 I think it would be better to use `content_type = "text/html; charset=UTF-8"` here:
https://github.com/rstudio/shiny/blob/43344d9a78a6f6641c90d5180ab4cb337fb0a458/R/middleware.R#L348-L353

We actually could remove the value for `content_type`, since we use `"text/html; charset=UTF-8"` as the default value for `httpResponse()` here:
https://github.com/rstudio/shiny/blob/43344d9a78a6f6641c90d5180ab4cb337fb0a458/R/middleware.R#L36-L39

wch:
I implemented the `charset=UTF-8` change in c5f2dece, after discussion with @jcheng5.

PR #3040, opened 2020-09-03
sliderInput: warn if value[2] > value[1] and sort value. Closes #3020

### LINKED ISSUE #3020: `sliderInput()` allows values[1] > values[2]
When using `value = c(10, 1)` in `sliderInput()`, `{shiny}` will turn both these values to `1`.

```r
library(shiny)
ui <- function(){
  tagList(
    sliderInput("a", "a", 1, 10, value = c(10, 1))
  )
}

server <- function(
  input, 
  output, 
  session
){
  observeEvent( input$a , {
    print(input$a)
  })
  
}

shinyApp(ui, server)
```

I would expect this to throw an error, but this might be more user-friendly to issue a warning, and to "reverse" the values so that the max of the rendered input is max(values), and the min of the rendered input is min(values)

aalucaci:


wch:
The code in this PR looks like a duplicate of #3040.

aalucaci:
Oh, mea culpa.
Closing this pull request and opening another one, hopefully this time with the right branch.

Kind regards,
Angela

Issue #2688, opened 2019-10-30
radioButtons cannot be reset


jcheng5:
This seems like it should work--hitting the Reset button should clear the radios. But it doesn't work, there's a JS error.

```
library(shiny)

ui <- fluidPage(
  radioButtons("radio", "Radios", letters[1:5]),
  actionButton("reset", "Reset")
)

server <- function(input, output, session) {
  observeEvent(input$reset, {
    updateRadioButtons(session, "radio", selected = character(0))
  })
  observe({
    print(input$radio)
  })
}

shinyApp(ui, server)
```

jcheng5:
This seems related too https://github.com/rstudio/shiny/issues/2266

armenic:
Interestingly, if I keep changing the `inputID` of `radioButtons()` within `renderUI` then the returned value in `input` gets reset, but the issue comes back again after changing `inputID` 9 times. Then it consistently returns NULL, no issues. Not sure where this 9 number comes from.

ColinFay:
Related: 

When updateSliderInput receives a `numeric(0)`, it sets the value of the slider to the `min` of the slider. I'm not sure if that's a desired behavior, and if not what would be the expected one 🤔

```r
ui <- fluidPage(
  sliderInput("slider", "slider", 1, 10, 4),
  actionButton("change", "Change")
)

server <- function(input, output, session) {
  
  
  observeEvent(input$change, {
    # This will make slider be equal to `min`
    updateSliderInput(session, "slider", value = numeric(0))
  })
  
  observeEvent( input$slider , {
    print(input$slider)
  })
  
  
}

shinyApp(ui, server)
```

Issue #2266, opened 2018-11-19
Radio buttons does not set answer to `NULL` if selected is `FALSE`


schloerke:
When a radio button is initialized, the radio button input value is not reset on display.  This should be set to `NULL` if selected is FALSE.

Steps to reproduce:
1. select a button and checkbox
2. click reset button
3. observe checkbox reset the value and the radiobutton value is not reset

Reprex:

```r
library(shiny)

shiny::shinyApp(
  fluidPage(
    actionButton("reset", "Reset Input"),
    uiOutput("radio_buttons"),
    verbatimTextOutput("radio_txt"),
    uiOutput("checkbox"),
    verbatimTextOutput("check_txt")
  ),
  function(input, output) {
    output$radio_txt <- renderPrint(input$radio_buttons)
    output$radio_buttons <- renderUI({
      input$reset
      radioButtons(
        "radio_buttons",
        "Input:",
        choices = c("Choice A", "Choice B"),
        selected = FALSE
      )
    })
    output$check_txt <- renderPrint(input$checkbox)
    output$checkbox <- renderUI({
      input$reset
      checkboxGroupInput(
        "checkbox",
        "Input:",
        choices = c("Choice C", "Choice D"),
        selected = FALSE
      )
    })
  }
)
```

jcheng5:
Hmmm, good one.

wch:
FWIW, I think the way to go is to use `character(0)` to select none. That's how it works for the update functions `updateCheckboxInput` and `updateSelectInput` (and the docs say that's how it works for `updateRadioButtons`, although #2688 says that's currently not true).

PR #3043, opened 2020-09-09
updateRadioButton with character(0)

ColinFay:
This will close two issues: 

#2688 and #2266: radioButton js binding didn't handle the cases where `data.value` is empty.

This also corrects a bug in `$escape` which couldn't handle empty arrays.

### DIFF
--- a/srcjs/input_binding_radio.js
+++ b/srcjs/input_binding_radio.js
@@ -15,9 +15,7 @@ $.extend(radioInputBinding, {
     }
   },
   setValue: function(el, value) {
-    if (value === undefined) {
-      return;
-    } else if (value.length === 0) {
+    if ($.isArray(value) && value.length === 0) {
       // Removing all checked item if the sent data is empty
       $('input:radio[name="' + $escape(el.id) + '"]').prop('checked', false);
     } else {
--- a/srcjs/input_binding_radio.js
+++ b/srcjs/input_binding_radio.js
@@ -9,15 +9,15 @@ $.extend(radioInputBinding, {
     // If none are checked, the input will return null (it's the default on load,
     // but it wasn't emptied when calling updateRadioButtons with character(0)
     if (value) {
-      return value
+      return value;
     } else {
-      return null
+      return null;
     }
   },
   setValue: function(el, value) {
-    if (value === undefined){
-      return
-    } else if (value.length === 0){
+    if (value === undefined) {
+      return;
+    } else if (value.length === 0) {
       // Removing all checked item if the sent data is empty
       $('input:radio[name="' + $escape(el.id) + '"]').prop('checked', false);
     } else {
--- a/srcjs/input_binding_radio.js
+++ b/srcjs/input_binding_radio.js
@@ -5,10 +5,23 @@ $.extend(radioInputBinding, {
   },
   getValue: function(el) {
     // Select the radio objects that have name equal to the grouping div's id
-    return $('input:radio[name="' + $escape(el.id) + '"]:checked').val();
+    var value = $('input:radio[name="' + $escape(el.id) + '"]:checked').val();
+    // If none are checked, the input will return null (it's the default on load,
+    // but it wasn't emptied when calling updateRadioButtons with character(0)
+    if (value) {
+      return value
+    } else {
+      return null
+    }
   },
   setValue: function(el, value) {
-    $('input:radio[name="' + $escape(el.id) + '"][value="' + $escape(value) + '"]').prop('checked', true);
+    if (value.length === 0){
+      // Removing all checked item if the sent data is empty
+      $('input:radio[name="' + $escape(el.id) + '"]').prop('checked', false);
+    } else {
+      $('input:radio[name="' + $escape(el.id) + '"][value="' + $escape(value) + '"]').prop('checked', true);
+    }
+
   },
   getState: function(el) {
     var $objs = $('input:radio[name="' + $escape(el.id) + '"]');
@@ -28,7 +41,6 @@ $.extend(radioInputBinding, {
   },
   receiveMessage: function(el, data) {
     var $el = $(el);
-
     // This will replace all the options
     if (data.hasOwnProperty('options')) {
       // Clear existing options and add each new one
--- a/srcjs/input_binding_radio.js
+++ b/srcjs/input_binding_radio.js
@@ -15,7 +15,9 @@ $.extend(radioInputBinding, {
     }
   },
   setValue: function(el, value) {
-    if (value.length === 0){
+    if (value === undefined){
+      return
+    } else if (value.length === 0){
       // Removing all checked item if the sent data is empty
       $('input:radio[name="' + $escape(el.id) + '"]').prop('checked', false);
     } else {
--- a/srcjs/utils.js
+++ b/srcjs/utils.js
@@ -240,7 +240,12 @@ function mergeSort(list, sortfunc) {
 
 // Escape jQuery selector metacharacters: !"#$%&'()*+,./:;<=>?@[\]^`{|}~
 var $escape = exports.$escape = function(val) {
-  return val.replace(/([!"#$%&'()*+,.\/:;<=>?@\[\\\]^`{|}~])/g, '\\$1');
+  // Sometime val comes empty, and Js throws a 'e.replace() is not a function' error
+  if (val.length)  {
+    return val.replace(/([!"#$%&'()*+,.\/:;<=>?@\[\\\]^`{|}~])/g, '\\$1');
+  } else {
+    return val
+  }
 };
 
 // Maps a function over an object, preserving keys. Like the mapValues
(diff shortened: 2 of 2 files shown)


srcjs/utils.js
```diff
@@ -240,7 +240,12 @@ function mergeSort(list, sortfunc) {
 
 // Escape jQuery selector metacharacters: !"#$%&'()*+,./:;<=>?@[\]^`{|}~
 var $escape = exports.$escape = function(val) {
-  return val.replace(/([!"#$%&'()*+,.\/:;<=>?@\[\\\]^`{|}~])/g, '\\$1');
+  // Sometime val comes empty, and Js throws a 'e.replace() is not a function' error
+  if (val.length)  {
```
cpsievert, line 244: This'll still throw an error if `val` is `undefined`. I'd either do `if (val)` or `if (typeof val === "string")`

srcjs/input_binding_radio.js
```diff
@@ -5,10 +5,23 @@ $.extend(radioInputBinding, {
...
+    // but it wasn't emptied when calling updateRadioButtons with character(0)
+    if (value) {
+      return value
+    } else {
+      return null
+    }
   },
   setValue: function(el, value) {
-    $('input:radio[name="' + $escape(el.id) + '"][value="' + $escape(value) + '"]').prop('checked', true);
+    if (value.length === 0){
```
cpsievert, line 18: For similar reasons to the other comment, we may also want/need to handle the case that `value` is `undefined`?

srcjs/utils.js
ColinFay: done ✅

I've chosen the `typeof` test, as what we tried to correct is `[].replace()`, and `if ([])` will try to do this 🙂

srcjs/input_binding_radio.js
ColinFay: Done! I've chose to return early if value is undefined, let me know!

wch (changes requested):

wch:
Could you also provide a small test app for this?

srcjs/utils.js
```diff
@@ -240,7 +240,12 @@ function mergeSort(list, sortfunc) {
 
 // Escape jQuery selector metacharacters: !"#$%&'()*+,./:;<=>?@[\]^`{|}~
 var $escape = exports.$escape = function(val) {
-  return val.replace(/([!"#$%&'()*+,.\/:;<=>?@\[\\\]^`{|}~])/g, '\\$1');
+  // Sometime val comes empty, and Js throws a 'e.replace() is not a function' error
```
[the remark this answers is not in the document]
ColinFay: @wch when running the current version of `{shiny}` with the example app at #2688, there is a JavaScript error: 

> shiny.min.js:3 Uncaught TypeError: e.replace is not a function

Screenshot 2020-09-11 at 08 42 12

When adding `console.log(val)` in `$escape`, I found that when running `updateRadioButtons()`, `val` is `[]`, hence throwing an error here.

ColinFay:
Will work on a test app in a few

srcjs/utils.js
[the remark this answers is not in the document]
cpsievert: Are you sure this is needed now that falsy values are properly accounted for in `.setValue()` (note that `updateRadioButton()` calls `receiveMessage()` which then calls `setValue()`)? Seems to me like it isn't (but I haven't verified)

srcjs/input_binding_radio.js
```diff
@@ -5,10 +5,25 @@ $.extend(radioInputBinding, {
...
   setValue: function(el, value) {
-    $('input:radio[name="' + $escape(el.id) + '"][value="' + $escape(value) + '"]').prop('checked', true);
+    if (value === undefined) {
+      return;
+    } else if (value.length === 0) {
+      // Removing all checked item if the sent data is empty
+      $('input:radio[name="' + $escape(el.id) + '"]').prop('checked', false);
+    } else {
+      $('input:radio[name="' + $escape(el.id) + '"][value="' + $escape(value) + '"]').prop('checked', true);
+    }
```
cpsievert, line 25: Note that an ideal test app should touch all 3 cases here

srcjs/utils.js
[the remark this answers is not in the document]
wch: I guess it makes sense that a `[]` is sent, since that's what `shiny:::toJSON(character(0))` returns.

That said, I still think it makes sense for that `$escape` to keep the original behavior. And I believe that with the new logic you've added for `setValue`, it will no longer call `$escape([])`, so the error would not happen even with the original version of `$escape`.

ColinFay:
@wch @cpsievert "test app" means a test in the testthat folder am I correct?

wch:
The test app can be something that you just include in a comment on this issue. We will probably include it in the shinycoreci-apps repo.

srcjs/input_binding_radio.js
ColinFay: @cpsievert I'm not sure when `value` could be undefined here?

ColinFay:
Here is a test app with three behaviours: updating with empty selection, classical update, and updating something else (i.e. not touching the value). 

I'm not sure how to test when `value` is `undefined` (not sure in which case that would happen), happy to modify this code to also check on this case :) 

```r
system("cd tools && yarn build")
pkgload::load_all()

ui <- fluidPage(
  radioButtons("radio", "Radios", letters[1:5]),
  actionButton("empty", "Empty"),
  actionButton("change", "Change"),
  actionButton("label", "Label")
)

  observeEvent(input$empty, {
    updateRadioButtons(session, "radio", selected = character(0))
  })
  
  observeEvent(input$change, {
    updateRadioButtons(session, "radio", selected = letters[1])
  })
  
  observeEvent(input$label, {
    updateRadioButtons(session, "radio", label = "New label")
  })
  
  observe({
    print(input$radio)
  })
}

shinyApp(ui, server)
```

srcjs/input_binding_radio.js
```diff
@@ -5,10 +5,25 @@ $.extend(radioInputBinding, {
...
+      return value;
+    } else {
+      return null;
+    }
   },
   setValue: function(el, value) {
-    $('input:radio[name="' + $escape(el.id) + '"][value="' + $escape(value) + '"]').prop('checked', true);
+    if (value === undefined) {
+      return;
+    } else if (value.length === 0) {
```
wch, line 20: This check feels weird to me. The behavior we want is that if `value` is `[]`, it does one thing, and if it's a string, it does another.

`value.length === 0` will be true if `value` is `[]`, but it will also be true if if `value` is `""`, because `.length` is also defined for strings.

Now, I don't think it's possible for `updateRadioButtons()` to send a value of `""`, so it won't cause problems, but this code is still a bit confusing.

It would be clearer if this first checked the type (with `$.isArray()`) and then the length.

srcjs/input_binding_radio.js
wch: If we don't know when it could be `undefined`, then let's take out the check. If some weird data gets sent to this function, better if it throws an error and we find out about it.

srcjs/input_binding_radio.js
ColinFay: Hey @wch, 

Ok, let me know if I get this straight: 

The browser receives a values, it can be: 

- an empty array, when `updateRadioButtons(session, "radio", selected = character(0))`, then we uncheck all the radioButtons
- a single value, when `updateRadioButtons(session, "radio", selected = letters[1])`, then we update the selected radiobutton
- an array of value, when doing `updateRadioButtons(session, "radio", selected = letters[1:5])` then shiny fails in the browser with `Uncaught TypeError: e.replace is not a function` => see https://github.com/rstudio/shiny/issues/3065
- an `""`, but I'm not sure when this happens? In any case, this might be handled as if `selected = character(0)`?

wch: I think it's reasonable to just do the following:

```js
if ($.isArray(value) && value.length === 0) {
  // This is the character(0) case 
  ...
} else {
  ...
}
```

This isn't so different from what you have right now. The differences are:

* No `undefined` check. If an `undefined` gets here, there's probably a bug somewhere, and it would be helpful for it to throw an error so that the bug can be more easily found.
* It just adds an array test to the `value.length===0` check. This is because, without the array check, it's confusing that `value.length` could be 0 if it's an array OR a string, and someone reading the code could be misled by it. When I looked at the code, I initially thought it was testing for `""`, because elsewhere in the function it's expected to be a string. Adding the `$.isArray()` test makes it unambiguous what that if-statement is really looking for.

In my limited testing, I wasn't able to get `updateRadioButtons` to send a `""` as the value. I think it gets removed on the R side first. I don't think we need a specific check for it.

ColinFay: @wch I've put this new test version in the last commit :)

wch:
I made some small updates and then squash-merged locally in 770ebc3.

Issue #3019, opened 2020-08-25
`sliderInput()` allows vector of length > 2 for `values`


ColinFay:
When adding a vector of length > 2 for `values` in `sliderInput`, `{shiny}` silently drops the values at length 3 and further. 

Using a vector of length > 2 for `values` here is probably an error from the user, either a typo or a misunderstanding of what should be used there, so it would be nice to have a warning that says that only the two first elements of `values` are used. 

```r
library(shiny)
ui <- function(){
  tagList(
    sliderInput("a", "a", 1, 10, value = 1:10)
  )
}

server <- function(
  input, 
  output, 
  session
){
  observeEvent( input$a , {
    print(input$a)
  })
  
}

shinyApp(ui, server)
```

PR #3047, opened 2020-09-11
sliderInput will warn if length(value) > 2 (#3019)

aalucaci:


wch:
I think it would make more sense to just `stop()` if there are more than two values. Also, the check can go closer to the top of the function, before it tries to do other stuff with `value`.

aalucaci:
Good point. Here is another try.

Kind regards,
Angela

PR #3053, opened 2020-09-15
Allow setting the state of RestoreContext

jcheng5:
This setter only sets public fields, but it's still necessary because
the RestoreInputSet R6 class is not exported.

(I needed this functionality for shinytableau config dialogs to do a
bookmark-like restore, but not at all based on querystring)


R/bookmark-state.R
```diff
@@ -217,6 +217,15 @@ RestoreContext <- R6Class("RestoreContext",
       self$dir <- NULL
     },
 
+    # Completely replace the state
+    set = function(active = FALSE, initErrorMessage = NULL, input = list(), values = new.env(parent = emptyenv()), dir = NULL) {
```
wch, line 221: Passing in a list for `values` feels more natural to me, and there's no risk of passing in an object that will be modified by other code. Any particular reason that `values` needs to be an environment, instead of taking a list and putting its contents into an environment?

jcheng5: I don't know, and in fact I don't know why it's an environment in the first place--I thought maybe the possibility of outside mutation was desirable for some reason? I'm happy to make the change, but the way it's written now lets you pass in `values = restoreContext$values` if you don't want to change anything.

I could accomplish both objectives by having the set method check if `values` is a list, and if so, call `values <- list2env2(values)`?

wch: It took me a while to find information about why it's an environment, but here it is:
https://github.com/rstudio/shiny/blob/46852e2051875b7241d9675fcba507b53f07ab2d/R/bookmark-state.R#L24-L33

I think it makes to do `values <- list2env2(values, parent = emptyenv())`.

jcheng5: Oh, nice find! I'll do it like you said, thanks.

jcheng5: Actually what if I do what you originally said and just insist that `values` is provided as a list?

wch: Sure, that's reasonable.

wch (approved):

wch:
Looks good!

jcheng5:
That force-push was just me resolving conflicts with a rebase. I'll merge if/when the CI checks pass.

PR #3055, opened 2020-09-15
Fix #2946: Ensure invalidation on freezeReactiveValue(input) freeze and new value

### LINKED ISSUE #2946: freezeReactiveValue() and indirect dependencies
In my opinion, `freezeReactiveValue()` has a crucial drawback (and I guess this also applies to `freezeReactiveVal()` though I haven't tested it yet):
Only reactive expressions depending *directly* on the object being freezed will quit with a silent error as done with `req(FALSE)`. Reactive expressions depending *indirectly* on the object being freezed won't realize that they should also quit with a silent error. Consider the following reproducible example which is a slightly modified version of the example from `?freezeReactiveValue`:

```r
library(shiny)

ui <- fluidPage(
  selectInput("data", "Data Set", c("mtcars", "pressure")),
  checkboxGroupInput("cols", "Columns (select 2)", character(0)),
  plotOutput("plot")
)

  observe({
    data <- get(input$data)
    # Sets a flag on input$cols to essentially do req(FALSE) if input$cols
    # is accessed. Without this, an error will momentarily show whenever a
    # new data set is selected.
    freezeReactiveValue(input, "cols")
    updateCheckboxGroupInput(session, "cols", choices = names(data))
  }, priority = 1)
  
  ### Additional code not from ?freezeReactiveValue:
  cols_reactive <- reactive({
    input$cols
  })
  ### 
  
  output$plot <- renderPlot({
    # When a new data set is selected, input$cols will have been invalidated
    # above, and this will essentially do the same as req(FALSE), causing
    # this observer to stop and raise a silent exception.
    ### Modified code:
    cols <- cols_reactive() # Original code: cols <- input$cols
    ### 
    data <- get(input$data)
    
    if (length(cols) == 2) {
      plot(data[[ cols[1] ]], data[[ cols[2] ]])
    }
  })
}

shinyApp(ui, server)

```

The difference to the original example from `?freezeReactiveValue` is that `output$plot` does not depend *directly* on `input$cols`, but only *indirectly* via `cols_reactive()`. Thus, after having selected two columns from `mtcars` and then switching to `pressure`, `freezeReactiveValue()` does not help in avoiding the (temporary) error. This temporary error might not be a problem in this simple example, but I have an app where I need to avoid it. Of course it would be possible to manually notify all indirect dependencies to quit with a silent error, but this is a tedious and error-prone task and in my opinion, it breaks the concept of reactivity.

I guess this drawback is already known, so I'm posting this as a feature request. I guess it will be quite complicated to fix this (so that also reactive expressions depending *indirectly* on the object being freezed will quit with a silent error), but it would be great if it could be managed somehow.

_**Edit 2020-09-16 by @jcheng5:** I added `priority = 1` to the example to make it clear that even with the observers executing in the most desirable order, this is still a problem._

jcheng5:
The goal of `freezeReactiveValue` is to protect reactives, observers, and (especially) outputs from seeing temporarily inconsistent states due to asynchronous operations that need to complete before the state can become consistent.

What is an inconsistent state?

Consider an app with two inputs: `input$dataset` and `input$columns`.

```r
selectInput("dataset", "Choose a dataset", c("pressure", "cars"))
selectInput("column", "Choose column", character(0))
```

The `columns` selector should show the users the columns available from the `input$dataset` dataset. To maintain this relationship, we use an observer.

```r
dataset <- reactive({
  switch(req(input$dataset),
    pressure = pressure,
    cars = cars,
    stop("Unexpected dataset value")
  )
})

observeEvent(input$dataset, {
  updateSelectInput(session, "column", choices = names(dataset()))
})
```

And here's an output that wants to use the dataset and column.

```r
output$summary <- renderPrint({
  req(input$column)
  summary(dataset()[,input$column])
})
```

The problem occurs when `input$dataset` changes (also on first load, but let's focus on the change case), say from `pressure` to `cars`. The session updates `input$dataset` to `cars`, but the value of `input$column` is going to be a column from `pressure` still. So the `renderPrint` is going to try to access `cars[,"temperature"]` which will cause an error.

A moment later, the `updateSelectInput` takes effect in the client, causing `input$column` to be updated. `renderPrint` runs again, causing the error to disappear and a successful render to happen in its place.

**The goal of `freezeReactiveValue` is to get rid of the flash.**

```r
observeEvent(input$dataset, {
  freezeReactiveValue(input, "column")
  updateSelectInput(session, "column", choices = names(dataset()))
})
```

The current implementation is supposed to work like this:

1. Updater calls `freezeReactiveValue`. This sets a `frozen` flag; while this flag is set, reads to the reactive value in question will throw the equivalent of `req(FALSE)`.
2. Updater calls `updateTextInput` or renders a UI containing the same input or whatever.
3. When the current reactive flush cycle ends, all `frozen` values are automatically thawed.
4. Whatever operation the updater initiated finishes, and changes the reactive value, causing all consumers to invalidate.

However, there are several issues in the current (CRAN) implementation that make it subtly unreliable.

## Bug 1: Ordering between updater and consumer matters

The first problem is that `freezeReactiveValue` doesn't ever trigger invalidation. If a consumer happens to run before an updater, then it can get into an error state, and freezing the reactive value afterwards won't notify the consumer. This is mostly a problem for outputs: it's not too bad for outputs to be in error states temporarily, but what's bad is for that error to make its way to the client.

1. `input$dataset = "cars"`
2. `output$summary` executes, resulting in an error
3. Updater executes, calling `freezeReactiveValue()`
4. (This is what's missing today:) `output$summary` re-executes, resulting in `req(FALSE)`

Ideally, updaters/outputs that call `freezeReactiveValue` would be set to a higher observer priority than any consumers. That would cause this not to be a problem at all, and it's also necessary to do this for non-output observers anyway (since an error can potentially be more serious in those cases).

Repro:

```r
library(shiny)

ui <- fluidPage(
  selectInput("dataset", "Choose a dataset", c("pressure", "cars")),
  selectInput("column", "Choose column", character(0)),
  verbatimTextOutput("summary")
)

  dataset <- reactive({
    switch(req(input$dataset),
      pressure = pressure,
      cars = cars,
      stop("Unexpected dataset value")
    )
  })
  
  observeEvent(input$dataset, {
    freezeReactiveValue(input, "column")
    updateSelectInput(session, "column", choices = names(dataset()))
  }, priority = -1)
  
  output$summary <- renderPrint({
    Sys.sleep(0.25)
    req(input$column)
    summary(dataset()[,input$column])
  })
}

shinyApp(ui, server)
```

Change the dataset and an error flashes by. I'm using `priority = -1` to make the problem a little more persistent, but it can occur without it.

The fix to this problem is to invalidate upon `freezeReactiveValue`. Possible downsides are that existing apps could have consumers that either expensively or side-effect-ily execute after the invalidation occurs, i.e. they do slow or side-effecty stuff before attempting to read the frozen value.

With that fix, you still get an error in the R console but the user never sees the error in the browser.

## Bug 2: Invalidation upon completion is unreliable due to dedupe

Similar scenario to last time, but we'll avoid Bug 1 and then go a little further.

1. `input$dataset = "cars"`
2. Updater executes, calling `freezeReactiveValue()` and `updateSelectInput`
3. `output$summary` executes, resulting in a `req(FALSE)`
4. Client sets numeric input
5. `output$summary` executes, resulting in success

The bug here is that step 5 can break down if `updateSelectInput` just so happens to set the input value to the same value it already has. In Shiny v1.5.0, that's a total no-op as far as the server is concerned.

Repro:

```r
library(shiny)

ui <- fluidPage(
  selectInput("dataset", "Choose a dataset", c("pressure", "pressure1", "cars")),
  selectInput("column", "Choose column", character(0)),
  verbatimTextOutput("summary")
)

  dataset <- reactive({
    switch(req(input$dataset),
      pressure = pressure,
      pressure1 = pressure,
      cars = cars,
      stop("Unexpected dataset value")
    )
  })
  
  observeEvent(input$dataset, {
    freezeReactiveValue(input, "column")
    updateSelectInput(session, "column", choices = names(dataset()))
  }, priority = 1)
  
  output$summary <- renderPrint({
    Sys.sleep(0.25)
    req(input$column)
    summary(dataset()[,input$column])
  })
}

shinyApp(ui, server)
```

Change the dataset from "pressure" to "pressure1" and you'll see the summary go blank.

The fix for this is that calling `freezeReactiveValue` should disable "ignore duplicate values" for the very next set. This needs to be implemented one way for `input`, another way for `reactiveValues`, and a third way for `reactiveVal`.

## Bug 3: #2946 Downstream reactive expressions don't react to their dependencies being frozen

The issue discusses it pretty well. Basically, invalidation on freeze is needed--same fix as for bug 1.


jcheng5:
OK. This PR now represents sort of a compromise, in an effort to fix the problems we know are actually happening to people, without attempting to fix problems that are theoretical and may not affect anyone (because attempting to fix the latter could conceivably break compatibility if people have figured out clever ways to make use of freeze/thaw).

1. `freezeReactiveValue(input)` now invalidates on freeze. (`freezeReactiveValue(rv)` and `freezeReactiveVal()` are unchanged.)
2. `freezeReactiveValue(input)` now causes the client to NOT dedupe the very next setting of the input. This ensures that reactive dependencies that read the frozen value, are guaranteed to get invalidated when it's safe to un-freeze.
3. `freezeReactiveValue(rv)` and `freezeReactiveVal()` emit deprecation messages, directing people to https://github.com/rstudio/shiny/issues/3063. The goal is to either get enough feedback to let us confidently apply the same fix to those scenarios, or, if nobody cares then we just drop that functionality.

Future work:

1. I haven't thought about `observeEvent(input$not_frozen, { input$frozen })`, hmmm, that sucks.
2. I would love not to thaw automatically at the end of the the flush cycle, it feels so arbitrary.

R/reactives.R
```diff
@@ -469,10 +473,15 @@ ReactiveValues <- R6Class(
...
     # be thrown.
-    freeze = function(key) {
+    freeze = function(key, invalidate = FALSE) {
       domain <- getDefaultReactiveDomain()
       rLog$freezeReactiveKey(.reactId, key, domain)
       setMeta(key, "frozen", TRUE)
+
+      if (invalidate) {
+        # Force an invalidation
+        self$set(key, NULL, force = TRUE)
```
wch, line 483: Setting it to `NULL` feels a little weird to me. Is there any way this value could leak out?

srcjs/shinyapp.js
```diff
@@ -696,6 +696,12 @@ var ShinyApp = function() {
     });
   });
 
+  addMessageHandler('frozen', function(message) {
+    for (let i = 0; i < message.ids.length; i++) {
+      exports.forgetLastInputValue(message.ids[i]);
```
wch, line 701: Will this handle inputs that have a type, like `x:num` (at least, I think that's what it is)?

srcjs/input_rate.js
```diff
@@ -260,6 +260,9 @@ var InputNoResendDecorator = function(target, initialValues) {
 
     this.lastSentValues = cacheValues;
   };
+  this.forget = function(name) {
+    delete this.lastSentValues[name];
```
jcheng5, line 264: Split the name here

jcheng5: @wch I'm trying to remember why we thought this was a problem? `name` isn't expected to be `name:type`, just the id.

jcheng5: (Oh I see your other comment, I think I just need to try it)

srcjs/shinyapp.js
jcheng5: Yes, these are handled. You can't call `forgetLastInput("x:num")` but you can call `forgetLastInput("x")` even if `x` has a type of `"num"`.

R/reactives.R
jcheng5: Discussed in person; even if it could leak out, it makes at least as much sense for it to leak as `NULL` versus leaking as the last-known-good value.

wch (approved):

PR #3058, opened 2020-09-16
Closes #2936 - `dateInput` iterating over vector of dates to avoid warning from `dateYMD`

### LINKED ISSUE #2936: datesdisabled in dateInput() supports a vector but it warns
The `datesdisabled` in `dateInput()` supports a vector of dates and the example in `help(dateInput)` also uses a vector input. However, it will throw a warning when the input's length is larger than 1, because it uses `dateYMD()` to corece the input. 

I think it should not throw warnings.

``` r
library(shiny)
as.character(
  dateInput(
    "date8",
    "Date:",
    value = "2012-02-29",
    datesdisabled = c("2012-03-01", "2012-03-02")
  )
) # this is the code I copy from the example of dateInput()'s doc
#> Warning in dateYMD(datesdisabled, "datesdisabled"): Expected `datesdisabled` to
#> be of length 1.
#> [1] "<div id=\"date8\" class=\"shiny-date-input form-group shiny-input-container\">\n  <label class=\"control-label\" for=\"date8\">Date:</label>\n  <input type=\"text\" class=\"form-control\" data-date-language=\"en\" data-date-week-start=\"0\" data-date-format=\"yyyy-mm-dd\" data-date-start-view=\"month\" data-initial-date=\"2012-02-29\" data-date-autoclose=\"true\" data-date-dates-disabled=\"[&quot;2012-03-01&quot;,&quot;2012-03-02&quot;]\" data-date-days-of-week-disabled=\"null\"/>\n</div>"
```

<sup>Created on 2020-06-24 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

andrewbaxter439:
Fix for Issue #2936. `dateYMD` expects object of length 1 and runs test before returning warning if >1. `dateInput` allows a vector of dates. Have changed `dateInput` to iterate over dates using `sapply` so as to not produce the warning which is not needed here (whilst preserving it for other cases).

Edit:
have run tests etc. as per Contribution guidelines. Also first pull request so do advise of parts omitted/incomplete!

### DIFF
--- a/R/input-date.R
+++ b/R/input-date.R
@@ -99,7 +99,7 @@ dateInput <- function(inputId, label, value = NULL, min = NULL, max = NULL,
   value <- dateYMD(value, "value")
   min <- dateYMD(min, "min")
   max <- dateYMD(max, "max")
-  datestisabled <- sapply(datesdisabled, dateYMD, "datesdisabled", USE.NAMES = FALSE)
+  datesdisabled <- sapply(datesdisabled, dateYMD, "datesdisabled", USE.NAMES = FALSE)
 
   value <- restoreInput(id = inputId, default = value)
 
--- a/R/input-date.R
+++ b/R/input-date.R
@@ -99,7 +99,7 @@ dateInput <- function(inputId, label, value = NULL, min = NULL, max = NULL,
   value <- dateYMD(value, "value")
   min <- dateYMD(min, "min")
   max <- dateYMD(max, "max")
-  datesdisabled <- dateYMD(datesdisabled, "datesdisabled")
+  datestisabled <- sapply(datesdisabled, dateYMD, "datesdisabled", USE.NAMES = FALSE)
 
   value <- restoreInput(id = inputId, default = value)
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -48,6 +48,8 @@ shiny 1.5.0.9000
 
 * Fixed #3033: When a `DiskCache` was created with both `max_n` and `max_size`, too many items could get pruned when `prune()` was called. (#3034)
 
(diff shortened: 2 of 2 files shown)


R/input-date.R
```diff
@@ -99,7 +99,7 @@ dateInput <- function(inputId, label, value = NULL, min = NULL, max = NULL,
   value <- dateYMD(value, "value")
   min <- dateYMD(min, "min")
   max <- dateYMD(max, "max")
-  datesdisabled <- dateYMD(datesdisabled, "datesdisabled")
+  datestisabled <- sapply(datesdisabled, dateYMD, "datesdisabled", USE.NAMES = FALSE)
```
cpsievert, line 102:
```suggestion
  datesdisabled <- sapply(datesdisabled, dateYMD, "datesdisabled", USE.NAMES = FALSE)
```

wch:
@cpsievert Could we just change the `dateYMD` function to not warn for vectors with length > 1?

cpsievert:
Hmm, yea, that seems reasonable, but there must've been a reason why we added that `length(date) > 1` check (I can't remember off the top of my head)?

wch:
It looks like it was originally added in f8d69ecb1fd7304316e83e605153d407713cb7a5. From that commit (and the related PR and issue) I don't know why the length check is needed, but if you can remember, let us know.

R/input-date.R
andrewbaxter439: *facepalm*

andrewbaxter439:
Thanks both, had erred on the side of not deleting things that I didn't know what there were there for. Does seem it's only implemented in a few instances, couldn't tell what the reason was. Have opened an alternative pull request (#3061) which actions the other fix - removing the >1 check, and reverting the `sapply` call to the original `dateYMD` call.

PR #3060, opened 2020-09-16
Make shiny.css bootstraplib aware; use jquerylib

### LINKED ISSUE #3059: inputPanel(), showNotification(), and other shiny.css isn't themable
Minimal example:

```r
library(shiny)
library(bootstraplib)
bs_theme_new()
bs_theme_base_colors(bg = "#202123", fg = "#B8BCC2")
bs_theme_accent_colors(primary = "#EA80FC")

ui <- fluidPage(
  "This inputPanel() has the wrong bg/fg color!", 
  inputPanel(textInput("txt", "Input label"))
)

```

Screen Shot 2020-09-16 at 1 03 12 PM

For a more "full-blown" example app, see:

cpsievert:
Closes #3059

## Testing notes

Install `remotes::install_github("rstudio/shiny")`, then deploy apps in https://github.com/rstudio/shinycoreci-apps/pull/73, make sure they render sensibly in different browsers


R/shinyui.R
```diff
@@ -80,6 +60,40 @@ renderPage <- function(ui, showcase=0, testMode=FALSE) {
...
+    return(list(src = c(href = "shared"), stylesheet = "shiny.css"))
+  }
+  scss_home <- system.file("www", "shared", "shiny_scss", package = "shiny")
+  scss_files <- file.path(scss_home, c("bootstrap.scss", "shiny.scss"))
+  outFile <- bootstrapSass(lapply(scss_files, sass::sass_file), theme = theme, "shiny")
+  list(src = c(file = dirname(outFile)), stylesheet = basename(outFile))
+}
+
+jqueryDependency <- function() {
+  jquerylib::jquery_core(getOption("shiny.jquery.version", 3))
```
wch, line 94: I think we should do the jquerylib stuff in a separate PR. This change points to a copy of jQuery on a CDN, which will cause breakage for people who are on private networks.

```
> jquerylib::jquery_core(getOption("shiny.jquery.version", 3))
List of 10
 $ name      : chr "jquery"
 $ version   : chr "3.5.1"
 $ src       :List of 2
  ..$ href: chr "https://code.jquery.com"
  ..$ file: chr "lib"
 $ meta      : NULL
 $ script    : chr "jquery-3.5.1.min.js"
 $ stylesheet: NULL
 $ head      : NULL
 $ attachment: NULL
 $ package   : chr "jquerylib"
 $ all_files : logi TRUE
 - attr(*, "class")= chr "html_dependency"
```

R/shinyui.R
```diff
@@ -42,31 +42,11 @@ renderPage <- function(ui, showcase=0, testMode=FALSE) {
...
-
-  shiny_deps <- list(
-    htmlDependency("json2", "2014.02.04", c(href="shared"), script = "json2-min.js"),
-    jquery(),
-    htmlDependency("shiny", utils::packageVersion("shiny"), c(href="shared"),
-      script = if (getOption("shiny.minified", TRUE)) "shiny.min.js" else "shiny.js",
-      stylesheet = "shiny.css")
+  shiny_deps <- c(
+    # TODO: is this still needed?
+    list(htmlDependency("json2", "2014.02.04", c(href="shared"), script = "json2-min.js")),
```
wch, line 47: Looks like json2 is needed only for IE8, which means we can get rid of it.

R/shinyui.R
```diff
@@ -80,6 +60,40 @@ renderPage <- function(ui, showcase=0, testMode=FALSE) {
...
+  )
+})
+
+shinyCssFile <- function(theme = getShinyOption("bootstrapTheme")) {
+  if (!is_bs_theme(theme)) {
+    return(list(src = c(href = "shared"), stylesheet = "shiny.css"))
+  }
+  scss_home <- system.file("www", "shared", "shiny_scss", package = "shiny")
+  scss_files <- file.path(scss_home, c("bootstrap.scss", "shiny.scss"))
+  outFile <- bootstrapSass(lapply(scss_files, sass::sass_file), theme = theme, "shiny")
```
wch, line 89: We can address this in a separate PR, but it feels a little funny that we have to manually construct the html dependency object for this.

`bs_dependencies` does that for the main bootstrap content, and it would be nice if we could do something similar for cases like this.

inst/www/shared/shiny_scss/shiny.scss
```diff
@@ -0,0 +1,436 @@
+/*************************************************
```
wch, line 1: Now that we have this file, we can get rid of shiny.css, right?

R/html-deps.R
```diff
@@ -24,6 +24,9 @@ createWebDependency <- function(dependency, scrubFile = TRUE) {
   if (!inherits(dependency, "html_dependency"))
     stop("Unexpected non-html_dependency type")
 
+  # Resolve tagFunction()
+  dependency <- htmltools::resolveDependencies(list(dependency))[[1]]
```
wch, line 28: I think we agreed to remove this.

R/shinyui.R
cpsievert: Oh interesting, I had thought that `file` took precedence over `href` -- if that's not true, then I'll remove the `href` in **jquerylib**

R/html-deps.R
cpsievert: Yep, this PR is rebased off of #3062 and I just haven't updated this one yet

R/shinyui.R
cpsievert: (We want to make sure we're doing the right thing in **jquerylib** because **bootstraplib** imports it)

inst/www/shared/shiny_scss/shiny.scss
cpsievert: We can if you want, but I was thinking it'd be better practice to have a build script to compile this to css that we can serve up statically if no `bs_theme()` is used, but I haven't gotten to that yet.

R/shinyui.R
wch: I still think we should revert this stuff and do the jquerylib stuff in another PR, so that we can focus on the details of that in one place (for example, possibly removing jquery files from shiny)

R/shinyui.R
cpsievert: Sounds good, let's do it in a follow up PR

wch:
We can also get rid of the json2-min.js file.

tools/updateShinyCSS.R
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env Rscript
+library(rprojroot)
+library(sass)
+
+scss <- find_package_root_file("inst/www/shared/shiny_scss/shiny.scss")
+shiny_css <- sass(
+  sass_file(scss),
+  options = sass_options(output_style = "expanded", source_comments = FALSE)
+)
+writeLines(shiny_css, "inst/www/shared/shiny.css")
```
wch, line 10: The dest path should also use rprojroot.

We should have a unit test to make sure the files match, so that we don't ship a wrong version of the css file. Need to watch out for `\n` issues on Windows though. Might be easiest to use `brio::write_lines(eol = "\n")`

The tools/README.md file should also have a mention of this script.

cpsievert: Seems like we could/should workaround that by doing the check via `git status` (like we do for built JS files), so I did that in 1c99f3d

wch: The reason for the separate JS check in GHA is that it requires outside resources (namely node and node packages). I think it makes sense to add this as a unit test since it only involves R code, and uses packages that are already installed. (bootstraplib has similar tests for the precompiled assets.)

Looking at the GHA node_js workflow, running this script there makes it so it needs to install/compile a bunch of R packages on every run in order to do the test when it's part of the node_js workflow. We might as well just do the test as part of the unit tests because the cost there is low -- instead of installing a bunch of packages, it just needs to compile the scss file and compare it to the css file.

One other thought: having the test in the unit tests will also ensure that someone the exact same output will be generated on all platforms (so that someone developing on Windows won't commit a file that's different from what would be generated on Linux/Mac).

tools/updateShinyCSS.R
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env Rscript
+library(rprojroot)
+library(sass)
+
+scss <- find_package_root_file("inst/www/shared/shiny_scss/shiny.scss")
+shiny_css <- sass(
```
wch, line 6: This should also have `cache=NULL`, just in case.

tests/testthat/test-built-files.R
```diff
@@ -0,0 +1,30 @@
...
+    target <- rprojroot::find_package_root_file("inst/www/shared/shiny.min.css")
+    expect_true(file.exists(target))
+    pre_build_time <- file.mtime(target)
+    pre_build_content <- readLines(target)
+    source(build_script, local = TRUE)
+    expect_true(file.mtime(target) > pre_build_time)
+    expect_identical(readLines(target), pre_build_content)
+  } else {
+    skip("Not able to verify whether shiny.css has been built")
+  }
```
wch, line 29:
```suggestion
  new_css <- sass::sass(
    sass::sass_file(system.file("www/shared/shiny_scss/shiny.scss", package = "shiny")),
    cache = NULL,
    options = sass::sass_options(output_style = "compressed"),
  )
  # Remove class and attributes
  new_css <- as.character(new_css)

  pkg_css_file <- system.file("www/shared/shiny.min.css", package = "shiny")
  pkg_css <- readChar(pkg_css_file, file.size(pkg_css_file), useBytes = TRUE)

  expect_identical(new_css, pkg_css)
```

wch (approved):

PR #3076, opened 2020-09-24
Split server.R into multiple files

wch:
The server.R file was getting unwieldy, so this splits server.R into

* server.R
* server-resource-paths.R
* runapp.R
* viewer.R

I used the method from [this article](https://devblogs.microsoft.com/oldnewthing/20190918-00/?p=102901) to preserve the line history, even though the file has been split up.

UPDATE: whether line changes are tracked across this method seems to depend on the version of git used, so now I think this is a bad idea.


wch:
Hmm, history seems to have been lost in some files, but not all. Taking a look...

PR #3078, opened 2020-09-24
Split server.R into multiple files, take 2

wch:
This is a redo of #3076, but with a different method -- by renaming the file to two different names on two separate branches, and then resolving the conflict on merge. I think this should work better for tracking history.


cpsievert (approved):
Nice! https://github.com/rstudio/shiny/blame/wch-rename1/R/viewer.R

wch:
Great, I just wanted to make sure that this wouldn't cause any merge conflicts with your work!

PR #3062, opened 2020-09-16
Localized bootstraplib themes

cpsievert:
This PR makes it possible to pass `{bootstraplib}` themes to the `theme` argument of `navbarPage()`, `fluidPage()`, `bootstrapPage()`, `bootstrapLib()`, etc. When this happens, the Bootstrap HTML dependency, as well as other HTML dependencies that want their styles to derive from Bootstrap defaults (e.g., `sliderInput()`, `selectInput()`, and `dateInput()`), are all generated and resolved at render-time.

```r
library(shiny)
ui <- fluidPage(
  theme = bootstraplib::bs_theme(bg = "black", fg = "white", primary = "violet"),
  dateInput("date", "Date", value = Sys.Date()),
  dateRangeInput("date", "Date", start = Sys.Date(), end = Sys.Date() + 7),
)
shinyApp(ui, function(input, output) {})
```

More generally, you can include `bootstrapLib(theme)` in an `htmlTemplate()` and/or a `tagList()` and things "just work" (for the most part) thanks to the lazy nature of the underlying `tagFunction()`.

```r
library(shiny)
library(bootstraplib)

ui <- htmlTemplate(
  "template.html",
  theme = bs_theme(bg = "black", fg = "white", primary = "purple"),
  date1 = dateInput("date", "Date", value = Sys.Date()),
  date2 = dateRangeInput("date", "Date", start = Sys.Date(), end = Sys.Date() + 7)
)

shinyApp(ui, function(input, output) {})
```

```html
<!DOCTYPE html>
<html>
  <head>
    {{ headContent() }}
    {{ bootstrapLib(theme) }}
  </head>
  <body>
    {{ date1 }}
    {{ date2 }}
  </body>
</html>
```

At the moment, there's at least two cases where things fall down

1. The 'themable' component (e.g., `dateInput()`) appears before the `bootstrapLib()` dependency in the tag tree. For example, `dateInput()` doesn't pick up the right `primary` color here:

```r
ui <- fluidPage(
  dateInput("foo", ""),
  bootstrapLib(bootstraplib::bs_theme(bg = "black", fg = "white", primary = "purple"))
)
shinyApp(ui, function(input, output) {})
```

2. The user calls something like `as.tags()` on the `bootstrapLib()` dependency prior to render-time (at least, in this case, we can throw an informative warning):

```r
ui <- fluidPage(
  htmltools::as.tags(
    bootstrapLib(
      bootstraplib::bs_theme(bg = "black", fg = "white", primary = "purple")
    )
[...]

Issue #2953, opened 2020-06-30
Improvements to shiny options


cpsievert:
The documentation on `?shinyOptions` is currently wrong (introduced in #2575):

> There are a number of global options that affect Shiny's behavior. These can be set globally with `options()` or locally (for a single app) with `shinyOptions()`.

This isn't actually true for the `shiny.*` options, which currently only work when set globally (i.e., shiny reads them with `getOption()`, not `getShinyOption()`). We should either fix the docs, or make the necessary changes so that the documentation is actually true (ideally the latter). In addition, maybe it makes sense for `shinyOptions()` to add the `shiny.` prefix (it none already exists) so that `shinyOptions(host = ...)` would work for setting `shiny.host`.

PR #3080, opened 2020-09-25
Add session-level shinyOptions()

wch:
This PR adds support for `shinyOptions()` that are set at the session level, app level, and global level. Previously, we only had support for app- and global-level options.

It also:
* Adds `getCurrentAppState()`, which returns an environment that contains data for the currently-running app. The app-level options are now set with this mechanism, and in the future it will be used to store and access other app-level data.
* Renames `consumeAppOptions` to `captureAppOptions` and clarifies what it does and why it's there. Also adds unit tests for this.
* Cleans up and adds annotations to `runApp()` so that it's clearer what's going on in there.
* Fixes the documentation for `shinyOptions`, so that the values set via `options` are separate from those from `shinyOptions`.

Here's a test application. Not sure we can use it with shinycoreci-apps, since it involves starting and stopping an application, and then checking some values.

```R
library(shiny)
shinyOptions(foo = NULL)
values <- list()
values["global_in"] <- list(getShinyOption("foo"))
shinyOptions(foo = "global")

s <- shinyApp(
  basicPage(),
    values["session_in"] <<- list(getShinyOption("foo"))
    shinyOptions(foo = "session")
    # shinyOptions(foo = NULL)
    # values["session_out"] <<- list(getShinyOption("foo"))
    onStop(function() {
      browser()
      values["session_out"] <<- list(getShinyOption("foo"))
    })

    observe(stopApp())
  },
  onStart = function() {
    values["app_in"] <<- list(getShinyOption("foo"))

    shinyOptions(foo = "app")

    onStop(function() {
      values["app_out"] <<- list(getShinyOption("foo"))
    })
  }
)

# Run the app
runApp(s)
values["global_out"] <- list(getShinyOption("foo"))

expect_identical(
  values,
  list(
    global_in = NULL,
    app_in = "global",
    session_in = "app", 
    session_out = "session",
    app_out = "app",
    global_out = "global"
  )
)
```

### DIFF
--- a/R/app-state.R
+++ b/R/app-state.R
@@ -1,8 +1,6 @@
 #' @include globals.R
 NULL
 
-# TODO: implement for mock session?
-
 # The current app state is a place to read and hang state for the
 # currently-running application. This is useful for setting options that will
 # last as long as the application is running.
--- a/R/app-state.R
+++ b/R/app-state.R
@@ -15,7 +15,8 @@ initCurrentAppState <- function(appobj) {
   }
   .globals$appState <- new.env(parent = emptyenv())
   .globals$appState$app <- appobj
-  .globals$appState$options <- list()
+  # Copy over global options
+  .globals$appState$options <- .globals$options
 }
 
 getCurrentAppState <- function() {
--- a/R/app-state.R
+++ b/R/app-state.R
@@ -1,0 +1,27 @@
+#' @include globals.R
+NULL
+
+# TODO: implement for mock session?
+
+# The current app state is a place to read and hang state for the
+# currently-running application. This is useful for setting options that will
+# last as long as the application is running.
+
+.globals$appState <- NULL
+
+initCurrentAppState <- function(appobj) {
+  if (!is.null(.globals$appState)) {
+    stop("Can't initialize current app state when another is currently active.")
+  }
+  .globals$appState <- new.env(parent = emptyenv())
+  .globals$appState$app <- appobj
+  .globals$appState$options <- list()
+}
+
+getCurrentAppState <- function() {
+  .globals$appState
+}
+
+clearCurrentAppState <- function() {
+  .globals$appState <- NULL
+}
--- a/R/runapp.R
+++ b/R/runapp.R
@@ -102,16 +102,6 @@ runApp <- function(appDir=getwd(),
     .globals$running <- FALSE
   }, add = TRUE)
 
-  # Enable per-app Shiny options, for shinyOptions() and getShinyOption().
-  oldOptionSet <- .globals$options
-  on.exit({
-    .globals$options <- oldOptionSet
-  },add = TRUE)
-
-  # A unique identifier associated with this run of this application. It is
-  # shared across sessions.
-  shinyOptions(appToken = createUniqueId(8))
-
   # Make warnings print immediately
   # Set pool.scheduler to support pool package
   ops <- options(
@@ -121,11 +111,9 @@ runApp <- function(appDir=getwd(),
   )
   on.exit(options(ops), add = TRUE)
 
-  # Set up default cache for app.
-  if (is.null(getShinyOption("cache"))) {
-    shinyOptions(cache = MemoryCache$new())
-  }
-
+  # ============================================================================
+  # Global onStart/onStop callbacks
+  # ============================================================================
   # Invoke user-defined onStop callbacks, before the application's internal
   # onStop callbacks.
   on.exit({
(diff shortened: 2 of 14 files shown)


R/app-state.R
```diff
@@ -0,0 +1,26 @@
...
+
+.globals$appState <- NULL
+
+initCurrentAppState <- function(appobj) {
+  if (!is.null(.globals$appState)) {
+    stop("Can't initialize current app state when another is currently active.")
+  }
+  .globals$appState <- new.env(parent = emptyenv())
+  .globals$appState$app <- appobj
+  # Copy over global options
```
cpsievert, line 16:
```suggestion
  # Copy over app-level global options (i.e., shinyOptions())
```

wch: The original comment is correct - it copies the global options and stores them in the app-level options.

R/runapp.R
```diff
@@ -135,11 +119,44 @@ runApp <- function(appDir=getwd(),
...
+
+  # Set up default cache for app.
+  if (is.null(getShinyOption("cache"))) {
+    shinyOptions(cache = MemoryCache$new())
+  }
+
+  # Extract appOptions (which is a list) and store them as shinyOptions, for
+  # this app. (This is the only place we have to store settings that are
+  # accessible both the UI and server portion of the app.)
+  applyCapturedAppOptions(appParts$appOptions)
```
cpsievert, line 152: To me this would be clearer if this was just `if (!is.null(opts)) do.call(shinyOptions, opts)`, but I'm not sure if there's a good reason to use the helper elsewhere

R/shiny-options.R
```diff
@@ -127,30 +149,76 @@ shinyOptions <- function(...) {
...
+    return(session$options)
+  }
+
+  app_state <- getCurrentAppState()
+  if (!is.null(app_state)) {
+    return(app_state$options)
+  }
 
-  expr
+  return(.globals$options)
```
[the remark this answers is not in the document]
wch: That specific code isn't quite right, since a value could be cleared at a more specific scope. If someone does that, we don't want it to fall through to the value at a larger scope (if we did that, then if an option were set at the global scope, there would be no way to unset it in an app or session level scope.)

I thought about making a function for it, but the little details for each use case made it complicated enough that I decided against it.

R/runapp.R
wch: I think it's good to have the `applyCapturedAppOptions` function call just to make it clear that it's the counterpart to `captureAppOptions`. That function is also called from one other place, which is another reason to make it clear that it's the counterpart to `captureAppOptions`.

The purpose of this code is pretty confusing and it took me a while to wrap my head around it (even though I was the one who originally wrote it!), so anything to help is a good thing.

tests/testthat/test-options.R
```diff
@@ -70,3 +16,47 @@ test_that("Shared secret", {
...
   expect_error(checkSharedSecret(1:10))
 })
+
+
+test_that("Capturing options at creation time", {
+  # These are tests of captureAppOptions. Specific options should be captured
+  # and stored at app creation time and immediately reset. If another app is
+  # created, it will also capture these options at creation time.
+  # When these apps are started, they will apply the options.
+  shinyOptions(bookmarkStore = 123)
```
schloerke, line 26: Also add a value (and test) for `appDir`

tests/testthat/test-options.R
```diff
@@ -70,3 +16,47 @@ test_that("Shared secret", {
   expect_false(checkSharedSecret(NULL))
   expect_error(checkSharedSecret(1:10))
 })
+
```
schloerke, line 19:
```suggestion

test_that("Captured options contain expected elements", {
  shinyOptions(bookmarkStore = 123)
  shinyOptions(appDir = normalizePath("../")) # stomped
  caps <- captureAppOptions()
  
  expect_equal(sort(names(caps)), c("appDir", "bookmarkStore")
  expect_equal(caps$appDir, getwd())
  expect_equal(caps$bookmarkStore, 123)

  # verify that options are reset
  expect_equal(getShinyOption("bookmarkStore"), NULL)
  expect_equal(getShinyOption("appDir"), NULL)
})
```

schloerke (approved):
LGTM given testing suggestion

tests/testthat/test-options.R
```diff
@@ -17,6 +17,19 @@ test_that("Shared secret", {
   expect_error(checkSharedSecret(1:10))
 })
 
+test_that("Captured options contain expected elements", {
+  shinyOptions(bookmarkStore = 123)
+  shinyOptions(appDir = normalizePath("../")) # stomped
+  caps <- captureAppOptions()
+  
+  expect_equal(sort(names(caps)), c("appDir", "bookmarkStore")
```
cpsievert:
```suggestion
  expect_equal(sort(names(caps)), c("appDir", "bookmarkStore"))
```

R/shiny-options.R
```diff
@@ -8,31 +8,85 @@ getShinyOption <- function(name, default = NULL) {
...
+#'   There are a number of global options that affect Shiny's behavior. These
+#'   can be set globally with `options()` or locally (for a single app) with
+#'   `shinyOptions()`.
+#'
+#' \describe{
+#' \item{bootstraplib (defaults to `FALSE`)}{When a global **bootstraplib** theme
+#'   is set (via [bootstraplib::bs_theme_new()]), should **shiny** compile it
+#'   (via [bootstraplib::bootstrap()]) and include the result with [bootstrapPage()]
+#'   as well as use it to set better default styles for input widgets (e.g., [selectInput()])?}
+#' }
```
cpsievert, line 87: This is a fantastic overhaul of the docs, thank you! Only nit here is that we likely won't be documenting a `bootstraplib` `shinyOption()`, so this section might end up being empty after https://github.com/rstudio/shiny/pull/3062, but maybe this provides motivation move the existing `getOption()` calls to `getShinyOption()` (as we made a note to do in https://github.com/rstudio/shiny/issues/2953)?

wch: I'll take out the `bootstraplib` docs for now, and I'll add in documentation for `cache`, since it should be there.

As for the `shinyOptions` and `options` fall-through, I can do that in another PR.

cpsievert: Ohh, actually, `shiny.testmode` should be transferred to this section in this PR, so at least that gets rid of the empty section issue

wch: `shiny.testmode` actually should still be in the `options` section.

There's an internally-used shinyOption named `testmode`, but its purpose is actually to transfer the value of the `test.mode` argument from `runApp()` (as in `runApp(test.mode=T)`) to both the UI and server code. That was one of the early uses of `shinyOptions`, where the purpose was to capture some information at the application level and share it with both the UI and server code.

I could change the name of the internally-used shinyOption `testmode` to disambiguate it from the global option `shiny.testmode`.

wch: Although, now that I look at it, I think we can actually remove the `test.mode` option from `runApp()`, since shinytest doesn't actually use it. I'll merge this and then clean all that up in another PR.

PR #3085, opened 2020-10-02
More generalized approach to real-time theming

cpsievert:
Requires https://github.com/rstudio/bootstraplib/pull/120


R/shiny-options.R
```diff
@@ -246,10 +246,16 @@ shinyOptions <- function(...) {
 captureAppOptions <- function() {
   options <- list(
     appDir = getwd(),
-    bookmarkStore = getShinyOption("bookmarkStore")
+    bookmarkStore = getShinyOption("bookmarkStore"),
+    bootstrapTheme = getShinyOption("bootstrapTheme"),
```
wch, line 250: Is it necessary to capture these options?

cpsievert: I was thinking it might be for `themeDependencyFuncs` to avoid a situation where creating two apps then running one of them would call all the registered functions (for both apps), but I suppose that's not a problem as long as we [do the registration in `tagFunction()` and check to make sure `shiny::isRunning()`](https://github.com/rstudio/bootstraplib/pull/120/files#diff-68e15f5a86c76d37b4d079274c90aa4fR137)

R/bootstrap.R
```diff
@@ -117,39 +120,63 @@ is_bs_theme <- function(x) {
...
+#'       primary(get_primary())
+#'     }, ignoreInit = TRUE)
+#'
+#'   }
+#'   shinyApp(ui, server)
 #' }
 #'
 #' @export
 getCurrentTheme <- function() {
+  # TODO: we still might need to elevate this to a reactive read
```
wch, line 150: One obstacle to making this reactive: app-level vs. session-level scoping. In most cases, when this is called, the session doesn't yet exist.

tools/documentation/pkgdown.yml
```diff
@@ -161,6 +161,7 @@ reference:
       - createWebDependency
       - resourcePaths
       - registerInputHandler
+      - registerThemeDependency
```
cpsievert, line 164: Is this really needed? I don't think pkgdown requires `@keywords internal` functions here (but maybe our custom pkgdown site does?)

R/bootstrap.R
```diff
@@ -111,45 +127,110 @@ is_bs_theme <- function(x) {
...
+#'       primary(get_primary())
+#'     }, ignoreInit = TRUE)
 #'
+#'   }
+#'   shinyApp(ui, server)
+#' }
+#' @keywords internal
 #' @export
-getCurrentTheme <- function() {
+getCurrentTheme <- function(session = getDefaultReactiveDomain()) {
```
cpsievert, line 164:
```suggestion
getCurrentTheme <- function() {
```

R/bootstrap.R
```diff
@@ -111,45 +127,110 @@ is_bs_theme <- function(x) {
 #' Intended for use by Shiny developers to create Shiny bindings with intelligent
 #' styling based on the [bootstrapLib()]'s `theme` value.
 #'
+#' @param session A Shiny session object.
+#'
```
cpsievert, line 131:
```suggestion
```

tools/documentation/pkgdown.yml
wch: Hm, that might have been from before it had `@keywords internal`. I'll remove it.

R/bootstrap.R
```diff
@@ -127,73 +125,26 @@ is_bs_theme <- function(x) {
...
-  invisible()
 }
 
 #' Register a theme dependency
 #'
 #' This function registers a function that returns an [htmlDependency()] or list
-#' of such objects. If `setCurrentTheme()` is called from within a Shiny
-#' session, the function will be re-executed, and the resulting html dependency
-#' will be sent to the client.
+#' of such objects. If `seesion$setCurrentTheme()` is called, the function will
```
cpsievert, line 146:
```suggestion
#' of such objects. If `session$setCurrentTheme()` is called, the function will
```

wch (approved):

PR #3087, opened 2020-10-05
Added lang attribute for accessibility and search-engine parsers

### LINKED ISSUE #2844: Shiny HTML element has no lang attribute

schloerke:
Fixes #2844 
Fixes #2920

#2920 main comment:
> * Added the `lang` argument to `shinyApp()` that specifies document-level language within the app for the accessibility of screen readers and search-engine parsers.
> * Users can pass [ISO 639-1 language code](https://www.w3schools.com/tags/ref_language_codes.asp) for this argument.
> * Defaulting to `lang = ""` when no input is given by users. I was pondering the default  lang value set to either "en" or auto-locale-detect method; however, that can lead to unexpected issues. For example, many people use English R and locale while creating shiny apps in their own languages. We cannot assume only based upon users' locale which can lead to other accessibility side effects for screen readers. Thus, defaulting to empty string is a safer way.
>
> \- `@jooyoungseo`

Followup changes to #2920: 

* Removed if statements
* Assign lang unconditionally
* Removed `lang` from `basicPage()` args
* Use `@inheritParams` for `@param lang`
* Added helper method to create the `html` start tag. 
  * Avoids `<html lang="">` when lang is `NULL` or empty string.

### DIFF
--- a/R/shinyui.R
+++ b/R/shinyui.R
@@ -41,7 +41,9 @@ renderPage <- function(ui, showcase=0, testMode=FALSE) {
     ui <- htmlTemplate(
       system.file("template", "default.html", package = "shiny"),
       lang = lang,
-      body = ui
+      body = ui,
+      # this template is a complete HTML document
+      document_ = TRUE
     )
   }
 
--- a/R/shinyui.R
+++ b/R/shinyui.R
@@ -25,6 +25,9 @@ withMathJax <- function(...) {
 }
 
 renderPage <- function(ui, showcase=0, testMode=FALSE) {
+  # Check if ui has lang attribute; otherwise, NULL
+  lang <- attr(ui, "lang", exact = TRUE)
+
   # If the ui is a NOT complete document (created by htmlTemplate()), then do some
   # preprocessing and make sure it's a complete document.
   if (!inherits(ui, "html_document")) {
@@ -38,6 +41,7 @@ renderPage <- function(ui, showcase=0, testMode=FALSE) {
     # Put the body into the default template
     ui <- htmlTemplate(
       system.file("template", "default.html", package = "shiny"),
+      lang = lang,
       body = ui
     )
   }
--- a/R/shinyui.R
+++ b/R/shinyui.R
@@ -25,8 +25,7 @@ withMathJax <- function(...) {
 }
 
 renderPage <- function(ui, showcase=0, testMode=FALSE) {
-  # Check if ui has lang attribute; otherwise, NULL
-  lang <- attr(ui, "lang", exact = TRUE)
+  lang <- getLang(ui)
 
   # If the ui is a NOT complete document (created by htmlTemplate()), then do some
   # preprocessing and make sure it's a complete document.
--- a/inst/template/default.html
+++ b/inst/template/default.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-{{ htmlStartTag(lang) }}
+<html{{ if (isTRUE(nzchar(lang))) paste0("lang=\"", lang, "\"") }}>
 <head>
 {{ headContent() }}
 </head>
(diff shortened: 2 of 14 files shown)


R/shinyui.R
```diff
@@ -38,7 +40,10 @@ renderPage <- function(ui, showcase=0, testMode=FALSE) {
     # Put the body into the default template
     ui <- htmlTemplate(
       system.file("template", "default.html", package = "shiny"),
-      body = ui
+      lang = lang,
+      body = ui,
+      # this template is a complete HTML document
+      document_ = TRUE
```
wch, line 46: Just checking: the template previously was autodetected so that this behaved like it was TRUE, right?

inst/template/default.html
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html>
+<html{{ if (isTRUE(nzchar(lang))) paste0("lang=\"", lang, "\"") }}>
```
wch, line 2: Just checking: if `lang` is NULL, would any extra spaces be inserted here?

R/shinyui.R
schloerke: Correct. I'll be removing this line though

inst/template/default.html
schloerke: Nope

```
❯❯ isTRUE(nzchar(NULL))
[1] FALSE
```

tests/testthat/test-test.R
```diff
@@ -161,7 +161,7 @@ test_that("app template works with runTests", {
     )
   ))
 
-  lapply(combos, function(combo) {
+  for (combo in combos) {local({
```
wch, line 164: Just curious: why the change?

inst/template/default.html
schloerke: And everything between the brackets (and including the brackets) will be removed

wch (approved):
Looks good! I just had a few small questions.

tests/testthat/test-test.R
schloerke: `testthat` says that the failure occurs at line 164 and not 172 or the actual failure line

R/shinyui.R
schloerke: I'm going to keep this line in.

Prior behavior had `document_ = "auto"` by default.  It would hit these lines of code in `htmltools`: https://github.com/rstudio/htmltools/blob/f589c77bc941428d4bf2cec3ca5f15399ce18f47/R/template.R#L63-L65 where `htmltools` would test for some form of `<html>` to set `document_`.

To make sure the `lang` logic does not corrupt the template `document_` logic, I am explicitly setting `document_ = TRUE`.

inst/template/default.html
wch: Sounds good - I was mostly interested in making sure that the `{{ }}` didn't result in any extra spaces.

PR #3092, opened 2020-10-07
Implement reactiveConsole()

### LINKED ISSUE #2518: Top level reactivity simulation
I think it would be useful to have some exported API to make it easy to simulate reactivity at the console:

```R
simulation <- new.env(parent = emptyenv())
simulation$id <- NULL

simulation_on <- function() {
  if (!is.null(simulation$id)) {
    stop("Simulation already active", call. = FALSE)
  }
  
  options(shiny.suppressMissingContextError = TRUE)
  simulation$id <- addTaskCallback(function(...) {
    shiny:::flushReact()
    TRUE
  })
  invisible()
}

simulation_off <- function() {
  if (is.null(simulation$id)) {
    stop("Simulation not active", call. = FALSE)
  }
  
  options(shiny.suppressMissingContextError = FALSE)
  removeTaskCallback(simulation$id)
  invisible()
}
```

Obviously this is only an approximation because it will only flush after a top level event, but this seems like

hadley:
Fixes #2518

Given the nearby functions in the namespace, I'm not sure that this is a good name.


wch (approved):

R/reactives.R
```diff
@@ -1377,35 +1377,34 @@ observe <- function(x, env=parent.frame(), quoted=FALSE, label=NULL,
...
-  values <- reactiveValues(value = initialValue)
-  makeActiveBinding(symbol, env=env, fun=function(v) {
-    if (missing(v))
-      values$value
-    else
-      values$value <- v
-  })
+  }
+
+  val <- reactiveVal(initialValue)
```
jcheng5, line 1406:
```suggestion
  val <- reactiveVal(initialValue, label = symbol)
```
(for reactlog purposes)

R/reactives.R
```diff
@@ -1441,6 +1440,29 @@ setAutoflush <- local({
...
+#' @examples
+#' reactiveConsole(TRUE)
+#' x <- reactiveVal(10)
+#' y <- observe({
+#'   message("The value of x is ", x())
+#' })
+#' x(20)
+#' x(30)
+#' reactiveConsole(FALSE)
+reactiveConsole <- function(enabled) {
```
jcheng5, line 1461: Time based reactives (i.e. `invalidateLater()`) won't currently work--but they could...

hadley: `timerCallbacks$executeElapsed()` doesn't handle time based reactives?

jcheng5: Well it does, but, it's only invoked right at the moment the task callback is called. We could use `later` to invoke it "spontaneously".

hadley: Ah, yeah, makes sense. That doesn't block this PR though, right?

jcheng5: Not at all.

PR #3096, opened 2020-10-14
Make template test less brittle and failures more verbose

schloerke:
* Add `registerThemeDependency()` to `"Extending Shiny"` section of docs
* Do not mix output from calling `runTests()` with the printed result of `runTests()`
	* While calling `runTests()`, output is created. The test is to only test the result of `runTests()`.

### DIFF
--- a/tools/documentation/pkgdown.yml
+++ b/tools/documentation/pkgdown.yml
@@ -158,7 +158,6 @@ reference:
     desc: Functions that are intended to be called by third-party packages that extend Shiny.
     contents:
       - createWebDependency
-      - registerThemeDependency
       - resourcePaths
       - registerInputHandler
       - removeInputHandler
--- a/tools/documentation/pkgdown.yml
+++ b/tools/documentation/pkgdown.yml
@@ -158,6 +158,7 @@ reference:
     desc: Functions that are intended to be called by third-party packages that extend Shiny.
     contents:
       - createWebDependency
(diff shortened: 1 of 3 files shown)


wch (approved):

tools/documentation/pkgdown.yml
```diff
@@ -158,6 +158,7 @@ reference:
     desc: Functions that are intended to be called by third-party packages that extend Shiny.
     contents:
       - createWebDependency
+      - registerThemeDependency
```
cpsievert, line 161: Hopefully no one will have to know this function exists (it's `@keywords internal`). The reason it's exported is so that `bootstraplib::bs_dependency_defer()` can use it under-the-hood

schloerke: Ah, ok. Then I'ma change `documentation` logic. Currently since it's exported, it's forcing a pkgdown entry.

schloerke:
Oh! The "brittle" tests were prolly the error found in https://github.com/rstudio/shiny/pull/3095

Still going to do some more work to make the error messaging to us better

tools/documentation/pkgdown.yml
schloerke: I've removed `registerThemeDependency` from the public docs. Also updated the error message to tell us where to make the fixes

cpsievert (approved):

PR #3097, opened 2020-10-15
Get session$setCurrentTheme() working on IE11 & improve timing of disabling old stylesheets

cpsievert:
## Testing notes

Install `remotes::install_github("rstudio/shiny")`, then run:

```r
bs_theme_preview(bs_theme())
```

Now change the background color to a dark color and pay careful attention to the update. Before this change, you should notice a [FOUC](https://en.wikipedia.org/wiki/Flash_of_unstyled_content), and after this change, it should go away. 

Also, an example like this can now work on IE11

```r
library(shiny)
library(bootstraplib)
light <- bs_theme()
dark <- bs_theme(bg = "black", fg = "white", primary = "purple")
ui <- fluidPage(
  theme = light,
  checkboxInput("dark_mode", "Dark mode", value = FALSE)
)
  observe({
    session$setCurrentTheme(
      if (isTRUE(input$dark_mode)) dark else light
    )
  })
}
shinyApp(ui, server)

```


srcjs/output_binding_html.js
```diff
@@ -107,23 +107,25 @@ function renderDependency(dep) {
...
       var this_href = href + "/" + encodeURI(stylesheet);
+      $head.append(
+        // Force client to re-request the stylesheet if we've already seen it
+        // (without this unique param, the request might get cached)
+        link.attr("href", restyle ? (this_href + "?restyle=" + new Date().getTime()) : this_href)
+      );
+      if (!restyle) return;
       // If a styleSheet with this href is already active, then disable it
-      var re = new RegExp(this_href + "$");
+      var re = new RegExp(this_href.split("/").join("\\/") + "$");
```
wch, line 120: Instead of the split and join, maybe use encodeURIComponent or encodeURI?

srcjs/output_binding_html.js
```diff
@@ -107,23 +107,25 @@ function renderDependency(dep) {
...
       for (var i = 0; i < document.styleSheets.length; i++) {
         var h = document.styleSheets[i].href;
         if (!h) continue;
-        if (restyle && h.match(re)) document.styleSheets[i].disabled = true;
+        if (h.split("?restyle=")[0].match(re)) {
+          setTimeout(function() { document.styleSheets[i].disabled = true; }, 10);
```
wch, line 125: How about also adding `document.styleSheets[i].shiny_remove = true;`, and then after a `setTimeout(, 11)`, a function iterates over `document.styleSheets` and removes all items with that attribute?

srcjs/output_binding_html.js
cpsievert: What I'd really want is a way to escape all the regex special characters, but I think I've found a reasonable to go about the same thing without regexing

srcjs/output_binding_html.js
```diff
@@ -107,23 +107,48 @@ function renderDependency(dep) {
...
+    $head.append(links);
+
+    // Disable and remove old stylesheets on the next tick
+    // The timeout amount here is currently a total guess...
+    // At first I had 10 milliseconds, but encountered some FOUC
+    // with bootstraplib's real-time themer. In the future, if we encounter
+    // other examples of FOUC we may want to consider increasing the timeout
+    if (restyle) {
+      setTimeout(function() {
+        for (var i = 0; i < document.styleSheets.length; i++) {
```
wch, line 144: I see a potential issue here with calling `sheet.ownerNode.remove()` while iterating over `document.styleSheets`. I think there's a good chance this will alter `document.styleSheets` as you're iterating over it, which is not a good idea, since that means that this loop could skip over elements. (That's why I had suggested adding a callback that took a separate final pass after all the sheets were disabled, and removed any elements that had been disabled.)

Also, is `document.styleSheets` OK to use? See:
https://caniuse.com/mdn-api_documentorshadowroot_stylesheets
https://developer.mozilla.org/en-US/docs/Web/API/DocumentOrShadowRoot/styleSheets.

srcjs/output_binding_html.js
```diff
@@ -107,23 +107,55 @@ function renderDependency(dep) {
...
+          if (!sheet._shiny_restyle) return null;
+          sheet.disabled = true;
+          // disabled=true doesn't work on IE11, but this does :shrug:?
+          if (navigator.appName === 'Microsoft Internet Explorer') {
+            sheet.cssText = "";
+          }
+          return sheet.ownerNode;
+        });
+        $.map(staleNodes, function(node) { if (node) $(node).remove(); });
+      }, 100);
```
wch, line 157: This assumes that all the stylesheets would have loaded within 100ms, which may not be true, especially on slow connections.

srcjs/output_binding_html.js
```diff
@@ -107,23 +107,55 @@ function renderDependency(dep) {
...
+        var url = link.attr("href") + "?restyle=" + new Date().getTime();
+        xhr.open('GET', url);
+        xhr.onreadystatechange = function() {
+          if (this.readyState === 4 && this.status === 200) {
+            xhr.onreadystatechange = null;
+            var id = "shiny_restyle_" + link.attr("href").replace(/\W/g,'_');
+            var style_old = $head.find("style#" + id);
+            var style_new = $("<style>").attr("id", id).html(this.responseText);
+            $head.append(style_new);
+            style_old.remove();
```
wch, line 140: `find_old_sheet` would be a function that does essentially the same as lines 119-124.

```
var old_sheet = find_old_sheet(link.attr("href"));

if (style_old) setTimeout(() => style_old.remove(), 100);
if (old_sheet) setTimeout(() => { old_sheet.disabled = true; old_sheet.remove(); }, 100);
```

srcjs/output_binding_html.js
```diff
@@ -107,23 +107,55 @@ function renderDependency(dep) {
...
+          sheet._shiny_restyle = true;
+        }
+        // Request the new stylesheet (with a timestamp to prevent caching)
+        // and insert a new inline style. Note that this could also done by inserting
+        // a the new <link> (w/ timestamp in href), then disabling the 'stale'
+        // document.styleSheets, but that approach doesn't work with IE11 (and this one does)
+        var xhr = new XMLHttpRequest();
+        var url = link.attr("href") + "?restyle=" + new Date().getTime();
+        xhr.open('GET', url);
+        xhr.onreadystatechange = function() {
```
wch, line 133: I think it would be better to have this xhr stuff happen only when needed (in IE), because the `link` method seems more correct.

srcjs/output_binding_html.js
```diff
@@ -107,23 +107,68 @@ function renderDependency(dep) {
...
+      // This inline <style> based approach works for IE11
+      function refreshStyle(href, oldSheet) {
+        var xhr = new XMLHttpRequest();
+        xhr.open('GET', href);
+        xhr.onload = function() {
+          var id = "shiny_restyle_" + href.split("?restyle")[0].replace(/\W/g, '_');
+          var oldStyle = $head.find("style#" + id);
+          var newStyle = $("<style>").attr("id", id).html(xhr.responseText);
+          $head.append(newStyle);
+          if (oldStyle) setTimeout(() => oldStyle.remove(), 10);
```
wch, line 145: What if `oldStyle` doesn't exist?

srcjs/output_binding_html.js
```diff
@@ -107,23 +107,68 @@ function renderDependency(dep) {
...
+        var oldSheet = findSheet(link.attr("href"));
+        // Add a timestamp to the href to prevent caching
+        var href = link.attr("href") + "?restyle=" + new Date().getTime();
+        // Use inline <style> approach for IE, otherwise use the more elegant
+        // <link> -based approach
+        if (isIE()) {
+          refreshStyle(href, oldSheet);
+        } else {
+          link.attr("href", href);
+          link.attr("onload", () => {
```
wch, line 129: It would help to have a comment here explaining why the `setTimeout` is needed

srcjs/output_binding_html.js
```diff
@@ -107,23 +107,68 @@ function renderDependency(dep) {
...
+        var old = null;
+        for (var i = 0; i < document.styleSheets.length; i++) {
+          var sheet = document.styleSheets[i];
+          // The sheet's href is a full URL
+          if (typeof sheet.href === "string" && sheet.href.indexOf(href) > -1) {
+            old = sheet;
+            break;
+          }
+        }
+        return old;
```
wch, line 161:
```suggestion
        for (var i = 0; i < document.styleSheets.length; i++) {
          var sheet = document.styleSheets[i];
          // The sheet's href is a full URL
          if (typeof sheet.href === "string" && sheet.href.indexOf(href) > -1) {
            return sheet;
          }
        }
        return null;
```

srcjs/utils.js
```diff
@@ -346,3 +346,9 @@ function updateLabel(labelTxt, labelNode) {
   }
 
 }
+
+// https://stackoverflow.com/a/22551342/1583084
+function isIE() {
```
wch, line 351: Can you move this into browser.js, replacing `browser.isIE`?

And while you're in there, you might as well update `getIEVersion()` to work with IE11.

srcjs/output_binding_html.js
cpsievert: Even if the `.find()` comes up empty, `oldStyle.remove()` won't error out since it's a jQuery selection, so I suppose we could either remove the `if` or change it to `if (oldStyle.length > 0) setTimeout(() => oldStyle.remove(), 10);`?

wch (approved):

PR #3098, opened 2020-10-18
Standardise coercion to functions/reactives

hadley:
Currently renderPlot() also handles reactives specially, so it is inconsistent with `renderTable()`. This isn't documented, so I suspect it might be ok to use coerceToFunc() instead, but it would be a breaking change.


hadley:
@wch any interest in finishing this off? I think it would also be nice to add formula support for concise lambdas.

Issue #3105, opened 2020-10-20
Screen readers don't read radioButtons() labels


eliocamp:
The radio buttons created with radioButtons() don't work well with screen readers. I tired both with Orca and ChromeVox and neither of them read their label. 

To see the difference, this code creates two groups of radio buttons. Launch the app in Chrome with ChromeVox installed and navigate between them with tab. 

The when navigating to the first one it speaks "selected normal radio button selected" which means that there is no context of that "normal" means. When navigating to the second one it speaks "Distribution type radio button group normal radio button selected", which lets the user hear the context of the radio button.

```r
aria_radio <-  function (inputId, label, choices = NULL, selected = NULL, inline = FALSE, 
                         width = NULL, choiceNames = NULL, choiceValues = NULL) {
  args <- shiny:::normalizeChoicesArgs(choices, choiceNames, choiceValues)
  selected <- shiny::restoreInput(id = inputId, default = selected)
  selected <- if (is.null(selected)) 
    args$choiceValues[[1]]
  else as.character(selected)
  if (length(selected) > 1) 
    stop("The 'selected' argument must be of length 1")
  
  options <- aria_generateOptions(inputId, selected, inline, "radio",
                                     args$choiceNames, args$choiceValues)
  divClass <- "form-group shiny-input-radiogroup shiny-input-container"
  if (inline) shiny::divClass <- paste(divClass, "shiny-input-container-inline")
  
  
  shiny::tags$div(id = inputId, style = if (!is.null(width)) paste0("width: ", shiny::validateCssUnit(width), ";"),
                  class = divClass, role = "radiogroup", 
                  `aria-labelledby` = paste0(inputId, "label"),                                   # this is new
                  shiny::tags$label(label, class = "control-label", class = if (is.null(label)) "shiny-label-null", 
                                    `for` = inputId, 
                                    id = paste0(inputId, "label")                                         # this is also new.
                                    ),
                  options)
}

aria_generateOptions <- function (inputId, selected, inline, type = "checkbox", choiceNames, 
                                  choiceValues, session = shiny::getDefaultReactiveDomain()) {
  options <- mapply(choiceValues, choiceNames, FUN = function(value, 
                                                              name) {
    inputTag <- shiny::tags$input(type = type, name = inputId, 
                                  value = value)
    if (value %in% selected) {
      inputTag$attribs$checked <- "checked"
    } else {
    }
    
    if (type == "radio") {
      inputTag$attribs$role <- "radio"        # This is also needed
[...]

PR #3106, opened 2020-10-22
Improving radioButtons accessibility: Closing #3105

jooyoungseo:
This PR addresses #3105.

After this patch, ...:

1. Choices generated from `radioButtons()` are semantically grouped together for assistive technologies (e.g., screen readers).
1. Group-level label passed to `label` argument in `radioButtons()` is now tied to each choice, meaning that screen readers announce the group label along with each radio option name when tabbing around.

Test has been carried out with various screen readers, giving satisfactory results.

Note that there is no break for other Shiny functionalities, nor does it slowdown Shiny performance. (it simply improves the accessibility).

## Reprex

``` r
library(shiny)
ui <- fluidPage(
  radioButtons("dist", "Distribution type:",
               c("Normal" = "norm",
                 "Uniform" = "unif",
                 "Log-normal" = "lnorm",
                 "Exponential" = "exp")),
  plotOutput("distPlot")
)

    dist <- switch(input$dist,
                   norm = rnorm,
                   unif = runif,
                   lnorm = rlnorm,
                   exp = rexp,
                   rnorm)

    hist(dist(500))
  })
}

shinyApp(ui, server)
```


cpsievert:
Thanks @jooyoungseo! I imagine we should also be making similar changes to `checkboxGroupInput()`?

jooyoungseo:
@cpsievert, no problem!

Yes, I think we may want to apply the same patch to the checkbox group, which I need to verify later on.

What about we open that PR separately so that I will be able to pull and test?

cpsievert:
@jooyoungseo sounds good, thanks

cpsievert (approved):

R/input-radiobuttons.R
```diff
@@ -108,6 +108,11 @@ radioButtons <- function(inputId, label, choices = NULL, selected = NULL,
     style = if (!is.null(width)) paste0("width: ", validateCssUnit(width), ";"),
     class = divClass,
     shinyInputLabel(inputId, label),
+    # Improve accessibility (#3105)
+    # https://www.w3.org/TR/2017/WD-wai-aria-practices-1.1-20170628/examples/radio/radio-1/radio-1.html
+    role = "radiogroup",
+    # The `aria-labelledby` is tied to id value in `shinyInputLabel()` to be announced to screen readers and in Braille:
+    `aria-labelledby` = paste0(inputId, "-label"),
```
wch, line 115: Probably safer to extract this value from the object returned by `shinyInputLabel(inputId, label)`

cpsievert:
Thanks @jooyoungseo! We'll take this over the finish line in https://github.com/rstudio/shiny/pull/3186

PR #3114, opened 2020-10-27
Warn when running in a package directory

wch:
I've tested this with a DESCRIPTION file that has `Type: Package`, and a DESCRIPTION file without that. It warns in the first case, and not in the second.

### DIFF
--- a/R/shinyapp.R
+++ b/R/shinyapp.R
@@ -345,6 +345,17 @@ loadSupport <- function(appDir=NULL, renv=new.env(parent=globalenv()), globalren
     appDir <- findEnclosingApp(".")
   }
 
+  descFile <- file.path.ci(appDir, "DESCRIPTION")
+  if (file.exists(descFile) &&
+      identical(as.character(read.dcf(descFile, fields = "Type")), "Package"))
+  {
+    warning(
+      "Loading R/ subdirectory for Shiny application, but the DESCRIPTION file ",
+      "says this directory contains an R package. Sourcing files in R/ may cause ",
+      "unexpected behavior."
+    )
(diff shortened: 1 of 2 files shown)


R/shinyapp.R
```diff
@@ -345,6 +345,17 @@ loadSupport <- function(appDir=NULL, renv=new.env(parent=globalenv()), globalren
     appDir <- findEnclosingApp(".")
   }
 
+  descFile <- file.path.ci(appDir, "DESCRIPTION")
+  if (file.exists(descFile) &&
+      identical(as.character(read.dcf(descFile, fields = "Type")), "Package"))
```
cpsievert, line 350: I'm pretty sure `Type` isn't required and defaults to `Package`, so we should also warn if it's missing?

wch: Hm, the problem is that Shiny apps can have a DESCRIPTION file with no `Type` field.

How about we also check for a NAMESPACE file?

cpsievert: > Hm, the problem is that Shiny apps can have a DESCRIPTION file with no Type field.

Packages, even on CRAN, can as well (plotly doesn't, for example)

> How about we also check for a NAMESPACE file? I think almost all packages have NAMESPACE files these days.

Sounds good to me

wch: >> Hm, the problem is that Shiny apps can have a DESCRIPTION file with no Type field.

> Packages, even on CRAN packages, can as well (plotly doesn't, for example)

Right, my point was that if we warned if there was a missing Type field, many legitimate Shiny apps would get a warning.

PR #3118, opened 2020-10-29
Use cache objects from cachem package

wch:
The `memoryCache` and `diskCache` from Shiny were migrated to [cachem](https://github.com/wch/cachem), and this PR uses that package.


hadley:
I think you'll probably need to leave shim `diskCache()` and `memoryCache()` functions that signal deprecation then call the appropriate cachem function.

wch:
Oops, I meant to do that. Thanks for catching it.

R/deprecated.R
```diff
@@ -0,0 +1,90 @@
...
+  max_age = Inf,
+  max_n = Inf,
+  evict = c("lru", "fifo"),
+  missing = key_missing(),
+  exec_missing = FALSE,
+  logfile = NULL)
+{
+  shinyDeprecated("cachem::cache_mem", version = "1.5.1")
+
+  cachem::cache_disk(
```
cpsievert, line 81:
```suggestion
  cachem::cache_mem(
```

PR #3127, opened 2020-11-03
Add `bindCache()` and `bindEvent()`

wch:
This removes `cachedReactive()` and adds the following two functions:

* `bindCache()`
* `bindEvent()`

The `bindCache()` function takes a reactive expression or render function, along with one or more _cache key expressions_, and returns a cached version of the object.

The `bindEvent()` function takes a reactive expression, render function, or observer, along with one or more _event expressions_, and returns a version of that object that fires only when the event expression is invalidated.

It also updates `observe()` and `reactive()` so that they accept quosures.

*****

To add caching to a reactive expression (imagine that the computation is an expensive one, not just multiplication):

```R
r <- reactive({
    input$x * input$y
  }) %>%
  bindCache(input$x, input$y)
```

To a `render` function:

```R
output$txt <- renderText({
    paste0("input$x is ", input$x)
  }) %>%
  bindCache(input$x, input$y)
```

`bindEvent()` can be used for these two kinds of objects, as well as observers:

```R
observe({
    message("input$x is ", input$x)
  }) %>%
  bindEvent(input$x)

# Equivalent to code with the existing observeEvent()
observeEvent(input$x, {
  message("input$x is ", input$x)
})
```

`bindCache()` and `bindEvent()` can be composed to get both caching and event-triggering. This can be useful if, for example, you a calculation to occur only when a button is pressed:

```R
r <- reactive({
    input$x * input$y
  }) %>%
  bindCache(input$x, input$y) %>%
  bindEvent(input$go)
```

Here's a full app example for `bindCache()`:

```R
shinyApp(
  ui = fluidPage(
    sliderInput("x", "x", 1, 10, 5),
    sliderInput("y", "y", 1, 10, 5),
    div("x * y: "),
    verbatimTextOutput("txt")
  ),
    r <- reactive({
      # The value expression is an _expensive_ computation
      message("Doing expensive computation...")
      Sys.sleep(2)
      input$x * input$y
    }) %>%
      bindCache(input$x, input$y)

    output$txt <- renderText(r())
  }
)
```

And another using both `bindCache()` and `bindEvent()`:

```R
shinyApp(
  ui = fluidPage(
    sliderInput("x", "x", 1, 10, 5),
    sliderInput("y", "y", 1, 10, 5),
    actionButton("go", "Go"),
    div("x * y: "),
    verbatimTextOutput("txt")
  ),
    r <- reactive({
      message("Doing expensive computation...")
      Sys.sleep(2)
      input$x * input$y
    }) %>%
      bindCache(input$x, input$y) %>%
      bindEvent(input$go)
      # The cached, eventified reactive takes a reactive dependency on
      # input$go, but doesn't use it for the cache key. It uses input$x and
      # input$y for the cache key, but doesn't take a reactive depdency on
      # them, because the reactive dependency is superseded by addEvent().

    output$txt <- renderText(r())
  }
)
```


R/with-cache.R
```diff
@@ -0,0 +1,522 @@
...
+#'   between the sessions.
+#'
+#'   It is also possible to scope the cache to the session, with
+#'   `cache="session"`. This removes the risk of information leaking between
+#'   sessions, but then one session cannot benefit from computations performed in
+#'   another session.
+#'
+#'   Finally, it is possible to pass in caching objects directly to
+#'   `withCache()`. This can be useful if, for example, you want to use a
+#'   particular type of cache with specific cached reactives.
```
hadley, line 146: I think it's worth specifically mentioning the disk cache here as a way of sharing the cache across processes.

R/with-cache.R
```diff
@@ -0,0 +1,522 @@
...
+#'   In general, `key` should use the same reactive inputs as `value`, but the
+#'   computation should be simpler. If there are other (non-reactive) values
+#'   that are consumed, such as external data sources, they should be used in
+#'   the `key` as well. Note that if the `key` is large, it can make sense to do
+#'   some sort of reduction on it so that the serialization and hashing of the
+#'   cache key is not too expensive.
+#'
+#'   Remember that the key is _reactive_, so it is not re-executed every single
+#'   time that someone accesses the cached reactive. It is only re-executed if
+#'   it has been invalidated by one of the reactives it depends on.. For
```
hadley, line 78:
```suggestion
#'   it has been invalidated by one of the reactives it depends on. For
```

R/with-cache.R
```diff
@@ -0,0 +1,522 @@
...
+#'
+#' `withCache()` adds caching to the following kinds of objects used in Shiny:
+#'
+#' * [reactive()]` expressions.
+#' * `render*` functions, like [renderText()], [renderTable()], and so on.
+#'
+#' It is often used in conjunction with [withEvent()].
+#'
+#'
+#' Ordinary [reactive()] expressions will cache their most recent value. This
```
hadley, line 13: This description is a bit long I think; you probably want to keep it to 2-3 paragraphs.

R/with-cache.R
```diff
@@ -0,0 +1,522 @@
...
+#' can make computation more efficient, because time-consuming code can execute
+#' once and the result can be used in multiple different places without needing
+#' to re-execute the code.
+#'
+#' When `withCache()` is used on a reactive expression, any number of previous
+#' values can be cached. You must provide one or more expressions that are used
```
hadley, line 19: Mention that total size is capped?

R/with-cache.R
```diff
@@ -0,0 +1,522 @@
...
+#' and hashed, and the result is used as the cache **key**.
+#'
+#' There is also a **value** expression, which is the expression passed to the
+#' original reactive.
+#'
+#' For each possible **key**, there should be one possible **value** returned by
+#' the original reactive expression; a given key should not correspond to
+#' multiple possible values from the reactive expression.
+#'
+#' To use `withCache(), the key should be fast to compute, and the value should
```
hadley, line 30:
```suggestion
#' To use `withCache()` the key should be fast to compute, and the value should
```

R/with-cache.R
```diff
@@ -0,0 +1,522 @@
...
+#' expensive (so that it benefits from caching).  To see the value should be
```
hadley, line 31:
```suggestion
#' expensive (so that it benefits from caching).  To see if the value should be
```

R/with-cache.R
```diff
@@ -0,0 +1,522 @@
...
+#' computed, a cached reactive evaluates the key, and then serializes and hashes
+#' the result. If the resulting hashed key is in the cache, then the cached
+#' reactive simply retrieves the previously calculated value and returns it; if
+#' not, then the value is computed and the result is stored in the cache before
+#' being returned.
+#'
+#' The key should be fast to compute, and ideally it will return an object that
```
hadley, line 38: Duplicates previous para

R/with-cache.R
```diff
@@ -0,0 +1,522 @@
...
+#'   some sort of reduction on it so that the serialization and hashing of the
+#'   cache key is not too expensive.
+#'
+#'   Remember that the key is _reactive_, so it is not re-executed every single
+#'   time that someone accesses the cached reactive. It is only re-executed if
+#'   it has been invalidated by one of the reactives it depends on.. For
+#'   example, suppose we have this cached reactive:
+#'
+#'   ```
+#'   r <- cachedReactive(
```
hadley, line 82: Needs update

R/with-cache.R
```diff
@@ -0,0 +1,522 @@
...
+#'
+#' Note that if the cached reactive is passed to [withEvent()], then the key
+#' expression will no longer be reactive; instead, the event expression will
+#' be reactive.
+#'
+#'
+#' @section Async with cached reactives:
+#'
+#'  With a cached reactive expression, the key and/or value expression can be
+#'  _asynchronous_. In other words, they can be promises -- not regular R
```
hadley, line 107:
```suggestion
#'  _asynchronous_. In other words, they can be promises --- not regular R
```

R/with-cache.R
```diff
@@ -0,0 +1,522 @@
...
+#'  promises, but rather objects provided by the \pkg{promises}  package, which
```
hadley, line 108: Needs link to package website

R/with-cache.R
```diff
@@ -0,0 +1,522 @@
...
+#'   means that it shares a cache with all user sessions connected to the
+#'   application (within the R process). This is done with the `cache`
+#'   parameter's default value, `"app"`.
+#'
+#'   With an app-level cache scope, one user can benefit from the work done for
+#'   another user's session. In most cases, this is the best way to get
+#'   performance improvements from caching. However, in some cases, this could
+#'   leak information between sessions. For example, if a user sees that a cached
+#'   reactive returns its value very quickly, they may be able to infer that
+#'   someone else has already used it with the same values. Or if the cache key
```
hadley, line 135: I think the cache key problem should come first (with an example) because it's way more important.

R/with-cache.R
```diff
@@ -0,0 +1,522 @@
...
+#'  Similarly, if the key is a promise (in other words, if it is asynchronous),
+#'  then the entire cached reactive must be asynchronous, since the key must be
+#'  computed asynchronously before it knows whether to compute the value or the
+#'  value is retrieved from the cache. Anything that consumes the cached
+#'  reactive must therefore expect it to return a promise.
+#'
+#'
+#' @section Cache scope:
+#'
+#'   By default, a cached reactive is scoped to the running application. That
```
hadley, line 125: I don't know how important this is, but I was left wondering if there's one overall cache, or one cache per reactive (i.e. is the bound on size per object or per app?)

hadley (commented):
Unsolicited drive-by review since I'm writing about this rn.

R/with-cache.R
```diff
@@ -0,0 +1,522 @@
...
+
+  class(res) <- c("reactive.cache", class(res))
+  res
+}
+
+#' @export
+withCache.shiny.render.function <- function(x, ..., cache = "app") {
+  keyFunc <- make_quos_func(enquos(...))
+
+  valueExprHash <- "TODO"
```
jcheng5, line 393: This is going to need to stir in the `renderXXX` invocation as well

PR #3138, opened 2020-11-10
Update render functions to accept quosures

wch:
Some render functions already accept quosures as of #3127. This PR updates the ones that remain.


schloerke (approved):

wch:
Going with #3139 instead

PR #3143, opened 2020-11-12
added ... support to icon(); fixes #3140

### LINKED ISSUE #3140: Add the .noWS argument to icons
All HTML tags created with the `htmltools::tags` list can take an optional `.noWS` parameter that's used to suppress whitespace. There is no way to achieve this whitespace suppression using the `icon()` function that creates an `i` tag. The only way currently is to manually create an `i` tag without the use of `icon()`. 

Example:

Using `icon()` inside a div, I can get the following:

```
> htmltools::tags$div("foo", shiny::icon("bar"))
<div>
  foo
  <i class="fa fa-bar" role="presentation" aria-label="bar icon"></i>
</div>
```

I would like to be able to achieve the following:

```
> htmltools::tags$div("foo", shiny::icon("bar", .noWS = "before"))
<div>
  foo<i class="fa fa-bar" role="presentation" aria-label="bar icon"></i>
</div>
```

I can do that if I use `tags$i()` instead of `icon()`.

daattali:


daattali:
fixes #3140 (trying to get the "Linked issues" on the right hand to link)

NEWS.md
```diff
@@ -58,6 +58,8 @@ shiny 1.5.0.9000
 
 * Shiny now uses `cache_mem` from the cachem package, instead of `memoryCache` and `diskCache`. (#3118)
 
+* Added support for `...` argument in `icon()`. (#3140)
```
wch, line 61:
```suggestion
* Closed #3140: Added support for `...` argument in `icon()`. (#3143)
```

daattali:
@wch fixed

wch (approved):

PR #3144, opened 2020-11-12
Don't unwrap quosures

### LINKED ISSUE #3139: rlang's quasi-quotation is evaluated too early in reactive expressions
Minimal example:

```r
library(shiny)

ui <- fluidPage(actionButton("x", "Press me"))

  observeEvent(input$x, {
    foo <- list(input$x)
    code <- rlang::expr(print(!!!foo))
    rlang::eval_tidy(code)
  })
  
}

shinyApp(ui, server)
```

On startup, the app crashes with

```r
Listening on http://127.0.0.1:3445
Warning: Error in enquo: object 'foo' not found
  51: enquo
  50: observeEvent [/Users/cpsievert/github/shiny/R/reactives.R#2281]
  49: server [#3]
Error in enquo(handlerExpr) : object 'foo' not found
```

Suggesting that the `rlang::expr(print(!!!foo))` is being evaluated at the wrong time with the wrong environment. I've confirmed that this was introduced by #3127

wch:
Closes #3139. This PR changes `reactive`, `observe`, and `render` functions so that they do not unwrap quosures on creation.

For example, this app now works:

```R
library(shiny)
library(dplyr)

shinyApp(
  fluidPage(
    checkboxGroupInput("vars", "Vars", names(mtcars)),
    tableOutput("tbl")
  ),
  function(input, output) {
    output$tbl <- renderTable({
      req(input$vars)
      mtcars %>% select(!!!input$vars)
    })
  }
)
```

It is possible to use quosures by using the `inject()` function from the upcoming version of rlang (https://github.com/r-lib/rlang/pull/591) to inject a quosure into the call:

```R
library(shiny)
library(dplyr)

shinyApp(
  fluidPage(
    checkboxGroupInput("vars", "Vars", names(mtcars)),
    tableOutput("tbl")
  ),
  function(input, output) {
    q <- quo({
      req(input$vars)
      mtcars[, input$vars, drop = FALSE]
    })
    
    output$tbl <- inject(renderTable(!!q))
  }
)
```


R/reactives.R
```diff
@@ -1838,7 +1817,8 @@ reactivePoll <- function(intervalMillis, session, checkFunc, valueFunc) {
     # firing and hold onto resources.
     if (re_finalized) {
       o$destroy()
-      rm(o, envir = parent.env(environment()))
+      # Note that rlang::as_function() inserts an additional parent environment.
+      rm(o, envir = parent.env(parent.env(environment())))
```
schloerke, line 1821: what is this removing `o` from? The `reactivePoll()` function environment?

Why not capture the environment value before 1814 and pass it in at 1821?

R/shinywrappers.R
```diff
@@ -583,15 +578,8 @@ createRenderPrintPromiseDomain <- function(width) {
...
-    deprecatedEnvQuotedMessage()
-    if (!quoted) expr <- enexpr(expr)
-    q <- new_quosure(expr, env)
-  } else {
-    q <- enquo(expr)
-  }
-
-  func <- quoToFunction(q, "renderText")
+  expr <- get_quosure(expr, env, quoted)
+  func <- quoToFunction(expr, "renderText")
```
schloerke, line 582: This cadence is repeated a bunch. Is this easy to wrap up into a helper method?

schloerke: (`expr` is never used afterwards either)

R/utils-lang.R
```diff
@@ -0,0 +1,148 @@
+
+# Return the ... arguments of the caller, as a list of quoted expressions.
+dot_exprs <- function() {
+  match.call(
+    definition  = sys.function(sys.parent()),
+    call        = sys.call(sys.parent()),
+    expand.dots = FALSE,
+    envir       = parent.frame(2L)
+  )$...
```
schloerke, line 9: This creeps me out. Changing to `[[`

```suggestion
  )[["..."]]
```

R/utils-lang.R
```diff
@@ -0,0 +1,148 @@
...
+    x
+  } else {
+    x
+  }
+}
+
+# Need this here until it is part of rlang.
+blast <- function(expr, env = parent.frame()) {
+  rlang::eval_bare(rlang::enexpr(expr), env)
+}
```
schloerke, line 88: If this is here, do we need the dev version of rlang?

tests/testthat/test-reactivity.r
```diff
@@ -659,6 +659,79 @@ test_that("suspended/resumed observers run at most once", {
...
+  flushReact()
+  expect_identical(val, 12)
+
+  # quosures can be used in reactive()
+  val <- NULL
+  a <- 1
+  f <- observe({ val <<- rlang::eval_tidy(rlang::quo(!!a + 10)) })
+  a <- 2
+  flushReact()
+  expect_identical(val, 12)
```
schloerke, line 709: Do we need any `!!!` tests?

tests/testthat/test-reactivity.r
```diff
@@ -684,10 +757,12 @@ test_that("reactive() accepts quoted and unquoted expressions", {
...
 
 
-  # Check that environment is correct - parent environment should be this one
+  # Check that environment is correct - parent of parent environment should be
+  # this one. Note that rlang::as_function() injects an intermediate
+  # environment.
   this_env <- environment()
   fun <- reactive(environment())
-  expect_identical(isolate(parent.env(fun())), this_env)
+  expect_identical(isolate(parent.env(parent.env(fun()))), this_env)
```
schloerke, line 765: Because of the extra layer, do you think others will also run into issues? Or is that _their problem_?

The `as_function()` idea seems small enough that we could implement it or even alter the final function environment after it is created.

schloerke (changes requested):

R/reactives.R
wch: It does remove `o` from the `reactivePoll()` function environment.

Storing the environment vs. calling `parent.env(parent.env(environment()))`... Seems like six of one, half a dozen of the other.

schloerke: Storing the environment variable is safe and would allow rlang to change their implementation.

Your proposed change above is no different than the previous code, but now we depend on an outside function to create the environments.

R/shinywrappers.R
wch: I wanted to make it clear that what you get out of `get_quosure()` is a quosure. The `quoToFunction` function is relatively straightforward, but `get_quosure()` is kind of magical, and I wanted to keep the magical part contained.

schloerke: Thanks!

R/utils-lang.R
wch: We need the dev version of rlang for other stuff to work properly (namely, `as_function()`).

Also, this isn't in the dev version of rlang yet but I'm told it will be very soon. https://github.com/r-lib/rlang/pull/591

tests/testthat/test-reactivity.r
wch: There is a remote, remote possibility that others could run into this.

I don't think we can change the way that `as_function()` works. I previously used `new_function()`, which doesn't insert another environment, but that can't handle cases where a quosure has another quosure in into it. For examples, see:
  https://gist.github.com/wch/812f7c6bb0f9c5b7d73194e3e5cbf758

R/reactives.R
wch: Fair enough, I've changed it to store the environment.

tests/testthat/test-reactivity.r
wch: I've added `!!!` tests.

schloerke (approved):

PR #3146, opened 2020-11-14
withCache slight doc fix: "should expensive" -> "should be slow to compute"

daattali:
Subject to your opinion, but IMO the average shiny user isn't familiar with the "expensive"="slow" terminology. Also changed "to see" to "to determine"

Edited in GitHub, so did not re-generate the docs


daattali:
Great new features by the way :)

wch:
Thanks - we've made these changes (among many others) in 493ef59d.

PR #3149, opened 2020-11-16
Make `bindCache()` work with `renderPlot()`

wch:
I previously thought that it would be really complicated to make `bindCache()` work with `renderPlot()` (and therefore replace the `renderCachedPlot` code), but it turns out to be not as bad as I thought. Most of the complexity in `renderCachedPlot` was due to caching, but that's already taken care of in `bindCache`.

Now, the complicated stuff in `bindCache.shiny.renderPlot` involves replacing the width/height with the width and height from `sizePolicy`.

Still need to:

- [ ] Figure out how to document the `sizePolicy` param, since it's only used by the `bindCache.shiny.renderPlot` S3 method (not in the `bindCache` generic).
- [ ] Replace our implementation of `enquos0` with one from rlang. @lionel- tells me it will be added soon.

Sample app:

```R
shinyApp(
  fluidPage(
    sidebarLayout(
      sidebarPanel(
        sliderInput("n", "Number of points", 4, 32, value = 8, step = 4)
      ),
      mainPanel(plotOutput("plot"))
    )
  ),
    output$plot <- renderPlot({
        message("Computing...")
        seqn <- seq_len(input$n)
      }) %>%
      bindCache(input$n)
  }
)
```


R/bind-event.R
```diff
@@ -244,8 +244,9 @@ bindEvent.shiny.render.function <- function(x, ..., ignoreNULL = TRUE, ignoreIni
     )
   }
 
-  class(res) <- c("shiny.render.function.event", class(res))
-  res
+  renderFunc <- addAttributes(renderFunc, renderFunctionAttributes(valueFunc))
+  class(renderFunc) <- c("shiny.render.function.event", class(valueFunc))
```
cpsievert, line 248: Why is this necessary?

R/shinywrappers.R
```diff
@@ -229,6 +229,22 @@ as.tags.shiny.render.function <- function(x, ..., inline = FALSE) {
   useRenderFunction(x, inline = inline)
 }
 
+# Get relevant attributes from a render function object.
+renderFunctionAttributes <- function(x) {
+  attrs <- c("outputFunc", "outputArgs", "hasExecuted", "cacheHint")
```
cpsievert, line 234: Does it matter that this appears to be missing `srcref`? Would it make sense to do something like:

```r
attrs <- names(attributes(renderText({})))
```

So we can avoid hard-coding the attributes?

R/bind-event.R
wch: It's so that we can throw an error if someone tries to call `bindCache` on it.

R/shinywrappers.R
wch: The purpose of this function is to pull out the attributes from a render function so that we can hoist them into a function that wraps the original render function. We don't want to carry along the `srcref` attribute - that would probably cause problems.

Maybe:

```R
attrs <- setdiff(names(attributes(x)), c("srcref", "wholeSrcref", "srcfile", "class"))
```

Although I do wonder if there are some other weird attributes that we might have missed here. Now that I think about it some more, I think we should leave the code as is, because the specified are the attributes we want to hoist up; we don't use others, so bringing them along could cause problems with no real benefits.

R/render-plot.R
```diff
@@ -173,7 +180,14 @@ renderPlot <- function(expr, width = 'auto', height = 'auto', res = 72, ...,
   outputFunc <- plotOutput
   if (!identical(height, 'auto')) formals(outputFunc)['height'] <- list(NULL)
 
-  markRenderFunction(outputFunc, renderFunc, outputArgs, cacheHint = FALSE)
+  markedFunc <- markRenderFunction(
+    outputFunc,
+    renderFunc,
+    outputArgs,
+    cacheHint = list(userExpr = get_expr(expr), res = res)
```
cpsievert, line 187: I think we should add `outputName`, `altWrapper()`, and `args` here? Also, could change `args <- list(...)` to `args <- rlang::list2(...)`?

wch: - We don't use `outputName` in any other cache hints, so I don't want to add it here. This allows there to be multiple plots in a page that share the same cache, which could actually useful. (I know we use `outputName` in `renderCachedPlot`, but I think it's better to not use it.)
- If we pass `altWrapper`, we'll need to make a change so that the `cacheHint` is dynamically evaluated (and reactive). Currently the `cacheHint` is pre-hashed when `bindCache` is called, instead of every time the the key is invalidated. I think it makes sense to do this change.
- The `args <- list(...)` from `renderCachedPlot` doesn't actually happen in `bindCache`. We might want to expose that as a parameter (instead of using `...`) because those are already used for the key expression.

wch: Actually, now that I look at it, making the cache hint dynamic (for `altWrapper`) will add in enough complexity that I'd like to do it in a separate PR.

Currently, `cacheHint` just treated as a list of regular objects when it is hashed. If it is to be dynamic, then we may need to inspect each element to see if it's a quosure (or function, maybe?) and if so, evaluate it.

wch:
There are a bunch of challenges with making `cacheHint` dynamic (so we can pass in `altWrapper`), and it would introduce a fair amount of additional complexity. The challenges are not insurmountable, but after a discussion with @schloerke, I don't think that we need to make it dynamic.

Some of the challenges:

- How should dynamic items in the `cacheHint` be specified? If we accept quosures, should they be unwrapped immediately, or should we require them to be `inject()`ed?
- Handling async reactive objects in `cacheHint` adds complexity.

The reason that I don't think they need to be dynamic can be illustrated with `altWrapper`. It makes sense for the user to be responsible for passing in `altWrapper` as the cache key expression to `bindCache` (instead of us automatically putting it in the `cacheHint`). Here's an example of where the `alt` is a reactive expression:

```R
alt <- reactive(paste("input$txt is ", input$txt))
output$plot <- renderPlot(
    plot(seq_len(input$y)),
    alt = alt
  )
```

In this case, the plot uses `input$y`, and the alt text uses `input$txt`. If `bindCache` is used on this, it makes sense for the use to use it like this:

```R
  bindCache(input$y, alt())
```

This would have the same effect in this case:

```R
  bindCache(input$y, input$txt)
```

The plot expression uses `input$y`, and we already expect users to tell `bindCache()` that it should use `input$y` as part of the key. Like the plot expression, `alt` also influences the output, and it's perfectly reasonable for users to have to inform `bindCache()` of that.

PR #3153, opened 2020-11-17
Have selectize-plugin-a11y redefine itself on destroy, closes #3125

### LINKED ISSUE #3125: 'selectize-plugin-a11y' is causing 'updateSelectInput()' to break when using a DT with `filter` option
Sample code to reproduce using latest dev version of shiny:

```
library(shiny)

ui <- fluidPage(
  selectInput("test", "test", choices = LETTERS[1:5]),
  actionButton("go", "go"),
  DT::DTOutput("table")
)

  observeEvent(input$go, {
    updateSelectInput(session, "test", choices = letters[6:10])
  })

  output$table <- DT::renderDT({
    DT::datatable(mtcars, filter = "top")
  })
}

shinyApp(ui, server)
```

When you click "go" to update the selectInput, it become a plain HTML selectbox (see screenshot below) and a JS error in the console comes up:

image

```
VM35:2 Uncaught Error: Unable to find "selectize-plugin-a11y" plugin
    at L.a.loadPlugin (<anonymous>:2:4581)
    at L.a.require (<anonymous>:2:4888)
    at L.a.initializePlugins (<anonymous>:2:4450)
    at new L (<anonymous>:2:10781)
    at HTMLSelectElement.<anonymous> (<anonymous>:3:5531)
    at Function.each (jquery.min.js:2)
    at S.fn.init.each (jquery.min.js:2)
    at S.fn.init.a.fn.selectize (<anonymous>:3:5249)
    at exports.InputBinding._selectize (input_binding_select.js:199)
    at exports.InputBinding.receiveMessage (input_binding_select.js:62)
```

Notes:

- The problem does not happen if the `filter` argument of DT is not used. It's possible that other javascript libraries would also break the new selectInput due to similar issues.
- The problem does not exist in the current CRAN version 1.5.0

cpsievert:
Closes #3125


inst/www/shared/selectize/accessibility/js/selectize-plugin-a11y.js
```diff
@@ -145,7 +145,14 @@ Selectize.define("selectize-plugin-a11y", function (options) {
     var original = self.destroy;
     return function () {
       self.accessibility.liveRegion.$region.remove();
-      return original.apply(this, arguments);
+      var destroyed = original.apply(this, arguments);
+      // Redefine plugin after destroying the instance so that selectizeInput()'s
+      // receiveMessage() can call destroy and update the instance without
+      // needing to reload this plugin
+      Selectize.define("selectize-plugin-a11y", selectize_accessibility_plugin);
```
wch, line 152: Is this line necessary? It looks like it's doing the same thing as 158.

cpsievert: Yea, this line is the main point of the PR -- it makes sure the plugin is defined again after `destroy()` is called here 

https://github.com/rstudio/shiny/blob/5b12980b7a7012b0bff51892d536b8b4fc69fa6a/srcjs/input_binding_select.js#L56

and here

https://github.com/rstudio/shiny/blob/5b12980b7a7012b0bff51892d536b8b4fc69fa6a/srcjs/input_binding_select.js#L200

wch: What's strange is that calling `.destroy()` on an instance causes the global `Selectize` object to be altered (and then requires us to modify the `Selectize` object to alter it again).

wch: When I add a breakpoint and inspect the `Selectize.plugins` object right after `Selectize.define("selectize-plugin-a11y", ...)` is called, I see that `selectize-plugin-a11y` is listed:

image


However, if I wait until the JS console is idle and then look at it again, then the plugin does not show up. So there appears to be a bug where `selectize-plugin-a11y` is getting removed from `Selectize.plugins`.

image

wch: I think the problem is that `selectize.min.js` is getting loaded twice. The first copy gets the accessibility plugin, and the second does not.

image

wch: With a breakpoint in `renderDependencies()` in `output_binding_html.js`, I can see that there's a `selectize-js: "0.12.4"` baked into the web page, but then there's also an dynamic "selectize", version: "0.12.0"`. So I think the issue is that there are two dependencies with different names, `selectize-js`, and `selectize`.


image

wch:
I'm pretty sure the cause of the problem is that DT is bringing in its own copy of selectize, and the `htmlDependency` has a different name, `"selectize"`, instead of Shiny's `"selectize-js"`.

https://github.com/rstudio/DT/blob/5aa48f10bed1db1bc3633e56a740128c27f8ef34/R/datatables.R#L615-L618

cpsievert:
Closing in favor of https://github.com/rstudio/shiny/pull/3155

PR #3156, opened 2020-11-19
Minor bindCache doc tweaks

hadley:
Some formatting fixes, and some improvements (IMO) to the text.

### DIFF
--- a/R/bind-cache.R
+++ b/R/bind-cache.R
@@ -12,7 +12,7 @@ utils::globalVariables(".GenericCallEnv", add = TRUE)
 #' `bindCache()` will cache all previous values (as long as they fit in the
 #' cache) and they can be shared across user sessions. This allows
 #' `bindCache()` to dramatically improve performance when used correctly.
-#'
+
 #' @details
 #'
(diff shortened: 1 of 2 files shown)


R/bind-cache.R
```diff
@@ -4,26 +4,19 @@ utils::globalVariables(".GenericCallEnv", add = TRUE)
...
 #'
-#' * [reactive()]` expressions.
-#' * `render*` functions, like [renderText()], [renderTable()], and so on.
+#' Ordinary [reactive()] expressions automatically cache their _most recent_
+#' value, which helps to  avoid redundant computation in downstream reactives.
+#' `bindCache()` will cache all previous values (as long as they fit in the
+#' cache) and they can be shared across user sessions. This allows
+#' `bindCache()` to dramatically improve performance when used correctly.
 #'
-
```
wch, line 12: These extra blank lines are because of RStudio messing up formatting when I have it reflow the text (https://github.com/rstudio/rstudio/issues/8378). If you tell RStudio reflow the text, it'll only do it for the current block (and won't mess up the other blocks).

hadley: Should be better now

PR #3163, opened 2020-11-23
Replace `renderCachedPlot()` code with `bindCache()`

wch:
This replaces the underlying implementation for `renderCachedPlot()`, with `bindCache(renderPlot())`.


R/render-cached-plot.R
```diff
@@ -216,256 +217,19 @@ renderCachedPlot <- function(expr,
...
-          img$height <- NULL
-
-          img
-        })
-      }
+  inject(
+    bindCache(
+      renderPlot(!!expr, res = res, alt = alt, outputArgs = outputArgs),
+      !!cacheKeyExpr,
+      sizePolicy = sizePolicy
```
schloerke, line 230: `cache` is not used anymore

schloerke: Neither is `width` or `height`

schloerke: `...` are also ignored

schloerke (changes requested):
Either deprecate or document new changes for `...`, `width`, `height`, and `cache`.

R/render-cached-plot.R
wch: Thanks, it now passes along `cache` and `...`. The `width` and `height` are meant to be ignored.

wch:
Regarding docs: there's (now) no change in the behavior of the args, so I don't think the documentation needs to be changed.

R/render-cached-plot.R
```diff
@@ -216,256 +217,20 @@ renderCachedPlot <- function(expr,
...
-  # This is so that the expr doesn't re-execute by itself; it needs to be
-  # triggered by the cache key (or width/height) changing.
-  isolatedFunc <- function() isolate(func())
-
-  args <- list(...)
-
-  if (!is.null(width) || !is.null(height)) {
-    warning("Unused argument(s) 'width' and/or 'height'. ",
-            "'sizePolicy' is used instead.")
-  }
```
schloerke, line 231: Restore this code, please

schloerke (approved):
Given height and width value check code being restored, LGTM

R/render-cached-plot.R
wch: Done

PR #3061, opened 2020-09-16
Remove test for date object length>1 (Closes #2936)

andrewbaxter439:
This does the alternative fix to the issue of `dateInput` warning about date objects length>1. @cpsievert @wch in case that helps!

PR #3165, opened 2020-11-23
Make sure dateInput() warns on empty string

### LINKED ISSUE #3162: dateInput() doesn't warn when `min=""`
Originally from https://github.com/rstudio/shinycoreci-apps/issues/85.

This should produce a warning, but it doesn't:

```R
dateInput("x", "x", min = "")
#> <div id="x" class="shiny-date-input form-group shiny-input-container">
#>   <label class="control-label" id="x-label" for="x">x</label>
#>   <input type="text" class="form-control" aria-labelledby="x-label" title="Date format: yyyy-mm-dd" data-date-language="en" data-date-week-start="0" data-date-format="yyyy-mm-dd" data-date-start-view="month" data-min-date data-date-autoclose="true" data-date-dates-disabled="null" data-date-days-of-week-disabled="null"/>
#> </div>

```

In the generated HTML `data-min-date` is there, but it doesn't have a value.

cpsievert:
Closes #3162

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -1612,6 +1612,8 @@ URLencode <- function(value, reserved = FALSE) {
 # so that strings like "2016-08-9" are expanded to "2016-08-09"
 dateYMD <- function(date = NULL, argName = "value") {
   if (!length(date)) return(NULL)
+  # as.Date("") yields NA, but we want it to produce an error (so we get the warning)
+  date[!nzchar(date)] <- " "
   tryCatch(date <- format(as.Date(date), "%Y-%m-%d"),
     error = function(e) {
       warning(
--- a/R/utils.R
+++ b/R/utils.R
@@ -1605,16 +1605,18 @@ URLencode <- function(value, reserved = FALSE) {
   if (reserved) encodeURIComponent(value) else encodeURI(value)
 }
 
-# Make user-supplied dates are either NULL or can be coerced
-# to a yyyy-mm-dd formatted string. If a date is specified, this
-# function returns a string for consistency across locales.
-# Also, `as.Date()` is used to coerce strings to date objects
-# so that strings like "2016-08-9" are expanded to "2016-08-09"
+# Make sure user-supplied dates are either NULL or can be coerced to a
+# yyyy-mm-dd formatted string. If a date is specified, this function returns a
+# string for consistency across locales. Also, `as.Date()` is used to coerce
+# strings to date objects so that strings like "2016-08-9" are expanded to
+# "2016-08-09". If any of the values result in error or NA, then the input
+# `date` is returned unchanged.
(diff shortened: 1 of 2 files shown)


R/utils.R
```diff
@@ -1612,6 +1612,8 @@ URLencode <- function(value, reserved = FALSE) {
 # so that strings like "2016-08-9" are expanded to "2016-08-09"
 dateYMD <- function(date = NULL, argName = "value") {
   if (!length(date)) return(NULL)
+  # as.Date("") yields NA, but we want it to produce an error (so we get the warning)
+  date[!nzchar(date)] <- " "
   tryCatch(date <- format(as.Date(date), "%Y-%m-%d"),
```
wch, line 1617: Instead of special-casing `""`, how about something like this:

```suggestion
  tryCatch({
      date <- format(as.Date(date), "%Y-%m-%d")
      if (is.na(date)) stop()
    },
```

That way, any value that would turn into `NA` will result in the warning. (For example, if input is `NA`, then this would catch it and give a warning.)

Note: I haven't tested this code in action.

wch: Another note: we need to make sure this works correctly for vector inputs (see #3061). My initial suggestion above didn't take that into account, but I've since fixed it.

wch:
This is the line from the test app that led to this issue and PR: 
https://github.com/rstudio/shinycoreci-apps/blob/a03c6c43ec6a72a3acd4658cec721bfea78179ca/apps/157-date-format/app.R#L52

But looking at the history of `dateYMD`, I don't see a time where it ever would have issued a warning for that input. Maybe I'm missing something?

cpsievert:
LGTM, thank you!

> But looking at the history of dateYMD, I don't see a time where it ever would have issued a warning for that input. Maybe I'm missing something?

Yea, I think you're right, I just checked-out https://github.com/rstudio/shiny/commit/67d3a504aefd5d2644a052381b75472267dfe9e7, and doesn't throw a warning for the `""` case 🤷‍♂️

wch:
Ah, it looks like it's due to a change in behavior of `as.Date("")` in R 4.0.3. From https://cran.r-project.org/doc/manuals/r-release/NEWS.html:

> `as.Date(<char>)` now also works with an initial `""`, thanks to Michael Chirico's [PR #17909](https://bugs.r-project.org/bugzilla3/show_bug.cgi?id=17909).

In R 4.0.3:

```R
as.Date("")
#> [1] NA
```

In prior versions of R:

```R
as.Date("")
#> Error in charToDate(x) : 
#>   character string is not in a standard unambiguous format
```


This is also why the tests are failing on R 3.6 and below. I've changed `dateYMD` so that if there are any errors or NA's from the input, then the original input is returned unchanged. This makes the behavior better specified, and more consistent across versions.

PR #3172, opened 2020-12-01
Replace %OR% with %||% from rlang

wch:
The internal `%OR%` function currently skips over the first argument if it is `NULL` or `NA`, but this behavior can be a bit confusing -- typically we only care about `NULL`. This PR replaces `%OR%` with `%||%` from rlang, which only respects `NULL`.

Also, it emits warnings if the first object is a function or environment, but that's not an uncommon use case. For example, we use it here with functions :
https://github.com/rstudio/shiny/blob/bde5a8829558c1f281647ac1c0c7c9d5b6ae2fcf/R/shiny.R#L619-L629

Here are examples of warnings:

```R
identity %OR% 5
#> Warning in is.na(x): is.na() applied to non-(list or vector) of type 'closure'
#> function (x) 
#> x
#> <bytecode: 0x7fd4ae840fc0>
#> <environment: namespace:base>
new.env() %OR% 5
#> Warning in is.na(x): is.na() applied to non-(list or vector) of type
#> 'environment'
#> <environment: 0x7fd44ec95408>
```

I did a search for usage of `%OR%` in the code base, and I don't think there are any cases where the first argument would be `NA`:

<details>
<summary>Search results</summary>

```
R/bind-cache.R:
  716        session <- getDefaultReactiveDomain()
  717  
  718:       width  <- session$clientData[[paste0('output_', outputName, '_width')]]  %OR% 0
  719:       height <- session$clientData[[paste0('output_', outputName, '_height')]] %OR% 0
  720  
  721        rect <- sizePolicy(c(width, height))
  ...
  768          req(FALSE)
  769        }
  770:       pixelratio <- session$clientData$pixelratio %OR% 1
  771  
  772        list(fitDims(), pixelratio)

R/bind-event.R:
  190    valueFunc <- wrapFunctionLabel(valueFunc, "eventReactiveValueFunc", ..stacktraceon = TRUE)
  191  
  192:   label <- label %OR%
  193      sprintf('bindEvent(%s, %s)', attr(x, "observable", exact = TRUE)$.label, quos_to_label(qs))
  194  
  ...
  263    # Note that because the observer will already have been logged by this point,
  264    # this updated label won't show up in the reactlog.
  265:   x$.label <- label %OR% sprintf('bindEvent(%s, %s)', x$.label, quos_to_label(qs))
  266  
  267    initialized <- FALSE

R/bootstrap-layout.R:
  441  
  442    children <- list(...)
  443:   childIdx <- !nzchar(names(children) %OR% character(length(children)))
  444    attribs <- children[!childIdx]
  445    children <- children[childIdx]
  ...
  524  
  525    children <- list(...)
  526:   childIdx <- !nzchar(names(children) %OR% character(length(children)))
  527    attribs <- children[!childIdx]
  528    children <- children[childIdx]

R/bootstrap.R:
  210      ),
  211      stylesheet = c(
  212:       theme %OR% "css/bootstrap.min.css",
  213        # Safely adding accessibility plugin for screen readers and keyboard users; no break for sighted aspects (see https://github.com/paypal/bootstrap-accessibility-plugin)
  214        "accessibility/css/bootstrap-accessibility.css"
  ...
  869          div <- markTabAsSelected(div)
  870        } else {
  871:         tabValue <- div$attribs$`data-value` %OR% div$attribs$title
  872          if (identical(selected, tabValue)) {
  873            foundSelected <<- TRUE

R/deprecated.R:
   12                              version = NULL) {
   13  
   14:   if (getOption("shiny.deprecation.messages") %OR% TRUE == FALSE)
   15      return(invisible())
   16  

R/graph.R:
   11    msg <- paste0(
   12      sQuote(package),
   13:     if (is.na(version %OR% NA)) "" else paste0("(>= ", version, ")"),
   14      " must be installed for this functionality."
   15    )

R/image-interact.R:
   94  
   95    # Try to extract vars from brush object
   96:   xvar      <- xvar      %OR% brush$mapping$x
   97:   yvar      <- yvar      %OR% brush$mapping$y
   98:   panelvar1 <- panelvar1 %OR% brush$mapping$panelvar1
   99:   panelvar2 <- panelvar2 %OR% brush$mapping$panelvar2
  100  
  101    # Filter out x and y values
  ...
  232  
  233    # Try to extract vars from coordinfo object
  234:   xvar      <- xvar      %OR% coordinfo$mapping$x
  235:   yvar      <- yvar      %OR% coordinfo$mapping$y
  236:   panelvar1 <- panelvar1 %OR% coordinfo$mapping$panelvar1
  237:   panelvar2 <- panelvar2 %OR% coordinfo$mapping$panelvar2
  238  
  239    if (is.null(xvar))

R/imageutils.R:
    5    # In the next shiny release, this option will likely be superseded in
    6    # favor of a fully customizable graphics device option
    7:   if ((getOption('shiny.useragg') %OR% FALSE) && is_available("ragg")) {
    8      pngfun <- ragg::agg_png
    9    } else if (capabilities("aqua")) {
   10      # i.e., png(type = 'quartz')
   11      pngfun <- grDevices::png
   12:   } else if ((getOption('shiny.usecairo') %OR% TRUE) && is_available("Cairo")) {
   13      pngfun <- Cairo::CairoPNG
   14    } else {

R/knitr.R:
   26  #' @rdname knitr_methods
   27  knit_print.shiny.appobj <- function(x, ...) {
   28:   opts <- x$options %OR% list()
   29    width <- if (is.null(opts$width)) "100%" else opts$width
   30    height <- if (is.null(opts$height)) "400" else opts$height

R/middleware.R:
  310        list(
  311          onHeaders = function(req) {
  312:           maxSize <- getOption('shiny.maxRequestSize') %OR% (5 * 1024 * 1024)
  313            if (maxSize <= 0)
  314              return(NULL)

R/reactives.R:
 2455      if (now >= v$when) {
 2456        # Mod by 999999999 to get predictable overflow behavior
 2457:       v$trigger <- isolate(v$trigger %OR% 0) %% 999999999 + 1
 2458        v$when <- NULL
 2459      } else {

R/render-plot.R:
  124          # If !execOnResize, don't invalidate when width/height changes.
  125          dims <- if (execOnResize) getDims() else isolate(getDims())
  126:         pixelratio <- session$clientData$pixelratio %OR% 1
  127          do.call("drawPlot", c(
  128            list(
  ...
  163        function(result) {
  164          dims <- getDims()
  165:         pixelratio <- session$clientData$pixelratio %OR% 1
  166          result <- do.call("resizeSavedPlot", c(
  167            list(name, shinysession, result, dims$width, dims$height, altWrapper(), pixelratio, res),
  ...
  627      # First look for log base in scale, then coord; otherwise NULL.
  628      list(
  629:       x = get_log_base(xscale$trans) %OR% coord$xlog %OR% NULL,
  630:       y = get_log_base(yscale$trans) %OR% coord$ylog %OR% NULL
  631      )
  632    }

R/shiny.R:
  488            # "json" unless requested otherwise. The only other valid value is
  489            # "rds".
  490:           format <- params$format %OR% "json"
  491  
  492            values <- list()
  ...
  619      getSnapshotPreprocessOutput = function(name) {
  620        fun <- attr(private$.outputs[[name]], "snapshotPreprocess", exact = TRUE)
  621:       fun %OR% identity
  622      },
  623  
  ...
  626      getSnapshotPreprocessInput = function(name) {
  627        fun <- private$.input$getMeta(name, "shiny.snapshot.preprocess")
  628:       fun %OR% identity
  629      },
  630  
  ...
 1412        }
 1413  
 1414:       tmp_info <- private$outputInfo[[name]] %OR% list(name = name)
 1415  
 1416        # cd_names() returns names of all items in clientData, without taking a
 ....
 1458        # https://github.com/rstudio/htmltools/issues/161
 1459        parse_css_colors <- function(x) {
 1460:         htmltools::parseCssColors(x %OR% "", mustWork = FALSE)
 1461        }
 1462  
 ....
 1920                    return(httpResponse(
 1921                      200,
 1922:                     download$contentType %OR% getContentType(filename),
 1923                      # owned=TRUE means tmpdata will be deleted after response completes
 1924                      list(file=tmpdata, owned=TRUE),

R/shinyapp.R:
  570    # knit_print.shiny.appobj, but I am trying to make the most conservative
  571    # change possible due to upcoming release.
  572:   opts <- x$options %OR% list()
  573    width <- if (is.null(opts$width)) "100%" else opts$width
  574    height <- if (is.null(opts$height)) "400" else opts$height

R/shinyui.R:
  143    allowed_methods <- "GET"
  144    if (is.function(ui)) {
  145:     allowed_methods <- attr(ui, "http_methods_supported", exact = TRUE) %OR% allowed_methods
  146    }
  147  

R/shinywrappers.R:
  432  
  433        # If contentType not specified, autodetect based on extension
  434:       contentType <- imageinfo$contentType %OR% getContentType(imageinfo$src)
  435  
  436        # Extra values are everything in imageinfo except 'src' and 'contentType'

R/utils.R:
  114  }
  115  
  116: `%OR%` <- function(x, y) {
  117    if (is.null(x))
  118      y

tests/testthat/test-test-server.R:
  567  get_mocked_publics <- function(instance, generator) {
  568    publics <- ls(instance, all.names = TRUE)
  569:   actives <- names(generator$active) %OR% character(0)
  570    # Active bindings are considered fields.
  571    methods_or_fields <- publics[!(publics %in% actives)]

```
</details>

This PR also removes the `%AND%` function, which isn't used anywhere, and is also kind of confusing. Instead of `x %AND% y`, it works just as well to do `if(x) y` (other than the `NA` handling, which I think is a bad idea anyway).


R/render-plot.R
```diff
@@ -626,8 +630,8 @@ find_panel_info_api <- function(b) {
 
     # First look for log base in scale, then coord; otherwise NULL.
     list(
-      x = get_log_base(xscale$trans) %OR% coord$xlog %OR% NULL,
-      y = get_log_base(yscale$trans) %OR% coord$ylog %OR% NULL
+      x = get_log_base(xscale$trans) %NA_OR% coord$xlog %NA_OR% NULL,
```
wch, line 633: `get_log_base()` (defined a few lines above) returns either a number or `NA`. Similarly `coord$xlog` is returned by `ggplot2::summarise_coord()`, and it does the same thing.

R/shiny.R
```diff
@@ -618,14 +618,14 @@ ShinySession <- R6Class(
...
-      fun %OR% identity
+      fun %||% identity
     },
 
     # Get the snapshotPreprocessInput function for an input name. If no preprocess
     # function has been set, return the identity function.
     getSnapshotPreprocessInput = function(name) {
       fun <- private$.input$getMeta(name, "shiny.snapshot.preprocess")
-      fun %OR% identity
+      fun %||% identity
```
cpsievert, line 628: So, IIUC, this currently throws a warning (with `%OR%`)?

R/image-interact.R
```diff
@@ -93,10 +93,10 @@ brushedPoints <- function(df, brush, xvar = NULL, yvar = NULL,
...
 
   # Try to extract vars from brush object
-  xvar      <- xvar      %OR% brush$mapping$x
-  yvar      <- yvar      %OR% brush$mapping$y
-  panelvar1 <- panelvar1 %OR% brush$mapping$panelvar1
-  panelvar2 <- panelvar2 %OR% brush$mapping$panelvar2
+  xvar      <- xvar      %||% brush$mapping$x
+  yvar      <- yvar      %||% brush$mapping$y
+  panelvar1 <- panelvar1 %||% brush$mapping$panelvar1
+  panelvar2 <- panelvar2 %||% brush$mapping$panelvar2
```
cpsievert, line 99: I wouldn't be that surprised if there are some user apps out there that have mistakenly passed `NA` and are relying it falling through to `brush$mapping`. That being said, I want to live in a `%||%` world, so I think it'd be worth the short-term pain

R/shiny.R
schloerke: Correct. (The `%OR%` version throws a warning)

R/shiny.R
```diff
@@ -33,7 +33,7 @@ NULL
 #'   as.promise
 #' @importFrom rlang quo enquo as_function get_expr get_env new_function enquos
 #'   eval_tidy expr pairlist2 new_quosure enexpr as_quosure is_quosure inject
-#'   enquos0 zap_srcref
+#'   enquos0 zap_srcref %||%
```
schloerke, line 36:
```suggestion
#'   enquos0 zap_srcref %||% is_na
```

R/render-plot.R
```diff
@@ -610,6 +610,10 @@ find_panel_info_api <- function(b) {
   coord  <- ggplot2::summarise_coord(b)
   layers <- ggplot2::summarise_layers(b)
 
+  `%NA_OR%` <- function(x, y) {
+    if (is.na(x)) y else x
```
cpsievert, line 614:
```suggestion
    if (isTRUE(is.na(x))) y else x
```

schloerke, line 614: ?rlang::is_na

> ‘is_na()’ is a scalar predicate and always returns a scalar
     boolean, ‘TRUE’ or ‘FALSE’. If its input is not scalar, it returns
     ‘FALSE’.

```suggestion
    if (is_na(x)) y else x
```

schloerke (approved):
Does this need a news entry due to the changed behavior?
LGTM Given `rlang::is_na()` change

R/shiny.R
```diff
@@ -1457,7 +1457,7 @@ ShinySession <- R6Class(
       # that string isn't a valid CSS color, will return NA)
       # https://github.com/rstudio/htmltools/issues/161
       parse_css_colors <- function(x) {
-        htmltools::parseCssColors(x %OR% "", mustWork = FALSE)
+        htmltools::parseCssColors(x %||% "", mustWork = FALSE)
```
cpsievert, line 1460: I can't remember if it's actually possible for `x` to have `NA`s, but we considered that possibility when writing this code (read the comment above). Should we just fix this in htmltools? https://github.com/rstudio/htmltools/issues/161

R/utils.R
```diff
@@ -113,20 +113,6 @@ isWholeNum <- function(x, tol = .Machine$double.eps^0.5) {
...
-`%AND%` <- function(x, y) {
-  if (!is.null(x) && !isTRUE(is.na(x)))
-    if (!is.null(y) && !isTRUE(is.na(y)))
-      return(y)
-  return(NULL)
-}
-
 `%.%` <- function(x, y) {
   paste(x, y, sep='')
 }
```
cpsievert, line 118: While we're at it, it'd be nice to get rid of this too (and use `paste0()` instead)?

R/image-interact.R
wch: Yeah, if anyone is doing that, I think they should fix it. I suppose we could warn in that case, though.

R/shiny.R
cpsievert: Here's a PR to make `parseCssColors()` better handle `NA` values  https://github.com/rstudio/htmltools/pull/194

R/render-plot.R
wch: FWIW, in practice, the `x` that gets passed to this function will only ever be a single-element vector. (Note the scoping for this function.) If it's not a single-element vector, it's probably good if we find out about it by having a warning occur.

cpsievert: Barret's suggestion is better

wch: Although... I suppose that given the name of the function, it does make sense to look specifically for a single NA using `is_na()`.

cpsievert (approved):
Assuming we take on that change to `parseCssColors()`

R/shiny.R
wch: Thanks, good catch!

R/utils.R
wch: Done in b0a1a06b.

PR #3173, opened 2020-12-02
Add cacheWriteHook and cacheReadHook

wch:
This adds support for `cacheWriteHook` and `cacheReadHook` for the objects passed to `bindCache()`. The purpose is to solve issues like https://github.com/ramnathv/htmlwidgets/issues/396.

The `cacheWriteHook` can modify values before the stored in the cache, and the `cacheReadHook` can modify values after they're retrieved from the cache, but before they're sent to the browser. These functions can also be used for side effects, as is done in https://github.com/ramnathv/htmlwidgets/pull/397, where it calls `createWebDependency()`.

Note that `bindCache.reactive` also adds support for these hooks, although there currently isn't a simple way to add the hooks to a `reactive()`. One would need to specifically add them as attributes to a reactive. I wonder if the `bindCache()` function itself should also allow users to pass in hooks like this -- currently, the author of a render function (not the end user) is the one who would provide these hooks.

This PR is made against the `wch-cache-read-hook` branch, because it uses changes from there, and because including that branch's changes would result in a lot of spurious changes shown in this PR. If/when we merge that PR (or if we don't), this branch can be retargeted against master.

### DIFF
--- a/R/bind-cache.R
+++ b/R/bind-cache.R
@@ -473,9 +473,6 @@ bindCache.reactiveExpr <- function(x, ..., cache = "app") {
   cacheHint <- digest(extractCacheHint(x), algo = "spookyhash")
   valueFunc <- wrapFunctionLabel(valueFunc, "cachedReactiveValueFunc", ..stacktraceon = TRUE)
 
-  cacheWriteHook <- attr(x, "cacheWriteHook", exact = TRUE) %||% identity
-  cacheReadHook  <- attr(x, "cacheReadHook",  exact = TRUE) %||% identity
-
   # Don't hold on to the reference for x, so that it can be GC'd
   rm(x)
   # Hacky workaround for issue with `%>%` preventing GC:
@@ -489,7 +486,7 @@ bindCache.reactiveExpr <- function(x, ..., cache = "app") {
     cache <- resolve_cache_object(cache, domain)
     hybrid_chain(
       keyFunc(),
-      generateCacheFun(valueFunc, cache, cacheHint, cacheReadHook, cacheWriteHook)
+      generateCacheFun(valueFunc, cache, cacheHint, cacheReadHook = identity, cacheWriteHook = identity)
     )
   })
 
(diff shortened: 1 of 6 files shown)


cpsievert:
>  I wonder if the bindCache() function itself should also allow users to pass in hooks like this

I'd be in favor of not doing this until we have a real need/motivation for it.

cpsievert (approved):

wch:
I removed `cacheReadHook` and `cacheWriteHook` support for `reactive()`, since there's not an interface for people to add them -- they have to do it by adding those attributes to the reactive object. If we do want add support for this in the future, we want to do it with `bindCache(readHook = f, writeHook = g)`, instead of having people add attributes manually.

For posterity, here's a set of tests that would work for that interface:

```R
test_that("bindCache() writeHook and readHook for reactives", {
  write_hook_n <- 0
  read_hook_n  <- 0
  n <- 0
  v <- reactiveVal(1)

  r <- reactive({ n <<- n+1; v() })
  r <- r %>% bindCache(v(),
    cache = cachem::cache_mem(),
    writeHook = function(value) {
      write_hook_n <<- write_hook_n + 1
      paste0(value, ",w")
    },
    readHook = function(value) {
      read_hook_n <<- read_hook_n + 1
      paste0(value, ",r")
    }
  )

  vals <<- character()
  o <- observe({
    vals <<- r()
  })

  # Writing to cache
  flushReact()
  expect_identical(vals, 1)
  expect_identical(write_hook_n, 1)
  expect_identical(read_hook_n, 0)

  v(2)
  flushReact()
  expect_identical(vals, 2)
  expect_identical(write_hook_n, 2)
  expect_identical(read_hook_n, 0)

  # Reading from cache
  v(1)
  flushReact()
  expect_identical(vals, "1,w,r")
  expect_identical(write_hook_n, 2)
  expect_identical(read_hook_n, 1)

  v(2)
  flushReact()
  expect_identical(vals, "2,w,r")
  expect_identical(write_hook_n, 2)
  expect_identical(read_hook_n, 2)
})
```

PR #3174, opened 2020-12-02
Add `devmode()` features

schloerke:
Added:
* `devmode(enable, verbose)`
* `in_devmode()`
* `with_devmode(devmode, code)`
* `devmode_inform(message, .frequency = "regularly", .frequency_id = message, ...)`
* `register_devmode_option(option, message, on)`
* `get_devmode_option(option, default = NULL, on = waiver(), message = waiver())`
* These devmode options:

```r
register_devmode_option(
  "shiny.autoreload",
  "Turning on shiny autoreload. To disable, call `options(shiny.autoreload = FALSE)`",
  TRUE
)

register_devmode_option(
  "shiny.minified",
  "Using full shiny javascript file. To use the minified version, call `options(shiny.minified = TRUE)`",
  FALSE
)

register_devmode_option(
  "shiny.fullstacktrace",
  "Turning on full stack trace. To disable, call `options(shiny.fullstacktrace = FALSE)",
  TRUE
)
```

Other updates:
* Removed `./R/htmltools.R` as it is covered in `./R/reexports.R`
* Consolidated all imports to `./R/shiny-package.R`
* Soft deprecate shinyUI and shinyServer
* Add lifecycle svgs to use in macros docs
* Provide shiny deprecated messages using `rlang::inform()`
* Use `lifecycle::deprecated()`
* Import `rlang::is_false()`
* Better error messaging for failing template tests


schloerke:
(current windows failure is unrelated to changes made in PR)

DESCRIPTION
```diff
@@ -95,7 +95,8 @@ Imports:
     glue (>= 1.3.2),
     bslib (>= 0.2.2.9002),
     cachem,
-    ellipsis
+    ellipsis,
+    lifecycle
```
wch, line 99: Require at least 0.2

R/deprecated.R
```diff
@@ -1,7 +1,55 @@
...
+}
+#' @describeIn soft-deprecated Passes along a message directly to [rlang::warn()] with `.frequency = "regular"`
+shinySoftDeprecatedMessage <- function(
+  ...
+) {
+  if (is_false(getOption("shiny.deprecation.messages"))) {
+    return(invisible())
+  }
+
+  rlang_warn(..., frequency = "regular")
```
wch, line 43: I think `"always"` makes more sense here, and other places we set the `frequency`.

R/deprecated.R
```diff
@@ -1,7 +1,55 @@
...
+#' @describeIn soft-deprecated Passes arguments to [lifecycle::deprecate_soft()]
+shinySoftDeprecated <- function(
+  when, what, with = NULL, details = NULL, id = NULL,
+  env = rlang::caller_env(1)
+) {
+  if (is_false(getOption("shiny.deprecation.messages"))) {
+    return(invisible())
+  }
+
+  lifecycle::deprecate_soft(when, what, with = with, details = details, id = id, env = env)
```
wch, line 33: Might need to change this. From the docs:

> deprecate_soft() warns only if the deprecated function is called from the global environment

wch: On second thought, I think we should continue to emit messages instead of warnings. So `rlang_inform`?

R/server.R
```diff
@@ -76,6 +76,15 @@ registerClient <- function(client) {
 #' @export
 #' @keywords internal
 shinyServer <- function(func) {
+  shinySoftDeprecated(
```
wch, line 79: I think we'll only ever want to display this message in dev mode.

R/shiny-package.R
```diff
@@ -0,0 +1,5 @@
+  ## usethis namespace: start
+  ## usethis namespace: end
+  #' @importFrom lifecycle deprecated
+  #' @importFrom rlang is_false
+  NULL
```
wch, line 5: Let's move in all the other import stuff from shiny.R

R/shinyui.R
```diff
@@ -132,6 +132,15 @@ shinyDependencyCSS <- function(theme) {
 #' @keywords internal
 #' @export
 shinyUI <- function(ui) {
+  shinySoftDeprecated(
```
wch, line 135: Same as `shinyServer`: only ever print message in dev mode.

R/deprecated.R
```diff
@@ -1,7 +1,55 @@
...
+  )
+}
+
 #' Print message for deprecated functions in Shiny
 #'
 #' To disable these messages, use `options(shiny.deprecation.messages=FALSE)`.
 #'
+#' @inheritParams lifecycle::deprecate_soft
+#' @keywords internal
+#' @describeIn soft-deprecated Passes arguments to [lifecycle::deprecate_soft()]
```
wch, line 24: Maybe use `shiny-deprecated` instead of `soft-deprecated`

wch:
For `dev_mode()`

  - Maybe there's a better name than `dev_mode()`? Also, maybe listen to an option like `options(shiny.devmode=TRUE)`, so that people can set it in their `.Rprofile` without causing shiny to load?
  - Display warnings about upcoming deprecations (like reactive with `env` and `quoted`)
  - Disable various forms of caching, in sass and bslib
  - Give hints about cache keys (not yet), and print out if missing one or more
  - Deprecate renderDataTable - use DT instead
  - `shinyUI()`, `shinyServer()` functions - can’t ever get rid of these, but we can message in dev mode
    - Call these functions superseded, in tidyverse terminology.
  - Automatically turn on `shiny.minified = FALSE`

schloerke:
`dev_mode()` TODO:

- [ ] Maybe there's a better name than `dev_mode()`? Also, maybe listen to an option like `options(shiny.devmode=TRUE)`, so that people can set it in their `.Rprofile` without causing shiny to load?
- [ ] Display warnings about upcoming deprecations (like reactive with `env` and `quoted`)
- [ ] Disable various forms of caching, in sass and bslib
- [ ] Give hints about cache keys (not yet), and print out if missing one or more
- [ ] Deprecate renderDataTable - use DT instead
- [ ] `shinyUI()`, `shinyServer()` functions - can’t ever get rid of these, but we can message in dev mode
  - [ ] Call these functions superseded, in tidyverse terminology.
- [ ] Automatically turn on `shiny.minified = FALSE`

R/dev-mode.R
```diff
@@ -0,0 +1,243 @@
...
+  on_msg,
+  on_value = NULL,
+  off_value = NULL
+) {
+  getOption(
+    option,
+    {
+      if (in_shiny_dev_mode()) {
+        shiny_dev_mode_msg(msg = on_msg)
+        enabled_value
```
wch, line 214: Should be `on_value` and `off_value`.

R/dev-mode.R
```diff
@@ -0,0 +1,243 @@
...
+
+#' @describeIn  shiny_dev_mode Helper method to retrieve an `options()` value. If the `option` is not set, `on_value` or `off_value` will be returned depending on the result of `in_shiny_dev_mode()`.  When Shiny Developer Mode is on, `on_msg` will be displayed using `shiny_dev_mode_msg()`
+#' @param option Option to look for in `options()`
+#' @param on_msg Message to display once every 8 hours if `option` is not set and `in_shiny_dev_mode()` returns `TRUE`.
+#' @param on_value Default value to use when `option` is not set and `in_shiny_dev_mode()` returns `TRUE`.
+#' @param off_value Default value to use when `option` is not set and `in_shiny_dev_mode()` returns `FALSE`.
+#' @export
+#' @examples
+#'
+#' # Used within `shiny::runApp(launch.browser)`
```
wch, line 196: Let's not set `shiny.launch.browser` because it doesn't seem like a dev mode related setting.

wch:
How about this API:

```R
register_devmode_option(
  "shiny.autoreload",
  on = TRUE,
  message = "Turning on shiny autoreload. To disable, call `options(shiny.autoreload = FALSE)`"
)

get_devmode_option <- function(name, default = NULL, on = waiver()) {
  # Do stuff here
}

get_devmode_option("shiny.autoreload")
get_devmode_option("shiny.autoreload", default = FALSE)
get_devmode_option("shiny.autoreload", default = FALSE, on = TRUE)
```

schloerke:
approach has been updated. Top comment has been cleaned up.

R/deprecated.R
```diff
@@ -1,44 +1,44 @@
...
+    msg <- paste0(msg, "\n", "Please use `", with, "` instead.")
+  }
+  if (!is.null(details)) {
+    msg <- paste0(msg, "\n", details)
   }
 
-  # Similar to .Deprecated(), but print a message instead of warning
-  message(msg)
+  # lifecycle::deprecate_soft(when, what, with = with, details = details, id = id, env = env)
+  rlang::inform(message = msg, .frequency = "always", .frequency_id = msg)
```
wch, line 27: It would be nice if the output were sent to `stderr()` (with `inform(.file = stderr())`, so that it comes out red in RStudio. This is what it looks like now:

image

Although maybe there's some reason that `inform()` defaults to stdout?

R/devmode.R
```diff
@@ -0,0 +1,251 @@
...
+    return()
+  }
+
+  rlang::inform(
+    message = paste0("shiny devmode - ", message),
+    .frequency = .frequency,
+    .frequency_id = .frequency_id,
+    ...
+  )
+}
```
wch, line 128: The formatting needs to be a bit different: no `\n` between the message and the "once every 8 hours" part, and there should be a `\n` after that.

image

R/devmode.R
```diff
@@ -0,0 +1,251 @@
...
+#'
+#' The methods explained in this documentation file act independently from the rest of Shiny but are included to provide blue prints for your own packages. If your package already has (or willing to take) a dependency on Shiny, we recommend using the exported methods for consistent behavior.
+#'
+#' If your package should not take a dependency on Shiny, we recommending re-implementing two functions:
+#'
+#' 1. `in_devmode()`: This function should return `TRUE` if `getOption("shiny.devmode")` is set. In addition, we strongly recommend that it also checks to make sure `testthat` is not testing.  This can be achieved using:
+#' ```r
+#' in_devmode <- function() {
+#'   isTRUE(getOption("shiny.devmode", FALSE)) &&
+#'     # !testthat::is_testing()
```
wch, line 46: Having this comment in the middle of the code is a bit confusing.

R/devmode.R
```diff
@@ -0,0 +1,251 @@
...
+    shiny.devmode = devmode,
+    shiny.devmode.verbose = verbose
+  )
+}
+
+
+#' @describeIn devmode Determines if Shiny is in Developer Mode. If the `getOption("shiny.devmode")` is set to `TRUE` and not in testing inside `testthat`, then Shiny Developer Mode is enabled.
+#' @section Avoiding direct dependency on shiny:
+#'
+#' The methods explained in this documentation file act independently from the rest of Shiny but are included to provide blue prints for your own packages. If your package already has (or willing to take) a dependency on Shiny, we recommend using the exported methods for consistent behavior.
```
wch, line 38: I think the first sentence should be the explanation of why `devmode()` exists. Something like, "developer mode enables a number of options to make a developer's life easier, like enabling non-minified JS and printing messages about deprecated functions and options."

We'll also need an explanation of how it searches for these values (e.g., first it looks here, then there, then ...), and why the `devmode()` function exists.

R/deprecated.R
schloerke: In our case, the deprecations are "bad". So being red is a good thing.

I believe the typical case it to replicate a `message()` call.

R/devmode.R
```diff
@@ -0,0 +1,251 @@
...
+    list(
+      shiny.devmode = devmode,
+      shiny.devmode.verbose = verbose
+    ),
+    code
+  )
+}
+
+
+#' @describeIn devmode If Shiny Developer Mode and verbosity are enabled, displays a message once every 8 hrs (by default)
```
wch, line 102: I actually think that the default of once per 8 hours is counterproductive for the use cases we have in mind. If people use dev mode, they want to be told about issues in their code.

Another reason that it's problematic is that it makes it difficult to reproduce the problematic message, which makes it hard track down the cause. For example, I was running into some issue for which `rlang::warn()` was emitting a warning once every 8 hours. When I was trying to pinpoint the code that caused it, I had to restart R every time I ran code that triggered the message, because it wouldn't be triggered again without a restart.

Imagine that someone uses (for example) `as.list.reactivevalues` in their code. Then when they run their code, they see the message once, but it doesn't appear when they re-run their code. Then they restart and manage to track it down and fix it. If there are multiple calls to that function, they will only be alerted to the problem once per restart, which makes it more difficult to fix.

schloerke: Deprecation messages are always displayed. 

The only messages that are cached for 8 hours are the messages from `get_devmode_option(key, default)`

wch: Ah, makes sense!

R/devmode.R
schloerke: https://github.com/r-lib/rlang/pull/1076

R/deprecated.R
schloerke: 13767b7

wch (approved):

R/devmode.R
schloerke: Closed in https://github.com/r-lib/rlang/commit/6b78803e7dff1b75d0b9d9587f20b0d5f37368fa

R/devmode.R
```diff
@@ -0,0 +1,292 @@
...
+) {
+  if (!is.null(message)) {
+    stopifnot(length(message) == 1 && is.character(message))
+  }
+  registered_devmode_options$set(
+    option,
+    list(on = on, message = message)
+  )
+}
+registered_devmode_options
```
wch, line 204: Delete

R/devmode.R
```diff
@@ -0,0 +1,292 @@
...
+#' get_devmode_option("shiny.minified", TRUE) # TRUE if Dev mode is off
+#' is_minified <- with_devmode(TRUE, {
+#'   get_devmode_option("shiny.minified", TRUE)
+#' })
+#' is_minified # FALSE
+#'
+get_devmode_option <- function(
+  option,
+  default = NULL,
+  on = waiver(),
```
wch, line 221: How about, don't provide a default value, and down below, use `if (missing(devmode_default))` ? In the future we may be able to revisit this with some kind of sentinel value.

R/devmode.R
```diff
@@ -0,0 +1,292 @@
...
+
+is_waiver <- function(x) {
+  inherits(x, "waiver")
+}
+waiver <- function() {
+  structure(
+    list(),
+    class = "waiver"
+  )
+}
```
wch, line 272: File an issue on rlang about having sentinel values.

R/devmode.R
```diff
@@ -0,0 +1,292 @@
...
+#' \enumerate{
+#' \item `in_devmode()`: This function should return `TRUE` if `getOption("shiny.devmode")` is set. In addition, we strongly recommend that it also checks to make sure `testthat` is not testing.  This can be achieved using:
+#' ```r
+#' in_devmode <- function() {
+#'   isTRUE(getOption("shiny.devmode", FALSE)) &&
+#'     !identical(Sys.getenv("TESTTHAT"), "true")
+#' }
+#' ```
+#'
+#' \item `get_devmode_option(option, default, on, message)`:
```
wch, line 45: How about:

```R
get_devmode_option(name, default, devmode_default, devmode_message)
```

R/devmode.R
```diff
@@ -0,0 +1,292 @@
...
+#' @export
+#' @examples
+#' # Used within `shiny::runApp(launch.browser)`
+#' get_devmode_option("shiny.minified", TRUE) # TRUE if Dev mode is off
+#' is_minified <- with_devmode(TRUE, {
+#'   get_devmode_option("shiny.minified", TRUE)
+#' })
+#' is_minified # FALSE
+#'
+get_devmode_option <- function(
```
wch, line 218: Mke it clear that `devmode_default` and `devmode_message` generally should not be used; they should be set with `register_devmode_option()` (but doing it this way requires a Shiny dependency).

R/devmode.R
```diff
@@ -29,39 +29,58 @@ devmode <- function(
 #' @describeIn devmode Determines if Shiny is in Developer Mode. If the `getOption("shiny.devmode")` is set to `TRUE` and not in testing inside `testthat`, then Shiny Developer Mode is enabled.
 #' @section Avoiding direct dependency on shiny:
 #'
-#' The methods explained in this documentation file act independently from the rest of Shiny but are included to provide blue prints for your own packages. If your package already has (or willing to take) a dependency on Shiny, we recommend using the exported methods for consistent behavior.
+#' The methods explained in this help file act independently from the rest of Shiny but are included to provide blue prints for your own packages. If your package already has (or is willing to take) a dependency on \pkg{shiny}, we recommend using the exported Shiny methods for consistent behavior.
```
wch, line 32: "Note that if you use the exported Shiny methods, calling them will cause the Shiny package to load. This may be undesirable if your code will be used in (for example) R Markdown documents without `runtime: shiny`."

R/devmode.R
```diff
@@ -29,39 +29,58 @@ devmode <- function(
...
 #'
-#' If your package can not take a dependency on Shiny, we recommending re-implementing two functions:
+#' If your package can _not_ take a dependency on Shiny, we recommending re-implementing these two functions:
```
wch, line 34: Wrap long lines.

R/devmode.R
```diff
@@ -29,39 +29,58 @@ devmode <- function(
...
 #' }
 #' ```
 #'
-#' \item `get_devmode_option(option, default, on, message)`:
-#' This function should return the global `option` when it is set, similar to `getOption(option, default)`. When the global option `option` is not set, this function should return the default value (`default`) when `in_devmode()` is `FALSE` and the default Dev Mode value (`on`) when `in_devmode()` is `TRUE`. We strongly recommend displaying a message (`message`) to the developer once every 8 hours using `rlang::inform(message, .frequency = "regularly", .frequency_id = message, .file = stderr())` if returning the `on` default Dev Mode option value. This will keep the author up to date as to which behaviors are being altered. To allow developers a chance to disable Dev Mode messages, the message should be skipped if `getOption("shiny.devmode.verbose", TRUE)` returns `FALSE`.
+#' \item `get_devmode_option(name, default, devmode_default, devmode_message)`:
+#'
+#' This function is similar to `getOption(name, default)`, but when the option is not set the default value changes depending on the Dev Mode.  `get_devmode_option()` should be implemented as follows:
```
wch, line 49:
```suggestion
#' This function is similar to `getOption(name, default)`, but when the option is not set, the default value changes depending on the Dev Mode.  `get_devmode_option()` should be implemented as follows:
```

R/devmode.R
```diff
@@ -29,39 +29,58 @@ devmode <- function(
...
+#' * If not in Dev Mode, return `getOption(name, default)`
+#' * Get the global option `getOption(name)` value
+#' * If option value is set, return the value
+#' * Notify the developer that the default value is different
+#' * Return the Dev Mode default value
```
wch, line 54:
```suggestion
#' * If not in Dev Mode:
#'   * Return `getOption(name, default)`.
#' * If in Dev Mode:
#'   * Get the global option `getOption(name)` value.
#'   * If the global option value is set: 
#'     * Return the value.
#'   * If the global option value is not set:
#'     * Notify the developer that the Dev Mode default value will be used.
#'     * Return the Dev Mode default value.
```

R/devmode.R
```diff
@@ -170,42 +193,47 @@ registered_devmode_options <- Map$new()
...
+    list(devmode_default = devmode_default, devmode_message = devmode_message)
   )
 }
-registered_devmode_options
 
-#' @describeIn devmode Provides a consistent way to change default `getOption()` behavior when Developer Mode is enabled. This method is very similar to [getOption()] where the globally set option takes precedence. However, when the global option is not set, `default` will be returned when `in_devmode()` is `FALSE`, or `on` will be returned if `in_devmode()` is `TRUE`.
-#'
+
+#' @describeIn devmode Provides a consistent way to change the expected [getOption()] behavior when Developer Mode is enabled. This method is very similar to [getOption()] where the globally set option takes precedence. `get_devmode_option()` is implemented as follows:
+#' * Get the global option `getOption(name)` value
```
wch, line 230: Same as above

wch (approved):

R/devmode.R
schloerke: https://github.com/r-lib/rlang/issues/1079

R/deprecated.R
```diff
@@ -1,44 +1,44 @@
...
 #'
-#' @param new Name of replacement function.
-#' @param msg Message to print. If used, this will override the default message.
-#' @param old Name of deprecated function.
-#' @param version The last version of Shiny before the item was deprecated.
+#' @param version Shiny version when the function was deprecated
+#' @param what Function with possible arguments
+#' @param with Possible function with arguments that should be used instead
+#' @param details Additional information to be added after a new line to the displayed message
 #' @keywords internal
```
cpsievert, line 10: I suppose we should mention the breaking changes in NEWS (and, more importantly, add NEWS items?)

R/devmode.R
```diff
@@ -0,0 +1,354 @@
+#' Shiny Developer Mode
+#'
+#' @description \lifecycle{experimental}
+#'
+#'   Developer mode enables a number of options to make a developer's life
+#'   easier, like enabling non-minified JS and printing messages about
+#'   deprecated functions and options.
+#'
+#'   Please see the function descriptions for more details.
```
cpsievert, line 9: What function descriptions?

R/deprecated.R
schloerke: `shinyDeprecated` was never exported. It just had an Rd file.

R/devmode.R
```diff
@@ -0,0 +1,354 @@
...
+#'
+#' @keywords internal
+#' @describeIn devmode Function to set two options to enable/disable Shiny
+#'   Developer Mode and Developer messages
+#' @param devmode Logical value which should be set to `TRUE` to enable Shiny
+#'   Developer Mode
+#' @param verbose Logical value which should be set to `TRUE` display Shiny
+#'   Developer messages
+#' @export
```
cpsievert, line 18: Do you want to link this to the other functions (i.e., `@seealso in_devmode`)?

R/deprecated.R
schloerke: Added in 4091b39

R/devmode.R
schloerke: (... this whole file produces a single help file)

cpsievert: Ah, ok

R/deprecated.R
cpsievert: Thanks!

cpsievert (approved):

R/deprecated.R
```diff
@@ -59,7 +59,7 @@ diskCache <- function(
   exec_missing = FALSE,
   logfile = NULL)
 {
-  shinyDeprecated("cachem::cache_disk", version = "1.5.1")
+  shinyDeprecated("1.5.1", "diskCache()", "cachem::cache_disk()")
```
wch, line 62:
```suggestion
  shinyDeprecated("1.6.0", "diskCache()", "cachem::cache_disk()")
```

Also, I just realized that it should include a check  for `exec_missing` -- I don't think anyone used that argument, but it doesn't do anything now. It should be a warning, even outside of dev mode. I'm not sure the best way to do this. Maybe something like this?

```R
if (!missing(exec_missing)) {
  warning( ... )
}
```

R/deprecated.R
```diff
@@ -89,7 +89,7 @@ memoryCache <- function(
   exec_missing = FALSE,
   logfile = NULL)
 {
-  shinyDeprecated("cachem::cache_mem", version = "1.5.1")
+  shinyDeprecated("1.5.1", "diskCache()", "cachem::cache_mem()")
```
wch, line 92:
```suggestion
  shinyDeprecated("1.6.0", "diskCache()", "cachem::cache_mem()")
```

This also should have a check for `exec_missing` like above.

schloerke:
macOS failures are due to https://github.com/rstudio/webdriver/issues/59 (existing before PR). Merging

PR #3176, opened 2020-12-03
Use `fastmap::faststack()` and remove `Stack` class

wch:
This PR replaces the internal `Stack` class with `fastmap::faststack()`, and uses `fastqueue()` in places where it makes sense (we previously did direct list manipulation).


R/bookmark-state.R
```diff
@@ -448,7 +445,7 @@ RestoreInputSet <- R6Class("RestoreInputSet",
 )
 
 
-restoreCtxStack <- Stack$new()
+restoreCtxStack <- fastmap::faststack()
```
schloerke, line 448: Would this possibly have issues in binaries if `fastmap` udpated it's c bindings?  

Before, the R6 object was _safe_.  Now, with info being offload to C in `fastmap`, is this a concern?

We could have it be a placeholder and update the value in `.onLoad()`

Ex:

```r
restoreCtxStack <- NULL

.onLoad <- function() {
  restoreCtxStack <- fastmap::faststack()
}
```

schloerke (approved):
LGTM pending question

R/bookmark-state.R
wch: Good catch -- `faststack()` doesn't call any C code and the object returned should be completely self-contained, but you're right that it's safer to create in `.onLoad`.

PR #3182, opened 2020-12-03
Create session$currentThemeDependency only when needed

wch:
This fixes https://github.com/rstudio/shinycoreci-apps/issues/89, where a new `session$currentThemeDependency` `reactiveVal` gets created every time a session starts, but they never get cleaned up.

Test app:

```R
library(shiny)
library(bslib)
library(reactlog)
reactlog_enable()

light <- bs_theme()
dark <- bs_theme(bg = "black", fg = "white", primary = "purple")
ui <- fluidPage(
  theme = light,
  checkboxInput("dark_mode", "Dark mode", FALSE),
  verbatimTextOutput("txt"),
  reactlog_module_ui()
)
  observe({
    session$setCurrentTheme(
      if (isTRUE(input$dark_mode)) dark else light
    )
  })
  output$txt <- renderPrint({
    vars <- bs_get_variables(session$getCurrentTheme(), c("bg", "fg"))
    str(as.list(vars))
  })
  reactlog_module_server()
}
shinyApp(ui, server)
```

Previously the reactive graph for this app looked like this after reloading three times:

image

After the change, it looks like this:

image

### DIFF
--- a/R/shiny.R
+++ b/R/shiny.R
@@ -628,16 +628,6 @@ ShinySession <- R6Class(
       fun %||% identity
     },
 
-    # Used in getCurrentTheme()/setCurrentTheme() to ensure their reactive 
-    # dependency is initialized. Note that we don't initialize this reactiveVal()
-    # inside the initialize() method because at that point there isn't a
-    # reactive domain yet, and so it can persist after the app has exited.
-    ensureCurrentThemeDependency = function() {
-      if (is.null(private$currentThemeDependency)) {
-        private$currentThemeDependency <- reactiveVal(0)
-      }
-    },
-
     # See cycleStartAction
     startCycle = function() {
       # TODO: This should check for busyCount == 0L, and remove the checks from
@@ -737,6 +727,13 @@ ShinySession <- R6Class(
 
       private$testMode <- getShinyOption("testmode", default = FALSE)
       private$enableTestSnapshot()
+
+      # This `withReactiveDomain` is used only to satisfy the reactlog, so that
+      # it knows to scope this reactiveVal to this session.
+      # https://github.com/rstudio/shiny/pull/3182
+      withReactiveDomain(self,
+        private$currentThemeDependency <- reactiveVal(0)
+      )
 
       private$registerSessionEndCallbacks()
 
@@ -1313,8 +1310,6 @@ ShinySession <- R6Class(
     },
 
     getCurrentTheme = function() {
-      private$ensureCurrentThemeDependency()
-
       private$currentThemeDependency()
       getCurrentTheme()
     },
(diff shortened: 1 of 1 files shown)


R/shiny.R
```diff
@@ -628,6 +628,15 @@ ShinySession <- R6Class(
...
     },
 
+    # Ensure that `private$currentThemeDependency` is initialized. We don't do
+    # this in the `initialize()` method because then the reactiveVal doesn't
+    # have a reactive domain, and can persist after the app has exited.
+    ensureCurrentThemeDependency = function() {
+      if (is.null(private$currentThemeDependency)) {
+        private$currentThemeDependency <- reactiveVal(0)
+      }
+    },
```
cpsievert, line 638: For readability purposes, I think it'd make more sense to put this definition between `getCurrentTheme` and `setCurrentTheme`

R/shiny.R
```diff
@@ -628,6 +628,15 @@ ShinySession <- R6Class(
       fun %||% identity
     },
 
+    # Ensure that `private$currentThemeDependency` is initialized. We don't do
+    # this in the `initialize()` method because then the reactiveVal doesn't
+    # have a reactive domain, and can persist after the app has exited.
```
cpsievert, line 633:
```suggestion
    # Used in getCurrentTheme()/setCurrentTheme() to ensure their reactive 
    # dependency is initialized. Note that we don't initialize this reactiveVal()
    # inside the initialize() method because at that point there isn't a
    # reactive domain yet, and so it can persist after the app has exited.
```

R/shiny.R
```diff
@@ -1305,6 +1312,8 @@ ShinySession <- R6Class(
     },
 
     getCurrentTheme = function() {
+      private$ensureCurrentThemeDependency()
+
       private$currentThemeDependency()
       getShinyOption("bootstrapTheme")
```
cpsievert, line 1318: ?

```suggestion
      shiny::getCurrentTheme()
```

R/shiny.R
```diff
@@ -1314,6 +1323,8 @@ ShinySession <- R6Class(
       # bootstrapTheme, (2) re-executes any registered theme dependencies, and
       # (3) sends the resulting dependencies to the client.
 
+      private$ensureCurrentThemeDependency()
+
       # Note that this will automatically scope to the session.
       shinyOptions(bootstrapTheme = theme)
```
cpsievert, line 1329:
```suggestion
      # Note that this will automatically scope to the session.
      setCurrentTheme()
      
      private$ensureCurrentThemeDependency()
```

cpsievert: I would also remove the whitespace below so the `currentThemeDependency` logic is near each other

cpsievert (approved):
Looks good besides the minor nitpicks

wch:
In light of #3080, I don't think we should add the `ensureCurrentThemeDependency()` method, since its main purpose is to hide something from the reactlog, but it adds complexity. Creating the `reactiveVal` itself is not an expensive operation.

PR #3184, opened 2020-12-04
Fix logic for `ShinySession$startCycle`

wch:
The previous code could result in a call stack that keeps increasing in depth as it polls.


R/shiny.R
```diff
@@ -632,7 +632,7 @@ ShinySession <- R6Class(
     startCycle = function() {
       # TODO: This should check for busyCount == 0L, and remove the checks from
       # the call sites
```
jcheng5, line 634:
```suggestion
```

PR #3194, opened 2020-12-07
Add a warning message when value < min | value > max in sliderInput

### LINKED ISSUE #1552: sliderInput doesn't warn on invalid value
The docs for the sliderInput says about `value`: "A warning will be issued if the value doesn't fit between min and max." I don't get a warning when running an app with an out-of-range `value`.

The value properly gets rounded up/down to a valid setting, but without a warning.

### LINKED ISSUE #2909: sliderInput: NULL value should be disallowed and throw an error, or it should be documented that it's allowed
The documentation for sliderInput's value parameter is:

>The initial value of the slider. A numeric vector of length one will create a regular slider; a numeric vector of length two will create a double-ended range slider. A warning will be issued if the value doesn't fit between min and max.

The docs says that either a 1-element or 2-element numeric vector is the expected input. But `NULL` also seems to work without defining its beahviour. `NULL` should either be diallowed and throw an error, or the documentation should say it's allowed.

### LINKED ISSUE #2910: sliderInput: documentation for `value` is wrong - no warning is issued when value is outside of [min, max]
The docs for `value` say:

> The initial value of the slider. A numeric vector of length one will create a regular slider; a numeric vector of length two will create a double-ended range slider. A warning will be issued if the value doesn't fit between min and max.

However, setting a value outside of min and max does not produce any messages anywhere.

wch:
This supersedes #3105, since I'm not able to push to the branch for that PR. Closes #2910, closes #2909, closes #1552.

I removed the check for `updateSliderInput` because there's no guarantee that any of `min`, `max`, and `value` are present, and the benefit of checking there is smaller.

### DIFF
--- a/R/input-slider.R
+++ b/R/input-slider.R
@@ -89,14 +89,7 @@ sliderInput <- function(inputId, label, min, max, value, step = NULL,
                     version = "0.10.2.2")
   }
 
-  if (isTRUE(min(value) < min | max(value) > max)) {
-    warning(noBreaks. = TRUE,
-      sprintf(
-        "Trying to set a `value` outside of the [`min`, `max`] range (`value` = %s, `min` = %s, `max = %s). `value` will be set to `max`.",
-        paste(value, collapse = ","), min, max
-      )
-    )
-  }
+  assert_slider_value(value = value, min = min, max = max, fun = "sliderInput")
 
   dataType <- getSliderType(min, max, value)
 
@@ -302,6 +295,18 @@ findStepSize <- function(min, max, step) {
 
   } else {
     1
+  }
+}
+
+# Throw a warning if ever `value` is not in the [`min`, `max`] range
+assert_slider_value <- function(value, min, max, fun){
+  if (isTRUE(min(value) < min || max(value) > max)) {
+    warning(noBreaks. = TRUE, call. = FALSE,
+            sprintf(
+              "In %s(): Trying to set a `value` outside of the [`min`, `max`] range (`value` = %s, `min` = %s, `max = %s). `value` will be set to `max`.",
+              fun, paste(value, collapse = ","), min, max
+            )
+    )
   }
 }
 
--- a/R/input-slider.R
+++ b/R/input-slider.R
@@ -89,7 +89,7 @@ sliderInput <- function(inputId, label, min, max, value, step = NULL,
                     version = "0.10.2.2")
   }
 
-  assert_slider_value(value = value, min = min, max = max, fun = "sliderInput")
+  check_slider_value(min, max, value, "sliderInput")
 
   dataType <- getSliderType(min, max, value)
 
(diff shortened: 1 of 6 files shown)


R/input-slider.R
```diff
@@ -296,6 +298,37 @@ findStepSize <- function(min, max, step) {
...
+  if (is.null(min)   || is_na(min) ||
+      is.null(max)   || is_na(max) ||
+      is.null(value) || is_na(value))
+  {
+    stop(call. = FALSE,
+      sprintf("In %s(): `min`, `max`, and `value` cannot be NULL or NA.", fun)
+    )
+  }
+
+  if (!isTRUE(min(value) >= min)) {
```
schloerke, line 312: `value` can be a vector. If of size = 2, then it would always fail.

```r
❯❯ isTRUE(c(T, T))
[1] FALSE
```

```suggestion
  if (any(min(value) < min)) {
```

R/input-slider.R
```diff
@@ -296,6 +298,37 @@ findStepSize <- function(min, max, step) {
...
+    warning(call. = FALSE,
+      sprintf(
+        "In %s(): `value` should be greater than or equal to `min` (value = %s, min = %s).",
+        fun, paste(value, collapse = ", "), min
+      )
+    )
+  }
+
+  if (!isTRUE(max(value) <= max)) {
```
schloerke, line 321:
```suggestion
  if (any(max(value) > max)) {
```

(Do not need to be worried about NA values or length 0 values at this point)

R/input-slider.R
```diff
@@ -296,6 +298,37 @@ findStepSize <- function(min, max, step) {
...
+# Throw a warning if ever `value` is not in the [`min`, `max`] range
+validate_slider_value <- function(min, max, value, fun) {
+  if (is.null(min)   || is_na(min) ||
+      is.null(max)   || is_na(max) ||
+      is.null(value) || is_na(value))
+  {
+    stop(call. = FALSE,
+      sprintf("In %s(): `min`, `max`, and `value` cannot be NULL or NA.", fun)
+    )
+  }
```
schloerke, line 310: If we're inspecting this.... 

Won't catch `numeric(0)`

```suggestion
  if (
    length(min) == 0 ||
    length(max) == 0) ||
    length(value) == 0 ||
    is.null(min)   || is_na(min) ||
    is.null(max)   || is_na(max) ||
    is.null(value) || is_na(value))
  {
    stop(call. = FALSE,
      sprintf("In %s(): `min`, `max`, and `value` cannot be NULL, NA, or empty.", fun)
    )
  }
```

tests/testthat/test-input-slider.R
```diff
@@ -0,0 +1,79 @@
...
+  expect_warning(sliderInput('s', 's', x-1,  x+1,  x-2))
+  expect_warning(sliderInput('s', 's', x-1,  x+1,  c(x-2, x)))
+  expect_warning(sliderInput('s', 's', x-1,  x+1,  c(x, x+2)))
+
+  expect_error(sliderInput('s', 's', x-1,  x+1))
+  expect_error(sliderInput('s', 's', x-1,  x+1,  NULL))
+  expect_error(sliderInput('s', 's', x-1,  NULL, x))
+  expect_error(sliderInput('s', 's', NULL, x+1,  x))
+  expect_error(sliderInput('s', 's', NULL, NULL, x))
+})
```
schloerke, line 79: Let's test for length of inputs.

```suggestion

  # Size
  ## length 0
  expect_error(sliderInput('s', 's', x-1, x+1, numeric(0)))
  expect_error(sliderInput('s', 's', x-1, numeric(0), x))
  expect_error(sliderInput('s', 's', numeric(0), x+1, x))
  ## length 1
  expect_silent(sliderInput('s', 's', x-1, x+1, x))
  ## length 2
  expect_silent(sliderInput('s', 's', x-1, x+1, c(x, x)))
  ## length 3+
  expect_error(sliderInput('s', 's', x-1, x+1, c(x, x, x)))
  expect_error(sliderInput('s', 's', x-1, c(x, x, x), x))
  expect_error(sliderInput('s', 's', c(x, x, x), x+1, x))
})
```

Should we also test for an error if min/max are a vector of values?

schloerke (changes requested):

R/input-slider.R
wch: `value` is passed to `min()`, so it returns just a single value.

wch:
I've incorporated (and modified) some of the changes, and an additional unit test for `value` vectors that have NA.

PR #3198, opened 2020-12-10
Resend CSS styles when relevant element(s) mutate.

### LINKED ISSUE #3196: Thematic styling can inherit error colors
Test app:

```R
library(thematic)
library(shiny)
library(ggplot2)

ui <- fluidPage(
  plotOutput("ggplot"),
  checkboxInput("err", "Throw an error", value = TRUE)
)
  output$ggplot <- renderPlot({
    if (input$err) stop("Error here")

    dat <- data.frame(x = rnorm(50), y = rnorm(50))
    ggplot(dat, aes(x, y)) +
      geom_point()
  })
}

thematic_shiny()
shinyApp(ui, server)
```

If you uncheck "Throw an error", the plot will appear normally. However, if you re-check it, resize the window, then uncheck it, then the plot will be red.

image

Possible solutions

* If there's an error class on the div, don't recompute the style and send to the server.
* Put error messages in a child div, and put the error class on that child div.

cpsievert:
Closes #3196


wch:
We need to make sure this works for dynamic UI.

Also, what happens if someone resizes the window when there's an error message?

cpsievert:
@wch 7d03d6c adds the mutation observer during bind which partially solves the issue for an example like this:

```r
library(thematic)
library(shiny)
library(ggplot2)

ui <- fluidPage(
  uiOutput("plots"),
  checkboxInput("err", "Throw an error", value = TRUE)
)
  output$plots <- renderUI({
    tagList(
      plotOutput("gg1"),
      plotOutput("gg2")
    )
  })
  output$gg1 <- renderPlot({
    if (input$err) stop("Error here")
    dat <- data.frame(x = rnorm(50), y = rnorm(50))
    ggplot(dat, aes(x, y)) +
      geom_point()
  })
  output$gg2 <- renderPlot({
    if (input$err) stop("Error here")
    dat <- data.frame(x = rnorm(50), y = rnorm(50))
    ggplot(dat, aes(x, y)) +
      geom_point()
  })
}

thematic_shiny()
shinyApp(ui, server)
```


I also decided to not send info during error (https://github.com/rstudio/shiny/pull/3198/files#diff-6670b74d30d60045efc8276e9b95b59407209477d00b8d5c40c7f0ec600c20c3R375) since, if we do send on error, you'd get a "flicker" of red before getting the expected result

srcjs/init_shiny.js
```diff
@@ -332,13 +338,48 @@ function initShiny() {
...
+    if (cl.contains("shiny-theme-observer")) return;
+
+    var reportTheme = cl.contains('shiny-image-output') || cl.contains('shiny-plot-output') || cl.contains('shiny-report-theme');
+    if (!reportTheme) return;
+
+    var observerCallback = new Debouncer(null, () => doSendTheme(el), 100);
+    var observer = new MutationObserver(() => observerCallback.normalCall());
+    var config = {attributes: true, attributeFilter: ['style', 'class']};
+    observer.observe(el, config);
+    cl.add("shiny-theme-observer"); // TODO: remove this class on unbind?
```
wch, line 370: If I'm understanding correctly, the `"shiny-theme-observer"` class exists only to indicate that there's a mutation observer. Instead of this class, I think it might make more sense to add a `.data()` value to the jQuery object.

wch:
I think there are still two issues with this potential fix with respect to the real-world problem that I'm hoping to fix:

1. This will cause flashing. The plot will be drawn twice when the div transitions from an error state to a non-error state; the first plot will be red, then the next will be normal.
1. This won't fix the case of when `bindCache` is used on a plot, since the invalidated CSS properties won't cause a redraw. If we put those properties into the cache key, we should be careful to take a dependency on them only when they're actually used. I'm not sure offhand how to do that, though.

For my particular use case, if we fix (1), then (2) won't be a problem. To do that, I think the most straightforward thing is to not recompute the CSS properties when we have a known Shiny error class on the wrapper div.

cpsievert:
@wch I filed a new issue about the caching issue here https://github.com/rstudio/shiny/issues/3208. It seems serious enough that I think we should consider a fix in the 1.6, but handle it in a separate PR.

wch (approved):

srcjs/init_shiny.js
```diff
@@ -332,13 +338,49 @@ function initShiny() {
...
+  function maybeAddThemeObserver(el) {
+    var $el = $(el);
+    if ($el.data("shiny-theme-observer")) return;
+
+    var cl = el.classList;
+    var reportTheme = cl.contains('shiny-image-output') || cl.contains('shiny-plot-output') || cl.contains('shiny-report-theme');
+    if (!reportTheme) return;
+
+    var observerCallback = new Debouncer(null, () => doSendTheme(el), 100);
+    var observer = new MutationObserver(() => observerCallback.normalCall());
```
wch, line 368: I just realized this will completely break Shiny apps on IE <= 10. There are things in Shiny that currently don't work on those older versions, but this would make apps not work at all. So maybe we should check for the existence of `MutationObserver` before doing this stuff.

srcjs/init_shiny.js
```diff
@@ -332,13 +338,49 @@ function initShiny() {
...
+    var config = {attributes: true, attributeFilter: ['style', 'class']};
+    observer.observe(el, config);
+    $el.data("shiny-theme-observer", true); // TODO: remove this flag on unbind?
```
wch, line 371: Instead of setting to `true`, this could be set to `observer`, which would also provide an interface for disconnecting the observer.

wch (approved):

PR #3195, opened 2020-12-07
All session parameters from the update* functions now default to `getDefaultReactiveDomain()`

wch:
Supersedes #3021 (since I'm not able to push to that branch). Closes #629.

PR #3199, opened 2020-12-10
Make `validate_session_object()` work with modules

wch:
PR #3195 introduced a problem: The `update*Input` functions would no longer work with modules. This fixes it.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -64,11 +64,11 @@ shiny 1.5.0.9000
 
 * Closed #3140: Added support for `...` argument in `icon()`. (#3143)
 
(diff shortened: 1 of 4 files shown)


cpsievert (approved):

cpsievert:
Oh, hmm, why is this test consistently failing on R < 4.0? https://github.com/rstudio/shiny/pull/3199/checks?check_run_id=1532608873#step:14:128

wch:
Yeah, that was due to #3194, I think. Trying to test on rstudio.cloud, but currently can't install devtools there due to libgit2 issues.

PR #3204, opened 2020-12-13
fix typo in docs

malcolmbarrett:
I actually can't seem to install several of the shiny-related packages right now (sass and bslib) and can't run roxygen on the pkg as a result, but it looks like you have the /document action for this repo, so I'll let you all handle it 😇


cpsievert (approved):

cpsievert:
Thank you @malcolmbarrett!

Issue #3202, opened 2020-12-11
session$setCurrentTheme should throw an error when trying to dynamically update to a different bootstrap version.


nstrayer:
### Issue
If you're using `session$setCurrentTheme()` and `bslib::bs_theme()` together to do dynamic theming, if there is a version mismatch between the original theme and the newly updated dynamic one (such as forgetting to set `version` in the update code) the theme change will silently change. 

### Proposed fix
Check to make sure that a dynamic update does not try and change the bootstrap version and throw an error if it does.

### Demo app

The following demonstrates the problem. Switching the theme won't work when the version is set to 4 because the initial chosen version was 3. Once you switch the version to 3, then the bootswatch changing starts working. 

__Steps to reproduce:__

- Change theme radio button to "flatly"
  - Nothing will happen
![image](https://user-images.githubusercontent.com/6764693/101956249-406aaf00-3bcd-11eb-864b-e4d413ed1080.png)

- Change the version radio button to '3' 
  - Theme changes now respected
![image](https://user-images.githubusercontent.com/6764693/101956281-4e203480-3bcd-11eb-91ce-e7afccc3abe6.png)

`app.R`
```R
library(shiny)
library(bslib)

# Setting version to 3 here
my_theme <- bs_theme(bootswatch = "darkly", version = "3")
ui <- fluidPage(
  theme = my_theme,
  radioButtons("current_theme", "Choose a theme for bslib",
               choices = c("darkly", "flatly")),
  radioButtons("bs_version", "Bootstrap version", choices = c("4", "3")),
)

server <- function(input, output, session) {
  observe({
    session$setCurrentTheme(
      bs_theme(bootswatch = input$current_theme, version = input$bs_version)
    )
  })
}

shinyApp(ui, server)
```

cpsievert:
Closed via #3210

PR #3210, opened 2020-12-16
Make sure setCurrentTheme() doesn't try and change bootstrap versions

nstrayer:
Based on https://github.com/rstudio/shiny/issues/3202

This PR adds a check to make sure that the new theme requested by `setCurrentTheme()` has the same version as the already loaded theme.

### Demo app

Before this PR the following app would not throw any errors when the user selects version 3. This would cause problems if the user then tried to change the bootswatch. With this PR shiny will throw an error on version switching. 

`app.R`
```R
library(shiny)
library(bslib)

# Setting version to 3 here
my_theme <- bs_theme(bootswatch = "darkly", version = "4")
ui <- fluidPage(
  theme = my_theme,
  radioButtons("current_theme", "Choose a theme for bslib",
               choices = c("darkly", "flatly")),
  radioButtons("bs_version", "Bootstrap version", choices = c("4", "3")),
)

  observe({
    session$setCurrentTheme(
      bs_theme(bootswatch = input$current_theme, version = input$bs_version)
    )
  })
}

shinyApp(ui, server)
```

### DIFF
--- a/R/shiny.R
+++ b/R/shiny.R
@@ -1322,7 +1322,12 @@ ShinySession <- R6Class(
       # bootstrapTheme, (2) re-executes any registered theme dependencies, and
       # (3) sends the resulting dependencies to the client.
 
-      # Make sure that the theme switching is just styles (e.g. bootswatch) rather than a different version
+      if (!is_bs_theme(theme)) {
+        stop("`session$setCurrentTheme()` expects a `bslib::bs_theme()` object.", call. = FALSE)
+      }
+
+      # Switching Bootstrap versions has weird & complex consequences
+      # for the JS logic, so we forbid it
       current_version <- bslib::theme_version(getCurrentTheme())
       next_version <- bslib::theme_version(theme)
       if (!identical(current_version, next_version)) {
--- a/R/shiny.R
+++ b/R/shiny.R
@@ -1323,14 +1323,15 @@ ShinySession <- R6Class(
       # (3) sends the resulting dependencies to the client.
 
       # Make sure that the theme switching is just styles (e.g. bootswatch) rather than a different version
-      mismatched_bootstrap_versions <- bslib::theme_version(theme) != bslib::theme_version(getCurrentTheme())
-      if(mismatched_bootstrap_versions){
-        stop(paste0("Requested theme uses a different bootstrap version (",
-                    bslib::theme_version(theme),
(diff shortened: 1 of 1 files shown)


R/shiny.R
```diff
@@ -1322,6 +1322,17 @@ ShinySession <- R6Class(
...
+      # Make sure that the theme switching is just styles (e.g. bootswatch) rather than a different version
+      mismatched_bootstrap_versions <- bslib::theme_version(theme) != bslib::theme_version(getCurrentTheme())
+      if(mismatched_bootstrap_versions){
+        stop(paste0("Requested theme uses a different bootstrap version (",
+                    bslib::theme_version(theme),
+                    ") than the currently loaded theme (",
+                    bslib::theme_version(getCurrentTheme()),
+                    "). Dynamic switching of bootstrap versions is not supported."),
+             call. = FALSE)
+      }
```
cpsievert, line 1334:
```suggestion
      current_version <- bslib::theme_version(getCurrentTheme()) 
      next_version <- bslib::theme_version(theme)
      if (!identical(current_version, next_version)) {
        stop(
          "Requested theme uses a different bootstrap version (",
          next_version, ") than the currently loaded theme (", current_theme,
           "). Dynamic switching of bootstrap versions is not supported.",
           call. = FALSE
         )
      }
```

R/shiny.R
```diff
@@ -1322,6 +1322,18 @@ ShinySession <- R6Class(
...
+      # Make sure that the theme switching is just styles (e.g. bootswatch) rather than a different version
+      current_version <- bslib::theme_version(getCurrentTheme()) 
+      next_version <- bslib::theme_version(theme)
+      if (!identical(current_version, next_version)) {
+        stop(
+          "Requested theme uses a different bootstrap version (",
+          next_version, ") than the currently loaded theme (", current_theme,
+           "). Dynamic switching of bootstrap versions is not supported.",
+           call. = FALSE
+         )
```
cpsievert, line 1334:
```suggestion
        stop(
          "session$setCurrentTheme() cannot be used to change the Bootstrap version ",
          "from ", current_version, " to ", next_version, ". ",
           "Try setting `bs_theme(version = ", next_version, ")`."
           call. = FALSE
         )
```

cpsievert (approved):
LGTM once latest suggestion is applied

PR #3212, opened 2020-12-17
Fix selectInput not closing dropdown when Enter is pressed

### LINKED ISSUE #3207: selectInput: when item is selected by pressing Enter, dropdown does not go away
In this basic example, if an item is selected by pressing Enter on the keyboard, the dropdown box doesn't go away, and you need to press Esc to make it disappear.

```R
shinyApp(
  ui = fluidPage(
    selectInput("variable", "Variable:",
                c("Cylinders" = "cyl",
                  "Transmission" = "am",
                  "Gears" = "gear")),
    tableOutput("data")
  ),
    output$data <- renderTable({
      mtcars[, c("mpg", input$variable), drop = FALSE]
    }, rownames = TRUE)
  }
)
```

What it looks like after selecting Transmission with the keyboard:

image

wch:
Fixes #3207.

This PR removes code in selectize-plugin-a11y where it triggers a `click` when Enter is pressed. I am not 100% sure that this will not cause problems for screen readers -- it seems likely that the code is there for a reason. However, I also know that the code causes the problem in #3207, and removing it fixes the problem. Any advice on why that code was there in the first place, and better ways to fix it would be appreciated. (@jooyoungseo, if you have any input, we'd appreciate it!)

This PR also adds some comments explaining why IDs are assigned to each selection item.


jooyoungseo:
Aw, I will also take a close look this weekend! Sorry for any inconvenience.

wch:
I've changed the PR so that if Enter is pressed, it will close the dropdown and then set focus to the selectize input.

jooyoungseo:
@wch, not closing the selectize combo-box after enter key does not provide screen readers with noticeable benefit. I have tested that against this PR with various screen readers in different browsers.

I assume the original author wanted to make sure enter key work using the click trigger, but I agree with you that we should take it out unless it gives us any significant accessibility advantages.

I am sorry that I did not notice this buggy behavior earlier.

jooyoungseo:
Maybe the original author wanted to explicitly make the selectize remain open after selection for screen reader users to not lose their focus; however, screen readers use a focus mode (AKA, forms mode) for combo-box interaction and it switches back to document mode (AKA, virtual cursor) when either users press ESC key or the form is collapsed. And, the focus stays in the selected item. In sum, there is no issue without the explicit trigger.

wch:
I suspect the original author did the `.click()` in order to keep the focus on the input object. With selectize, however, the default behavior is to always open the dropdown when the input gets focus, and giving it focus without opening the dropdown requires a strange workaround, like the one that I did.

With the previous code, when Enter is pressed, I found that the screen reader's cursor stayed on the same area in the window, and it started reading numbers from the table that was uncovered when the dropdown went away. But that was in Chrome on Mac, and I gather that it's not a widely used combination for visually-impaired users.

jooyoungseo:
As you know, the reason why screen readers start reading the numbers in table is because of the `aria-live = "polite"` for dynamic content. FYI, Chrome has some accessibility issues on Mac, so Safari might be the most accessible browser for Mac screen reader (i.e., VoiceOver). I don't see any difference between before and after this PR in terms of aria-live announcement on my end.

wch:
Sorry, I meant that with the code in a previous version of this PR, it started reading the table after selecting the item. In that version of the code, I simply removed the code where the Enter wold trigger a `.click()`, and I did not add any code to make the focus go to the selectize input. Gary pointed out that that behavior was problematic for accessibility.

At any rate, I'm glad that the voice announcement behavior now is the same as before the PR. That means that both the voice announcements and the visual behavior are both behaving well.

jooyoungseo:
Aw, I got it! (since I did not test the previous version of this PR, I got a bit confused. I am sorry) Yes, as far as I tested with the latest commit, the screen reader behavior is the same as you would expect for sighted users.

I understand what Gary meant concerning the focus issue (that's a great highlight). Taken together, I feel very comfortable using the latest version of this PR.

Thank you so much for your hard work, Winston!

R/input-select.R
```diff
@@ -224,7 +228,7 @@ selectizeIt <- function(inputId, select, options, nonempty = FALSE) {
       type = 'application/json',
       `data-for` = inputId, `data-nonempty` = if (nonempty) '',
       `data-eval` = if (length(res$eval)) HTML(toJSON(res$eval)),
-      if (length(res$options)) HTML(toJSON(res$options)) else '{}'
+      HTML(toJSON(res$options))
```
cpsievert, line 231: Why this change?

cpsievert: Oh, I see, is it because `res$options` will always have positive length now?

cpsievert (approved):

R/input-select.R
wch: I changed this when I was trying a different method for fixing things, but it ended up being independent from the other stuff in this PR. Still, it simplified the code and I thought it was a good idea so I decided to keep it.

The `else` case of the original if-else was (apparently) to handle the case where `res$options` was either an empty list or `NULL` -- the `NULL` would get translated to a `null` in JSON. But now, with the code earlier in this function, if the incoming options were NULL, it gets turned into an empty named list, which then gets translated to `{}` in JSON. So the special case handling isn't necessary here.

wch: Although, now that you mention it, I see one potential issue: if `options` was an empty unnamed list, then the `toJSON()` would convert it to `[]` instead of `{}`. So I'll change the code above so that instead of `if (is.null(options))`, it will be `if (length(options) == 0)`.

wch: Oh, but in the case of an empty list or NULL, it would get turned into a named list anyway by the next `if` statement, where it adds the accessibility plugin to `options$plugins`.

PR #3221, opened 2020-12-18
bump dev version to 1.5.0.9006

schloerke:
Quickly signifies that Shiny Dev Mode is available

Related: https://github.com/rstudio/shiny/pull/3174


cpsievert (approved):

wch:
I merged, but now I notice that that node_js job said there were differences in the built .js files. They also differ from what I get when I build on my machine. I think you may need to re-run `yarn` to update your packages.

schloerke:
Yup! I'll go after it on Monday

PR #3226, opened 2020-12-21
Drop GHA pr-commands and add GHA Rituals

schloerke:
* Auto document
* Auto build js files
* Use current GitHub actor to do commit
* Push to master or PR if any changes found


.github/workflows/R-CMD-check.yaml
```diff
@@ -45,37 +45,43 @@ jobs:
...
+          saveRDS(pak::pkg_deps_tree("local::.", dependencies = TRUE), ".github/r-depends.rds")
 
       - name: Cache R packages
-        uses: actions/cache@v1
+        uses: actions/cache@v2
         with:
           path: ${{ env.R_LIBS_USER }}
-          key: ${{ matrix.config.os }}-r-${{ matrix.config.r }}-1-${{ hashFiles('.github/depends.Rds') }}
-          restore-keys: ${{ matrix.config.os }}-r-${{ matrix.config.r }}-1-
+          key: ${{ runner.os }}-${{ steps.install-r.outputs.installed-r-version }}-1-${{ hashFiles('.github/r-depends.rds') }}
```
wch, line 64: I think this should still use `matrix.config.os`?

schloerke: Yes

.github/workflows/R-CMD-check.yaml
```diff
@@ -86,8 +92,8 @@ jobs:
         uses: actions/cache@v1
```
wch, line 92: Might as well move to `v2` here.

.github/workflows/rituals.yaml
```diff
@@ -0,0 +1,139 @@
...
+        shell: Rscript {0}
+        run: |
+          install.packages("pak", repos = "https://r-lib.github.io/p/pak/dev/")
+          saveRDS(pak::pkg_deps_tree("local::.", dependencies = TRUE), ".github/r-depends.rds")
+
+      - name: Cache R packages
+        uses: actions/cache@v2
+        with:
+          path: ${{ env.R_LIBS_USER }}
+          key: ${{ runner.os }}-${{ steps.install-r.outputs.installed-r-version }}-1-${{ hashFiles('.github/r-depends.rds') }}
```
wch, line 61: `matrix.config.os`

wch:
The reason that pak failed on Mac-devel is because it couldn't build textshaping:
https://github.com/rstudio/shiny/runs/1595657866?check_suite_focus=true#step:10:481

I believe that's why you switched back to using remotes. But it looks like it also fails to install with remotes:
https://github.com/rstudio/shiny/runs/1595770963?check_suite_focus=true#step:9:5382

The difference is that, with remotes, the error is just silently ignored. 


With pak, it couldn't uncompress bslib on Windows. I'm not sure why that is. I'll try re-running that job and see if it magically gets better.
https://github.com/rstudio/shiny/runs/1595657925?check_suite_focus=true#step:10:365

schloerke:
Closing in favor of https://github.com/rstudio/shiny/pull/3230

PR #3236, opened 2021-01-04
Don't change the return value of bootstrapPage() if bslib isn't relevant

### LINKED ISSUE #3235: The return value of bootstrapPage() has changed in an unfortunate way
With the current CRAN version of shiny (1.5.0):

```r
> str(bootstrapPage())
List of 3
 $ : NULL
 $ : NULL
 $ : list()
 - attr(*, "class")= chr [1:2] "shiny.tag.list" "list"
 - attr(*, "html_dependencies")=List of 1
  ..$ :List of 10
  .. ..$ name      : chr "bootstrap"
...
```

With the development version (> 1.5.0):

```r
> str(bootstrapPage())
List of 4
 $ :function ()  
  ..- attr(*, "srcref")= 'srcref' int [1:8] 83 15 126 3 15 3 83 126
  .. ..- attr(*, "srcfile")=Classes 'srcfilecopy', 'srcfile' <environment: 0x7fbefe65d950> 
  ..- attr(*, "class")= chr "shiny.tag.function"
 $ : NULL
 $ : NULL
 $ : list()
 - attr(*, "class")= chr [1:2] "shiny.tag.list" "list"
```

This means, user code that does something like this will now break:

```r
x <- bootstrapPage()
tagAppendAttribute(x[[3]], class = "navbar")
```

We can do better to not break this code if `{bslib}` isn't being used, but the structure will likely have to change if `{bslib}` is being used.

Some relevant discussion/use-case https://github.com/radiant-rstats/radiant.data/pull/26#issuecomment-754259304

cpsievert:
Closes #3235


R/bootstrap.R
```diff
@@ -43,19 +43,27 @@ bootstrapPage <- function(..., title = NULL, responsive = deprecated(), theme =
...
   )
 
-  ui <- setLang(ui, lang)
+  # If theme is a bslib::bs_theme() object, bootstrapLib() needs to come first
+  # (so other tags, when rendered via tagFunction(), know about the relevant
+  # theme). However, if theme is anything else, we intentionally avoid changing
+  # the tagList() contents to avoid breaking user code that makes assumptions
+  # about the return value https://github.com/rstudio/shiny/issues/3235
+  if (is_bs_theme(theme)) {
+    ui <- tagList(bootstrapLib(theme), ui)
```
wch, line 61: This is a slight change of the structure, right? It adds one more level of nesting. It would be better to keep the existing structure, especially since that's the one we'll use in the future.

R/bootstrap.R
```diff
@@ -43,19 +43,27 @@ bootstrapPage <- function(..., title = NULL, responsive = deprecated(), theme =
   }
 
   ui <- tagList(
-    bootstrapLib(theme),
```
wch, line 46: This can be replaced with:

```R
if (is_bs_theme(theme)) bootstrapLib(theme),
```

(See comment below for why we'd do it this way)

cpsievert: I think we'd still have the same problem because of this?

```r
> str(tagList(NULL, "a"))
List of 2
 $ : NULL
 $ : chr "a"
 - attr(*, "class")= chr [1:2] "shiny.tag.list" "list"
```

cpsievert: I suppose we could do a `do.call(tagList, dropNulls(ui))` though?

R/bootstrap.R
cpsievert: Yea, it's adds another level to the tree, wasn't sure which direction I liked better, but I get on-board with a flatter structure

R/bootstrap.R
cpsievert: Actually I'm not sure that's a good idea either since that could break existing code for similar reasons

wch: Oh right, the `NULL` would be in there. Something like this could work and is reasonably straightforward:

```R
  args <- list(
    if (!is.null(title)) tags$head(tags$title(title)),
    if (is.character(theme)) {
      if (length(theme) > 1) stop("`theme` must point to a single CSS file, not multiple files.")
      tags$head(tags$link(rel="stylesheet", type="text/css", href=theme))
    },
    # remainder of tags passed to the function
    list(...)
  )

  if (is_bs_theme(theme)) {
    args <- c(bootstrapLib(theme), args)
    ui <- do.call(tagList, args)
  } else {
    ui <- do.call(tagList, args)
    ui <- attachDependencies(ui, bootstrapLib())
  }
```

cpsievert: I landed on something slightly different before seeing this e117d7e

Would you prefer to do the `do.call()` in the non-bslib case as well

R/bootstrap.R
```diff
@@ -43,19 +43,27 @@ bootstrapPage <- function(..., title = NULL, responsive = deprecated(), theme =
...
   )
 
-  ui <- setLang(ui, lang)
+  # If theme is a bslib::bs_theme() object, bootstrapLib() needs to come first
+  # (so other tags, when rendered via tagFunction(), know about the relevant
+  # theme). However, if theme is anything else, we intentionally avoid changing
+  # the tagList() contents to avoid breaking user code that makes assumptions
+  # about the return value https://github.com/rstudio/shiny/issues/3235
+  if (is_bs_theme(theme)) {
+    ui <- do.call(tagList, c(bootstrapLib(theme), ui))
```
wch, line 61: I started writing a comment explaining this potentially-confusing usage of a tagList as a regular list (of arguments), and how it could be changed back to a normal function call (without `do.call`) in the future if we decide to get rid of the `else` code path, but it was getting long and I think it's just easier to write the code the way I did above. (My goal is to avoid situations where people are afraid to touch the code because they don’t understand what it does.)

PR #3249, opened 2021-01-11
Use url checker

schloerke:
* Auto update links: https://github.com/rstudio/shiny/pull/3249/files#diff-19de1df58f452c6bbed8dbd2554145ebb03ba62be7da8c99f8d21ce4938f0cc5R85-R99
  * Will throw an error if man files are changed. (Which requires manual updates to R)
  * Will auto update links to news, readme, etc.

### DIFF
--- a/.github/workflows/rituals.yaml
+++ b/.github/workflows/rituals.yaml
@@ -73,6 +73,7 @@ jobs:
           pak::local_install_dev_deps(upgrade = TRUE)
           pak::pkg_install("sessioninfo")
           pak::pkg_install("devtools")
+          pak::pkg_install("r-lib/urlchecker")
 
       - name: Session info
         shell: Rscript {0}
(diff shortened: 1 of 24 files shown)


.github/workflows/rituals.yaml
```diff
@@ -81,6 +82,21 @@ jobs:
...
+          if [ -n "$(git status --porcelain man)" ]
+          then
+            git status --porcelain
+            >&2 echo "Updated links found in files above"
+            >&2 echo 'Run `urlchecker::url_update()` to fix links locally'
+            exit 1
+          fi
+          # Add locally changed urls
+          git add .
+          git commit -m 'Update links (GitHub Actions)' || echo "No link changes to commit"
```
cpsievert, line 98: Is there any point to committing the changes?

cpsievert (approved):

.github/workflows/rituals.yaml
schloerke: I don't like to commit something that can be automated.  Like documentation updates

wch: I'm uneasy about having it change URLs without someone seeing the redirected target first. It's not too implausible that a redirect would end up at the wrong place, say, if some service is temporarily down and has a generic redirect.

Also, am I correct that if roxygen is used, this won't change the original .R files?

schloerke: In the lines above (89-95), I'm checking for any man files were updated. If so, I throw an error as `urlchecker` does not alter `./R` files which would be updated in the roxygen step.

You are correct in that this is open to issues due to services being down.  Maybe we only do this step for `rc-v*` branches?

---------------------------

This goes back to the "outdated code should throw an error" vs "Update all code that can be auto-generated" dilemma. Currently, `rituals.yaml` is in the "throw if it can't be auto-generated" camp.

cpsievert: > Maybe we only do this step for rc-v* branches?

I like that idea

schloerke: Implemented in 26c9ac0

Shown that the url step is not run in https://github.com/rstudio/shiny/pull/3249/checks?check_run_id=1691570478

wch (approved):

PR #3256, opened 2021-01-14
Run accessiblity plugin JS when DOM is loaded (via defer attribute) a…

cpsievert:
https://github.com/paypal/bootstrap-accessibility-plugin (added in #2911) has a bunch of problems with it which we ended up fixing in https://github.com/rstudio/bslib/pull/224, see there for details


tools/updateBootstrapAccessibilityPlugin.R
```diff
@@ -8,26 +8,9 @@ library(rprojroot)
 
 sha   <- "fbbf870eafc1ee5d4547fabbeea71778ddbe2166"
 dest_dir  <- rprojroot::find_package_root_file("inst/www/shared/bootstrap/accessibility")
+unlink(dest_dir, recursive = TRUE)
+dir.create(dest_dir)
 
-min_js_url       <- sprintf("https://raw.githubusercontent.com/paypal/bootstrap-accessibility-plugin/%s/plugins/js/bootstrap-accessibility.min.js", sha)
-css_url       <- sprintf("https://raw.githubusercontent.com/paypal/bootstrap-accessibility-plugin/%s/plugins/css/bootstrap-accessibility.css", sha)
-url <- c(min_js_url, css_url)
```
wch, line 14: It would help to have a comment about why the css files aren't copied over.

tools/updateBootstrapAccessibilityPlugin.R
```diff
@@ -8,26 +8,9 @@ library(rprojroot)
...
-)
-
-# Copying css file:
-dir.create(file.path(dest_dir, "css"), recursive = TRUE)
-file.copy(
-  file.path(tempdir(), "bootstrap-accessibility.css"),
-  file.path(dest_dir, "css"),
-  overwrite = TRUE
-)
+url <- sprintf("https://raw.githubusercontent.com/paypal/bootstrap-accessibility-plugin/%s/plugins/js/bootstrap-accessibility.min.js", sha)
```
wch, line 15: The content of the JS file should be wrapped in a `$()` so that it runs on load. (We'll need a patch for this)

wch (approved):

PR #3264, opened 2021-01-22
Use `rlang::hash()` instead of `digest()`

wch:
To be merged after 1.6.0 is released.

This changes shiny to use `rlang::hash()` instead of `digest::digest()`. The former is very fast, and reduces the number of dependencies needed by Shiny.

- [x] Update NEWS (after 1.6.0 is released, to avoid merge conflict)


schloerke (approved):
Very exciting!

PR #3273, opened 2021-01-28
Perform type check before value check

hadley:
If you don't validate the input types before validating the values, you get uninformative errors from min/max. (Which comes up in Mastering Shiny).

I also moved `getSliderType()` into input-slider.R since it's the only place that it's used, and mildly improved the error message.

I'm happy to add some tests, but I think the existing tests will need some refactoring, as there are acres of tests that are neither particularly sensitive nor specific.

### DIFF
--- a/R/input-slider.R
+++ b/R/input-slider.R
@@ -95,9 +95,8 @@ sliderInput <- function(inputId, label, min, max, value, step = NULL,
     )
   }
 
-  validate_slider_value(min, max, value, "sliderInput")
-
   dataType <- getSliderType(min, max, value)
+  validate_slider_value(min, max, value, "sliderInput")
 
   if (is.null(timeFormat)) {
     timeFormat <- switch(dataType, date = "%F", datetime = "%F %T", number = NULL)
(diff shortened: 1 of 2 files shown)


hadley:
After #3274, I could take a look the input checking tests if you wanted.

wch:
Suppose someone calls `sliderInput()` without min/max/value. Before the change, this is the error:

```
Error in validate_slider_value(min, max, value, "sliderInput") : 
  argument "min" is missing, with no default
```

After the change this is the error:

```
Error in dropNulls(list(value, min, max)) : 
  argument "value" is missing, with no default
```

The old error message isn't perfect, but at least it lets the user know it's coming from the slider. Maybe just `force()` those arguments at the top of `sliderInput()`?

wch:
Hm, it looks like using `force()` also results in an uninformative error:

```R
sliderInput("x", "x")
#> Error in force(min) : argument "min" is missing, with no default
```

But simply evaluating them does result in something useful. For example, if the function has `inputId; label; min; max; value` at the top, then the error is:

```R
sliderInput("x", "x")
#> Error in sliderInput("x", "x") : 
#>   argument "min" is missing, with no default
```

`force()` is more informative when reading the code, but the bare variable names result in a better error message.

hadley:
I can take a look at that specific case.

Do you want better tests too?

wch:
Sure, better tests wouldn't hurt. We should also have tests for the most common case, which is when the args are missing.

PR #3276, opened 2021-01-29
Add size xl to modalDialog

msaltieri:


cpsievert:
Thanks, this has now been addressed via https://github.com/rstudio/shiny/pull/3410

PR #3288, opened 2021-02-10
Update to Font-Awesome 5.15.2

### LINKED ISSUE #3286: Upgrade Font Awesome icons to latest version 5.15.2
Font Awesome is currently included in version 5.13.0 (see #2891).

The later versions (5.13.1, 5.14.0, 5.15.0, 5.15.1, 5.15.2) have added some more Covid-19-related icons (and a few others), which might be very useful in many Shiny apps built these days.

See [all added/updated icons](https://fontawesome.com/icons?d=gallery&s=solid&v=5.13.1,5.14.0,5.15.0,5.15.1,5.15.2&m=free) and the [release notes](https://fontawesome.com/changelog/latest).

@wch Maybe you would like give it a go once more? :)

wch:
Closes #3286.


tools/updateFontAwesome.R
```diff
@@ -100,7 +100,7 @@ v4_icon_df <- data.frame(
 )
 
 # Get info on icons in version 5.
-metadata <- fromJSON(find_package_root_file("data-raw", "fontawesome", "icons.json"))
+metadata <- fromJSON(file.path(source_dir, "metadata", "icons.json"))
 styles <- lapply(metadata, `[[`, "styles")
```
schloerke, line 105: Near line 167, can we add this line: 

```r
cat('# Generated by tools/updateFontAwesome.R: do not edit by hand\n', file = font_awesome_r)
```

And update the following line to 

```r
cat('font_awesome_brands <- c(\n  "', file = font_awesome_r, append = TRUE)
```

R/font-awesome.R
```diff
@@ -4,7 +4,6 @@ font_awesome_brands <- c(
   "accusoft",
   "acquisitions-incorporated",
   "adn",
-  "adobe",
```
schloerke, line 7: What happens when an icon is asked for an it doesn't exist?

schloerke (approved):
LGTM given update to adding the "do not edit" comment line and rerunning the update script.

tools/updateFontAwesome.R
wch: Done, thanks for the suggestion!

R/font-awesome.R
wch: Looks like it'll generate the HTML, but it obviously won't display in the web page.

```
as.character(shiny::icon("foo", lib = "font-awesome"))
#> [1] "<i class=\"fa fa-foo\" role=\"presentation\" aria-label=\"foo icon\"></i>"
```

The auto-generated file is only used to figure out whether to use a `fab` prefix:

https://github.com/rstudio/shiny/blob/a6347341e3968a1b7b66f840efbc606e673c1a40/R/bootstrap.R#L1546-L1548

FloEdelmann:
Awesome, thanks @wch for this fast response!

PR #3298, opened 2021-02-18
Proposal for new feature: conditionalReactive

stefanedwards:
This PR proposes a new method that allows finer control over when an upstream, invalidated reactive should invalidate downstream reactives. Basically, it wraps

    # in a server-function
    value <- reactiveVal()
    observe({
      oldvalue <- isolate(value())
      if (oldvalue != input$newvalue) {
        value(input$newvalue)
      }
    })

into

    value <- conditionalReactive(reactive(input$newvalue))

or, simply avoid triggering a downstream reactive if the input value has not change.

```
output$result <- renderX({
  value <- conditionalReactive(reactive(input$newvalue))
  # do stuff, but only if value has *changed* from previously, and not purely because something triggered `input$newvalue` to re-send an event with the same value
})
```
without the need of additional variables/reactives to keep track of new and old values.

**Usage:**

```
conditionalReactive(
  x,
  checkFun = NULL,
  fire.on.NULL = c("never", "change", "always"),
  fire.on.NA = c("never", "change", "always"),
  label = NULL
)
```

`fire.on.NULL` and `fire.on.NA` allows us to select different strategies if the new value is `NULL` or `NA`. In the first example (`if (oldvalue != input$newvalue)`, if the new value is `NULL` or `NA`, the if-statement will error or not behave as expected (as the expression either returns a logical vector with length not-equal to 1, or simply `NA`). With these two arguments, the saves the user from hassle of checking for `NA`s and `NULL`s - while allowing them to decide e.g. whether a `NULL` should be considered different from the previous `NULL`.

`checkFun` allows us to fine-tune what is considered an inequality between the new and old value. One could decide that not only should the new value be different from the previous, but it should be different by a margin, before recalculating, i.e.

```
conditionalReactive(r, checkFunc = function(x,y) {abs(x-y) > 5})
```

**Other scenarios:**

Say a user must change 2 or more input values (e.g. `x` and `y`) before allowing invalidating downstreams reactives:

```
# without conditionalReactive
xy <- reactiveVal(c(0,0))
observe({
  newvalue <- c(input$x, input$y)
  if (all(newvalue != xy()) # xy is isolated by bindEvent
    xy(newvalue)
}) %>% bindEvent(input$x, input$y)

# with conditionalReactive
conditionalReactive(reactive(c(input$x, input$y)))
```

**R6 classes, complex data structures, and side-effects**
In instances where a recalculating can be computational consuming, shiny now offers `bindCache`. However, if the dependencies of the current reactive result are unclear (many, many inputs), listing these in `bindCache` is not attractive.
In cases of reference-like objects like R6, sharing the result among different users is a bad idea (i.e. don't cache if evaluation has side-effects).

**Annoy your users**
If you want to your users to click 3 times on an actionButton before moving on,
```
reactive({
  # wait for it...
  conditionalReactive(input$go, checkFunc = function(x,y) (x-y) >= 3)
  # go!
})
```

**Avoids re-calculating old result when debounced**
In DataTable, the user can select cells/rows/column, but if the input is combined with `debounce`, you can get situations where the input is re-sent with the same input, if a user selects then deselects the same item before the debounce fires. Additionally, for DataTable input, the input-reactive (e.g. `input$dt_cells_selected`) is `NULL` before the table is loaded, so any handling of the input has to deal with `NULL`s before considering whether the action is worth updating e.g. a render.

```
output$plot <- renderPlot({
  cells_selected <- conditionalReactive(reactive(input$dt_cells_selected))
  req(nrow(cells_selected) > 0)
})
```

**Comparison to existing features of Shiny**
Some of these scenarios might be solves through the use of `req`, `validate`, and `bindCache`, but by wrapping the conditionality into a method allows the code to be more readable, and the method takes care of some of the recurring obstacles of dealing with `NULL`s, `NA`s, or inputs with different lengths.


wch:
Hi, thanks for taking the time to put this together. We have been gradually collecting examples of useful higher-order reactives that could be useful enough for users to merit a new function in Shiny. We are planning revisit the topic at some point in the future and decide what to do with these ideas. In other words, we won't merge this now, but we will take a look again in the future.

PR #3301, opened 2021-02-18
Add shiny:restyle event for listening to instances where session() updates stylesheet(s)

cpsievert:
Will be useful for real-time theming integration in flexdashboard https://github.com/rstudio/flexdashboard/pull/308


srcjs/output_binding_html.js
```diff
@@ -174,6 +174,9 @@ function renderDependency(dep) {
         sheet.disabled = true;
         if (browser.isIE) sheet.cssText = "";
         $(sheet.ownerNode).remove();
+        $(document).trigger({
+          type: 'shiny:restyle', oldSheet: sheet
```
wch, line 178: What is `oldSheet` used for?

wch:
Just a note: we should also document this in the Shiny JS events article.

cpsievert:
Sorry, I was actually thinking of closing this on Friday since I found a way to do what I needed without this. 

That being said, I suppose this could maybe useful for something else at some point 🤷

srcjs/output_binding_html.js
cpsievert: Mainly so can get at the `href` so you which `<link>` was updated

cpsievert: That being said, maybe we could add the `name` and `version` of the relevant HTML dependency

PR #3302, opened 2021-02-19
Modify `icon()` function to call `fontawesome::fa_i()` for equivalent functionality

rich-iannone:
This PR migrates the dependency files for Font Awesome icon support (font files, CSS, JS) over to the fontawesome package and modifies the `shiny::icon()` function so that it calls `fontawesome::fa_i()` to generate the `<i>` tag object. Other scripts that create supporting Font Awesome objects have been moved over to the fontawesome package and are removed here.

The fontawesome package has been placed in `Imports` and `Remotes` (for now, due to be in CRAN shortly). Manual testing has been done using this branch and the release candidate of fontawesome. The demo app `138-icon-fontawesome` works with these builds.


DESCRIPTION
```diff
@@ -80,6 +78,7 @@ Imports:
     mime (>= 0.3),
     jsonlite (>= 0.9.16),
     xtable,
+    fontawesome,
```
wch, line 81: It would be good to use the upcoming CRAN version as a requirement, in case users have the old version installed.

DESCRIPTION
```diff
@@ -113,6 +112,8 @@ Suggests:
     ragg,
     showtext,
     sass
+Remotes:
+    rstudio/fontawesome
```
wch, line 116: Since it's going to be on CRAN soon, I think we can remove this and just merge once it's there.

R/bootstrap.R
```diff
@@ -1492,78 +1492,79 @@ downloadLink <- function(outputId, label="Download", class=NULL, ...) {
...
-#'   the v5.13.0 set are supported with the v4 naming convention) and
-#'   [Glyphicons](https://getbootstrap.com/components/#glyphicons)
-#'   libraries. Note that the "fa-" and "glyphicon-" prefixes should not be used
-#'   in icon names (i.e. the "fa-calendar" icon should be referred to as
-#'   "calendar")
-#' @param class Additional classes to customize the style of the icon (see the
+#' of a button, or as an icon for a [tabPanel()] within a [navbarPage()].
+#'
+#' @param name The name of the icon. Icons are drawn from the [Font Awesome
+#'   Free](https://fontawesome.com/) (currently icons from the v5.13.0 set are
```
wch, line 1498: The version number here is wrong -- but more generally, since the version can change independently of Shiny, perhaps we should link to the {fontawesome} package page, and list the version number there. Or link to the function that would implement https://github.com/rstudio/fontawesome/issues/30.

If those changes in fontawesome need to happen after the CRAN release, then we'll have to do something else in the meantime so that the incorrect version number is not shown here.

R/bootstrap.R
```diff
@@ -1492,78 +1492,79 @@ downloadLink <- function(outputId, label="Download", class=NULL, ...) {
...
 
-  # determine stylesheet
+  # Stop function if `lib` value is not allowed
   if (is.null(prefix)) {
-    stop("Unknown font library '", lib, "' specified. Must be one of ",
-         paste0('"', names(prefixes), '"', collapse = ", "))
+    stop(
+      "Unknown font library '", lib, "' specified. Must be one of ",
+      paste0('"', names(prefixes), '"', collapse = ", "),
+      call. = FALSE
```
wch, line 1544: I think this `call.=FALSE` will make the message more confusing, since users won't see where the error is coming from.

man/icon.Rd
```diff
@@ -7,33 +7,33 @@
 icon(name, class = NULL, lib = "font-awesome", ...)
```
wch, line 7: Why did this file go away?

R/bootstrap.R
rich-iannone: I agree and it's gone now.

man/icon.Rd
rich-iannone: The .Rd file? I still see it.

R/bootstrap.R
rich-iannone: How about we take out the version number for now? I can make changes in both packages' docs shortly after the release.

DESCRIPTION
rich-iannone: Minimum version requirement has been added here. Thanks!

man/icon.Rd
wch: Oh sorry, I was confused.

R/bootstrap.R
wch: Sure, that sounds reasonable.

rich-iannone: This is now changed.

cpsievert:
@rich-iannone the unit tests are failing because some of the tabset panel logic relies on this behavior:

```r
> cat(as.character(icon(NULL)))
<i class="" role="presentation" aria-label=" icon"></i>
```

We should probably do an early return for that case...

rich-iannone:
Thanks @cpsievert , didn't know about that one and was puzzled by the CI results. I've pushed a change based on your recommendation.

DESCRIPTION
```diff
@@ -116,7 +115,7 @@ Suggests:
 Remotes:
     r-lib/rlang,
     rstudio/bslib,
-    rstudio/htmltools
+    rstudio/htmltools#212
```
cpsievert, line 118: Needed for the `tagGetAttribute()` usage, which isn't necessary, but it'd be nice to move in that direction

cpsievert:
> Would it be possible to get the tabPanel to take a fontawesome::fa() object as the icon?

Yea, but I'm not sure I want to allow `icon` to be arbitrary `HTML()`, so I've sent @rich-iannone a PR to have `fa()` return a `shiny.tag` rather than an `HTML()` string  https://github.com/rstudio/fontawesome/pull/59/files

wch (approved):

PR #3311, opened 2021-03-02
Update to jQuery 3.6.0

wch:
This PR also makes it so the jQuery version is stored in a variable `version_jquery`, and that value is used when building the documentation. In the future, only the version number in the script needs to be updated, and nothing else needs to be edited by hand.


tools/updatejQuery.R
```diff
@@ -6,10 +8,19 @@ jq_cdn_download <- function(version) {
...
 )
 
+dest_file <- rprojroot::find_package_root_file("R", "version_jquery.R")
+
+writeLines(
+  c(
+    "# Generated by tools/updatejQuery.R; do not edit by hand",
+    sprintf('version_jquery <- "%s"', version)
+  ),
+  find_package_root_file("R", "version_jquery.R")
```
schloerke, line 25:
```suggestion
  rprojroot::find_package_root_file("R", "version_jquery.R")
```

tools/updatejQuery.R
```diff
@@ -6,10 +8,19 @@ jq_cdn_download <- function(version) {
...
+)
```
schloerke, line 26: (untested) Update jquery version to srcts
```suggestion
)

# Update TypeScript installation
withr::with_dir(
  rprojroot::find_package_root_file("srcts"),
  {
  	exit_code <- system(paste0("yarn add --dev jquery@", version))
  	if (exit_code != 0) stop("yarn could not install jquery")
  }
)
```

tools/updatejQuery.R
```diff
@@ -6,10 +8,19 @@ jq_cdn_download <- function(version) {
...
-jq_cdn_download("3.5.1")
+jq_cdn_download(version)
 
 download.file(
   "https://raw.githubusercontent.com/jquery/jquery/master/AUTHORS.txt",
   "inst/www/shared/jquery-AUTHORS.txt"
 )
 
+dest_file <- rprojroot::find_package_root_file("R", "version_jquery.R")
+
```
schloerke, line 19:
```suggestion
```

schloerke (changes requested):
If we can also update the TypeScript jquery version too, that'd be great!

schloerke (approved):

PR #3315, opened 2021-03-04
Various improvements to tab panels

### LINKED ISSUE #1823: tabSetpanel first argument
Migrating from shiny 1.0.3 to 1.0.4 tabSetpanel changed in a way it took a while to figure out the issue, see example below. Although caused by bad programming on my side by not declaring id specifically, others may experience similar problems.

### Works in 1.0.3 and 1.0.4
```
library(shiny)    
ui <- shinyUI(basicPage( 
  
  textOutput("text"),
  tabsetPanel(id = "tabs",
              tabPanel("Tab 1", value = "1", "Tab 1 content"),
              tabPanel("Tab 2", value = "2", "Tab 2 content")
  )
))

  output$text <- renderText({paste0("Viewing tab \"", input$tabs, "\"")})
}

```

### Works in 1.0.3 and crashes in 1.0.4 with "Error : $ operator is invalid for atomic vectors" on start
```
library(shiny)    
ui <- shinyUI(basicPage( 
  
  textOutput("text"),
  tabsetPanel("tabs",
              tabPanel("Tab 1", value = "1", "Tab 1 content"),
              tabPanel("Tab 2", value = "2", "Tab 2 content")
  )
))

  output$text <- renderText({paste0("Viewing tab \"", input$tabs, "\"")})
}

```

### LINKED ISSUE #1928: tabsetPanel doesn't ignore NULLs
```{R}
tabsetPanel(tabPanel("blah"), if (FALSE) tabPanel("alsoblah"))
```
It generates HTML for the NULL element. 
```
<div class="tabbable">
  <ul class="nav nav-tabs" data-tabsetid="5123">
    <li class="active">
      <a href="#tab-5123-1" data-toggle="tab" data-value="blah">blah</a>
    </li>
    <li>
      <a href="#tab-5123-2" data-toggle="tab"></a><!-- this is linking to NULL -->
    </li>
  </ul>
  <div class="tab-content" data-tabsetid="5123">
    <div class="tab-pane active" data-value="blah" id="tab-5123-1"></div>
    tab-5123-2
  </div>
</div>
```
My first thought is to add `dropNulls` [here](https://github.com/rstudio/shiny/blob/master/R/bootstrap.R#L672), but I suppose an informative warning or error message might be appropriate.

### LINKED ISSUE #2927: tabsetPanel() doesn't work with list of panels
It would be nice if `tabsetPanel()` could work more easily with a list of panels, so you can program with it without needing to know `do.call()`:

```R
panels <- list(
  tabPanel("Plot", "plot"),
  tabPanel("Summary", "summary"),
  tabPanel("Table", "table")
)

ui <- fluidPage(
  tabsetPanel(panels)
)

ui <- fluidPage(
  do.call(tabsetPanel, panels)
)
```

This is somewhat related to #1928 (in the sense that if you fix issue you should also fix this one)

I think you have two main options to fix:

* Automatically splice list inputs (like htmltools)
* Use `list2()` instead of list to get explicit `!!!` splicing

### LINKED ISSUE #3313: tabsetPanel() et al give confusing error messages when they receive incorrect input
For example:

```r
library(shiny)

ui <- fluidPage(
    tabsetPanel(
        "1",
        tabPanel("B", "b")
    )
)
```

```r
Error in divTag$attribs : $ operator is invalid for atomic vectors
```

This is especially confusing given that strings may be provided to `navlistPanel()`

### LINKED ISSUE #3320: Tab panel functions don't produce Bootstrap 4 compatible markup
And as a result, `tabsetPanel()` et al. don't work well with Bootstrap 4 "plugin" frameworks like Bootswatch 4 or AdminLTE 3, which make strong assumptions about nav (as well as other) markup. In addition to working better with these frameworks, if `tabPanel()` markup produced the right markup when it's aware that Bootstrap 4 is relevant (via it's `{bslib}` integration), then folks can also use tabs without the compatibility layer that comes by default with `bs_theme()`

```r
library(shiny)
library(bslib)

ui <- fluidPage(
  theme = bs_theme() %>% bs_remove("bs3compat"),
  tabsetPanel(
    id = "tabset", type = "pills",
    tabPanel("A", "a"),
    tabPanel("B", "b"),
    navbarMenu(
      "More letters", 
      tabPanel("C", "c"),
      tabPanel("D", "d")
    )
  )
)

  observe(print(input$tabset))
}

shinyApp(ui, server)
```

### LINKED ISSUE #3321: Tab panel behavior when shiny tags are supplied
It's strange that providing HTML tags to `...` inside `navbarPage()` 'works' in the sense that it creates an empty nav item that doesn't control anything, and also includes the tag on every tab. We should throw a warning in this case, and point to `header`/`footer` as a more official/better way to include content on each tab

```r
shinyApp(
  navbarPage(
    "Title",
    div("other stuff 1"),
    tabPanel("a", "tab a"),
    div("other stuff 2"),
    tabPanel("b", "tab b"),
    div("other stuff 3")
  ),
  function(input, output) { }
)
```

### LINKED ISSUE #3322: Make it easier to include header/footer content in tabsetPanel()/navlistPanel()
`navbarPage()` has `header`/`footer` arguments for including content that should appear on every 'page'. This is a quite useful feature that is missing from `tabsetPanel()` and `navlistPanel()`. For example, if I wanted to add a link break bewteen the nav and content, I currently have to do that on every tab:

```r
tabsetPanel(
  tabPanel("a", br(), "a"),
  tabPanel("b", br(), "b")
)
```

It'd be much easier to do:

```r
tabsetPanel(
  header = br(),
  tabPanel("a",  "a"),
  tabPanel("b", "b")
)
```

This would also open the possibility of including inputs/outputs that should appear on every tab:

```r
tabsetPanel(
  header = textInput("foo", "bar"),
  tabPanel("a",  "a"),
  tabPanel("b", "b")
)
```

cpsievert:
This PR closes numerous issues with `tabsetPanel()`/`navlistPanel()`/`navbarPage()`. See the respective issue links for examples:

1. `!!!` may now be used to splice `tabPanel()`s into them (closes #2927)
2. The `tabsetPanel()` and `navlistPanel()` functions gain `header`/`footer` arguments, akin to the already existing `header`/`footer` arguments in `navbarPage()`, making it easier to include content that should appear on every tab (closes #3322). 
3. `NULL` values are now dropped instead of producing an empty nav item (closes #1928).
4. More informative error when non-`tabPanel()`/`shiny.tag` objects are supplied to `...` (closes #3313, closes #1823).
5. New informative warning when `shiny.tag` object(s) are supplied to `...`. In this case we will continue to create an "empty" nav item and include the content on every tab, but the warning will mention the (new) `header`/`footer` args, which is likely what the user wants (closes #3321).
6. Bootstrap 4+ HTML/CSS markup is now produced when used with `theme = bslib::bs_theme()` (closes #3320). At least for now, you'll still want `{bslibs}`'s BS3 compatibility layer to provide sensible styles in `navlistPanel()` and `navbarPage()`, but perhaps in follow up PRs we can explore completely removing the compatibility layer dependency for those as well. The main downside to this change in behavior is that `tabsetPanel()` now returns `tagFunction()`s, meaning that user code which makes assumptions about the return value may break, but hopefully by providing more intelligent tag modification functions (see, for example, https://github.com/rstudio/htmltools/pull/202), this won't be much of an issue. 

### Testing notes

Install `remotes::install_github("rstudio/shiny")` then verify there is no regression in the example app above as well as the one below (in the app below, when you click add 'dynamic' tab, a tab should be inserted between Static1 and Static2 in the dropdown)

```r
library(shiny)
library(bslib)

ui <- fluidPage(
  theme = bs_theme(),
  sidebarLayout(
    sidebarPanel(
      actionButton("add", "Add 'Dynamic' tab"),
      actionButton("remove", "Remove 'Foo' tab")
    ),
    mainPanel(
      tabsetPanel(
        id = "tabs",
        tabPanel("Hello", "This is the hello tab"),
        tabPanel("Foo", "This is the foo tab"),
        navbarMenu(
          "Static", 
          tabPanel("Static 1", "Static 1", value = "s1"),
          tabPanel("Static 2", "Static 2", value = "s2")
        )
      )
    )
  )
)
  observeEvent(input$add, {
    insertTab(
      inputId = "tabs",
      tabPanel("Dynamic", "Dynamic"),
      target = "s2"
    )
  })
  observeEvent(input$remove, {
    removeTab(inputId = "tabs", target = "Foo")
  })
}

shinyApp(ui, server)
```


R/bootstrap.R
```diff
@@ -953,56 +959,155 @@ buildTabset <- function(tabs, ulClass, textFilter = NULL, id = NULL,
...
+        # (or else the .tab('show') method won't update the .active classes properly)
+        items <- tabset$navList
+        items$children <- lapply(items$children, function(x) {
+          # x should be a tagFunction() due to the else block below
+          x <- if (inherits(x, "shiny.tag.function")) x() else x
+          # Replace <li class="nav-item"><a class="nav-link"></a></li>
+          #    with <a class="dropdown-item"></a>
+          navItem <- x$children[[1]]
+          navItem$attribs$class <- "dropdown-item"
+          navItem
```
cpsievert, line 995: The main unfortunate thing about all of this is that BS4 doesn't support `<li><a></a></li>` inside of dropdowns, which apparently they brought back in BS5

https://getbootstrap.com/docs/4.6/components/navs/#tabs-with-dropdowns
https://getbootstrap.com/docs/5.0/components/navs-tabs/#tabs-with-dropdowns

Might be worth asking for a backport into BS4, but I doubt we'll see it

R/bootstrap.R
```diff
@@ -938,10 +944,10 @@ buildTabset <- function(tabs, ulClass, textFilter = NULL, id = NULL,
             tabs = tabs, textFilter = textFilter)
 
   tabNavList <- tags$ul(class = ulClass, id = id,
-                  `data-tabsetid` = tabsetId, lapply(tabs, "[[", 1))
+                  `data-tabsetid` = tabsetId, !!!lapply(tabs, "[[", "liTag"))
 
   tabContent <- tags$div(class = "tab-content",
-                  `data-tabsetid` = tabsetId, lapply(tabs, "[[", 2))
+                  `data-tabsetid` = tabsetId, !!!lapply(tabs, "[[", "divTag"))
```
cpsievert, line 950: Should we consider extracting non-tabPanel() content and putting it above/below the `tab-content`?

R/bootstrap.R
```diff
@@ -953,56 +959,168 @@ buildTabset <- function(tabs, ulClass, textFilter = NULL, id = NULL,
...
+  if (isTabPanel(divTag)) {
+    return(buildNavItem(divTag, tabsetId, index))
+  }
+
+  warning(
+    "Expected a collection tabPanel()s",
+    if (is.null(textFilter)) " and tabPanelMenus().",
+    if (!is.null(textFilter)) ", tabPanelMenus(), and/or character strings.",
+    call. = FALSE
+  )
```
cpsievert, line 988: Note that in `navbarPage()` we actually (un-officially) support `shiny.tag` `divTag` objects

R/bootstrap.R
wch: Here's an example that we should think about -- what _should_ happen with code like this?

```R
shinyApp(
  navbarPage(
    "Title",
    div("other stuff 1"),
    tabPanel("a", "tab a"),
    div("other stuff 2"),
    tabPanel("b", "tab b"),
    div("other stuff 3")
  ),
  function(input, output) { }
)
```

cpsievert: I decided to revert back to supporting this off-label usage with a warning in https://github.com/rstudio/shiny/pull/3315/commits/2e33634726004fe6f16830c8a34a0effaef5b0eb

(Thankfully this has only ever worked with tags, so we can throw we a better message in any other case)

R/bootstrap.R
```diff
@@ -757,18 +777,39 @@ tabsetPanel <- function(...,
...
+          return(NULL)
+        }
+        warning(
+          "`tabsetPanel(card = TRUE)` requires Bootstrap 4 or higher, ",
+          "so the app has been upgraded from Bootstrap 3 to 4. ",
+          "To remove this warning, either supply `theme = bslib::bs_theme()` ",
+          "to the app's page layout or set `card = FALSE`.",
+          call. = FALSE
+        )
+        bootstrapLib(bslib::bs_theme(version = 4))
```
cpsievert, line 801: This is probably a little too magical -- any tags that would appear above this `tagFunction()` would end up generating BS3 markup....

R/shiny-package.R
```diff
@@ -14,7 +14,7 @@
 #'    quo enquo as_function get_expr get_env new_function enquos
 #'    eval_tidy expr pairlist2 new_quosure enexpr as_quosure is_quosure inject
 #'    enquos0 zap_srcref %||% is_na
-#'    is_false
+#'    is_false list2 abort
```
cpsievert, line 17: No longer using

```suggestion
#'    is_false list2
```

R/html-deps.R
```diff
@@ -50,7 +50,7 @@ processDeps <- function(tags, session) {
   names(dependencies) <- NULL
 
   list(
-    html = doRenderTags(ui),
+    html = doRenderTags(tagList(ui)), # TODO: doRenderTags() shouldn't fail on a tagFunction()
```
cpsievert, line 53: https://github.com/rstudio/htmltools/pull/204 fixes this issue more generally

R/bootstrap.R
```diff
@@ -949,56 +964,173 @@ buildTabset <- function(tabs, ulClass, textFilter = NULL, id = NULL,
...
+      `data-toggle` = "tab",
+      `data-value` = value,
+      icon,
+      title
+    )
+  )
+}
+
+# TODO: something like this should exist in htmltools
+tagHasClass <- function(x, class) {
```
wch, line 1127: The complexity of this code inspired this issue:
https://github.com/rstudio/htmltools/issues/205

tests/testthat/test-tabPanel.R
```diff
@@ -1,4 +1,87 @@
...
+  navbarPage(...)
+}
+tabset_panel <- function(...) {
+  withPrivateSeed(set.seed(100))
+  tabsetPanel(...)
+}
+
+expect_snapshot2 <- function(...) {
+  if (getRversion() < "3.6.0") {
+    skip("Skipping non-deterministic snapshots on R < 3.6")
```
wch, line 19: This message confused me for a bit -- I think the default RNG method was changed in 3.6, but it was still deterministic.

```suggestion
    skip("Skipping snapshots on R < 3.6 because of different RNG method")
```

PR #3328, opened 2021-03-09
Allow trailing commas in more places

hadley:
I grepped for list(...) and replaced with rlang::list2(...). This also enables !!! which is generally not important for Shiny because it automatically splices lists/tagLists, but I doubt it will affect any existing code.

Let me know if you want me to document this, or add a news bullet.


cpsievert:
Since there was some overlap with messaging for #3315 I took a stab at a NEWS item (and fixed the merge conflict).

cpsievert (approved):

NEWS.md
```diff
@@ -9,6 +9,8 @@ shiny 1.6.0.9000
 
 ### New features and improvements
 
+* All uses of `list(...)` have been replaced with `rlang::list2(...)`. By effect, `{rlang}`'s `!!!` operator may now be used to "splice" a list of argument values into `...`. We think this'll be particularly useful for passing a list of `tabPanel()` to their consumers (i.e., `tabsetPanel()`, `navbarPage()`, etc). For example, `tabs <- list(tabPanel("A", "a"), tabPanel("B", "b")); navbarPage(!!!tabs)`. (#3315 and #3328)
```
hadley, line 12:
```suggestion
* All uses of `list(...)` have been replaced with `rlang::list2(...)`. This means that you can use trailing `,` with out warning and use rlang's `!!!` operator to splice" a list of argument values into `...`. We think this'll be particularly useful for passing a list of `tabPanel()` to their consumers (i.e., `tabsetPanel()`, `navbarPage()`, etc). For example, `tabs <- list(tabPanel("A", "a"), tabPanel("B", "b")); navbarPage(!!!tabs)`. (#3315 and #3328)
```

Is that true about splicing though? I thought `tagList()` already did auto-splicing?

cpsievert: Yea, I believe all of htmltools uses `dots_list()` now. I couldn't find any mention of it in the NEWS though so I figured it wouldn't hurt to say it just generally works now with any `...`?

cpsievert: Also, shouldn't the suggestion be `without error` instead of `with out warning`?

hadley: Ooops yeah. But I meant doesn't splicing already happen without `!!!`?

cpsievert: Ohh, technically no, but in practice the difference usually important

```r
> str(tagList(list(tags$a(), tags$div())))
List of 1
 $ :List of 2
  ..$ :List of 3
  .. ..$ name    : chr "a"
  .. ..$ attribs : Named list()
  .. ..$ children: list()
  .. ..- attr(*, "class")= chr "shiny.tag"
  ..$ :List of 3
  .. ..$ name    : chr "div"
  .. ..$ attribs : Named list()
  .. ..$ children: list()
  .. ..- attr(*, "class")= chr "shiny.tag"
 - attr(*, "class")= chr [1:2] "shiny.tag.list" "list"
> str(tagList(!!!list(tags$a(), tags$div())))
List of 2
 $ :List of 3
  ..$ name    : chr "a"
  ..$ attribs : Named list()
  ..$ children: list()
  ..- attr(*, "class")= chr "shiny.tag"
 $ :List of 3
  ..$ name    : chr "div"
  ..$ attribs : Named list()
  ..$ children: list()
  ..- attr(*, "class")= chr "shiny.tag"
 - attr(*, "class")= chr [1:2] "shiny.tag.list" "list"
```

NEWS.md
```diff
@@ -9,6 +9,8 @@ shiny 1.6.0.9000
 
 ### New features and improvements
 
+* All uses of `list(...)` have been replaced with `rlang::list2(...)`. This means that you can use trailing `,` without error and use rlang's `!!!` operator to "splice" a list of argument values into `...`. We think this'll be particularly useful for passing a list of `tabPanel()` to their consumers (i.e., `tabsetPanel()`, `navbarPage()`, etc). For example, `tabs <- list(tabPanel("A", "a"), tabPanel("B", "b")); navbarPage(!!!tabs)`. (#3315 and #3328)
```
hadley, line 12:
```suggestion
* All uses of `list(...)` have been replaced with `rlang::list2(...)`. This means that you can use trailing `,` without error and use rlang's `!!!` operator to "splice" a list of argument values into `...`. (#3315 and #3328)
```

Since it doesn't provide any new benefit, I think it's better to not over advertise it.

cpsievert: `!!!` inside `tabsetPanel()` is new (was merged earlier today), but `!!!` inside `tagList()` isn't.

That said, I'd be OK with removing it, but I'll leave that up to @wch

wch (approved):

PR #3329, opened 2021-03-09
Improve tab layout documentation

hadley:
Nothing major here, just a bunch more cross-linking and some extra consistency in argument and return value documentation.

### DIFF
--- a/R/bootstrap.R
+++ b/R/bootstrap.R
@@ -388,11 +388,9 @@ collapseSizes <- function(padding) {
 #'   menu within the navbar that in turns includes additional tabPanels (see
 #'   example below).
 #'
-#' @seealso [tabPanel()], [tabsetPanel()],
-#'   [updateNavbarPage()], [insertTab()],
-#'   [showTab()]
-#'
+#' @seealso [updateNavbarPage()], [insertTab()], [showTab()]
 #' @family layout functions
+#' @family tab layouts
 #'
 #' @examples
 #' navbarPage("App Title",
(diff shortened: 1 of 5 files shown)


R/bootstrap.R
```diff
@@ -772,29 +751,20 @@ tabsetPanel <- function(...,
...
-#' @param id If provided, you can use `input$`*`id`* in your
-#'   server logic to determine which of the current navlist items is active. The
-#'   value will correspond to the `value` argument that is passed to
-#'   [tabPanel()].
-#' @param selected The `value` (or, if none was supplied, the `title`)
-#'   of the navigation item that should be selected by default. If `NULL`,
-#'   the first navigation will be selected.
+#' @inheritParams tabsetPanel
+#' @param ... [tabPanel()] elements to include in the navbar.
+#'   Plain strings will be converted to headers.
```
cpsievert, line 756: This should mention that strings are only allowed in `navlistPanel()` (otherwise, it's an error)

cpsievert: It should also probably also mention that, in that case, `"---"` is rendered as a divider

R/bootstrap.R
```diff
@@ -342,20 +342,13 @@ collapseSizes <- function(padding) {
 #' Create a page with a top level navigation bar
 #'
 #' Create a page that contains a top level navigation bar that can be used to
-#' toggle a set of [tabPanel()] elements.
+#' toggle a set of [tabPanel()] elements. `navbarMenu()` can be used to create
+#' an embedded menu within the navbar that in turns includes additional
+#' `tabPanels`.
```
cpsievert, line 347: I've always found it a bit awkward that `navbarMenu()` is documented on the same page as `navbarPage()`, even though it can be used with any tab layout. What do you think of documenting it alongside`?tabPanel` ?

R/bootstrap.R
hadley: Isn't `----` deprecated?

R/bootstrap.R
hadley: Yeah, I think that's an improvement.

PR #3333, opened 2021-03-11
When refreshing a stylesheet, schedule a report CSS values once the sheet is loaded

cpsievert:
Should fix timing issues with https://connect.rstudioservices.com/explore_your_weather


srcts/src/main.ts
```diff
@@ -4061,6 +4062,17 @@ function main(): void {
           $(sheet.ownerNode).remove();
         };
 
+        // There doesn't appear to be proper way to wait until new styles have been
+        // _applied everywhere_ so we repeatedly call bindAll() every .1s for 10 seconds.
+        // (bindAll() is used as an indirect way of calling doSendTheme() which resends CSS
+        // info stored in input values accessible to shiny::getCurrentOutputInfo()).
+        /* global Shiny */
+        let scheduleCssReporter = function () {
+          let handle = setInterval(Shiny.bindAll, 100);
```
cpsievert, line 4071: Instead of `Shiny.bindAll` lets just send the CSS info (`sendImageSize`) and throttle calls to it.

wch:
For future reference, with this PR, when I load a Shiny app, it calls `sendImageSize()` 403 times.

image


Changing the app from light to dark mode results in about 400 more calls to that function.

wch:
It looks like the debouncing reduces the number of calls to `doSendImageSize()` to 101 times when the app is loaded. But that still seems pretty expensive, considering that each call to `doSendImageSize()` does a fair bit of work, and triggers `shiny:visualchange`:

https://github.com/rstudio/shiny/blob/16e0d9e355713e6765d69c9203cb29ee2af4eeb7/srcts/src/main.ts#L6843-L6878

wch:
@cpsievert I recall that there can be a delay between when the CSS is loaded and when it is actually applied. I also remember that the delay can vary. Do you remember why it can take so long -- up to 10 seconds?

I'm considering the following:

* Use a MutationObserver to re-send image size information when an image/plot output changes size (instead of listening for the `window` `resize` event). This would also trigger the `shiny:visualchange` event. We'd need some sort of workaround for IE11, but it wouldn't need to be perfect.
* To detect CSS style changes is trickier, since there's no observer that detects those changes. I wonder if, in the `onload` call back for CSS file, we can set some arbitrary CSS style property via JS, and then trigger a callback after that is applied. If the CSS style set by JS always takes effect after the file-based CSS styles, that could work. But we'll have to test to make sure.

cpsievert:
> Do you remember why it can take so long -- up to 10 seconds?

10 seconds is a somewhat arbitrary decision. Same with the 0.1sec interval (which combined is leading to the 101 calls to `doSendImageSize()`). Although, it's important to keep the interval somewhat small so that you're not waiting forever to get an update. Maybe it's make sense to implement some sort of exponential back-off and/or clear the interval if the value has changed (I'd probably need help implementing the later)

wch:
With the new method I've implemented, `doSendImageSize` gets called 3 times on load, and 2 times when switching styles.

image

srcts/src/main.ts
```diff
@@ -4051,8 +4051,10 @@ function main(): void {
             const newStyle = $("<style>").attr("id", id).html(xhr.responseText);
 
             $head.append(newStyle);
-            setTimeout(() => oldStyle.remove(), 500);
-            setTimeout(() => removeSheet(oldSheet), 500);
+
+            oldStyle.remove();
+            removeSheet(oldSheet);
+            sendImageSize();
```
cpsievert, line 4057: Would be good to have a comment here about this working because (as far as we know), in this case, styles are applied syncronously

srcts/src/main.ts
```diff
@@ -4091,24 +4093,61 @@ function main(): void {
...
+            // that they are loaded. This seems to be true in the browsers we
+            // have tested.
             link.attr("onload", () => {
-              setTimeout(() => removeSheet(oldSheet), 500);
+              const dummy_id = "dummy-" + Math.floor(Math.random() * 999999999);
+              const css_string =
+                "#" + dummy_id + " { " +
+                "color: #a7c920; " +       // An arbitrary color for the transition
+                "transition: 0.1s all; " +
+                "visibility: hidden; " +
```
cpsievert, line 4123: If we wanted to be extra careful, maybe add the `!important` here as well?

srcts/src/main.ts
```diff
@@ -4091,24 +4093,61 @@ function main(): void {
...
+              const base64_css_string =
+                "data:text/css;base64," + btoa(css_string);
+
+              let $dummy_link = $("<link rel='stylesheet' type='text/css' />");
+              $dummy_link.attr("href", base64_css_string);
+
+              let $dummy_el = $("<div id='" + dummy_id + "'></div>")
+              $dummy_el.one("transitionend", () => {
+                $dummy_el.remove();
+                $dummy_link.remove();
```
cpsievert, line 4136: I think we should be using `removeSheet()` here -- this won't necessarily disable the stylesheet

cpsievert:
LGTM once merge conflicts are addressed, thanks!

wch (approved):

PR #3334, opened 2021-03-12
bootstrapPage() now includes jQuery so static rendering works as expected

### LINKED ISSUE #3316: tabsetPanel()s don't work outside of shiny
```r
fluidPage(
  tabsetPanel(
    tabPanel("A", "a"),
    tabPanel("B", "b")
  )
) %>% htmltools::browsable()
```

Doesn't work because jQuery doesn't get included

```
bootstrap.min.js:6 Uncaught Error: Bootstrap's JavaScript requires jQuery
    at bootstrap.min.js:6
bootstrap-accessibility.min.js:1 Uncaught ReferenceError: $ is not defined
    at Object.parcelRequire.aQMl (bootstrap-accessibility.min.js:2)
    at f (bootstrap-accessibility.min.js:1)
    at parcelRequire.aQMl (bootstrap-accessibility.min.js:1)
    at bootstrap-accessibility.min.js:1
```

This works though because `bs_theme()` bundles jQuery

```r
fluidPage(
  theme = bslib::bs_theme(),
  tabsetPanel(
    tabPanel("A", "a"),
    tabPanel("B", "b")
  )
) %>% htmltools::browsable()
```

cpsievert:
Closes #3316

### DIFF
--- a/R/shinyui.R
+++ b/R/shinyui.R
@@ -47,27 +47,8 @@ renderPage <- function(ui, showcase=0, testMode=FALSE) {
     )
   }
 
-  jquery <- function() {
-    version <- getOption("shiny.jquery.version", 3)
-    if (version == 3) {
-      return(htmlDependency(
-        "jquery", version_jquery,
-        c(href = "shared"),
-        script = "jquery.min.js"
-      ))
-    }
-    if (version == 1) {
-      return(htmlDependency(
-        "jquery", "1.12.4",
-        c(href = "shared/legacy"),
-        script = "jquery.min.js"
-      ))
-    }
-    stop("Unsupported version of jQuery: ", version)
-  }
-
   shiny_deps <- c(
-    list(jquery()),
+    list(jqueryDependency()),
     shinyDependencies()
   )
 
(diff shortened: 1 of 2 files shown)


R/shinyui.R
```diff
@@ -82,6 +63,27 @@ renderPage <- function(ui, showcase=0, testMode=FALSE) {
   enc2utf8(paste(collapse = "\n", html))
 }
 
+jqueryDependency <- function() {
+  version <- getOption("shiny.jquery.version", 3)
+  if (version == 3) {
+    return(htmlDependency(
+      "jquery", version_jquery,
+      src = "www/shared",
```
wch, line 71: Do we want to do something like:

```R
src = c(file=filepath, href=url)
```

There are some cases where `file` is used and others where `href` is used, but I can't remember the details -- we should find that out before we make the change.

cpsievert: Yea, good point. By including that, I was able to confirm the `href` is preferred when rendering via `shinyApp()` (BTW, `bootstrapLib()` also defines both `file` and `href`)

wch (approved):

PR #3341, opened 2021-03-22
Add card argument to tabsetPanel()

cpsievert:
Adds a `card` argument to `tabsetPanel()`, which when set to `TRUE`, [puts the tabs/pills in the header of a BS4 card component](https://getbootstrap.com/docs/4.1/components/card/#navigation). This helps to easily add padding between the nav and the content while also reducing the appearance of the tabs/content floating in space.


cpsievert:
Closing in favor of #3343.

PR #3343, opened 2021-03-23
Proof of concept for card(), getNavList(), and getNavContent()

cpsievert:
An alternative approach to achieving https://github.com/rstudio/shiny/pull/3341 (i.e., placing a tabset within a `card()`), but offers two main benefits:

1. A new `card()` function provides a more general `wellPanel()`-like function (has to be used with `bslib::bs_theme()`).
2. New generic `getNavList()` and `getNavContent()` functions to extract the nav/content containers. This could come in handy for `navtreePanel()`/`navlistPanel()`, especially when considering how folks want to control layout with CSS grid instead of Bootstrap's grid .

The main downside to this approach (compared to #3341) is that `card = TRUE` is easier/more discoverable

```r
tabs <- tabsetPanel(
  tabPanel("A", "a"),
  tabPanel("B", "b")
)
ui <- fluidPage(
  theme = bslib::bs_theme(),
  card(
    getNavContent(tabs),
    header = getNavList(tabs)
  )
)

if (interactive()) {
}
```


NAMESPACE
```diff
@@ -77,6 +79,7 @@ export(brushOpts)
 export(brushedPoints)
 export(callModule)
 export(captureStackTraces)
+export(card)
```
cpsievert, line 82: Move this to bslib

PR #3357, opened 2021-04-07
Make external libs builder leveraging esbuild

schloerke:
Should minifying the altered external libraries.


srcts/package.json
```diff
@@ -46,6 +46,7 @@
     "setup_build_shiny": "yarn run lint && yarn run typescript-check",
     "build_shiny": "yarn run setup_build_shiny && yarn run bundle_shiny",
```
cpsievert, line 47: And then the `tools` script could stay with using `yarn build`?


```suggestion
    "build_shiny": "yarn run setup_build_shiny && yarn run bundle_shiny && yarn bundle_external_libs",
```

cpsievert (approved):

srcts/package.json
schloerke: I added it into `yarn build`, rather than `yarn build_shiny`

PR #3361, opened 2021-04-08
Don't double-wrap quosure when quoted=TRUE

### LINKED ISSUE #3360: quosures don't work with quoted = TRUE
For example 

```r
library(shiny)
ui <- fluidPage(textOutput("txt1"))
  x <- "hello"
  my_quo <- rlang::quo(x)
  output$txt1 <- renderText(my_quo, quoted = TRUE)
}
shinyApp(ui, server)
```

```r
Error in cat: argument 1 (type 'language') cannot be handled by 'cat'
```

cpsievert:
Closes #3360

### DIFF
--- a/R/utils-lang.R
+++ b/R/utils-lang.R
@@ -94,7 +94,9 @@ get_quosure <- function(x, env, quoted) {
     if (!quoted) {
       x <- eval(substitute(substitute(x)), parent.frame())
     }
-    x <- new_quosure(x, env)
+    if (!rlang::is_quosure(x)) {
+      x <- new_quosure(x, env)
+    }
 
   } else {
     x <- eval(substitute(substitute(x)), parent.frame())
--- a/R/utils-lang.R
+++ b/R/utils-lang.R
@@ -94,7 +94,7 @@ get_quosure <- function(x, env, quoted) {
     if (!quoted) {
       x <- eval(substitute(substitute(x)), parent.frame())
     }
-    if (!rlang::is_quosure(x)) {
+    if (!is_quosure(x)) {
       x <- new_quosure(x, env)
     }
 


wch:
I think that if a user passes a quosure and uses `quoted=TRUE`, it should throw an error/warning, and not just accept the quosure. If the user really wants to pass a quosure to the reactive/observer/render function, that's what `inject()` is for.


Here's the example app from #3360:

```R
library(shiny)
ui <- fluidPage(textOutput("txt1"))
  x <- "hello"
  my_quo <- rlang::quo(x)
  output$txt1 <- renderText(my_quo, quoted = TRUE)
}
shinyApp(ui, server)
#> Error in cat: argument 1 (type 'language') cannot be handled by 'cat'
```

The goal in the long run is to deprecate the `quoted` and `env` arguments, instead of expanding what `quoted=TRUE` allows you to pass in. So I think that if `quoted==TRUE`, it should make sure `x` is a language object and NOT a quosure. Something like this:

```R
if (!is.language(x)) || is.quosure(x)) {
  stop( ... )
}
```

jcheng5:
@wch Thanks for the clarification, I'm totally fine with `rlang::inject` instead of `quoted = TRUE`.

Unfortunately, for render functions, `rlang::inject` doesn't work either:

```r
> quoA <- rlang::quo(1)
> quoB <- rlang::quo(1 + !!quoA)
> rlang::inject(shiny::renderText(!!quoB))()
Error: Base operators are not defined for quosures.
Do you need to unquote the quosure?

  # Bad:
  lhs + myquosure

  # Good:
  lhs + !!myquosure
```

Note that this *isn't* a problem for reactives and observers:

```r
> shiny::isolate(rlang::inject(shiny::reactive(!!quoB))())
2
```

@cpsievert and I tracked it down to the fact that only render functions use `quoToFunction`, which uses `rlang::new_function`; whereas reactives/observers use `rlang::as_function`, and these seem to have subtly different behaviors for reasons I don't understand.

```r
# This shows the difference between as_function and new_function;
# the former can handle nested quosures, the latter cannot.

quoA <- rlang::quo(1)
quoB <- rlang::quo(1 + !!quoA)

func1 <- rlang::as_function(quoB)
func1()
func2 <- rlang::new_function(NULL, rlang::quo_get_expr(quoB), rlang::quo_get_env(quoB))
func2()
```

I *think* that implementing `quoToFunction` with `rlang::as_function` will fix the problem but we have not tried it.

cpsievert:
Closing since #3373 fixes the real issue we were dealing with in shinymeta

Issue #1828, opened 2017-08-25
Internationalization: custom disconnect message?


nicholaelaw:
I asked this in the discussion group and got no answer yet. I couldn't find the origin in sources neither. 

Is it possible to change the notification after disconnecting? I mean this thing here:

![image](https://user-images.githubusercontent.com/10719482/29705200-9276ec10-89af-11e7-8e31-12885e71be4f.png)

This is essentially the first thing the user will see when returning to the app after a while of inactivity. Particularly so on mobile devices where the browser could not maintain connection when screen is turned off. 

**tl;dr**: I want to make this message 1) readable by people who don't know English (translation), and 2) a bit more friendly (custom HTML, perhaps. Pictures of cats. Anything other than a "this isn't working" screen.).

jcheng5:
It's not currently customizable but that is something we should do. It would actually be a change to the server software (Shiny Server, Connect, and ShinyApps.io).

nicholaelaw:
Well then should I submit a feature request in Shiny Server repo? Is this feature under consideration?

tcash21:
Also agree this would be hugely helpful. Any updates or workarounds we can use in the meantime. Just want to show the user something a bit informative with a message on what to do next.

michaelhogersnplm:
@tcash21 an alternative is to use the shiny:disconnected state, the below may require some extra tinkering. I recommend to use JavaScript to toggle the visibility of a modal for example, say ```$("#modalIdentifier").toggle()```. 

```
$(document).on('shiny:disconnected', function(event) {
// use JavaScript code here to show a customized pop-up
// ```$("#modalIdentifier").toggle()```
});
```

daattali:
This isn't a perfect solution, but it's what I've been using for years. This is an example of what one of my apps looks like when disconnected:

![image](https://user-images.githubusercontent.com/952340/83962495-e8dd3680-a86b-11ea-84db-60f823789024.png)

I achieved it by adding the following CSS: (you can customize the text or the colours and style)

```
#ss-connect-dialog {
  position: absolute;
  top: 50px;
  bottom: auto;
  left: 50%;
  margin-left: -150px;
  font-size: 20px;
  width: 300px;
  background: #fff;
  color: #444;
  padding: 20px 30px;
  text-align: center;
}
#ss-connect-dialog::before {
  content: 'An error occurred. Please refresh the page and try again.';
}
#ss-connect-dialog label, #ss-connect-dialog a {
  display: none;
}
```

[...]

PR #3365, opened 2021-04-12
Allow to specify disconnect overlay template

dseynaev:
This PR adds an overridable `<template>` that, when present, is used to fill the grey-out div when it is generated.

This can be used to customize the content of the disconnect overlay (without the need for extra JS from the user).

This is a POC implementation of my suggestion in https://github.com/rstudio/shiny/issues/1828

The only downside is that this does not properly support IE (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template#html).


wch:
Can you provide an example of how the application author would use this?

dseynaev:
I've been using the following example for testing:

```
shiny::shinyApp(
    ui = function(request) {	
      shiny::tags$body(
          shiny::fluidPage(),
          shiny::tags$template(
              id = "shiny-disconnected-overlay-template",
              shiny::tags$div("some content")
          )
      )
    },
      stop("something happened")
    }
)
```

wch:
If I understand the code correctly, wouldn't this also work (without the call to `body()`)?

```R
shinyApp(
  ui = fluidPage(
    "Page content",
    tags$template(
      id = "shiny-disconnected-overlay-template",
      tags$div("some content")
    )
  ),
    stop("something happened")
  }
)
```


And I don't think the changes in shinyui.R are needed.

dseynaev:
that would also work. I put the `<template>` under `<body>` to have it at the same level as the `shiny-disconnected-overlay` `div (seemed to be the general practice, but the template spec is flexible enough), but this can be totally left to application author.

point taken about the changes to `shinyui.R`; removed them

dseynaev:
rebased to resolve conflicts

@wch anything else I can do from my side? ci was reporting a 403 trying to access my forked repo; not sure how to resolve that: it's public

PR #3372, opened 2021-04-19
Reduce complexity and 'black-boxed' nature of tab panel logic

cpsievert:
(To be merged after https://github.com/rstudio/htmltools/pull/215)

In #3315, a lot of complexity was added to `buildTabset()` (and the JS behind `insertTab()`, etc) due to [this Bootstrap 4+ regression](https://github.com/twbs/bootstrap/issues/33625) (at the time, I thought it was a "known" regression, but apparently it was an oversight). That regression has already been fixed in `{bslib}`'s `version = 4` (via https://github.com/rstudio/bslib/pull/306), hence the bump in `{bslib}`'s version requirement.


R/bootstrap.R
```diff
@@ -1026,119 +1026,72 @@ buildNavItem <- function(divTag, tabsetId, index) {
...
-          # Replace <li class="nav-item"><a class="nav-link"></a></li>
-          #    with <a class="dropdown-item"></a>
-          if (tagHasClass(x, "nav-item")) {
-            x <- x$children[[1]]
-            x$attribs$class <- "dropdown-item"
-          }
-          x
-        })
-        bs4NavItemDropdown(title, value, icon, active, items)
+    liTag = tagRenderHook(
```
wch, line 1029: Should be changed to `tagAddRenderHook` here and elsewhere.

R/bootstrap.R
```diff
@@ -1026,119 +1026,72 @@ buildNavItem <- function(divTag, tabsetId, index) {
...
-            x <- x$children[[1]]
-            x$attribs$class <- "dropdown-item"
-          }
-          x
-        })
-        bs4NavItemDropdown(title, value, icon, active, items)
+    liTag = tagAddRenderHook(
+      liTag(id, title, value, icon),
+      function(x) {
+        if (isTRUE(getCurrentVersion() >= 4)) {
```
wch, line 1032: Looking at `getCurrentVersion`, it looks like it can return a string in some cases. That function should be modified to always return a number.

Also, the name `getCurrentVersion` is confusing -- how about `getCurrentThemeVersion`?

wch (approved):

wch:
I believe that `tagQuery()` should preserve attribute order. Can you provide a minimal example where it doesn't?

cpsievert:
> I believe that tagQuery() should preserve attribute order

You're right. I checked with Barret and it should preserve HTML attribute order (it doesn't preserve ordering of the `$name`, `$attribs`, and `$children` elements, though)

I was getting confused by the fact that, in the old behavior, active vs non-active tabs will have different ordering of their attributes. If you think it's important we maintain that (inconsistent) ordering, I can do it, but it'll introduce more complexity

wch (approved):

cpsievert:
Merging since the (new) unit tests failures are caused by dev rlang, which will be addressed by #3379 (or in rlang itself)

PR #3373, opened 2021-04-22
Fix rlang::inject with render functions

### LINKED ISSUE #3374: rlang::inject() with renderFunction() don't work with nested quosures
```r
> quoA <- rlang::quo(1)
> quoB <- rlang::quo(1 + !!quoA)
> rlang::inject(shiny::renderText(!!quoB))()
Error: Base operators are not defined for quosures.
Do you need to unquote the quosure?

  # Bad:
  lhs + myquosure

  # Good:
  lhs + !!myquosure
```

As explained in more detail here https://github.com/rstudio/shiny/pull/3361#issuecomment-820672180, this isn't a problem for `reactive()` or `observe()` since those use `as_function()` instead of `new_function()`.

jcheng5:
Closes #3374

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -539,6 +539,9 @@ installExprFunction <- function(expr, name, eval.env = parent.frame(2),
 quoToFunction <- function(q, label, ..stacktraceon = FALSE) {
   q <- as_quosure(q)
   func <- as_function(q)
+  # as_function returns a function that takes `...`. We want one that takes no
+  # args.
+  formals(func) <- list()
   wrapFunctionLabel(func, label, ..stacktraceon = ..stacktraceon)
 }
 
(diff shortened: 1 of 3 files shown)


wch (approved):
Looks good! Can you add a NEWS item?

wch (approved):

R/utils.R
```diff
@@ -538,9 +538,7 @@ installExprFunction <- function(expr, name, eval.env = parent.frame(2),
 #' @export
 quoToFunction <- function(q, label, ..stacktraceon = FALSE) {
   q <- as_quosure(q)
-  # Use new_function() instead of as_function(), because as_function() adds an
-  # extra parent environment. (This may not actually be a problem, though.)
-  func <- new_function(NULL, get_expr(q), get_env(q))
+  func <- as_function(q)
```
wch, line 541: Note that this is causing a test failure because the old code creates a function with the signature `function()`, but the new code creates a function with the signature `function(...)`. I don't think that the added `...` currently cause any problems right now (other than the test failure), but in order to keep the code tighter, I think we should remove them. I can make the change.

PR #3381, opened 2021-05-05
glyphicons hyperlink should be pinned to Bootstrap 3 docs

cpsievert:
The current link (confusingly) links to the current version of Bootstrap, which has dropped glyphicons


cpsievert:
Actually #3302 will fix this

PR #3379, opened 2021-05-04
Install dev version of rlang

cpsievert:
Mainly to demonstrate that the CI failure in #3372 is introduced by the dev version of rlang

cc @schloerke

PR #3382, opened 2021-05-05
Make sure dev version of rlang is available

schloerke:
Followup to #3379


cpsievert (approved):

cpsievert (changes requested):
move the `options()` to `local_options()`

cpsievert (approved):

PR #3385, opened 2021-05-07
fix bookmarking bug #2297: don't break all bookmarking system if some URL params don't parse correctly

daattali:


daattali:
Another attepmt to fix this issue, after https://github.com/rstudio/shiny/pull/2770 was unsuccessful

daattali:
As a refresher: URL query params not parsing "correctly" can happen for a variety of reasons that don't mean something went wrong. For example, when sharing a link on facebook, facebook adds a query param that does not have quotes around it, so the json parser fails on it

schloerke (approved):

wch (approved):

Issue #3386, opened 2021-05-07
"213-tab-panels": Console error when adding Foo tab


MadhulikaTanuboddi:
1. Run https://github.com/rstudio/shinycoreci-apps/blob/master/apps/213-tab-panels/app.R 
2. Open the dev console
3. Click on "Add New Foo tab" button
4. Notice the Console error

<img width="1699" alt="213-tabpanels-AddFoo-Console-error" src="https://user-images.githubusercontent.com/5369655/117494366-140f5080-af29-11eb-88a9-915f6aa3b5e4.png">

PR #3387, opened 2021-05-10
Follow up to #3372: fix oversight in refactor

cpsievert:
Fixes https://github.com/rstudio/shiny/issues/3386


schloerke (approved):
I just reverted the change made in Rituals.yaml that will fix it. No need to merge if you don't want to.

PR #3388, opened 2021-05-12
Use bslib's new nav() api to implement tabPanel() and friends

### LINKED ISSUE #2310: windowTitle argument needs better default
In `titlePanel`, the `windowTitle` argument defaults to `title`. If the title is complicated, this can end up with arbitrary HTML, CSS, and script tags being inserted into the head, which can break shinyapps.io, Connect, and Shiny Server when they go to inject their SockJS scripts:

image

image

One possibility discussed with @aronatkins would be to only take the `title` as the default value if it's a string, or alternatively, walk the first level of the tree looking for strings (which would let you at least put a logo). We could emit a warning if the `title` is too complicated to extract a `windowTitle`, asking the user to provide one explicitly.

cpsievert:
https://github.com/rstudio/bslib/pull/314 moves this logic into bslib (where we will redesign and extend the API)

Closes #2310 (via `bslib::page_navbar()`)

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -114,7 +114,7 @@ Suggests:
     sass
 Remotes:
     r-lib/rlang,
-    rstudio/bslib#314,
+    rstudio/bslib,
     rstudio/htmltools
 URL: https://shiny.rstudio.com/
 BugReports: https://github.com/rstudio/shiny/issues
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -91,7 +91,7 @@ Imports:
     withr,
     commonmark (>= 1.7),
     glue (>= 1.3.2),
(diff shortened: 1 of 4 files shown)


DESCRIPTION
```diff
@@ -114,7 +114,7 @@ Suggests:
     sass
 Remotes:
     r-lib/rlang,
-    rstudio/bslib,
+    rstudio/bslib#314,
```
wch, line 117: Reminder: merge bslib PR and remove the number here.

cpsievert, line 117:
```suggestion
    rstudio/bslib,
```

wch (approved):

wch (approved):

PR #3392, opened 2021-05-14
tagify() dynamic UI before attempting to write the tags

### LINKED ISSUE #3391: processDeps() doesn't correctly resolve tags
All of dynamic UI (i.e., `insertUI()`, `uiOutput()`, etc) runs through the internal `processDeps()` function, and at the moment, it doesn't `htmltools:::tagify()` ui before sending it to the client. As a result, passing a tag-like object with an `as.tags()` method (i.e., htmlwidgets, `tagAddRenderHook()`, etc) through dynamic UI doesn't work the way you'd expect it to, for example:

```r
library(shiny)
library(leaflet)
ui <- fluidPage(uiOutput("m"))
  output$m <- renderUI(leaflet() %>% addTiles())
}
shinyApp(ui, server)
```

```r
Error in if: argument is of length zero
```

I don't think we've noticed this subtle issue before since, if you wrap the UI in a `list()` or `tagList()`, things just happen to work because `htmltools:::flattenTags()` winds up calling `as.tags()` on the contents of the list. However, this doesn't negate the fact `processDeps()` should be processing the top-level UI object with `htmltools::tagify()` (it already does this to `findDependencies()` but the `tagify()`-ing should be done up-front and passed along to `doRenderTags()`)

cpsievert:
Closes #3391 (see there for an example and explanation of the problem)

Also closes https://github.com/rstudio/shinycoreci-apps/issues/143


R/html-deps.R
```diff
@@ -40,21 +40,18 @@ createWebDependency <- function(dependency, scrubFile = TRUE) {
 
 # Given a Shiny tag object, process singletons and dependencies. Returns a list
 # with rendered HTML and dependency objects.
+# This implementation is very similar to renderTags(), but ignores
+# <head> handling (it should only be used after the user session has started)
 processDeps <- function(tags, session) {
-  ui <- takeSingletons(tags, session$singletons, desingleton=FALSE)$ui
+  tags <- getFromNamespace("tagify", "htmltools")(tags)
+  ui <- takeSingletons(tags, session$singletons, desingleton = FALSE)$ui
```
cpsievert, line 47: Note the similarity to renderTags() https://github.com/rstudio/htmltools/blob/684675/R/tags.R#L932-L934

wch (approved):

Issue #1545, opened 2017-01-18
When inserting javascript with insertUI, does it have to be wrapped in tags$head()? (Bug or feature?)


daattali:
Consider this code that adds JS code that simply shows an alert box:

```
ui <- fluidPage(
  actionButton("go", "Insert javascript with an alert")
)
server <- function(input, output, session) {
  observeEvent(input$go, {
    insertUI(
      selector = "head", where = "beforeEnd", immediate = TRUE,
      tags$script("alert('hello')")
    )
  })
}
shinyApp(ui = ui, server = server)
```

When you click the button, the javascript is added into the `<head>` tag of the DOM, but nothing happens.

If you wrap the inserted UI in `tags$head(...)`, then you do get the javascript alert to show up.

I don't know how the tags$head() works - I assumed that it simply inserts into the <head> tag, but that would mean that the original code should also work. I guess there is some extra magic that happens for tags$head() that doesn't happen when I add it manually. Is it correct that JS code (and maybe other resources?) have to be wrapped in `tags$head()` in order for them to work?

bborgesr:
This seems to be a minor bug. When you use `tags$head`, [this](https://github.com/rstudio/shiny/blob/master/inst/www/shared/shiny.js#L3396-L3404) is called. This somehow seems to avoid [this problem](http://stackoverflow.com/questions/2592092/executing-script-elements-inserted-with-innerhtml) (script elements inserted using `innerHTML` do not execute when they are inserted.)

jcheng5:
We could reimplement insertAdjacentHTML using jquery's after, before, prepend, and append functions.

daattali:
Sometimes I wonder if I'm the only person who runs into this issue bcause once a year I come across this problem, google for a solution, and always end up back in this thread :)

That's a subtle way to **ping** to see if this would perhaps make it into a future release if the "Effort: Low" label is correct

cpsievert:
> Sometimes I wonder if I'm the only person who runs into this issue 

@daattali not the only one :)

We should probably do what @jcheng5 suggested years ago, but in the meantime, a workaround is to leverage `htmlDependency()`'s `head` argument:

```r
ui <- fluidPage(
  actionButton("go", "Insert javascript with an alert")
)
server <- function(input, output, session) {
  observeEvent(input$go, {
    insertUI(
      selector = "head",
      ui = htmltools::htmlDependency(
        name = "fake-dep",
        version = "1.0",
        src = ".", package = "shiny",
[...]

PR #3395, opened 2021-05-18
Correctly render script tags defined as list() objects

### LINKED ISSUE #3345: htmlDependency used in dynamic UI breaks when `script` is provided as a named list instead of a string
Not sure if this should be filed under {htmltools} or {shiny} - I filed it here because the interaction with shiny is what's broken from what I can tell.

To use `htmlDependency` to load a javascript module, we need to use `script = list(src = "test.js", type = "module")`. This works fine when the dependency is resolved from the UI, but it seems to parse incorrectly when done dynamically. Example:

```
library(shiny)

ui <- fluidPage(
  htmltools::attachDependencies(
    "test",
    htmltools::htmlDependency(
      name = "test",
      version = "1.0.0",
      src = c(href = "https://cdn.jsdelivr.net/gh/daattali/htmldependencybug"),
      script = list(src = "test.js", type = "module")
    )
  )
)

}

shinyApp(ui, server)
```

The above code works - you'll see a javascript alert.

But if the content is loaded dynamically -- either using insertUI or using a uiOutput/renderUI -- then it breaks. Example:

```
library(shiny)

ui <- fluidPage(
  uiOutput("test")
)

  output$test <- renderUI({
    htmltools::attachDependencies(
      "test",
      htmltools::htmlDependency(
        name = "test",
        version = "1.0.0",
        src = c(href = "https://cdn.jsdelivr.net/gh/daattali/htmldependencybug"),
        script = list(src = "test.js", type = "module")
      )
    )
  })
}

shinyApp(ui, server)
```

The page loads but the javascript module isn't loaded. Looking at the page source, I can see that it's trying to load a script with src of `https://cdn.jsdelivr.net/gh/daattali/htmldependencybug/%5Bobject%20Object%5D` - this is what's causing this bug. It seems that shiny does not know how to deal with the `script` argument being a named list, even though a named list is  allowed according to the documentation.

cpsievert:
Closes #3345


srcts/src/main.ts
```diff
@@ -4155,8 +4156,23 @@ function main(): void {
...
+        const attrs = typeof x === "string" ? "src" : Object.keys(x);
+        let script = $("<script>");
+
+        for (let i = 0; i < attrs.length; i++) {
+          const nm = attrs[i];
+          const val = x[nm];
+
+          nm === "src"
+            ? script.attr("src", href + "/" + encodeURI(val))
+            : script.attr(nm, val);
```
cpsievert, line 4172: TODO: do we need to worry about security implications?

srcts/src/main.ts
```diff
@@ -4155,8 +4155,28 @@ function main(): void {
...
+        for (let i = 0; i < attrs.length; i++) {
+          const attr = attrs[i];
+          let val = x[attr];
+
+          if (attr === "src") {
+            val = href + "/" + encodeURI(val);
+          }
+          // If val isn't truthy (e.g., null), consider it a boolean attribute
+          val ? script.setAttribute(attr, val) : script.toggleAttribute(attr);
+        }
```
schloerke, line 4177: Leverage jquery for the for-loop;
`toggleAttribute()` is not IE friendly.

```suggestion
        $.map(attrs, function(val, attr) {

          if (attr === "src") {
            val = href + "/" + encodeURI(val);
          }
          // If val isn't truthy (e.g., NA), only set the key
          script.setAttribute(attr, val ? val : "")
        })
```

schloerke: https://caniuse.com/?search=toggleAttribute

schloerke:
Tested and works with @daattali 2nd example in https://github.com/rstudio/shiny/issues/3345#issue-839301034
Screen Shot 2021-05-19 at 5 45 54 PM

schloerke (approved):
LGTM given suggestion; Errors are unrelated to PR

daattali:
Do you know if I can install shiny from this PR to check for my more complex usecase that it works? `remotes::install_github("https://github.com/rstudio/shiny", ???)`

schloerke:
@daattali 

> Do you know if I can install shiny from this PR to check for my more complex usecase that it works? `remotes::install_github("https://github.com/rstudio/shiny", ???)`

```r
remotes::install_github("rstudio/shiny#3395")
```

schloerke:
@daattali Please let us know if there are any more changes needed and/or if you have more complicated reprex apps to test.

We'll merge after we receive your confirmation. 😃

daattali:
I can verify it works for my more complex usecase https://github.com/daattali/shinywc/issues/13

daattali:
(I'd also like to untactfuly hijack this PR to see if https://github.com/rstudio/shiny/issues/1545 can also be tackled as it seems potentially related)

cpsievert:
#1545 should be handled in a different PR -- it's not related to this same code path

srcts/src/main.ts
```diff
@@ -4155,8 +4155,28 @@ function main(): void {
...
+        for (let i = 0; i < attrs.length; i++) {
+          const attr = attrs[i];
+          let val = x[attr];
+
+          if (attr === "src") {
+            val = href + "/" + encodeURI(val);
+          }
+          // If val isn't truthy (e.g., null), consider it a boolean attribute
+          script.setAttribute(attr, val ? val : "");
+        }
```
wch, line 4177: I think you may be able to replace this with:

```suggestion
        for (let [attr, val] of Object.entries(x)) {
          if (attr === "src") {
            val = href + "/" + encodeURI(val);
          }
          // If val isn't truthy (e.g., null), consider it a boolean attribute
          script.setAttribute(attr, val ? val : "");
        }
```

See https://effectivetypescript.com/2020/05/26/iterate-objects/

@schloerke, @nstrayer, it would be great if you guys could weigh in on this.

schloerke: A polyfill for `Object.entries` will be required (IE / Phantom), but it is more succinct

(Changed the proposed to a `const` instead of a `let` in the for loop.)

```suggestion
        for (const [attr, val] of Object.entries(x)) {
          if (attr === "src") {
            val = href + "/" + encodeURI(val);
          }
          // If val isn't truthy (e.g., null), consider it a boolean attribute
          script.setAttribute(attr, val ? val : "");
        }
```

wch: With `const [attr, val]`, isn't that a problem for `val = href + "/" + encodeURI(val)`?

wch: Also, IMO, it's not worth adding a polyfill for this.

PR #3398, opened 2021-05-20
Use ggplot2::get_alt_text() if available to provide better default alt text

### LINKED ISSUE #3397: Use upcoming ggplot2 alt label
We're [working on](https://github.com/tidyverse/ggplot2/pull/4482) making it possible to add alt text to ggplots. It would be great if `renderPlot()` could use this. It could either inspect the object (if there's a specific ggplot2 code path), or if that's not possible use a technique similar to what I propose in https://github.com/yihui/knitr/issues/2001.

cpsievert:
Closes #3397 (via https://github.com/tidyverse/ggplot2/pull/4482)

```r
library(shiny)

ui <- fluidPage(
  plotOutput("p")
)

  output$p <- renderPlot({
    ggplot(mtcars, aes(wt, cyl)) + 
      geom_point() +
      labs(alt = "Foo bar")
  })
}

shinyApp(ui, server)
```


R/render-plot.R
```diff
@@ -333,10 +335,34 @@ custom_print.ggplot <- function(x) {
...
+  res <- list(
     build = build,
     gtable = gtable
-  ), class = "ggplot_build_gtable")
+  )
+  class(res) <- "ggplot_build_gtable"
+
+  # ggplot2::get_alt_text() was added in v3.3.4
+  # https://github.com/tidyverse/ggplot2/pull/4482
+  get_alt <- asNamespace("ggplot2")$get_alt_text
```
wch, line 346: `asNamespace()` has always felt a bit squishy to me; `getNamespace()` is slightly clearer:

```suggestion
  get_alt <- getNamespace("ggplot2")$get_alt_text
```

R/render-plot.R
```diff
@@ -36,12 +36,13 @@
...
-#'   (or a function referencing reactive values) that returns a character string.
-#'   NULL or "" is not recommended because those should be limited to decorative images
-#'   (the default is "Plot object").
+#' @param alt Alternate text for the HTML `<img>` tag if it cannot be displayed
+#'   or viewed (i.e., the user uses a screen reader). In addition to a character
+#'   string, the value may be a reactive expression (or a function referencing
+#'   reactive values) that returns a character string. By default,
+#'   `ggplot2::get_alt_text()` is used to get a sensible default for `ggplot2`
+#'   plots (other plots default to `"Plot Object"`). `NULL` or `""` is not
+#'   recommended because those should be limited to decorative images.
```
wch, line 45: The logic of the ggplot2 part is a bit hard to follow. Also, it would be good to explicitly explain the `NA` default value. Maybe something like this:

> If the value is `NA` (the default), then for ggplot2 plots, it will try to use `ggplot2::get_alt_text()` to extract alt text from the ggplot object; for other plots, `NA` will result in alt text "Plot object".

wch (approved):
Looks good, provided that the PR on the ggplot2 side stays pretty much the same.

PR #3404, opened 2021-05-24
insertTab() now handles position correctly when target is NULL

### LINKED ISSUE #3403: insertTab(target = NULL) messes up position
For example, I'd expect the positioning here to be CAB instead of ABC

```r
library(shiny)
ui <- fluidPage(
  tabsetPanel(id = "container", tabPanel("A", "a"), tabPanel("B", "b"))
)
  observe({
    insertTab("container", tabPanel("C", "c"), target = NULL, position = "before")
  })
}
shinyApp(ui, server)
```

Screen Shot 2021-05-24 at 1 25 13 PM

Note that you get the reverse problem when `position = "after"`

cpsievert:
Closes #3403, to be merged after https://github.com/rstudio/bslib/pull/328

We may also want to consider setting `target = NULL` in `insertTab()` since it leads to sensible default behavior


schloerke (changes requested):
I believe this _breaks_ `shiny::prependTab()` and `appendTab()`. The position values will need to be swapped

schloerke:
> We may also want to consider setting `target = NULL` in `insertTab()` since it leads to sensible default behavior

Sounds good to me! This is how `prependTab()` and `appendTab()` work internally

schloerke:
This PR will need to re-apply the changes within the respective TS file

schloerke (changes requested):
This PR will need to re-apply the changes within the respective TS file

schloerke (approved):

PR #3407, opened 2021-05-26
Remove Font Awesome 5 message about the `level-up` icon

schloerke:
Message:

> The `name` provided ('level-up') is deprecated in Font Awesome 5:
> * please consider using 'level-up-alt' or 'fas fa-level-up-alt' instead
> * use the `verify_fa = FALSE` to deactivate these messages


schloerke:
I don't know if there are other icons that would fall into this situation...

rich-iannone (approved):
Looks great!

schloerke:
I don't believe this needs a NEWS entry. Merging.

PR #3408, opened 2021-05-26
Install Cairo macOS devel brew dependency

schloerke:
> fatal error: 'X11/Intrinsic.h' file not found

via https://stackoverflow.com/a/66568545/591574


schloerke:
All green lights. Merging

PR #3412, opened 2021-06-02
Fix tab input value updating for BS4 dropdowns

cpsievert:
This fixes an oversight made in #3372, when we switched (back to) emit-ing `li a.dropdown-item` instead of `a.dropdown-item`. To verify the fix, notice how (before this change) selecting tab B results in an input value of `NULL` (when it should be `"B"`):

```r
library(shiny)

ui <- fluidPage(
  theme = bslib::bs_theme(version = 4),
  tabsetPanel(
    id = "tabs",
    tabPanel("A", "a"),
    navbarMenu("menu", tabPanel("B", "b"))
  )
)

  observe({
    print(input$tabs)
  })
}

shinyApp(ui, server)
```


srcts/src/main.ts
```diff
@@ -5977,7 +5977,7 @@ function main(): void {
         let anchors = isBS3()
```
schloerke, line 5977:
```suggestion
        // For future... These selectors may not work for nested anchors
        let anchors = isBS3()
```

srcts/src/main.ts
```diff
@@ -5961,7 +5961,7 @@ function main(): void {
       let anchor = isBS3()
```
schloerke, line 5961:
```suggestion
      // For future... These selectors may not work for nested anchors
      let anchor = isBS3()
```

schloerke (approved):

srcts/src/main.ts
```diff
@@ -5958,10 +5958,11 @@ function main(): void {
     },
     getValue: function (el) {
       // prettier-ignore
+      // For future... These selectors may not work for nested anchors
```
cpsievert, line 5961:
```suggestion
      // The BS4+ selectors may not work as is for dropdowns within dropdowns, but BS3+ dropped support for those anyway
```

srcts/src/main.ts
```diff
@@ -5974,10 +5975,11 @@ function main(): void {
 
       if (value) {
         // prettier-ignore
+        // For future... These selectors may not work for nested anchors
```
cpsievert, line 5978:
```suggestion
      // The BS4+ selectors may not work as is for dropdowns within dropdowns, but BS3+ dropped support for those anyway
```

PR #3413, opened 2021-06-09
Set selectize dropdownParent to "body" to prevent clipping

jjallaire:
To prevent clipping of the selectize drop-down we set the dropdownParent to "body". This might be necessary if e.g. overflow-x: scroll is set on it's container, which forces overflow-y to 'auto' (as per https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-y).

See option docs here: https://github.com/selectize/selectize.js/blob/master/docs/usage.md

Additional discussion of usage here: https://github.com/selectize/selectize.js/issues/192

### DIFF
--- a/R/input-select.R
+++ b/R/input-select.R
@@ -208,6 +208,15 @@ selectizeIt <- function(inputId, select, options, nonempty = FALSE) {
     options$plugins <- c(options$plugins, list('selectize-plugin-a11y'))
   }
 
+  # to prevent clipping of the selectize drop-down we set the dropdownParent
+  # to "body". This might be necessary if e.g. overflow-x: scroll is set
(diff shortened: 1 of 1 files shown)


jcheng5:
I paired with JJ on this, LGTM

cpsievert:
I was a little worried this would break [the recently added accessibility plugin](https://github.com/rstudio/shiny/blob/master/inst/www/shared/selectize/accessibility/js/selectize-plugin-a11y.js), but it seems OK

cpsievert (approved):
I was a little worried this would break [the recently added accessibility plugin](https://github.com/rstudio/shiny/blob/master/inst/www/shared/selectize/accessibility/js/selectize-plugin-a11y.js), but it seems OK

PR #3418, opened 2021-06-11
Export TypeScript type definitions to local folder

schloerke:
Am able to import this package directly and use within VS Code:
Screen Shot 2021-06-11 at 12 56 49 PM

In `package.json`, by setting `main` to `""`, then we explicitly state we are not exporting any logic, only types. This also prompted the pkg name `@types/shiny`.

In a different PR, we should set the package version from the R version.

(Everything in `src_d` is auto generated)


schloerke:
Merging now so that we may play with the changes in other packages. Probably not the final form of the package shape.

PR #3432, opened 2021-06-16
Match casing for plot alt text `"Plot object"`

schloerke:
Followup of #3398


cpsievert:
Whoops, let's revert to `"Plot object"` as the default (not `"Plot Object"`)

cpsievert (changes requested):
Whoops, let's revert to `"Plot object"` as the default (not `"Plot Object"`)

cpsievert:
Thanks!

PR #3436, opened 2021-06-17
TypeScript other distributed JS/CSS files

schloerke:
* Add license banner to CSS and JS compiled files
* Compile `shiny-autoreload.js`, `shiny-showcase.js`, `shiny-showcase.css`, and `shiny-testmode.js` from TypeScript or Sass
* Run `showcase.css` through a postCSS autoprefixer
* Run `shiny.min.css` through a postCSS autoprefixer

--------------

Needed to set the legacySliderType to where it was used to not upset jQuery.

--------------

Questions:
- [ ] Should `yarn build` call some form of `./tools/updateShinyCSS.R` ?
- [ ] Is `inst/shared/www/shiny_scss/bootstrap.scss` directly used other than as an htmldependency supplied to `{bslib}`?

* A: Keep the files there, generate bundle separately


srcts/extras/shiny-showcase.sass
```diff
@@ -0,0 +1,64 @@
+#showcase-well
+  border-radius: 0
```
cpsievert, line 2: Let's do scss instead

cpsievert (approved):

PR #3452, opened 2021-06-29
Use random inline styles to ensure transitionend fires everytime

### LINKED ISSUE #3451: Regression with session$setCurrentTheme() not refreshing stylesheets
After #3333, I've noticed some situations where `session$setCurrentTheme()` won't always disable old stylesheets. For example, note that with this example, sometimes when changing the theme, the `transitionend` event won't fire and in that case you'll see slate's border styles "leak" into the flatly theme:

```r
library(shiny)
library(bslib)
slate <- bs_theme(bootswatch = "slate")
flatly <- bs_theme(bootswatch = "flatly")
ui <- navbarPage(
  theme = slate,
  title = "Navbar title",
  tabPanel("A", "a"),
  tabPanel("B", "b"),
  header = selectInput("theme", "Choose theme", selected = "slate", c("slate", "flatly"))
)
  observeEvent(input$theme, {
    theme <- if (identical(input$theme, "flatly")) flatly else slate
    session$setCurrentTheme(theme)
  }, ignoreInit = TRUE)
}
shinyApp(ui, server)
```

(This can also be reproduced by changing Bootswatch themes in `bs_theme_preview()` -- you'll be able to see similar "leaking" of one theme into another)

cpsievert:
Closes #3451. Follow up to #3333


srcts/src/shiny/render.ts
```diff
@@ -233,37 +233,25 @@ function renderDependency(dep: HtmlDep) {
...
             $(document.body).append($dummyEl);
 
-            // Need to add the CSS with a setTimeout 0, to ensure that it
-            // takes effect _after_ the DOM element has been added. This is
-            // necessary to ensure that the transition actually occurs.
-            setTimeout(() => $head.append($dummyLink), 0);
+            // To ensure a transition actually happens, change the inline style _after_
+            // the DOM element has been added, and also use a new random color each time
+            // to prevent any potential caching done by the browser
+            const color = Math.floor(Math.random() * 16777215).toString(16);
```
wch, line 252: I think this needs a `"#"` in front of it.

wch (approved):

Issue #2521, opened 2019-07-03
Document and Allow specifying port range for random port selection


colearendt:
It seems that Shiny chooses random ports from 3000-8000 for connections. Some possible improvements that would be desirable:

- document that this is the random port range
- allow specifying a different random port range
- some mechanism for determining if the port is available before it is selected?
- a way to exclude ports from the range

The context here is RStudio Server and users who want to be restrictive around which ports are accessible. Some documentation would be nice, and optional control over the process through a server or package-install specification would probably be ideal. 

I'm not sure what standards are for this type of thing... if an environment variable would be workable, etc.

jcheng5:
We only listen on 127.0.0.1 by default, and RStudio Server proxies access through its own webserver. Can you be more specific about what kinds of restrictions customers are setting up? Are they trying to deny most ports even from 127.0.0.1?

colearendt:
Yeah, exactly. The customer I was working with this morning wanted to know exactly what ports to allow (even from 127.0.0.1).

jcheng5:
The range we're using is 3000-8000. We'll try up to 20 times in case something is using the port. It sounds like a red herring though TBH. Can they try using curl from the RSP Terminal pane to see if the port is responsive?

colearendt:
Ahh nice! That's good to know, thanks! I'm with you - I suspect that a bit of tweaking on their end (and the call where I explained all of this) will make this issue go away for them. I showed them how to curl from the RSP terminal pane to help establish the networking requirements, too.

Any value in documenting this port range, at least? Or is it something better left as an internal implementation detail that we can pull out of the source when needed?

wch:
Closed by #3456.

PR #3456, opened 2021-07-01
docs: update runapp port parameter docs

heds1:
To address #2521.

Also added three more TCP ports to reject, based on [this](https://chromium.googlesource.com/chromium/src.git/+/refs/heads/main/net/base/port_util.cc) Chrome source.


wch (approved):

wch:
Thanks!

PR #3459, opened 2021-07-02
Make disconnected overlay sizing styles stronger

nstrayer:
If you have an app with a rather innocent line of css that restricts sizes of children of the body such as: `body > div { max-width: 400px; max-height: 400px; }` this will cause the disconnected overlay div to be too small as well. 

This would be fine if it still covered the children, but because of the styles of `left, right, top, bottom` being set explicitly it will always be stuck to the upper-left of the viewport. Here's a minimum app example where the aforementioned styles were used to make the content horizontally centered on the screen and not allowed to get too big.

```
library(shiny)
shinyApp(
  ui = fluidPage(
    tags$head(
      tags$style(
        HTML("body > div { max-width: 400px; max-height: 400px; }
                    #content { margin: 0 auto; background: salmon; }")
      )
    ),
    div(id = "content", "Here's my content")
  ),
)
```

## Current behavior
If we run this app and then the disconnected overlay gets triggered (by simply killing the app process) we get this rather confusing result: 
image

## New behavior
And here's the result after the changes in this PR:
image

### DIFF
--- a/inst/www/shared/shiny_scss/shiny.scss
+++ b/inst/www/shared/shiny_scss/shiny.scss
@@ -92,6 +92,8 @@ pre.shiny-text-output {
   bottom: 0;
   left: 0;
   right: 0;
+  max-width: 100%;  /* Make sure width isn't restricted by styles on selectors like "body > div" */
+  max-height: 100%;
   background-color: $shiny-disconnected-bg;
   opacity: 0.5;
   overflow: hidden;


inst/www/shared/shiny_scss/shiny.scss
```diff
@@ -92,6 +92,8 @@ pre.shiny-text-output {
   bottom: 0;
   left: 0;
   right: 0;
+  max-width: 100%;  /* Make sure width isn't restricted by styles on selectors like "body > div" */
+  max-height: 100%;
```
cpsievert, line 96: `max-*` won't work with something like `body > div { width: 100px; height: 100px }`

```suggestion
  width: 100%;  /* Make sure width isn't restricted by styles on selectors like "body > div" */
  height: 100%;
```

cpsievert (changes requested):
See suggestion above

PR #3460, opened 2021-07-04
Get imputed URL for current app/session

hadley:
Code original written by @jcheng5 for https://github.com/r-lib/gargle/pull/157, but I think it makes the most sense for it to live in Shiny, given that y'all know the most about how to derive this url. I have zero clue how to test this. I made the following simple app, but then couldn't think how to deploy it + dev shiny in such away that it would run on shinyapps or RSC.

```R
library(shiny)
ui <- fluidPage(
  textOutput("url")
)
  output$url <- renderText(session$getUrl())
}
shinyApp(ui, server)
```

### DIFF
--- a/R/shiny.R
+++ b/R/shiny.R
@@ -2126,6 +2126,51 @@ ShinySession <- R6Class(
           }
         })
       }
+    },
+
+    getUrl = function() {
+      req <- self$request
+
+      url <-
+        # Connect
+        req[["HTTP_X_RSC_REQUEST"]] %||%
+        req[["HTTP_RSTUDIO_CONNECT_APP_BASE_URL"]] %||%
+        # ShinyApps.io
+        if (!is.null(req[["HTTP_X_REDX_FRONTEND_NAME"]])) {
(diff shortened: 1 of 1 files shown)


jcheng5:
There's actually (more reliable?) data available on `clientData`, as it's basically the browser telling the server what it's got in the URL bar:

https://github.com/rstudio/shiny/blob/7e4248bbca07cf12aebbcaac1b2e753756d303e1/srcts/src/shiny/init.ts#L460-L466

```r
library(shiny)

ui <- fluidPage(
  textOutput("url")
)

  output$url <- renderText({
    paste0(
      session$clientData$url_protocol,
      "//",
      session$clientData$url_hostname,
      ":",
      session$clientData$url_port,
      session$clientData$url_pathname,
      if (nzchar(session$clientData$url_search)) "?",
      session$clientData$url_search
    )
  })
}

shinyApp(ui, server)
```

I think I did not use this for gargle because I needed the URL while handling an HTTP request (i.e. in the UI), not from the server function.

PR #3468, opened 2021-07-23
Add `sustainEnvAndQuoted()`. Remove `getQuosure()`

schloerke:
Related: https://github.com/rstudio/shiny/pull/3462
Related: https://github.com/rstudio/shiny/pull/3466

`exprToFunction()` (and `installExprFunction()`) do not work in all situations and reaching up into the calling code is dangerous / should be avoided. Now, the dangerous code is isolated to `handleEnvAndQuoted()`, which ideally package authors will stop using. 

With the updated code, all handling will be done with quosures. woot.

-----------------------------------------

Shiny 1.6.0 code
```r
renderDouble <- function(expr, env = parent.frame(), quoted = FALSE) {
	func <- exprToFunction(expr, env, quoted)
	function() {
		rep(func(), 2)
	}
}
```

Shiny 1.7.0 migration code
```r
renderDouble <- function(expr, env = deprecated(), quoted = deprecated()) {
	q <- enquo0(expr)
	q <- handleEnvAndQuoted(q, expr, env, quoted)
    func <- quoToFunction(q)
	function() {
		rep(func(), 2)
	}
}
```

Ideal Shiny 1.7.0 code
```r
renderDouble <- function(expr) {
    func <- quoToFunction(enquo0(expr))
	function() {
		rep(func(), 2)
	}
}
```

-------------------------------------

Other changes
* Allow `shinyDeprected()` to know of `superseded` functions
* Document
	* Add superseded badge to `renderDataTable()`
	* Add superseded badge to `exprToFunction()`, `installExprFunction()`

Remaining:
- [ ] Document TODO `deprecated()` parameters
- [ ] Document upgrade/migration path for shinyRenderWidget
- [ ] Document `quoToFunction()` & `handleEnvAndQuoted()`
	- [ ] Explain why we're moving to quosures. One reason is that we don't have to reach to different levels using substitute. Once you have a quosure, it can be passed around safely. Normal NSE expressions are like bombs; quosures are safe to pass around.

PR #3472, opened 2021-07-28
`exprToFunction()` and `installExprFunction()` support quosures

schloerke:
Major changes:
* Removes `sustainEnvAndQuoted()` logic and docs from https://github.com/rstudio/shiny/pull/3468 
* Revert `getQuosure()`-like logic from https://github.com/rstudio/shiny/pull/3466 and https://github.com/rstudio/shiny/pull/3462 to now leverage `installExprFunction()` (as it now has quosure support).
* `exprToFunction()` is now able to allow for quosures
	* To use a quosure's expression as `expr`, call `exprToFunction(myquo, quoted = TRUE)` and it will ignore `env`.
	* To return the actual `myquo` value (as if it is an expression), set `quoted = FALSE`.
	* Therefore `installExprFunction()` also has quosure support
	* Therefore any existing `exprToFunction()` or `installExprFunction()` within other packages also has quosure support.

Other changes:
* No deprecation warning of `env` and `quoted` as quoted MUST be true to use a quosure. `env` can not check if is `missing()` as we can not know if it is 2,3,4... layers back that it is the _true_ missing `env`. (Could be done with many try-catches, but it feels VERY hacky and slow.)
* Tried using templates with `@templateVar`, but could not have the same template multiple times. See https://github.com/r-lib/roxygen2/issues/1241
* Updated reactives to use `installExprFunction()` for better debugging where `exprToFunction()` was used in v1.5.0 of shiny.


R/shinywrappers.R
```diff
@@ -2,6 +2,9 @@ utils::globalVariables('func', add = TRUE)
 
 #' Mark a function as a render function
 #'
+#' `r lifecycle::badge("superseded")` Please see [`createRenderFunction()`]
+#' for updated usage. (Shiny 1.1.0)
+#'
```
wch, line 7: Remove

R/shinywrappers.R
```diff
@@ -45,6 +48,7 @@ utils::globalVariables('func', add = TRUE)
 #'
 #' @seealso [createRenderFunction()], [quoToFunction()]
 #' @export
+#' @keywords internal
```
wch, line 51: Remove

R/shinywrappers.R
```diff
@@ -101,6 +111,7 @@ markRenderFunction <- function(
     cacheHint <- lapply(cacheHint, function(x) {
       if      (is.function(x)) formalsAndBody(x)
       else if (is.language(x)) zap_srcref(x)
+      else if (is_quosure(x)) quo_get_expr(x)
```
wch, line 114:
```suggestion
      else if (is_quosure(x)) zap_srcref(quo_get_expr(x))
```

R/shinywrappers.R
schloerke:
```suggestion
```

R/shinywrappers.R
schloerke:
```suggestion
```

R/bind-cache.R
```diff
@@ -310,19 +310,18 @@ utils::globalVariables(".GenericCallEnv", add = TRUE)
...
+#'     q <- enquo0(expr)
 #'
-#'     htmlwidgets::shinyRenderWidget(expr,
+#'     htmlwidgets::shinyRenderWidget(
+#'       q,
 #'       myWidgetOutput,
-#'       quoted = TRUE,
-#'       env = parent.frame(),
-#'       cacheHint = list(label = "myWidget", userExpr = expr)
+#'       cacheHint = list(label = "myWidget", userQuo = q)
```
cpsievert, line 318:
```suggestion
#'       quoted = TRUE
#'       cacheHint = list(label = "myWidget", userQuo = q)
```

R/bind-cache.R
```diff
@@ -310,19 +310,18 @@ utils::globalVariables(".GenericCallEnv", add = TRUE)
...
 #'     )
 #'   }
 #'   ```
 #'
 #'   If your `render` function sets any internal state, you may find it useful
-#'   in your call to [createRenderFunction()] or [markRenderFunction()] to use
+#'   in your call to [createRenderFunction()] (or [markRenderFunction()]) to use
```
cpsievert, line 324:
```suggestion
#'   in your call to [createRenderFunction()] to use
```

R/bind-cache.R
```diff
@@ -339,8 +338,8 @@ utils::globalVariables(".GenericCallEnv", add = TRUE)
 #'   effects or modify some external state, and they must re-execute each time
 #'   in order to work properly.
 #'
-#'   For developers of such code, they should call [createRenderFunction()] or
-#'   [markRenderFunction()] with `cacheHint = FALSE`.
+#'   For developers of such code, they should call [createRenderFunction()] (or
+#'   [markRenderFunction()]) with `cacheHint = FALSE`.
```
cpsievert, line 342:
```suggestion
#'   For developers of such code, they should call [createRenderFunction()] with `cacheHint = FALSE`.
```

R/reactives.R
```diff
@@ -2270,23 +2272,34 @@ maskReactiveContext <- function(expr) {
...
+  #   ..stacktraceon = TRUE
+  # )
+
+  # handlerFunc <- installExprFunction(
+  #   handlerExpr, "handlerFunc", handler.env, handler.quoted,
+  #   label = "observeEventHandler",
+  #   ..stacktraceon = TRUE
+  # )
+
+  # label <- funcToLabel(eventFunc, "observeEvent", label)
```
cpsievert, line 2302: Confirm with Winston that this is correct

R/shinywrappers.R
```diff
@@ -53,6 +53,12 @@ markRenderFunction <- function(
   cacheWriteHook = NULL,
   cacheReadHook = NULL
 ) {
+  # # Do not include this code
+  # # This method is called by the superseding function, createRenderFunction().
```
cpsievert, line 57: If you want to include this, let's expand on the details of why this func is "sort of" superceded but not actually

R/shinywrappers.R
```diff
@@ -101,6 +107,7 @@ markRenderFunction <- function(
     cacheHint <- lapply(cacheHint, function(x) {
       if      (is.function(x)) formalsAndBody(x)
       else if (is.language(x)) zap_srcref(x)
+      else if (is_quosure(x)) zap_srcref(quo_get_expr(x))
```
cpsievert, line 110:
```suggestion
      else if (is_quosure(x)) zap_srcref(quo_get_expr(x))
      else if (is.language(x)) zap_srcref(x)
```

R/utils-lang.R
```diff
@@ -181,182 +116,90 @@ quoToSimpleFunction <- function(q) {
...
-    env_is_present = env_is_present,
-    quoted_is_present = quoted_is_present,
-    x_is_quosure = x_is_quosure,
-    verbose = FALSE
-  )
+updateFunctionLabel <- function(label) {
+  if (all(is.language(label))) {
+    # Prevent immediately invoked functions like as.language(a()())
+    if (is.language(label) && length(label) > 1) {
+      return("wrappedFunction")
```
cpsievert, line 150: Let's think of a better name, maybe "anonymous"?

R/utils-lang.R
```diff
@@ -446,17 +311,75 @@ sustainEnvAndQuoted_ <- function(
...
+#' # Using `quoted = TRUE`
+#' tripleA <- renderTriple(q, quoted = TRUE)
+#' isolate(tripleA())
+#' #> "hello, hello, hello"
+#' # Using `rlang::inject()`
+#' tripleA <- rlang::inject(renderTriple(!!myquo))
+#' isolate(tripleA())
+#' #> "hello, hello, hello"
 #' @export
+#' @keywords internal
```
cpsievert, line 348:
```suggestion
```

tests/testthat/test-stacks.R
```diff
@@ -129,25 +129,25 @@ test_that("integration tests", {
...
     40L, 39L, 38L, 37L, 36L, 35L, 34L, 33L, 32L, 31L, 30L, 29L,
     28L, 27L, 26L, 25L, 24L, 23L, 22L, 21L, 20L, 19L, 18L, 17L,
     16L, 15L, 14L, 13L, 12L, 11L, 10L, 9L, 8L, 7L, 6L, 5L, 4L,
     3L, 2L, 1L))
   expect_equal(df$call, c("h", ".handleSimpleError", "stop",
-    "A", "B", "<reactive:C>", "..stacktraceon..", ".func", "withVisible",
+    "A", "B", "reactive", "<reactive:C>", "..stacktraceon..", ".func", "withVisible",
     "withCallingHandlers", "contextFunc", "env$runWith", "force",
     "domain$wrapSync", "promises::with_promise_domain",
     "withReactiveDomain", "domain$wrapSync", "promises::with_promise_domain",
```
schloerke, line 153: @jcheng5 These changes are necessary due to calling `installExprFunction()` which adds a function label. (Where before was `exprToFunction()`, which didn't wrap the function.)

Are these changes ok? I saw many messages warning about interactions with promises, but I believe the changes are expected changes.

R/render-plot.R
```diff
@@ -187,7 +192,7 @@ renderPlot <- function(expr, width = 'auto', height = 'auto', res = 72, ...,
     outputFunc,
     renderFunc,
     outputArgs,
-    cacheHint = list(userExpr = get_expr(q), res = res)
+    cacheHint = list(userExpr = installedFuncExpr(func), res = res)
```
wch, line 195: We should have a test for the correct `userExpr` for `renderPlot()` -- if in the future we make changes that result in the wrong cache hint, it could cause problems for users.

R/shinywrappers.R
```diff
@@ -10,7 +13,7 @@ utils::globalVariables('func', add = TRUE)
 #'
 #' Note that it is generally preferable to use [createRenderFunction()] instead
 #' of `markRenderFunction()`. It essentially wraps up the user-provided
-#' expression in the `transform` function passed to it, then pases the resulting
+#' expression in the `transform` function passed to it, then passes the resulting
 #' function to `markRenderFunction()`. It also provides a simpler calling
 #' interface.
```
wch, line 18:
```suggestion
#' interface. There may be cases where `markRenderFunction()` must be used instead of
#' [createRenderFunction()] -- for example, when the `transform` parameter of
#' [createRenderFunction()] is not flexible enough for your needs.
```

R/shinywrappers.R
```diff
@@ -100,6 +110,7 @@ markRenderFunction <- function(
     # For everything else, do nothing.
     cacheHint <- lapply(cacheHint, function(x) {
       if      (is.function(x)) formalsAndBody(x)
+      else if (is_quosure(x)) zap_srcref(quo_get_expr(x))
```
wch, line 113: We should make sure there are tests for quosures vs. expressions here.

R/shinywrappers.R
```diff
@@ -139,11 +150,27 @@ print.shiny.render.function <- function(x, ...) {
...
+#' Developer-facing utilities for implementing a custom `renderXXX()` function.
+#' Before using these utilities directly, consider using the [`htmlwidgets`
+#' package](http://www.htmlwidgets.org/develop_intro.html) to implement custom
+#' outputs (i.e., custom `renderXXX()`/`xxxOutput()` functions). That said,
+#' these utilities can be used more directly if a full-blown htmlwidget isn't
+#' needed and/or the user-supplied reactive expression needs to be wrapped in
+#' additional call(s).
+#'
+#' To implement a custom `renderXXX()` function, essentially 2 things are needed:
+#'   1. Capture the user's reactive expression as a function
```
wch, line 164:
```suggestion
#'   1. Capture the user's reactive expression as a function.
```

R/shinywrappers.R
```diff
@@ -176,12 +202,44 @@ print.shiny.render.function <- function(x, ...) {
...
+#' #> [1] "rlang, rlang, rlang"
+#'
+#' # Supplying quoted code without rlang::quo() requires installExprFunction()
+#' renderTripleLegacy(quote({ r("legacy"); r() }), quoted = TRUE)()
+#' #> [1] "legacy, legacy, legacy"
+#'
+#' # The legacy approach also supports with quosures (env is ignored in this case)
+#' r3 <- rlang::quo({ r("legacy-rlang"); r() })
+#' renderTripleLegacy(r3, quoted = TRUE)()
+#' #> [1] "legacy-rlang, legacy-rlang, legacy-rlang"
```
wch, line 238:
```suggestion
# User can supply quoted code via rlang::quo(). Note that evaluation of the
# expression happens when r2() is invoked, not when r2 is created.
q <- rlang::quo({ v() })
r2 <- rlang::inject(renderTriple(!!q))
v("rlang")
r2()
#> [1] "rlang, rlang, rlang"

# Supplying quoted code without rlang::quo() requires installExprFunction()
expr <- quote({ v() })
r3 <- renderTripleLegacy(expr, quoted = TRUE)
v("legacy")
r3()
#> [1] "legacy, legacy, legacy"

# The legacy approach also supports with quosures (env is ignored in this case)
q <- rlang::quo({ v() })
r4 <- renderTripleLegacy(q, quoted = TRUE)
v("legacy-rlang")
r4()
#> [1] "legacy-rlang, legacy-rlang, legacy-rlang"
```

wch: This also makes me think -- we need to explain somewhere why it's better to create a render function **without** `env` and `quoted` and have people use `inject()`, rather than just keep using `quoted` when they have quosures.

wch (approved):

tests/testthat/test-stacks.R
schloerke: Changes reverted. Closing

schloerke:
Merging to allow for shinycoreci-apps to run overnight

PR #3473, opened 2021-08-03
Fix Rituals workflow validating commits have not been made

schloerke:
Only perform logic if there is SOMETHING, not NOTHING


schloerke:
Merging early as logic check can only be performed on master branch

PR #3474, opened 2021-08-03
Fix bash logic in action step

schloerke:
Followup to #3473


schloerke:
Merging as logic is on master branch only

PR #3481, opened 2021-08-07
Enable `fileInput` to set the `capture` attribute

khaled-alshamaa:
https://www.w3.org/TR/html-media-capture/

### DIFF
--- a/R/input-file.R
+++ b/R/input-file.R
@@ -23,6 +23,14 @@
 #' @param buttonLabel The label used on the button. Can be text or an HTML tag
 #'   object.
 #' @param placeholder The text to show before a file has been uploaded.
+#' @param capture What source to use for capturing image, audio or video data.
+#'   This attribute facilitates user access to a device's media capture 
+#'   mechanism, such as a camera, or microphone, from within a file upload 
+#'   control.
+#'
+#'   A value of `user` indicates that the user-facing camera and/or microphone
+#'   should be used. A value of `environment` specifies that the outward-facing
+#'   camera and/or microphone should be used.
 #'
 #' @examples
 #' ## Only run examples in interactive R sessions
@@ -73,7 +81,8 @@
 #'
 #' @export
 fileInput <- function(inputId, label, multiple = FALSE, accept = NULL,
-  width = NULL, buttonLabel = "Browse...", placeholder = "No file selected") {
+  width = NULL, buttonLabel = "Browse...", placeholder = "No file selected",
+  capture = NULL) {
 
   restoredValue <- restoreInput(id = inputId, default = NULL)
 
@@ -101,6 +110,14 @@ fileInput <- function(inputId, label, multiple = FALSE, accept = NULL,
   if (length(accept) > 0)
     inputTag$attribs$accept <- paste(accept, collapse=',')
 
+  if (!is.null(capture)) {
+    if (tolower(capture) %in% c("user", "environment")) {
+      inputTag$attribs$capture <- tolower(capture)
+    } else {
+      # capture was previously a Boolean attribute which, if present, requested that the device's media capture device(s) such as camera or microphone be used instead of requesting a file input.
+      inputTag$attribs$capture <- NA
+    }
+  }
 
   div(class = "form-group shiny-input-container",
     style = css(width = validateCssUnit(width)),


khaled-alshamaa:
Sorry, I forget to add an associated entry to NEWS.md in the __New features an improvements__ section! Here it is:

* `fileInput()` can set the `capture` attribute to facilitates user access to a device's media capture mechanism, such as a camera, or microphone, from within a file upload control ([W3C HTML Media Capture](https://www.w3.org/TR/html-media-capture/)).

khaled-alshamaa:
I submitted this #3483 feature request issue for any further discussion.

schloerke:
Another example of docs: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/capture

cpsievert:
Thanks @khaled-alshamaa, but I don't think this'll be worth adding since you can achieve the same thing by doing:

```r
library(shiny)
# remotes::install_github('rstudio/htmltools')
library(htmltools)

tagAppendAttributes(
  fileInput("foo", "label"), 
  .cssSelector = "input",
  capture = "user"
)
```

R/input-file.R
```diff
@@ -101,6 +110,14 @@ fileInput <- function(inputId, label, multiple = FALSE, accept = NULL,
   if (length(accept) > 0)
     inputTag$attribs$accept <- paste(accept, collapse=',')
 
+  if (!is.null(capture)) {
+    if (tolower(capture) %in% c("user", "environment")) {
+      inputTag$attribs$capture <- tolower(capture)
+    } else {
+      # capture was previously a Boolean attribute which, if present, requested that the device's media capture device(s) such as camera or microphone be used instead of requesting a file input.
+      inputTag$attribs$capture <- NA
+    }
```
schloerke, line 119:
```suggestion
    inputTag$attribs$capture <- switch(
      tolower(capture[1]),
      # Only standard values are `user` and `environment`
      "user" = "user",
      "environment" = "environment",
      # Missing attr value is allowed
      NA
    )
```

While the code above works, I'm trying to steer internal code to avoid %in% and use `switch` when possible.

Reprex:

``` r
bench_mark <- function(capture) {
  bench::mark(
    switch = {
      switch(
            tolower(capture[1]),
            # Only standard values are `user` and `environment`
            "user" = "user",
            "environment" = "environment",
            # Missing attr value is allowed
            NA
          )
    },
    ifin = {
      capture2 <- tolower(capture[1])
      if (capture2 %in% c("user", "environment")) {
        capture2
      } else {
        NA
      }
    }
  )
}

bench_mark(c("Environment", "wrong"))
#> # A tibble: 2 × 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 switch       1.65µs   1.97µs   377007.        0B        0
#> 2 ifin         2.78µs   3.21µs   217854.        0B        0
bench_mark("user")
#> # A tibble: 2 × 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 switch       1.54µs   1.86µs   468372.        0B        0
#> 2 ifin         2.74µs   3.18µs   283665.        0B        0
bench_mark(NA)
#> # A tibble: 2 × 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 switch       1.28µs    1.5µs   587514.        0B     58.8
#> 2 ifin         2.29µs   2.65µs   349124.        0B      0
bench_mark("other")
#> # A tibble: 2 × 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 switch        1.6µs   1.89µs   452467.        0B      0  
#> 2 ifin         2.68µs   3.12µs   298898.        0B     29.9
```

<sup>Created on 2021-08-13 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

khaled-alshamaa: Dear @cpsievert, thanks for your response, but I don't think that it will work that way (at least the snippet that you shared) because the `fileInput` returning a complete block of HTML tags not a single tag.

Please note that tagAppendAttributes will append the attribute to the root tag in that block (i.e., in this case, the very first div tag `div class="form-group shiny-input-container"`)! Instead, we need that capture attribute much deeper in the `input` tag. 

Anyway, you may find a workaround solution I added in here #3483, which works well, but I believe that the Shiny framework should provide that fundamental attribute in an easier and straightforward way.

khaled-alshamaa:
> Thanks @khaled-alshamaa, but I don't think this'll be worth adding since you can achieve the same thing by doing:
> 
> ```r
> library(shiny)
> # remotes::install_github('rstudio/htmltools')
> library(htmltools)
> 
> tagAppendAttributes(
>   fileInput("foo", "label"), 
>   .cssSelector = "input",
>   capture = "user"
> )
> ```

Dear @cpsievert, thanks for your response, but I don't think that it will work that way (at least the snippet that you shared) because the `fileInput` returning a complete block of HTML tags not a single tag.

Please note that tagAppendAttributes will append the attribute to the root tag in that block (i.e., in this case, the very first div tag `div class="form-group shiny-input-container"`)! Instead, we need that capture attribute much deeper in the `input` tag. 

Anyway, you may find a workaround solution I added in here #3483, which works well, but I believe that the Shiny framework should provide that fundamental attribute in an easier and straightforward way.

khaled-alshamaa:
> Another example of docs: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/capture

Also here as well, I just stick with the W3C reference seeking for standard/official source:
https://developers.google.com/web/fundamentals/media/capturing-images

schloerke:
> > Thanks @khaled-alshamaa, but I don't think this'll be worth adding since you can achieve the same thing by doing:
> > ```r
> > library(shiny)
> > # remotes::install_github('rstudio/htmltools')
> > library(htmltools)
> > 
> > tagAppendAttributes(
> >   fileInput("foo", "label"), 
> >   .cssSelector = "input",
> >   capture = "user"
> > )
> > ```
> 
> Dear @cpsievert, thanks for your response, but I don't think that it will work that way (at least the snippet that you shared) because the `fileInput` returning a complete block of HTML tags not a single tag.

@khaled-alshamaa It will work with the next version of `htmltools`. The `.cssSelector = "input"` will search for all `input` tags within the tags. This parameter is added in next version of `htmltools` (releasing this month). You can use the dev version and learn about it here: https://rstudio.github.io/htmltools/articles/tagQuery.html

khaled-alshamaa:
> > > Thanks @khaled-alshamaa, but I don't think this'll be worth adding since you can achieve the same thing by doing:
> > > ```r
> > > library(shiny)
> > > # remotes::install_github('rstudio/htmltools')
> > > library(htmltools)
> > > 
> > > tagAppendAttributes(
> > >   fileInput("foo", "label"), 
> > >   .cssSelector = "input",
> > >   capture = "user"
> > > )
> > > ```
> > 
> > 
> > Dear @cpsievert, thanks for your response, but I don't think that it will work that way (at least the snippet that you shared) because the `fileInput` returning a complete block of HTML tags not a single tag.
> 
> @khaled-alshamaa It will work with the next version of `htmltools`. The `.cssSelector = "input"` will search for all `input` tags within the tags. This parameter is added in next version of `htmltools` (releasing this month). You can use the dev version and learn about it here: https://rstudio.github.io/htmltools/articles/tagQuery.html

I see

> > > Thanks @khaled-alshamaa, but I don't think this'll be worth adding since you can achieve the same thing by doing:
> > > ```r
> > > library(shiny)
> > > # remotes::install_github('rstudio/htmltools')
> > > library(htmltools)
> > > 
> > > tagAppendAttributes(
> > >   fileInput("foo", "label"), 
> > >   .cssSelector = "input",
> > >   capture = "user"
> > > )
> > > ```
> > 
> > 
> > Dear @cpsievert, thanks for your response, but I don't think that it will work that way (at least the snippet that you shared) because the `fileInput` returning a complete block of HTML tags not a single tag.
> 
> @khaled-alshamaa It will work with the next version of `htmltools`. The `.cssSelector = "input"` will search for all `input` tags within the tags. This parameter is added in next version of `htmltools` (releasing this month). You can use the dev version and learn about it here: https://rstudio.github.io/htmltools/articles/tagQuery.html

I see, thanks @schloerke for the clarification. But I still believe that such a fundamental attribute deserves a more simple and straightforward way to set in the Shiny framework instead of this a bit of tricky solution. Keep in mind the wide use case in the mobile app when developers integrate such kinds of inputs in their image recognition/classification functionality using deep learning or any other machine learning models ;-)

khaled-alshamaa:
Oops! What went wrong in the checking process and how I can resolve it? @schloerke your help is highly appreciated :-)

schloerke:
Timing is off because it was the first PR. Can you pull the latest, run `devtools::document()`, and then push?

khaled-alshamaa:
> Oops! What went wrong in the checking process and how I can resolve it? @schloerke your help is highly appreciated :-)

I believe it is related to the `fileInput` function signature, because we added a new parameter and that does not match the expected value in the checking procedure. Is there anything that I can do from my side? Please let me know.

schloerke:
The failure above a false positive. 

We'll hold off on this PR for a bit until Shiny 1.7.0 is on CRAN.

khaled-alshamaa:
> The failure above a false positive.
> 
> We'll hold off on this PR for a bit until Shiny 1.7.0 is on CRAN.

Gr8, so we are done in here, am I right? We will expect this feature to be available in the next version 1.7.0 of the Shiny framework. When is the planned schedule for that release?

Thank you @schloerke, @cpsievert and all the RStudio/Shiny team for your great job and support. I am looking forward for more contributions in this game changer utility in hands of data scientists :-)

cpsievert:
@khaled-alshamaa have you actually tested that this works? This seems to suggest that there is virtually no browser support https://caniuse.com/html-media-capture

khaled-alshamaa:
> @khaled-alshamaa have you actually tested that this works? This seems to suggest that there is virtually no browser support https://caniuse.com/html-media-capture

Yes @cpsievert, it works with most major mobile browsers which is the actual use case for such an attribute (i.e., a media capture device like a front-facing camera or a primary camera) as shown in the link you shared. Please notice that the related W3C recommendation was just published on Feb 1, 2018.

khaled-alshamaa:
> > @khaled-alshamaa have you actually tested that this works? This seems to suggest that there is virtually no browser support https://caniuse.com/html-media-capture
> 
> Yes @cpsievert, it works with most major mobile browsers which is the actual use case for such an attribute (i.e., a media capture device like a front-facing camera or a primary camera) as shown in the link you shared. Please notice that the related W3C recommendation was just published on Feb 1, 2018.

Also, please note that I already tested it on the Firefox for Android version 91.1.0 (Build #2015826187) and it works perfectly fine (the report you shared refer that is not supported in version 90).

khaled-alshamaa:
> The failure above a false positive.
> 
> We'll hold off on this PR for a bit until Shiny 1.7.0 is on CRAN.

@schloerke, I hope this PR find it's way to merge soon :-)

khaled-alshamaa:
Is there any chance to see this tiny update finds its way to the official Shiny framework? It is very helpful for any developer who is integrating ML/AI image recognition models as a part of a mobile/web application.

R/input-file.R
```diff
@@ -23,6 +23,14 @@
...
 #' @param placeholder The text to show before a file has been uploaded.
+#' @param capture What source to use for capturing image, audio or video data.
+#'   This attribute facilitates user access to a device's media capture 
+#'   mechanism, such as a camera, or microphone, from within a file upload 
+#'   control.
+#'
+#'   A value of `user` indicates that the user-facing camera and/or microphone
+#'   should be used. A value of `environment` specifies that the outward-facing
+#'   camera and/or microphone should be used.
 #'
```
wch, line 34:
```suggestion
#'
#'   By default on most phones, this will accept still photos or video. For
#'   still photos only, also use `accept="image/*"`. For video only, use
#'   `accept="video/*"`.
```

NEWS.md
```diff
@@ -28,6 +28,8 @@ shiny 1.6.0.9000
 
 * Addressed #2521: Updated the list of TCP ports that will be rejected by default in runapp.R, adding 5060, 5061 and 6566. Added documentation describing the port range (3000:8000) and which ports are rejected. (#3456)
 
+* `fileInput()` can set the `capture` attribute to facilitates user access to a device's media capture mechanism, such as a camera, or microphone, from within a file upload control ([W3C HTML Media Capture](https://www.w3.org/TR/html-media-capture/)).
```
wch, line 31:
```suggestion
* `fileInput()` can set the `capture` attribute to facilitates user access to a device's media capture mechanism, such as a camera, or microphone, from within a file upload control ([W3C HTML Media Capture](https://www.w3.org/TR/html-media-capture/)). (Thanks to khaled-alshamaa, #3481)
```

PR #3488, opened 2021-08-18
Adjust app port tests to use random port values

schloerke:
This change allows `revdepcheck` to run in parallel without errors


tests/testthat/test-app.R
```diff
@@ -211,6 +211,17 @@ test_that("Setting options in various places works", {
   op <- options(shiny.launch.browser = FALSE)
   on.exit(options(op), add = TRUE)
 
+  # If the port values are fixed, the tests fail while being tested concurrently
+  test_app_port <- httpuv::randomPort()
```
wch: Watch out for collisions between these port numbers.

schloerke:
Reviewed with @wch earlier. Merging.

PR #3492, opened 2021-08-24
Update GHA workflows to use latest versions

schloerke:
* Test on ubuntu 18,20
* Use dynamic r versions; 
* Match workflow to other packages within the shiny-verse


schloerke:
Merging as tests work

PR #3515, opened 2021-09-29
Opt-in to `C` collate order in test snapshots

schloerke:
** Meant to have commit say ``Enforce `en_US.UTF-8` ordering`` on first commit

Fixes https://github.com/rstudio/shinytest/issues/409

When snapshots values are not ordered using a consistent locale, then the order can change.  This PR enforces that the locale used for sorting is `en_US.UTF-8`

------------------

- [x] Shiny 1.7.0 patch releases are done
- [x] To be merged for Shiny 1.8.0 release


schloerke:
😩 Ugh. Windows utf-8 support is bad. 

If we switch to a `C` locale, then **EVERYONE**'s tests will break. That is worse. 😭 

----------------------

Holding off for now as there is not consistent solution

bersbersbers:
> If we switch to a `C` locale, then **EVERYONE**'s tests will break.

I don't get this part (are you talking about existing or future tests?), but do you think the `radix` method might be better, as discussed in https://github.com/r-lib/devtools/issues/2377?

R/utils.R
```diff
@@ -193,7 +193,11 @@ sortByName <- function(x) {
   if (length(x) == 0)
     return(x)
 
-  x[order(names(x))]
+  # Use `en_US.UTF-8` over `C`
+  #  https://github.com/rocker-org/rocker/issues/19
+  # Must provide consistent sort order
+  #  https://github.com/rstudio/shinytest/issues/40
+  withr::with_collate("en_US.UTF-8", x[order(names(x))])
```
bersbersbers, line 200: I think the 2nd issue URL is truncated - should be issues/409

tests/testthat/test-utils.R
```diff
@@ -171,6 +171,30 @@ test_that("sortByName works as expected", {
   # Make sure atomic vectors work
   expect_identical(sortByName(c(b=1, a=2)), c(a=2, b=1))
   expect_identical(sortByName(c(b=1, a=2, b=3)), c(a=2, b=1, b=3))
+
+  # Collate order is consistent
+  val <- list(a = 1, b = 2, A = 3, B = 4)
```
bersbersbers, line 176: These tests would be more complete when having `_a` as well (which makes life much harder, I know...)

bersbersbers:
You may to try also setting `Sys.setenv(LC_COLLATE = "en_US.UTF-8")` - this seems to fix the underscore issue regardless of any external `LC_COLLATE`.

R/utils.R
```diff
@@ -193,7 +193,11 @@ sortByName <- function(x) {
   if (length(x) == 0)
     return(x)
 
-  x[order(names(x))]
+  # Use `en_US.UTF-8` over `C`
+  #  https://github.com/rocker-org/rocker/issues/19
+  # Must provide consistent sort order
+  #  https://github.com/rstudio/shinytest/issues/40
```
schloerke, line 199:
```suggestion
  #  https://github.com/rstudio/shinytest/issues/409
```

schloerke:
The current approach of setting locale of `en_us.UTF-8` won't work as windows does not recognize it.

--------------------------

Talking through with the team some more, there are couple open approaches moving forward to try to reduce false-positive changes in existing tests:
* Suggest an import of `stringi` (heavy / unwanted package dependency) and use `stringi::stri_order(names(x), opts_collator = stringi::stri_opts_collator(....))` where `....` is rules that produce similar results to a collate order of `en_us.UTF-8`
* Make our own sorting function in R that produces a sorting order similar to `en_us.UTF-8`

Both of these options should also have `shinytest` check if the json files are different and produce a message that the content is semantically the same, but instruct the user that the newly ordered output should be adopted.

Both of these options have the least amount of impact to existing tests compared to using `order(method = "radix"` or setting the locale to `C`.

-------------------------

Another approach would be to send in a flag to the snapshot url function to opt in to using `radix` sorting. I like this approach as existing mechanism will continue to work as they currently do (broken in `shinytest`) and new mechanisms ex: `shinytest2` (not public yet) will be able to opt in to consistent sorting methods. Maybe `shinytest` could listen to a `DESCRIPTION` field like the `testthat` edition, ex `Config/shinytest/sort`.

There could even be a hybrid approach of opting in to sort with stringi...

bersbersbers:
> The current approach of setting locale of `en_us.UTF-8` won't work as windows does not recognize it.

Yes, that is a problem that I had noticed, too: I was surprised to see that R does not have any portable API for this. See https://github.com/rstudio/shinytest/issues/409#issuecomment-930403069 for my naive solution, but I guess the R world would benefit from a package with a curated lookup table from language identifier and encodings to locale strings for all supported platforms.

schloerke:
Notes for the PR:

Now using a flag to opt-in to the new sorting method. This is implemented using `order(x, method = "radix")` which uses a `C` locale. 

Even if two platforms share `en_us.UTF-8`, they may not sort consistently. See https://blog.zhimingwang.org/macos-lc_collate-hunt . 

Example from blog post:
```bash
# (macOS)
LC_ALL=en_US.UTF-8 sort <<<$'python-dev\npython3-dev'
#> python-dev
#> python3-dev
```
```bash
# (Linux)
LC_ALL=en_US.UTF-8 sort <<<$'python-dev\npython3-dev'
#> python3-dev
#> python-dev
```

---------------

Using the opt-in approach:
* Allows existing apps to stay consistent
* Allows users having difficulty to resolve the issue
* Allows newer packages (ex: `shinytest2`) to permanently opt-in

-----------------------

If ordering is a problem for a shiny application using shinytest, a field/value in the DESCRIPTION file `Config/shiny/snapshotSortC: 1` should be added. The DESCRIPTION file should be in the root of the app directory.

R/shiny.R
```diff
@@ -478,6 +478,44 @@ ShinySession <- R6Class(
...
+          # R CMD check runs under the `C` locale.
+          # However, before this parameter, existing snapshots were most likely not
+          #   under the `C` locale is would cause failures. This parameter allows
+          #   users to opt-in to the `C` locale.
+          # From ?sort:
+          #   However, there are some caveats with the radix sort:
+          #     If ‘x’ is a ‘character’ vector, all elements must share the
+          #   same encoding. Only UTF-8 (including ASCII) and Latin-1
+          #   encodings are supported. Collation always follows the "C"
+          #   locale.
```
wch, line 491: Add a note that shinytest2 will always set `sortC` to 1.

R/shiny.R
```diff
@@ -478,6 +478,44 @@ ShinySession <- R6Class(
...
+              }
+              "radix"
+            } else {
+              # No `sortC` param
+              if (!file.exists("DESCRIPTION")) {
+                "auto"
+              } else {
+                # read from description file?
+                tryCatch({
+                  info <- as.list(read.dcf("DESCRIPTION")[1, , drop = TRUE])
```
wch, line 505: Convert to a `shinyOption`. Make sure to document that it's relevant for shinytest, but not shinytest2.

schloerke:
@bersbersbers From feedback from Winston, the DESCRIPTION field has been converted to a shiny option. 

Within your apps, please set the option below to opt-in to consistent sorting with {shinytest}.
```r
options(shiny.snapshotsortc = TRUE)
```

wch (approved):

schloerke:
Correction. Within an app.R file, you'll need to call `shinyOptions(snapshotsortc = TRUE)`.   (I've updated the comment above.)

Similar to where we can set `options(shiny.testmode = TRUE)`, `options(shiny.snapshotsortc = TRUE)` may also be set.

PR #3537, opened 2021-11-03
HTML dependency improvements

wch:
Three main things in this PR:

* Shiny's internal HTML dependencies are now mounted dynamically instead of statically. `htmlDependency(src=c(href=x))` is no longer used; only `htmlDependency(src=c(file=x))`.
	* Previously, `script` and `link` tags looked like this:
	  ```html
	  <script src="shared/jquery.min.js"></script>
	  <link href="shared/shiny.min.css" rel="stylesheet" />
	  <script src="shared/shiny.js"></script>
	  ```
      but after this change they look like this: 
	  ```html
	  <script src="jquery-3.6.0/jquery.min.js"></script>
	  <link href="shiny-css-1.7.1.9001/shiny.min.css" rel="stylesheet" />
	  <script src="shiny-javascript-1.7.1.9001/shiny.js"></script>
	  ```

<br class="Apple-interchange-newline">

* HTML dependencies that are sent to dynamic UI now have better type checking, and no longer require a `dep.src.href` field.
* For dynamic UI rendering, the code path for a `restyle` has been extracted out so that it's not entangled with the normal case.


R/bootstrap.R
```diff
@@ -215,11 +215,10 @@ registerThemeDependency <- function(func) {
...
 bootstrapDependency <- function(theme) {
   htmlDependency(
-    "bootstrap", bootstrapVersion,
-    c(
-      href = "shared/bootstrap",
-      file = system.file("www/shared/bootstrap", package = "shiny")
-    ),
+    "bootstrap",
+    bootstrapVersion,
+    c(file = "www/shared/bootstrap"),
```
cpsievert, line 220: I don't feel super strongly about it, but I prefer this usage sense it simpler to read and way more commonly used

```suggestion
    src = "www/shared/bootstrap",
```

srcts/src/shiny/render.ts
```diff
@@ -116,166 +141,69 @@ function needsRestyle(dep: HtmlDep) {
...
-      });
-    }
+  if (stylesheetLinks.length !== 0) {
+    $head.append(stylesheetLinks);
   }
 
-  if (dep.script && !restyle) {
-    const scriptsAttrs = asArray(dep.script);
-    const scripts = $.map(scriptsAttrs, function (x) {
+  if (dep.script.length !== 0) {
```
cpsievert, line 198:
```suggestion
  if (dep.script.length > 0) {
```

srcts/src/shiny/render.ts
```diff
@@ -287,34 +215,17 @@ function renderDependency(dep: HtmlDep) {
...
-    }
-
-    const attach = $.map(attachments, function (attachment, key) {
+  if (dep.attachment.length !== 0) {
+    const attach = dep.attachment.map((attachment) => {
       return $("<link rel='attachment'>")
-        .attr("id", dep.name + "-" + key + "-attachment")
-        .attr("href", href + "/" + encodeURI(attachment));
+        .attr("id", dep.name + "-" + attachment.key + "-attachment")
+        .attr("href", encodeURI(attachment.href));
```
cpsievert, line 222: This could also be vanilla JS (as we're now doing for `<meta>`?)

srcts/src/shiny/render.ts
```diff
@@ -116,166 +141,69 @@ function needsRestyle(dep: HtmlDep) {
...
-              "#" + Math.floor(Math.random() * 16777215).toString(16);
-
-            setTimeout(() => $dummyEl.css("color", color), 10);
-          });
-
-          $head.append(link);
-        }
-      });
-    }
+  if (stylesheetLinks.length !== 0) {
```
cpsievert, line 194:
```suggestion
  if (stylesheetLinks.length > 0) {
```

srcts/src/shiny/render.ts
```diff
@@ -116,166 +141,69 @@ function needsRestyle(dep: HtmlDep) {
 }
 
 // Client-side dependency resolution and rendering
-function renderDependency(dep: HtmlDep) {
-  const restyle = needsRestyle(dep);
+function renderDependency(dep_: HtmlDep) {
+  const dep = simplifyHtmlDependency(dep_);
+
+  // Convert stylesheet objs to links early, because if `restyle` is true, we'll
+  // pass them through to `renderStylesheet` below.
```
cpsievert, line 148:
```suggestion
  // pass them through to `addStylesheetsAndRestyle` below.
```

srcts/src/shiny/render.ts
```diff
@@ -287,34 +215,17 @@ function renderDependency(dep: HtmlDep) {
...
-      $.each(attachments, function (index, attachment) {
-        const key = index + 1 + "";
-
-        tmp[key] = attachment;
-      });
-      attachments = tmp;
-    }
-
-    const attach = $.map(attachments, function (attachment, key) {
+  if (dep.attachment.length !== 0) {
```
cpsievert, line 218:
```suggestion
  if (dep.attachment.length > 0) {
```

cpsievert (approved):
LGTM pending minor suggestions

R/showcase.R
```diff
@@ -32,26 +32,45 @@ licenseLink <- function(licenseName) {
...
+      package="shiny",
+      script = "highlight.pack.js",
+      stylesheet = "rstudio.css"
+    ),
+    htmlDependency(
+      "showcase",
+      "0.1.0",
+      c(file = "www/shared"),
+      package = "shiny",
+      script = "shiny-showcase.js",
```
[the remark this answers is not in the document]
schloerke: It should as the main logic is inside a `on load` method: https://github.com/rstudio/shiny/blob/8ea97df3f29292cb47f33fccf55aa54296e23509/srcts/extras/shiny-showcase.ts#L310-L311

R/bootstrap.R
schloerke: Feels weird to me. I'm all for `src = FILEPATH`

R/bootstrap.R
```diff
@@ -215,11 +215,10 @@ registerThemeDependency <- function(func) {
...
+    package = "shiny",
```
schloerke, line 221: Thank you for putting these under `package = shiny`

R/shinyui.R
```diff
@@ -69,7 +70,6 @@ jqueryDependency <- function() {
     return(htmlDependency(
       "jquery", version_jquery,
       src = c(
-        href = "shared",
         file = "www/shared"
```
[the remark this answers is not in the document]
schloerke: I believe we'd need a single dependency that is `all_files = TRUE` for the `www/shared` folder.

Now that we're also not using `href` to define the location, should we move jQuery, jQuery legacy, shiny.js, and all other html deps that do not need to be in the `shared` folder out of `www/shared`? (... since they are directly imported, no need to _double_ import)

R/showcase.R
```diff
@@ -32,26 +32,45 @@ licenseLink <- function(licenseName) {
...
-      script = "showdown.js"),
-    htmlDependency("highlight.js", "6.2", c(href="shared/highlight"),
-      script = "highlight.pack.js")
+    htmlDependency(
+      "jqueryui",
+      "1.12.1",
+      c(file = "www/shared/jqueryui"),
+      package = "shiny",
+      script = "jquery-ui.min.js"
+    ),
```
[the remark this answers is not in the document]
wch: Done in 07e7bb88 -- it's probably a good idea to look at that commit in isolation.

R/showcase.R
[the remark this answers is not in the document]
wch: All this stuff is in the `showcaseHead()` function, which gets wrapped in `tags$head(showcaseHead())` when it's used later on, so it previously did appear in the `<head>`. Showcase mode all still appears to work fine in the browser, so I think we're good.

srcts/types/src/shiny/render.d.ts
```diff
@@ -8,21 +8,32 @@ declare function renderContent(el: BindScope, content: string | {
...
         href: string;
     };
-    meta?: string[] | string;
-    stylesheet?: string[] | string;
-    script?: Array<{
-        [key: string]: string;
-    }> | string[] | string | {
-        [key: string]: string;
+    meta?: {
+        [x: string]: string;
```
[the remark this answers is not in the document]
wch: The incoming format is something like:

```ts
{"x": "xval", "y": "yval"}
```

Which gets translated to:

```html
<meta name="x", content="xval">
<meta name="y", content="yval">
```

So we can't really make it more strict.

Note that the `MetaItem[]` format is different. It's normalized to have a structure similar to `ScriptItem[]` and `StylesheetItem[]`, and it looks like this:

```ts
[{name: "x", content: "xval"}, {name: "y", content: "yval"}]
```

The benefits are (A) that it's consistent with the other types of structures, and (B) it also supports other `<meta>` attributes, in case they're wanted. The normalized format is what prism generates, so it could hypothetically make use of the other attributes.

R/bootstrap.R
wch: I guess now that we don't use `href` anywhere, there's less risk of confusion and it makes a bit more sense to go with unnamed strings.

R/shinyui.R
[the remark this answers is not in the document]
cpsievert: > I believe we'd need a single dependency that is all_files = TRUE for the www/shared folder.

I don't think that's good a idea: other packages also include libraries like jQuery, ionRangeSlider, etc., and by bundling them together, we'd open the possibility of loading two, possibly different versions, of the same library

> should we move jQuery, jQuery legacy, shiny.js, and all other html deps that do not need to be in the shared folder out of www/shared? 

Seems reasonable to me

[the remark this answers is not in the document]
wch: I don't think we should move them out of `shared` at this point. If we do that, it would break any apps that supply their own `index.html` with hard-coded paths, like the [`08_html`](https://github.com/rstudio/shiny/blob/ed547fdf406a0518d20c078ebfd547559c173b8d/inst/examples/08_html/www/index.html) example.

If we do make that change, we should do it in a separate PR and figure out a way to provide a good migration path for users.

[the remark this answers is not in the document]
wch: Oh, and I agree that we should should use `all_files=FALSE`. I don't think anyone would use the html dependencies in the top-level `www/shared/` dir for statically-rendered documents (that dir just contains jquery, shiny, and shiny-showcase files), but just in case, `all_files=FALSE` wouldn't hurt.

PR #3550, opened 2021-11-17
Update stack trace test expectations

cpsievert:
These changes are ultimately a result of https://github.com/r-lib/testthat/commit/1a71c67e7a567d63c740950c7fe427a23ccb7668

It seems these changes are safe to accept since these differences only apply when capturing stack traces in a testthat context (also, it seems like `extractStackTraces()` isn't 100% correct anyway and isn't actually used anywhere outside of the tests??)


cpsievert:
@schloerke any ideas what's going on with https://github.com/rstudio/shiny/runs/4244810868?check_suite_focus=true#step:11:87 ?

wch (approved):

schloerke:
Merging!

Issue #3346, opened 2021-03-24
runGitHub looks for the "master" branch by default, but GitHub is changing to "main"


johnpauls:
In October of 2020, GitHub changed its terminology so that what used to be the `master` branch is now the `main` branch. 

`runGitHub()` has a default where it looks for the app on the `master` branch. So, if you make a new repo, `runGitHub` will be looking for a branch that does not exist. If you know this is the issue, you can just use the `ref="main` argument to fix this, but if you don't know that is the problem it is not obvious what is going on.

daattali:
Seeing as most existing repos still use `master`, in my opinion simply changing the default code from master to main will create as many problems as it solves. I think it's a valid point to bring up though, so maybe having better error messages could be a not-clean-but-practical solution: if 'master' fails but there's a 'main' branch, then the error message can suggest to the user that maybe they need to use 'main'

johnpauls:
I think that makes sense. The other thing to do is to address this in the documentation so that users will know to check to see if they are using 'master" or "main" in their repos.

wch:
@daattali That sounds like a good approach. Anyone want to make a PR?

k-doering-NOAA:
Another solution could be to use `ref = "HEAD"` as in [remotes::install_github](https://github.com/r-lib/remotes/blob/fcad17b68b7a19d5363d64adfb0a0426a3a5b3db/R/install-github.R#L54). This should reference whichever branch is the default one for the repository.

@wch I can submit this as a PR if it seems like this is the way to go, although it is a very simple change. I also don't mind someone else making the change!

jcheng5:
@k-doering-NOAA That sounds like the perfect solution! If you wouldn't mind making a PR, we'd certainly appreciate it.

schloerke:
Closed in #3564

PR #3564, opened 2021-12-29
Use HEAD for ref instead of master in runGitHub()

k-doering-NOAA:
Addresses #3346 using solution proposed in #3346 (comment)


schloerke (approved):
Approval in https://github.com/rstudio/shiny/issues/3346#issuecomment-1002762459 . 

No other use of `"master"` in R files. ✅

schloerke:
`routine` failure is a false-positive. Checks pass. Merging.

Issue #3566, opened 2022-01-07
`"shiny.silent.error"` conditions are intercepted by user handlers


lionel-:
Originally reported in https://github.com/tidyverse/dplyr/issues/5552.

Shiny silent errors thrown within dplyr verbs no longer work as expected since dplyr 1.0. That's because starting from this version, dplyr catches all conditions of class `"error"` and wraps them in a new dplyr-specific error.

```r
class(rlang::catch_cnd(dplyr::mutate(mtcars, stop("foo"))))
#> [1] "dplyr:::mutate_error" "rlang_error"
#> [3] "error"                "condition"

# Same with `req()`
class(rlang::catch_cnd(dplyr::mutate(mtcars, shiny::req(NULL))))
#> [1] "dplyr:::mutate_error" "rlang_error"
#> [3] "error"                "condition"
```

Here are two common ways of wrapping errors.

```r
# Chained wrapping
chain <- function(expr) {
  withCallingHandlers(expr, error = function(cnd) {
    rlang::abort(
      "Chained.",
      parent = cnd,
      call = rlang::env_parent()
    )
  })
}
chain(1 + "")
#> Error in `chain()`:
#> ! Chained.
#> Caused by error in `1 + ""`:
#> ! non-numeric argument to binary operator

# Non-chained wrapping
wrap <- function(expr) {
  withCallingHandlers(expr, error = function(cnd) {
    rlang::abort(
      "Wrapped.",
      body = c(x = conditionMessage(cnd)),
      call = rlang::env_parent()
    )
  })
}
wrap(1 + "")
#> Error in `wrap()`:
#> ! Wrapped.
#> ✖ non-numeric argument to binary operator
```

Another possible pattern is to intercept errors and demote them to a warning. This would also interfere with operators like `req()`, causing strang warning outputs.

In principle Shiny shouldn't use `"error"` conditions to communicate or transfer control between end points since these conditions may be intercepted and handled by user code. Ideally it would use bare conditions that don't inherit from `"error"`. However this is currently hard to implement because of promises which only support `"error"` handling.

As a partial fix we could handle the chained error case by using `cnd_inherits()` instead of `inherits()` to detect silent errors. dplyr versions 1.0.0 up to current 1.0.7 use both sorts of wrapping but the next release of dplyr will exclusively use chained errors though, AFAIK.

PR #3567, opened 2022-01-07
Handle chained errors

lionel-:
Closes tidyverse/dplyr#5552
Part of #3566. See this issue for context.

This PR makes Shiny compatible with dplyr error wrapping by detecting silent errors using `cnd_inherits()` instead of `inherits()`. This makes it compatible with rlang chained errors created with the `parent` argument of `rlang::abort()`.

Non-chained wrapped errors will still cause unexpected behaviours with `req()` and friends. Also error handlers that systematically demote errors to warnings will still interfere with `"shiny.silent.error"` errors.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,6 +11,8 @@ shiny 1.7.1.9001
 
 ### Bug fixes
 
+* Closed tidyverse/dplyr#5552: Compatibility of dplyr 1.0 (and rlang chained errors in general) with `req()`, `validate()`, and friends.
+
 * Closed #2955: Input and output bindings previously attempted to use `el['data-input-id']`, but that never worked. They now use `el.getAttribute('data-input-id')` instead. (#3538)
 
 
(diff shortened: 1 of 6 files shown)


jcheng5:
Wow, thanks for doing this! I was expecting to have to do it myself.

Can you confirm for me that this will work for both pre- and post- rlang 1.0? (Specifically, that `$parent` works for both of those cases)

lionel-:
> Can you confirm for me that this will work for both pre- and post- rlang 1.0? (Specifically, that $parent works for both of those cases)

Yup `cnd_inherits()` works with both chained and normal errors. It recurses over defined `parent` fields. If there is no parent it is equivalent to `inherits()`.

To be clear, chained errors have been implemented in rlang for a long time, they are just now becoming more apparent. dplyr has been using them since the 1.0 version released in May 2020. From then on, dplyr has been using a mix of chained errors (which this PR fixes) and wrapped errors (still to be fixed). The next version of dplyr to be released immediately after rlang 1.0 will use only chained errors and so should work correctly with Shiny once this PR is merged.

Are the integration test failures something I should do about? I see a bunch of failures locally as well, with and without the changes.

jcheng5:
I'll worry about the integration test failures. Your code changes look great--I'll take it from here. Thanks again!

jcheng5 (approved):

PR #3569, opened 2022-01-10
Pass args from knit_print.shiny.render.function() down to it's use of the knit_print() generic

### LINKED ISSUE #3568: Render functions used inside inline Rmd code blocks no longer render inline
A minimal reprex of https://github.com/rstudio/shinycoreci-apps/issues/199, which was introduced by https://github.com/rstudio/htmltools/pull/308. Note when `rmarkdown::run()`-ing this example, the result is 3 lines of text instead of 1:

````
---
runtime: shiny
title: "Untitled"
output: html_document
---

Some `r shiny::renderText("inline")` text.
````

cpsievert:
Closes #3568

This is especially important now that (as of https://github.com/rstudio/htmltools/pull/308) `knit_print.shiny.tag()` uses the inline argument to determine whether or not an inline code block should be used to preserve HTML.


schloerke (approved):
(JS/CSS changes are auto compiled)

schloerke:
- [ ] Wait for auto compilation before merging

PR #3582, opened 2022-02-08
Fix #3581: Errors in throttled/debounced reactive expressions cause session to exit

### LINKED ISSUE #3581: Errors in throttled/debounced reactive expressions cause session to exit
Launch this app and press "Cause error". It should just cause red errors to appear in the two outputs, but instead, the session exits. This is because internally, observers are used to implement the throttle/debounce behavior, and they are being exposed to the error in the reactive.

See #3580, thanks @mmuurr for pointing this out!

```r
library(shiny)

ui <- fluidPage(
  actionButton("err", "Cause error"),
  textOutput("output1"),
  textOutput("output2"),
)

  r <- eventReactive(input$err, {
    stop("Error at ", Sys.time())
  })
  
  r_throttle <- r |> throttle(1000)
  r_debounce <- r |> debounce(1000)
  
  output$output1 <- renderText({ r_throttle() })
  output$output2 <- renderText({ r_debounce() })
}

shinyApp(ui, server)
```

jcheng5:
I'd like to talk to someone about those couple of TODOs before calling this ready to merge

### DIFF
--- a/R/reactives.R
+++ b/R/reactives.R
@@ -2472,11 +2472,14 @@ debounce <- function(r, millis, priority = 100, domain = getDefaultReactiveDomai
 
       # Ensure r() is called only after setting firstRun to FALSE since r()
       # may throw an error
-      r()
+      try(r(), silent=TRUE)
       return()
     }
     # This ensures r() is still tracked after firstRun
-    r()
+    # TODO: If this fails, should we still update v$when?
+    # TODO: Does it make sense to update v$when before, rather
+    #       than after, executing r()?
+    try(r(), silent=TRUE)
 
     # The value (or possibly millis) changed. Start or reset the timer.
     v$when <- getDomainTimeMs(domain) + millis()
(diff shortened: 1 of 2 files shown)


R/reactives.R
```diff
@@ -2472,11 +2472,14 @@ debounce <- function(r, millis, priority = 100, domain = getDefaultReactiveDomai
...
       # may throw an error
-      r()
+      try(r(), silent=TRUE)
       return()
     }
     # This ensures r() is still tracked after firstRun
-    r()
+    # TODO: If this fails, should we still update v$when?
+    # TODO: Does it make sense to update v$when before, rather
+    #       than after, executing r()?
```
wch, line 2481: I'll start with the caveat that it's possible I'm misunderstanding the current behavior. That said...
I think it makes sense to reset the timer even in the case of an error. Since `debounce` will propagate errors from `r()`(I think), it makes sense to treat an execution of `r()` that throws error just like any other execution of `r()`.

Suppose there's a debouncer with a 3 second limit. Imagine `r()` fired once at time=0 with a good value, and then kept firing once per second for 4 more seconds, but threw errors each time, and then it fired once with a good value. It would be strange if, at time=3, the debouncer propagated the error, but `r()` kept executing each second.

wch:
It would be good to have tests and a NEWS item.

PR #3593, opened 2022-02-24
Change size 'xl' of modalDialog to 'l' if Bootstrap 3

gsmolinski:
(Problem mentioned also in the issue #3631)

**Motivation**

`modalDialog` with `size` set to `xl` in Bootstrap 3 is not supported (class 'modal-xl' is not supported). Problem is that this is a default Bootstrap version for Shiny, there is also no information about this issue in documentation. Now the behavior may be surprising for the user if working with Bootstrap 3 - user by chosen size `xl` probably wants the biggest modal dialog, but gets medium size. My intention was to change the size into `l` (large, 'modal-lg') in this case.

**What was changed**

* I have changed documentation for parameter size with information that in Bootstrap 3 size `xl` will be changed into `l` as `xl` is not supported in Bootstrap 3.
* I have added JS code to (1) check the version of Bootstrap; (2) if Bootstrap 3 and first child has class 'modal-xl', remove this class and add class 'modal-lg'.
* I have added entry to the NEWS.md to the 'Bug fixes'.

**Risks**

* I'm not sure if the code used to check the Bootstrap version is reliable.
* I'm taking the _first child_ of modal dialog to change the class - if the structure will change, the code won't work.
* Tested only manually (I have run `devtools::check()` and `devtools::test()`, get some fails  / warnings, but I think not related to my changes).

**Other comments**

* I think the added code should be removed once the Shiny will get as a default Bootstrap 4 or higher version.


cpsievert:
Thanks for PR. I'm not particularly keen on shimming `xl` to also work with Bootstrap 3. I'd rather just document it properly

gsmolinski:
I understand; I have prepared another commit and hopefully I didn't mess something (with git). This one only documents that in Bootstrap 3 instead of size `xl`, size `m` will be displayed.

R/modal.R
```diff
@@ -43,7 +43,8 @@ removeModal <- function(session = getDefaultReactiveDomain()) {
 #' @param title An optional title for the dialog.
 #' @param footer UI for footer. Use `NULL` for no footer.
 #' @param size One of `"s"` for small, `"m"` (the default) for medium,
-#'   or `"l"` for large.
+#'   `"l"` for large or `"xl"` for extra large (will be changed to `"m"`
+#'   in Bootstrap 3).
```
cpsievert, line 47:
```suggestion
#'   `"l"` for large, or `"xl"` for extra large. Note that `"xl"` only 
#'    works with Bootstrap 4 and above (to opt-in to Bootstrap 4+, 
#'    pass [bslib::bs_theme()] to the `theme` argument of a page container 
#'    like [fluidPage()])
```

gsmolinski:
Should I `document()` now and commit again because of this update?

cpsievert:
yes that'd be great

gsmolinski:
Thank you for review, manual is updated.

NEWS.md
```diff
@@ -41,6 +41,8 @@ shiny development
 
 * Closed rstudio/shinytest2#222: When restoring a context (i.e., bookmarking) from a URL, Shiny now better handles a trailing `=` after `_inputs_` and `_values_`. (#3648)
 
+* Closed #3631: documenting that `modalDialog` with parameter `size` set to `xl` will be changed to `m` in Bootstrap 3.
+
```
cpsievert, line 45:
```suggestion
```

PR #3599, opened 2022-03-11
Update `shinyAppTemplate()` content to use `{shinytest2}`

schloerke:
Fixes https://github.com/rstudio/shinytest2/issues/122

--------------------------------

As far as I can tell, `{shinytest}` was never directly used by `{shiny}`.  `{shinytest}` was only created as a byproduct of `shinyAppTemplate()` and in a test. Therefore, I believe we do not _need_ the dependency on `{shinytest}` or on `{shinytest2}`. 

For a followup release, we can add in the dependency of `{shinytest2}` to be safe, but it is nice to avoid the chicken / egg / CRAN problem until `{shinytest2}` is on CRAN.

----------------------------------

Changes to `shinyAppTemplate()`:
* Supports legacy choice values of `"shinytest"` and `"testthat"`
* Added new choice of `"tests"` for a generic _test_ name
* `./tests/testthat/test-shinytest2.R` test file is always added with `"tests"` option


R/app_template.R
```diff
@@ -45,20 +42,20 @@
...
-#'   snapshot-based testing.
 #' * `tests/testthat.R` is a test runner for test files in the
-#'   `tests/testthat/` directory using the [testthat](https://testthat.r-lib.org/) package.
+#'   `tests/testthat/` directory using the
+#'   [shinytest2](https://rstudio.github.io/shinytest2/reference/test_app.html)
+#'   package.
 #' * `tests/testthat/test-examplemodule.R` is a test for an application's module server function.
 #' * `tests/testthat/test-server.R` is a test for the application's server code
+#' * `tests/testthat/test-shinytest2.R` is a test that uses the
+#'   [shinytest2](https://rstudio.github.io/shinytest/) package to do
```
wch, line 52:
```suggestion
#'   [shinytest2](https://rstudio.github.io/shinytest2/) package to do
```

DESCRIPTION
```diff
@@ -105,13 +105,14 @@ Suggests:
...
     magrittr,
-    shinytest (>= 1.4.0.9003),
     yaml,
     future,
     dygraphs,
     ragg,
     showtext,
     sass
+Suggests/future:
+    shinytest
```
wch, line 115: This should be shinytest2? (I'm not sure what `Suggests/future` is for)

schloerke: That's twice now! 🤦‍♂️

> I'm not sure what `Suggests/future` is for

For a followup release, we can add in the dependency of `{shinytest2}` to be safe, but it is nice to avoid the chicken / egg / CRAN problem until `{shinytest2}` is on CRAN.

schloerke:
```suggestion
    shinytest2
```

schloerke, line 115:
```suggestion
Suggests/future:
    shinytest2
```

schloerke: Added a TODO in the test file

tools/documentation/checkDocsCurrent.sh
```diff
@@ -4,13 +4,15 @@ set -e
 
 # Generate package docs in the working directory
 echo "Document..."
+PRE_STATUS=$(git status --porcelain)
```
wch, line 7: Please add a comment explaining why things are done this way (for GHA)

wch (approved):

tests/testthat/test-test-shinyAppTemplate.R
```diff
@@ -0,0 +1,42 @@
+# testthat::skip_on_cran()
+# TODO-future; Add {shinytest2} to Suggests in DESCRIPTION once {shinytest2} is on CRAN
+suppressWarnings(testthat::skip_if_not_installed("shinytest2"))
```
wch, line 3: Please add `shinytest2` to `Config/Needs/check`.

tools/documentation/checkDocsCurrent.sh
schloerke: This was fixed by r-lib/actions. I've reverted the changes as the original code is easier to understand

schloerke:
All checks pass. Merging

PR #3628, opened 2022-05-04
Add ability for autoreload ws to be at a custom URL

jcheng5:
With this change, the `<script>` tag that loads shiny-autoreload.js can include a `data-ws-url` attribute to direct the WebSocket to use a custom URL.


schloerke:
Where is this being set or who is setting this `data-ws-url` attr?

schloerke (approved):
Current changes are solid. Just don't know who is setting the value

PR #3630, opened 2022-05-09
insertUI() now uses jQuery to insert HTML; and thus, now supports execution of <script>

### LINKED ISSUE #1545: When inserting javascript with insertUI, does it have to be wrapped in tags$head()? (Bug or feature?)

cpsievert:
Closes #1545

## TODO

A proper (shinyjster) testing app


srcts/src/shiny/singletons.ts
```diff
@@ -23,20 +22,29 @@ function renderHtml(
...
   if (where === "replace") {
     $(el).html(processed.html);
   } else {
-    let elElements: HTMLElement[];
-
-    if (el instanceof HTMLElement) {
-      elElements = [el];
-    } else {
-      elElements = el.toArray();
+    switch (where.toLowerCase()) {
```
schloerke, line 32: Should we also add the `replace` situation into the cases below?

schloerke (approved):
LGTM. Up to you if you'd like to fold in the `replace` case to the switch

PR #3638, opened 2022-05-18
Attempt another bind when registering a binding outside a renderHtml() context

### LINKED ISSUE #3635: Bindings can't be registered asynchronously
Consider a basic input binding:

```R
# app.R
library(shiny)

incrementButton <- function(inputId, value = 0) {
  tagList(
    tags$head(tags$script(src = "increment.js")),
    tags$button(id = inputId,
                class = "increment btn btn-default",
                type = "button",
                as.character(value))
  )
}

ui <- fluidPage(
  incrementButton("foo"),
  textOutput("foo_value")
)

  output$foo_value <- renderText({
    paste("Clicked this many times:", input$foo)
  })
}


shinyApp(ui, server)
```

```js
// www/increment.js

$(document).on("click", "button.increment", function(evt) {

  // evt.target is the button that was clicked
  var el = $(evt.target);

  // Set the button's text to its current value plus 1
  el.text(parseInt(el.text()) + 1);

  // Raise an event to signal that the value changed
  el.trigger("change");
});

var incrementBinding = new Shiny.InputBinding();
$.extend(incrementBinding, {
  find: function(scope) {
    return $(scope).find(".increment");
  },
  getValue: function(el) {
    return parseInt($(el).text());
  },
    $(el).text(value);
  },
    $(el).on("change.incrementBinding", function(e) {
      callback();
    });
  },
  unsubscribe: function(el) {
    $(el).off(".incrementBinding");
  }
});
```

Now, if we register synchronously all is fine, but not if we register in a `setTimeout()`, the binding doesn't bind to the DOM:

```js
setTimeout(() => { Shiny.inputBindings.register(incrementBinding); }, 100);
```

That use case may seem contrived, but becomes a real problem if you're trying to load JS dependencies asynchronously (e.g., via requirejs) to define/register the binding:

```r
incrementButton <- function(inputId, value = 0) {
  tagList(
    tags$head(
      tags$script(src="https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.6/require.min.js"
      tags$script(src = "increment.js")
    ),
    tags$button(id = inputId,
                class = "increment btn btn-default",
                type = "button",
                as.character(value))
  )
}
```

```js
require.config({paths: {"lodash": "https://cdnjs.cloudflare.com/ajax/libs/lodash.js/4.17.4/lodash.min"}});
require(["lodash"], function(_) {
  const finalBinding = _.last([incrementBinding]);
  Shiny.inputBindings.register(finalBinding);
})
```

cpsievert:
Closes #3635


srcts/src/shiny/render.ts
```diff
@@ -72,8 +72,13 @@ function renderHtml(
   dependencies: HtmlDep[],
   where: WherePosition = "replace"
 ): ReturnType<typeof singletonsRenderHtml> {
-  renderDependencies(dependencies);
-  return singletonsRenderHtml(html, el, where);
+  (window as unknown)["Shiny"]._renderingHtml = true;
```
cpsievert, line 75: Avoid window, keep track recursive count

srcts/src/bindings/registry.ts
```diff
@@ -10,6 +11,10 @@ interface BindingObj<Binding> {
   name?: string;
 }
 
+const debouncedBind = debounce(10, () => {
```
cpsievert, line 14:
```suggestion
const debouncedBind = debounce(0, () => {
```

srcts/src/shiny/init.ts
```diff
@@ -150,6 +150,20 @@ function initShiny(windowShiny: Shiny): void {
...
+  // When future bindings are registered via dynamic UI, check to see if renderHtml()
+  // is currently executing. If it's not, it's likely that the binding registration
+  // is occurring a tick after renderHtml()/renderContent(), in which case we need
+  // to make sure the new bindings get a chance to bind to the DOM. (#3635)
+  const maybeBindOnRegister = debounce(0, () => {
+    if (!renderHtml.isExecuting()) {
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      windowShiny.bindAll!(document.documentElement);
+    }
+  });
```
cpsievert, line 162:
```suggestion
  const debouncedBindAll = debounce(0, () => windowShiny.bindAll!(document.documentElement))

  const maybeBindOnRegister = () => {
    if (!renderHtml.isExecuting()) debouncedBindAll();
  };
```

srcts/src/shiny/init.ts
```diff
@@ -150,6 +150,22 @@ function initShiny(windowShiny: Shiny): void {
...
+  const debouncedBindAll = debounce(0, () =>
+    windowShiny.bindAll?.(document.documentElement)
+  );
+
+  const maybeBindOnRegister = () => {
+    if (!renderHtml.isExecuting()) debouncedBindAll();
+  };
+
+  inputBindings.onRegister(maybeBindOnRegister, false);
+  outputBindings.onRegister(maybeBindOnRegister, false);
```
wch, line 167: A rough suggestion:

```suggestion
  let enqueuedCount = 0;
  const enqueueRebind = () => {
    enqueuedCount++;

    ShinyApp.actionQueue.enqueue(() => {
      enqueuedCount--;
      // If this function is scheduled more than once in the queue, don't do anything.
      // Only do the bindAll when we get to the last instance of this function in the queue.
      if (enqueuedCount === 0) {
        windowShiny.bindAll?.(document.documentElement)
      }
    });
  };

  inputBindings.onRegister(enqueueRebind, false);
  outputBindings.onRegister(enqueueRebind, false);
```

Notes:
* Expand on the comment, otherwise we'll forget what this is all about.
* Check to make sure this callback doesn't fire during init.

PR #3654, opened 2022-06-24
Use `ragg::agg_png` over `Cairo::CairoPNG` if available

### LINKED ISSUE #3626: R 4.2.0: Fontconfig error: Cannot load default config file
After installing R 4.2.0 on Windows, I get 

`Fontconfig error: Cannot load default config file`

on every Shiny program creating a plot. Some old-style default font is used for plotting (Chrome Browser).  No warning when ggplot or default plots are created/printed without shiny.

I am aware that this message is not coming from Shiny, but rather from the operating system, and it has to do with some path settings (https://askubuntu.com/questions/492033/fontconfig-error-cannot-load-default-config-file). The reported cases were old and all referred to Linuxes.

I also tested with an Ubuntu installation, where the warning does not come up.

cpsievert:
Closes #3626

See also https://github.com/rstudio/rsconnect/pull/598

## Testing notes

None needed. We should already have some pretty good coverage of this in our automated tests.

### DIFF
--- a/R/imageutils.R
+++ b/R/imageutils.R
@@ -52,33 +52,31 @@ startPNG <- function(filename, width, height, res, ...) {
   grDevices::dev.cur()
 }
 
-#' Run a plotting function and save the output as a PNG
+#' Capture a plot as a PNG file.
 #'
-#' This function returns the name of the PNG file that it generates. In
-#' essence, it calls `png()`, then `func()`, then `dev.off()`.
-#' So `func` must be a function that will generate a plot when used this
-#' way.
-#'
-#' For output, it will try to use the following devices, in this order:
-#' quartz (via [grDevices::png()]), then [Cairo::CairoPNG()],
-#' and finally [grDevices::png()]. This is in order of quality of
-#' output. Notably, plain `png` output on Linux and Windows may not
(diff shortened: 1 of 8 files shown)


R/imageutils.R
```diff
@@ -1,19 +1,23 @@
...
-  # In the next shiny release, this option will likely be superseded in
-  # favor of a fully customizable graphics device option
-  if ((getOption('shiny.useragg') %||% FALSE) && is_installed("ragg")) {
-    pngfun <- ragg::agg_png
+  # Prefer ragg over Cairo, but only consider them over a quartz (i.e., Mac)
+  # device if user opts into that behavior (if quartz isn't available,
+  # ragg/cairo are an opt-out when installed).
+  use_ragg <- getOption('shiny.useragg')
+  use_cairo <- getOption('shiny.usecairo')
+  pngfun <- if (isTRUE(use_ragg) && is_installed("ragg")) {
```
cpsievert, line 7: Now that I think of it, maybe we should default to `{ragg}` (when installed) even when a quartz device available. That would make the custom fonts story a whole lot simpler (i.e., install `{ragg}` and custom fonts "just work" everywhere)

Issue #3657, opened 2022-07-01
Client-side throttling, used for brushes, click, and hover handlers, is broken as of 0fb154c (in main branch)


dvg-p4:
### System details

Browser Version: RStudio and Google Chrome (but 99% sure it will replicate elsewhere)

Output of `sessionInfo()`:

```
R version 4.2.0 (2022-04-22)
Platform: aarch64-apple-darwin20 (64-bit)
Running under: macOS Monterey 12.4

Matrix products: default
LAPACK: /Library/Frameworks/R.framework/Versions/4.2-arm64/Resources/lib/libRlapack.dylib

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] shiny_1.7.1.9003

loaded via a namespace (and not attached):
 [1] Rcpp_1.0.8.3      withr_2.5.0       digest_0.6.29     later_1.3.0       mime_0.12         R6_2.5.1          jsonlite_1.8.0    lifecycle_1.0.1   xtable_1.8-4     
[10] magrittr_2.0.3    cachem_1.0.6      rlang_1.0.2       cli_3.3.0         promises_1.2.0.1  jquerylib_0.1.4   bslib_0.3.1       ragg_1.2.2        ellipsis_0.3.2   
[19] textshaping_0.3.6 tools_4.2.0       httpuv_1.6.5      fastmap_1.1.0     compiler_4.2.0    systemfonts_1.0.4 htmltools_0.5.2   sass_0.4.1 
```

### Example application *or* steps to reproduce the problem

 

```R
# Adapted from example in plotOutput documentation
library(shiny)
options(shiny.trace=TRUE)
shinyApp(
  ui = basicPage(
    plotOutput("plot", height=300,
               hover = hoverOpts(id = "plot_hover", delayType = "throttle"),
    ),
    verbatimTextOutput("plot_hoverinfo")
  ),
  server = function(input, output, session) {
    output$plot <- renderPlot({
      d <- cars
      plot(d$speed, d$dist)
    })
    output$plot_hoverinfo <- renderPrint({
      cat("Hover (throttled):\n")
      str(input$plot_hover)
    })
  }
)
```

To have a permanently unresponsive hover, click outside the plot first.
To have a hover permanently stuck in the first location you hovered over, that spams the server with the same
[...]

PR #3659, opened 2022-07-01
Fix throttle.ts to make Throttler objects work as intended

dvg-p4:
There has been some faulty logic in `throttle.ts` that appears to have been there since the file's creation. However, a syntax error within the same appears to have mitigated the problem and made it function *almost* as intended until recently. 0fb154c fixed the syntax error, revealing the greater flaw in the logic. 

This pull request fixes the logic fully, closing #3657.


srcts/src/time/throttle.ts
```diff
@@ -46,13 +44,28 @@ class Throttler<X extends AnyVoidFunction> implements InputRatePolicy<X> {
...
+    // set this.timerId to a newly-created timer, which will invoke a call
+    // with the most recently called args (if any) when it expires
+    this.timerId = setTimeout(() => {
+      // IE8 doesn't reliably clear timeout, so this additional
+      // check is needed
+      if (this.timerId === null) return;
+
+      this.$clearTimer();
+      // Do we have a call queued up? (check existence rather than length)
+      if (this.args) {
```
wch, line 64: Can you make this explicit? That will help reduce the possibility of confusion in the future about what counts as falsy.

```suggestion
      if (this.args === null) {
```

wch: Actually, now that I think about it, I still think this code could lead to confusion.

Specifically, when `args` is an array, that means that a call is scheduled, and when `args` is `null` means that no call is scheduled.

Elsewhere, in `normalCall`, this uses `this.timerId === null` to check whether a call is scheduled.

It would be clearer to consistently use the same condition.

wch:
Thanks for the PR. After looking at some more, I still find the code a bit confusing. I think it may be clearer (and more reliable to use the lodash `throttle` method and have a thin wrapper for it.

wch:
One more note: we are very close to releasing a new version of Shiny and it would be great if we could get a fix for this issue in there.

srcts/src/time/throttle.ts
dvg-p4: Actually, those are intentionally testing different things. `timerId` checks whether there was an invocation within the last interval; `args` checks if there’s a call waiting.

dvg-p4: Making the bill check explicit seems wise though

dvg-p4:
> Thanks for the PR. After looking at some more, I still find the code a bit confusing. I think it may be clearer (and more reliable to use the lodash `throttle` method and have a thin wrapper for it.

Do we already import lodash or would this be a new import?

srcts/src/time/throttle.ts
dvg-p4: Oh `isPending()` should be using `this.args` I'll fix that

srcts/src/time/throttle.ts
```diff
@@ -19,40 +19,62 @@ class Throttler<X extends AnyVoidFunction> implements InputRatePolicy<X> {
...
   immediateCall(...args: Parameters<X>): void {
     this.$clearTimer();
     this.args = args;
     this.$invoke();
   }
+
+  // Is there a call waiting to send?
   isPending(): boolean {
-    return this.timerId !== null;
+    return this.args !== null;
```
wch, line 44: Note that this changes the semantics of `isPending()`, compared to the old version, before this code was refactored:
https://github.com/rstudio/shiny/blob/v1.6.0/srcjs/input_rate.js#L90-L92

I think the new semantics are correct, but I'll need to check whether this will cause any problems in existing code.

srcts/src/time/throttle.ts
```diff
@@ -19,40 +19,62 @@ class Throttler<X extends AnyVoidFunction> implements InputRatePolicy<X> {
...
+    // set this.timerId to a newly-created timer, which will invoke a call
+    // with the most recently called args (if any) when it expires
+    this.timerId = setTimeout(() => {
+      // IE8 doesn't reliably clear timeout, so this additional
+      // check is needed
+      if (this.timerId === null) return;
+
+      this.$clearTimer();
+      // Do we have a call queued up? (check existence rather than length)
+      if (this.args !== null) {
```
wch, line 73: Should use `this.isPending()`

wch:
A few small things; I'll fix and merge.

PR #3663, opened 2022-07-07
Load script tags normally, instead of with jQuery's synchronous `XMLHttpRequest`

### LINKED ISSUE #789: Synchronous XMLHttpRequest on the main thread is deprecated
I'm getting the warning

```
Synchronous XMLHttpRequest on the main thread is deprecated because
of its detrimental effects to the end user's experience.
For more help, check http://xhr.spec.whatwg.org/.
```

when loading an app in the `shinyapps.io` service, due to the use of `c.async="false"` in column 16511 of `jquery.min.js` or `xml.async = "false";` in line 8567 of `jquery.js`.

It prevents the use of `$( document ).ready()` to execute js code when the document finishes loading.

wch:
Closes #789.

When a `script` tag is added to the document with jQuery, as in `$el.append(scriptTag)`, it does some work behind the scenes to make a _synchronous_ XMLHttpRequest, which is out of line with modern JS code.

For example, Safari displays this warning:
> Synchronous XMLHttpRequest on the main thread is deprecated because of its detrimental effects to the end user’s experience. For more help http://xhr.spec.whatwg.org/

This PR avoids the sync XHR request, by adding the `script` tag using plain JavaScript. After the tags are loaded, it re-binds all. This is probably not a great long-term solution, as it does one extra binding. The "proper" way to do this would be to do all of this asynchronously. That would probably require making changes several levels up the call stack.

Note that this PR is currently targeted against the `bindAfterRegister` branch, but we may want to re-target it to main.

### DIFF
--- a/srcts/src/shiny/render.ts
+++ b/srcts/src/shiny/render.ts
@@ -240,12 +240,6 @@ function renderDependency(dep_: HtmlDep) {
   // that way.)
   document.head.append(...scriptElements);
 
-  // After the scripts are all loaded, re-bind all.
-  Promise.allSettled(scriptPromises).then(() => {
-    shinyInitializeInputs(document.body);
-    shinyBindAll(document.body);
-  });
-
   dep.attachment.forEach((x) => {
     const link = $("<link rel='attachment'>")
       .attr("id", dep.name + "-" + x.key + "-attachment")
@@ -254,12 +248,22 @@ function renderDependency(dep_: HtmlDep) {
     $head.append(link);
   });
 
-  if (dep.head) {
-    const $newHead = $("<head></head>");
+  Promise.allSettled(scriptPromises).then(() => {
+    // After the scripts are all loaded, insert any head content. This may
+    // contain <script> tags with inline content, which we want to execute after
+    // the script elements above, because the code here may depend on them.
+    if (dep.head) {
+      const $newHead = $("<head></head>");
 
-    $newHead.html(dep.head);
-    $head.append($newHead.children());
-  }
+      $newHead.html(dep.head);
+      $head.append($newHead.children());
+    }
+
+    // Bind all
+    shinyInitializeInputs(document.body);
+    shinyBindAll(document.body);
+  });
+
   return true;
 }
 


wch:
The code here first (asynchronously) loads `<script>` tags, then inserts `<head>` content, then does `initializeInputs()` and `bindAll()`. This the same order that happens in the main branch.

However, there is one important difference: because the script loading is async, and the insertion of head content and the `initializeInputs()` + `bindAll()` happens after the script-loading promises resolve, there are other things that happen in the synchronous code path before the scripts are loaded. One of the synchronous things is that it calls `initializeInputs()` and `bindAll()` -- this is separate form the one in the promise chain mentioned above. 

So the order of events is roughly:
* Insert `<script>` tags into the page, and return a promise that resolves when they're all loaded.
* Meanwhile, after the `<script>` tags have been inserted but _before_ they load (and the promise resolves), the synchronous code continues on and calls `initializeInputs()` and `bindAll()` elsewhere.
* The scripts load and the promise resolves.
* The next step in the promise chain injects other content into the document head, then runs `initializeInputs()` and `bindAll()`.

The can result in errors when the first `initializeInputs()` expects the JS script to already be loaded.

In the example below, switching to the date input results in just this kind of error:

```R
ui <- fluidPage(
  fluidRow(
    column(3, wellPanel(
      selectInput("input_type", "Input type",
        c("text", "slider", "numeric", "checkbox",
          "checkboxGroup", "radioButtons", "selectInput",
          "selectInput (multi)", "date", "daterange"
        )
      )
    )),
    
    column(3, wellPanel(
      # This outputs the dynamic UI component
      uiOutput("ui")
    )),
    
    column(3,
      tags$p("Input type:"),
      verbatimTextOutput("input_type_text"),
      tags$p("Dynamic input value:"),
      verbatimTextOutput("dynamic_value")
    )
  )
)


  output$ui <- renderUI({
    if (is.null(input$input_type))
      return()
    
    # Depending on input$input_type, we'll generate a different
    # UI component and send it to the client.
    switch(input$input_type,
      "slider" = sliderInput("dynamic", "Dynamic",
        min = 1, max = 20, value = 10),
      "text" = textInput("dynamic", "Dynamic",
        value = "starting value"),
      "numeric" =  numericInput("dynamic", "Dynamic",
        value = 12),
      "checkbox" = checkboxInput("dynamic", "Dynamic",
        value = TRUE),
      "checkboxGroup" = checkboxGroupInput("dynamic", "Dynamic",
        choices = c("Option 1" = "option1",
          "Option 2" = "option2"),
        selected = "option2"
      ),
      "radioButtons" = radioButtons("dynamic", "Dynamic",
        choices = c("Option 1" = "option1",
          "Option 2" = "option2"),
        selected = "option2"
      ),
      "selectInput" = selectInput("dynamic", "Dynamic",
        choices = c("Option 1" = "option1",
          "Option 2" = "option2"),
        selected = "option2"
      ),
      "selectInput (multi)" = selectInput("dynamic", "Dynamic",
        choices = c("Option 1" = "option1",
          "Option 2" = "option2"),
        selected = c("option1", "option2"),
        multiple = TRUE
      ),
      "date" = dateInput("dynamic", "Dynamic"),
      "daterange" = dateRangeInput("dynamic", "Dynamic")
    )
  })
  
  output$input_type_text <- renderText({
    input$input_type
  })
  
  output$dynamic_value <- renderPrint({
    str(input$dynamic)
  })
}

shinyApp(ui, server)
```

The first time you switch to the date, it has these errors:
image

But on subsequent times switching to the date input, it works fine, because the JS files have already been loaded.

So we'll probably need another approach, like making the whole call stack async.

PR #3665, opened 2022-07-07
Fix sporadic dates (#3664)

### LINKED ISSUE #3664: dateInput and dateRangeInput send spurious values to server mid-typing
### System details

Browser Version: RStudio, Google Chrome

Output of `sessionInfo()`:

```
R version 4.2.0 (2022-04-22)
Platform: aarch64-apple-darwin20 (64-bit)
Running under: macOS Monterey 12.4

Matrix products: default
LAPACK: /Library/Frameworks/R.framework/Versions/4.2-arm64/Resources/lib/libRlapack.dylib

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] shiny_1.7.1.9003

 [1] Rcpp_1.0.8.3     withr_2.5.0      digest_0.6.29    later_1.3.0      mime_0.12       
 [6] R6_2.5.1         jsonlite_1.8.0   lifecycle_1.0.1  xtable_1.8-4     magrittr_2.0.3  
[11] cachem_1.0.6     rlang_1.0.2      cli_3.3.0        promises_1.2.0.1 jquerylib_0.1.4 
[16] bslib_0.3.1      ellipsis_0.3.2   tools_4.2.0      httpuv_1.6.5     fastmap_1.1.0   
[21] compiler_4.2.0   htmltools_0.5.2  sass_0.4.1      
```

```R
library(shiny)
shinyApp(
  ui = basicPage(
    textOutput("date_out"),
    dateInput("date_in", "Date input", "2000-01-01")
  ),
    output$date_out <- renderText(as.character(input$date_in))
  }
)
```

Click into the text field of the date input and press backspace; the date displayed will be 1999-12-31. Press backspace again; it will be today's date. Continue pressing backspace; it will be null when the field is empty.

`dateInput` (and `dateRangeInput`) eagerly try to parse incomplete date strings as they are being typed, and then send off the results to the server. This can have the effect of the server trying to process a date completely different from the one that the user intended to enter. In the case of a range, this can result in a min and max in the wrong order; and with a cleared-out field that the user intends to start typing a new date in, the server sees a null value. The server will then take actions with the bogus date before the user is even finished typing (some of which may prevent the user from completing their typing, e.g. updating the input to be in the "correct" order or throwing an error for a null date).

This might stem from JavaScripts officially undefined behavior when parsing nonstandard datestrings; I'd recommend ignoring typed input in this field until the user presses enter or unfocuses the field; or only updating when the current string matches the set format.

dvg-p4:
Closes #3664.

This pull request removes the handlers for `keyup` and `input` from `dateInput` and `dateRangeInput`. This prevents spurious updates while typing, but still sends the parser's best interpretation of the field when enter is pressed, focus is lost (i.e. the user clicks out), or a date selection is made in the GUI (due to the remaining `changeDate` and `change` handlers).

A value of `null` (for `dateInput`) or `NA` (for `dateRangeInput`) is still sent immediately when the field is cleared, but this can be more easily checked for and ignored on the server side. I suspect that `bootstrap-datepicker` is interpreting the empty string as a special value and firing off the `changeDate` event, and stopping that would probably be much more intensive than removing a few lines of code.

~~Not sure if y'all will want to shove this into 1.7.2, leave it for the next release, or if there's some crazy use case I can't imagine where you'd *want* "2001-01-0" to be immediately interpreted as the last day of 1999.~~

Edit: Merged in 1.7.2-1.7.4 changes, no conflicts except for minified JS as usual.

Edit 2: Merged in all changes up to the current development version 1.9.1.9000, again no conflicts except for minified JS.


dvg-p4:
I think that errors is inherited from the previous commit on the main branch that I just merged in--any idea what the deal with that is / how to fix it @wch ?

dvg-p4:
On further investigation, it seems that the problem was my own darn fault. I needed to update my roxygen2 from 7.2.0 to 7.2.1, so that devtools::check() would actually redocument the package, fixing the version number in package.json, so that the github actions wouldn't try to do that itself, get mad at itself for doing it, and fail the check. Probably.

dvg-p4:
...nope, looks like even with updated roxygen, devtools::check() still doesn't update package.json; I need to manually run `source .github/shiny-workflows/routine.sh`. Or maybe I could have just manually updated the package version.

dvg-p4:
@wch whether or not y'all accept this PR you should probably update the package.json version number on main

dvg-p4:
This will probably fail the R 3.5 test for the same reason as https://github.com/rstudio/shiny/pull/3683#issuecomment-1470033399 and main

dvg-p4:
It looks like some other checks also failed due to R 4.2.3 being released today but not quite making it to the expected repositories yet

dvg-p4:
Is there a way to rerun the tests here?

dvg-p4:
Looks like everyone else's PRs are also getting the same error on ubuntu-20.04 (4.1.3), so that's almost certainly not my fault.

The yarn stuff might be, but I just ran the standard commands -- `yarn install`; `yarn build`, with the latest yarn and npm installed; so I'm not sure what's going wrong there. It looks like the checksum for selectize is changing, despite being pinned to a specific commit?

dvg-p4:
And I'm also not sure why `license/cla Expected — Waiting for status to be reported` is hanging--I signed the CLA 2 years ago. Actually maybe that's the problem if you've updated it since, but the bot doesn't realize it and isn't giving me the right link to click.

image

@cpsievert any idea how to fix this?

dvg-p4:
Aha, the yarn issue may be due to https://github.com/yarnpkg/berry/issues/2399 / https://github.com/yarnpkg/berry/issues/2774#issuecomment-822472844, since I'm building on a Mac and the github checks are on Ubuntu.

...but, I get the same checksum when building on a Rocky 8 box, so maybe that's not it. Unless MacOS and Rocky8 match while Ubuntu is different, but that'd be strange.

dvg-p4:
Downgrading my node/npm to match the versions listed [here](https://github.com/rstudio/shiny/actions/runs/11019045984/job/30600677061?pr=3665#step:12:23) does not change my checksum for this package.

dvg-p4:
Hah, looks like that got the test to pass...might be worth looking into what exactly's wrong with the checksum, but I suspect it's a yarn bug rather than a shiny bug.

dvg-p4:
Remaining test failure is due to https://github.com/rstudio/shiny/issues/4133

gadenbuie (approved):
Thanks @dvg-p4 for the fix and for patiently keeping this PR up to date. We agree that the current behavior of sending eager updates is problematic and that your solution of not reporting updates on keydown is a good choice.

I have a few minor edits to comments and the news entry. I've also manually verified that we do have your CLA on file; I wasn't able to figure out what is happening with the cla-assistant but we don't need the check.

Thanks again!

PR #3666, opened 2022-07-08
Take 2: load script tags normally, instead of with jQuery's synchronous `XMLHttpRequest`

wch:
This is an alternate take on #3663. It converts `renderDependency()` to an `async` function, as well as functions would be on the call stack, all the way down to `renderContent()`.

In this PR, the functions which call `renderContent()` are _not_ converted to `async`. There are a few places where `renderContent` is called multiple times, and I think it might make sense to convert the caller functions to `async` and then `await` each `renderContent()` call (or just use a promise chain).

This version does not have the same initialization order problems that #3663 has -- the date input works fine.

Test app:

```R
library(shiny)

ui <- fluidPage(
  fluidRow(
    column(3, wellPanel(
      selectInput("input_type", "Input type",
        c("text", "numeric", "checkbox",
          "checkboxGroup", "radioButtons", "selectInput",
          "selectInput (multi)", "date", "daterange"
        )
      )
    )),
    column(3, wellPanel(
      uiOutput("ui")
    )),
    column(3,
      tags$p("Input type:"),
      verbatimTextOutput("input_type_text"),
      tags$p("Dynamic input value:"),
      verbatimTextOutput("dynamic_value")
    )
  )
)

  output$ui <- renderUI({
    if (is.null(input$input_type))
      return()
    
    switch(input$input_type,
      "slider" = sliderInput("dynamic", "Dynamic", min = 1, max = 20, value = 10),
      "text" = textInput("dynamic", "Dynamic", value = "starting value"),
      "numeric" =  numericInput("dynamic", "Dynamic", value = 12),
      "checkbox" = checkboxInput("dynamic", "Dynamic", value = TRUE),
      "checkboxGroup" = checkboxGroupInput("dynamic", "Dynamic",
        choices = c("Option 1" = "option1", "Option 2" = "option2"),
        selected = "option2"
      ),
      "radioButtons" = radioButtons("dynamic", "Dynamic",
        choices = c("Option 1" = "option1", "Option 2" = "option2"),
        selected = "option2"
      ),
      "selectInput" = selectInput("dynamic", "Dynamic",
        choices = c("Option 1" = "option1", "Option 2" = "option2"),
        selected = "option2"
      ),
      "selectInput (multi)" = selectInput("dynamic", "Dynamic",
        choices = c("Option 1" = "option1", "Option 2" = "option2"),
        selected = c("option1", "option2"),
        multiple = TRUE
      ),
      "date" = dateInput("dynamic", "Dynamic"),
      "daterange" = dateRangeInput("dynamic", "Dynamic")
    )
  })

  output$input_type_text <- renderText({
    input$input_type
  })

  output$dynamic_value <- renderPrint({
    str(input$dynamic)
  })
}

shinyApp(ui, server)
```


tsconfig.json
```diff
@@ -1,12 +1,13 @@
 {
   "declaration": true,
   "compilerOptions": {
-    "target": "ES5",
+    "target": "es2020",
```
wch, line 4: This is necessary to use `Promise.allSettled()`. It doesn't seem to change any of the generated code in shiny.js. I believe it's because this is just the TypeScript config, and the actual compilation/transpilation is being done by something else.

The `"moduleResolution": "node"` was necessary after switching to `es2020`.

srcts/src/shiny/shinyapp.ts
```diff
@@ -633,19 +633,19 @@ class ShinyApp {
...
+  ): Promise<void> {
     // Dispatch messages to handlers, if handler is present
     for (let i = 0; i < handlerOrder.length; i++) {
       const msgType = handlerOrder[i];
 
       if (hasOwnProperty(msgObj, msgType)) {
         // Execute each handler with 'this' referring to the present value of
         // 'this'
-        handlers[msgType].call(this, msgObj[msgType]);
+        await handlers[msgType].call(this, msgObj[msgType]);
```
schloerke, line 648: This now allows for message handlers to be `async`/`await`-able

srcts/src/shiny/render.ts
```diff
@@ -209,9 +213,28 @@ function renderDependency(dep_: HtmlDep) {
...
+      // <script> tags, browsers set async to true by default, which differs
+      // from static <script> tags in the html, which default to false.
+      //
+      // Refs:
+      // https://stackoverflow.com/a/8996894/412655
+      // https://jason-ge.medium.com/dynamically-load-javascript-files-in-order-5318ac6bcc61
+      //
+      // Note that one odd thing about these dynamically-created <script> tags
+      // is that even though the JS object's `x.script` property is true, it
+      // does NOT show up as a property on the <script> element.
```
wch, line 234: For future reference, this is what happens when I run this in Chrome. The dynamic `script` tag has `script.async` set to `true`, but it doesn't show it on the DOM element, unless it's explicitly set.


```R
let script = document.createElement('script');
#> <script>​</script>​

script.async
#> true
script
#> <script>​</script>​

script.async = true
script
#> <script async>​</script>​
```

Also, setting it to `false` makes no visible change on the DOM element.

```R
let script = document.createElement('script');
#> <script>​</script>​

script.async
#> true
script
#> <script>​</script>​

script.async = false
script.async
#> false
script
#> <script>​</script>
```

srcts/src/shiny/render.ts
```diff
@@ -220,9 +243,24 @@ function renderDependency(dep_: HtmlDep) {
...
+    scriptPromises.push(p);
+    scriptElements.push(script);
   });
 
+  // Append the script elements all at once, so that we're sure they'll load in
+  // order. (We didn't append them individually in the `forEach()` above,
+  // because we're not sure that the browser will load them in order if done
+  // that way.)
+  document.head.append(...scriptElements);
+  await Promise.allSettled(scriptPromises);
```
schloerke, line 262: Why `Promise.allSettled()` and not `Promise.all()`? (We are not using the result.)

schloerke: I still don't see why the promise is needed as long as the scripts are loaded in order.

If they're loaded in order, then the whole async/await changes can be removed.

srcts/src/shiny/render.ts
```diff
@@ -237,6 +275,7 @@ function renderDependency(dep_: HtmlDep) {
     $newHead.html(dep.head);
     $head.append($newHead.children());
```
schloerke, line 276: Needs to use the new `document.head.append()` approach?

srcts/src/shiny/render.ts
```diff
@@ -220,9 +243,24 @@ function renderDependency(dep_: HtmlDep) {
...
+  // Append the script elements all at once, so that we're sure they'll load in
+  // order. (We didn't append them individually in the `forEach()` above,
+  // because we're not sure that the browser will load them in order if done
+  // that way.)
+  document.head.append(...scriptElements);
+  await Promise.allSettled(scriptPromises);
+
   dep.attachment.forEach((x) => {
     const link = $("<link rel='attachment'>")
       .attr("id", dep.name + "-" + x.key + "-attachment")
```
schloerke, line 266: Should this be attached using the new `document.head.append()` approach?

srcts/src/shiny/shinyapp.ts
schloerke: This spooks me as we are adding the possibility that the handling of messages can occur in multiple steps, allowing other events to happen in between processing. 

Before, the _handshake_ was that all messages were handling in a single tick. Now, the execution could take many ticks and introduce many edge cases that are hard to reprex.

schloerke (changes requested):
I'd like to remove the async/await chagnes as much as possible. As long as we know the scripts are in order, I do not see the need for the promise logic

srcts/src/shiny/render.ts
wch: > Why `Promise.allSettled()` and not `Promise.all()`? (We are not using the result.)

The `.allSettled()` is there in case there's an error loading any script tags. If it used `.all()`, that would require all the loads to be successful before moving on.

wch: The `await` is necessary because we don't want to continue on this code path until all the scripts have run.

For example, we don't want run `bindInput` to initialize a slider element before the slider JS files have loaded.

wch:
> I'd like to remove the async/await chagnes as much as possible. As long as we know the scripts are in order, I do not see the need for the promise logic

The promises and awaits are necessary because the scripts inserted on the page can interact with other code in the application. See this comment in a previous attempt at a PR about the problems that can occur: https://github.com/rstudio/shiny/pull/3663#issuecomment-1178498378

(Those problems are the reason for this PR instead of the previous one.)

srcts/src/shiny/shinyapp.ts
wch: The `async`-ness is unfortunately necessary because `renderDependency` needs to be `async`, and that has implications all the way up the chain.

Adding the `await` here doesn't actually change whether or not a handler can be `async`. Even prior to this change, they could be `async`; it's just that they would run, create a promise, and then the code would move on without waiting for the promise to resolve.

srcts/src/shiny/shinyapp.ts
```diff
@@ -253,6 +258,15 @@ class ShinyApp {
     return socket;
   }
 
+  async startInMessageQueueLoop(): Promise<void> {
```
wch, line 264: Should wrap the content of this function in a `try`-`catch` and print any errors to the console.

wch:
Dang. Maybe it makes sense to keep the old synchronous functions (with deprecation messages), and add new async functions? We may be able to do it in a way that shares almost all code. 

FWIW, there are many instances of those functions that turn up in the code search, but they appear to come from just a handful of packages, like shinydashboard, shinydashboardPlus, and bs4dash.

jcheng5:
> Dang. Maybe it makes sense to keep the old synchronous functions (with deprecation messages), and add new async functions? We may be able to do it in a way that shares almost all code.

I was thinking that too.

> FWIW, there are many instances of those functions that turn up in the code search, but they appear to come from just a handful of packages, like shinydashboard, shinydashboardPlus, and bs4dash.

I think that's true for `renderHtml` but not for `renderContent`?

wch:
For future reference, @schloerke and I discussed the following: Instead of having two separate implementations of each function, like `renderHtml` and `renderHmtlAsync`, do something similar to the `hybrid_chain` function we use in R. It would use the [synchronous-promise](https://www.npmjs.com/package/synchronous-promise) package from npm. The interface would be something like this:

```ts
import { SynchronousPromise } from "synchronous-promise";

function promiseWrapper(type: "async" | "sync", fn) {
  if (type === "async") {
    return new Promise(fn);
  } else {
    return new SynchronousPromise(fn);
  }
}

function foo(x) {
  fooImpl("sync", x)
}

function fooAsync(x) {
  fooImpl("async", x)
}

function fooImpl(type: "async" | "sync", x) {
  promiseWrapper(type, (resolve, reject) => {

    // Do stuff here...
    if (type === "async") {
      resolve(doSomething());  // doSomethingAsync() returns a normal value
    } else {
      resolve(doSomethingAsync()); // doSomethingAsync() returns a Promise
    }
  }).then((value) => {
    // Do more stuff here
  });
}
```

There are some drawbacks to doing this:
* It brings in another package, one which is primarily meant to be used in tests.
* It can complicate code that uses, especially for functions which have early return statements.

I think the drawbacks outweigh the benefits here. If in the future we decide to do a lot more sync + async versions of functions, that might make it worthwhile.

srcts/src/shiny/render.ts
```diff
@@ -354,6 +367,209 @@ function addStylesheetsAndRestyle(links: HTMLLinkElement[]): void {
...
+    scriptPromises.push(p);
+    scriptElements.push(script);
+  });
+
+  // Append the script elements all at once, so that we're sure they'll load in
+  // order. (We didn't append them individually in the `forEach()` above,
+  // because we're not sure that the browser will load them in order if done
+  // that way.)
+  document.head.append(...scriptElements);
+  await Promise.allSettled(scriptPromises);
```
jcheng5, line 469: This will return an array of objects containing the status of the promises, what do we do if one or more of them fail?

jcheng5: Basically my goal is that one day when we show unhandled exceptions in the browser UI (not just in the JS console) I want these to be included. Maybe `window.addEventListener("error")` will already catch these

srcts/src/shiny/shinyapp.ts
```diff
@@ -253,6 +263,19 @@ class ShinyApp {
     return socket;
   }
 
+  async startActionQueueLoop(): Promise<void> {
+    // eslint-disable-next-line no-constant-condition
+    while (true) {
+      const action = await this.actionQueue.dequeue();
+
+      try {
+        action();
```
jcheng5, line 272: `await action();`

srcts/src/bindings/outputAdapter.ts
```diff
@@ -31,8 +31,8 @@ class OutputBindingAdapter {
   getId(): string {
     return this.binding.getId(this.el);
   }
-  onValueChange(data: unknown): void {
-    this.binding.onValueChange(this.el, data);
+  async onValueChange(data: unknown): Promise<void> {
+    await this.binding.onValueChange(this.el, data);
```
jcheng5, line 35:
```suggestion
  onValueChange(data: unknown): Promise<void> | void {
    return this.binding.onValueChange(this.el, data);
```

jcheng5 (changes requested):

srcts/src/shiny/render.ts
wch: Note that I've changed this to `Promise.all()`. It won't return an array, but in the `script.onerror` handler above, we can add code to log it to the to the UI.

srcts/src/shiny/render.ts
```diff
@@ -354,6 +367,209 @@ function addStylesheetsAndRestyle(links: HTMLLinkElement[]): void {
...
+      //
+      // Note that one odd thing about these dynamically-created <script> tags
+      // is that even though the JS object's `x.script` property is true, it
+      // does NOT show up as a property on the <script> element.
+      script.async = false;
+    }
+
+    Object.entries(x).forEach(function ([attr, val]) {
+      if (attr === "src") {
+        val = encodeURI(val);
```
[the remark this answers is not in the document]
wch: It'll get over-escaped. This escaping was already present, so it seems the expectation has been that it will receive an unescaped URL from the server side.

```js
encodeURI("%")
# '%25'
```

srcts/src/shiny/render.ts
```diff
@@ -354,6 +367,209 @@ function addStylesheetsAndRestyle(links: HTMLLinkElement[]): void {
...
+  // (like scripts) aren't added, because they would have been added in a
+  // previous run.
+  if (needsRestyle(dep)) {
+    addStylesheetsAndRestyle(getStylesheetLinkTags(dep));
+    return true;
+  }
+
+  if (hasDefinedProperty(htmlDependencies, dep.name)) return false;
+
+  registerDependency(dep.name, dep.version);
```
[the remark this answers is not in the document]
wch: I think this possible. I'd like to do it in a separate PR though.

srcts/src/shiny/render.ts
```diff
@@ -354,6 +367,212 @@ function addStylesheetsAndRestyle(links: HTMLLinkElement[]): void {
...
+
+    scriptPromises.push(p);
+    scriptElements.push(script);
+  });
+
+  // Append the script elements all at once, so that we're sure they'll load in
+  // order. (We didn't append them individually in the `forEach()` above,
+  // because we're not sure that the browser will load them in order if done
+  // that way.)
+  document.head.append(...scriptElements);
```
wch, line 471: Instead of making the `scriptElements` array and appending all together, they can be added in the loop above. (The existing comment referred to some previous version of the code.)

wch:
I've created a test app at https://gist.github.com/wch/0f9e850d92c2d006feb0b27ec1de944f.

jcheng5 (approved):

jcheng5:
Given the severity of the change, can you bump version and add NEWS for this before merging?

PR #3676, opened 2022-07-28
Switch to Font Awesome 6 icon names

ismirsehregal:
Same procedure as described here:

https://github.com/dreamRs/shinyWidgets/pull/516#issue-1320737774

### DIFF
--- a/R/bootstrap.R
+++ b/R/bootstrap.R
@@ -1263,7 +1263,7 @@ downloadLink <- function(outputId, label="Download", class=NULL, ...) {
 #' submitButton("Update View", icon = icon("arrow-rotate-right"))
 #'
 #' navbarPage("App Title",
-#'   tabPanel("Plot", icon = icon("bar-chart-o")),
+#'   tabPanel("Plot", icon = icon("chart-column")),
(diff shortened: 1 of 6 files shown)


cpsievert:
Thanks @ismirsehregal, do you know if any of the current examples will fail with `{fontawesome}` 0.3.0? If not, I don't think we need to make these changes right away, but if/when we do, we should be requiring `fontawesome >= 0.3.0` in the `DESCRIPTION`

ismirsehregal:
Hi, @cpsievert no the current examples won't fail with {fontawesome} 0.3.0.

This is just to avoid warnings like:

```
This Font Awesome icon ('level-up-alt') does not exist:
* if providing a custom `html_dependency` these `name` checks can 
  be deactivated with `verify_fa = FALSE`
```

Accordingly, it is not urgent to merge this PR.

Cheers

ismirsehregal:
The latest version of {[fontawesome](https://github.com/rstudio/fontawesome/releases/tag/v0.4.0)} allows to reference icons by several of their previous names. Leaving this open in case you still want to update the icon names, otherwise feel free to close the PR.

PR #3702, opened 2022-09-16
Allow selection of minimum and maximum view mode

MEO265:
Use the options of the bootstrap datepicker to limit the view mode


MEO265:
Anyone wanna check?

PR #3708, opened 2022-09-26
Make it easier to vertically stretch plotOutput(), imageOutput(), and uiOutput() 

cpsievert:
The main motivation for these changes is for `plotOutput()`, `imageOutput()`, and `uiOutput()` to be able to stretch when used with `bslib::card_body_fill()` (see https://github.com/rstudio/bslib/pull/452 for more on that), but one nice side effect is that `plotOutput()`/`imageOutput()` no longer needs `height: 100%` inside of `fillPage()`:

```r
shinyApp(
  fillPage(plotOutput("p"), theme = bslib::bs_theme()),
  function(input, output) { output$p <- renderPlot(hist(rnorm(100))) }
)
```

In more technical detail, here's a summary of the changes:

1. Changes the default value for `plotOutput()` (and `imageOutput()`)'s `height` and `width` arguments to `NULL` (the default size won't change, but the default is no longer be provided as an inline style).
2. When `height = NULL` (the new default), the output container (for `plotOutput()`/`imageOutput()`) gains a `vfill-item` class (effectively `flex: 1`), so that when it's a flex item, it'll stretch/grow.
3. `fillPage()` now puts a `vfill-container` on the `<body>` (effectively `display: flex` with `flex-direction: column`) so that it's direct children can more easily stretch vertically to fit the viewport (via `vfill-item` class).
4. `uiOutput()` gains a new `fill` argument, which when set to `TRUE`, will add both the `vfill-container` and `vfill-item` class (effectively relaying the potential to stretch down to it's children).

N.B. the implementation of `vfill-container`/`vfill-item` will go in bslib because their primary purpose is to support the new/upcoming `bslib::card()` API (and it's more important that they work sensibly when statically rendered rather than working out-of-the-box with `fillPage()`)

## TODO

- [ ] A testing app with `fillPage()`?


cpsievert:
Closing in favor of https://github.com/rstudio/shiny/pull/3715

PR #3710, opened 2022-10-03
Remove types-jquery.patch

wch:
This removes `types-jquery.patch`. What happens now is that `$` is ambient -- it just available everywhere, even when the import isn't explicit. I believe that if the following is present in any file (and it is in several ts files), then `$` is available in all files:

```ts
import $ from "jquery"
```

However, I believe that the normal way of making ambient imports know to TS is by putting this in a file:

```ts
/// <reference types="jquery" />

declare global {
  interface Window {
    $: typeof jQuery;
  }
}
```

(Although I'm not sure if the `declare global` part is really necessary.)

This is what we do in https://github.com/rstudio/shinylive/blob/d2cdc7b21436e72d601ab4b64dd63f5bc0e57285/src/Components/Terminal.tsx#L2-L13.

I agree that it would be nicer to have the `$` import be explicit in each file, but doing it with the patch is swimming against the current and it causes problems with importing shiny's JS components in other JS packages.

Note that if we do take this PR, we should probably also:
* Update src/README.md
* Add in `/// <reference types="jquery" />` somewhere


schloerke (changes requested):
Let's remove the patch and add R tests to find all `.ts` files that use `$(` or `$.` and make sure they `import $ from "jquery"`.

schloerke:
Added a test in the bundling code. Figured it made sense to test the jquery stuff within the TS code

jcheng5 (approved):

Issue #1921, opened 2018-01-08
Restoring bookmark state with two RDatas


samuelhuerga:
Hi all.

I have a big shiny app with several inputs. Some of them are RDatas. When I bookmark a state in server that has two RDatas loaded and try to restore it, just one (the last) is correctly restored.

A bit complicated reproducible example:

Shiny app:
```
library(shiny)
library(dplyr)
library(stringr)

ui <- function(request) {
  fluidPage(
    fileInput("file_a", "Enter first RData"),
    verbatimTextOutput("text_a"),
    fileInput("file_b", "Enter second RData"),
    verbatimTextOutput("text_b"),
    bookmarkButton()
  )
}
server <- function(input, output, session) {
  observeEvent(input$file_a,{
    load(input$file_a$datapath,envir = globalenv())
    print("A")
    print(ls(envir = globalenv()))
  })
  observeEvent(input$file_b,{
    load(input$file_b$datapath,envir = globalenv())
    print("B")
    print(ls(envir = globalenv()))
  })
  output$text_a <- reactive({
    input$file_a
    return (ls(envir = globalenv()) %>% str_subset("a"))
  })
  output$text_b <- reactive({
    input$file_b
    return(ls(envir = globalenv()) %>% str_subset("b"))
  })
}

shinyApp(ui, server, enableBookmarking = "server")
```
I create two RDatas from R:
```
a <- "a"
b <- "b"
save(a,file="file_a.Rdata")
save(b,file="file_b.Rdata")
```

And load them into the app and bookmark its state:
![image](https://user-images.githubusercontent.com/12885915/34694344-0497a17c-f4c7-11e7-97e6-965a6f3a02c7.png)

then, stop shiny app (to clear global environment), and restore its state:
![image](https://user-images.githubusercontent.com/12885915/34694400-47724344-f4c7-11e7-836c-4ef8d12ce48e.png)

I think it could be caused by how bookmarks are saved in server mode:
```
> readRDS("shiny_bookmarks/ee078afb8cfff9a4/input.rds")
$file_a
          name size type datapath
1 file_a.Rdata   56       0.Rdata

$file_b
          name size type datapath
1 file_b.Rdata   56       0.Rdata
```
Both RData inside input.rds has asociated same datapath: 0.Rdata, so, only last one is preserved.

I don't expect this behaviour. Is there a chance this could be fixed?

Thanks!

stkrogs:
[...]

PR #3713, opened 2022-10-18
Changed filename to use unique id

Gabriel2409:
When enabling bookmarks, if there is more than one file input, the extra files are not correctly saved. This is due to a name conflict of the uploaded filenames.  This PR should solve this issue


Gabriel2409:
This PR should close #1921

PR #3715, opened 2022-10-21
Add fill arguments to plotOutput(), imageOutput(), and uiOutput()

cpsievert:
See [here](https://github.com/rstudio/shiny/pull/3715/files#diff-51920e95310ebfbc1ae31709f3b95f89afffbf4f1a6e38e8b2b406e2fb6197eaR10) for an explanation of what this PR does

### Testing notes

No manual testing needed. With existing tests as well as new ones being added in https://github.com/rstudio/shinycoreci/pull/132, we'll have pretty good coverage on these changes


R/bootstrap.R
```diff
@@ -792,9 +792,9 @@ verbatimTextOutput <- function(outputId, placeholder = FALSE) {
 #' @name plotOutput
 #' @rdname plotOutput
 #' @export
-imageOutput <- function(outputId, width = "100%", height="400px",
+imageOutput <- function(outputId, width = "100%", height = "400px",
```
cpsievert, line 795: Shouldn't this be `auto`?

R/bootstrap.R
```diff
@@ -792,9 +792,9 @@ verbatimTextOutput <- function(outputId, placeholder = FALSE) {
...
                         click = NULL, dblclick = NULL, hover = NULL, brush = NULL,
-                        inline = FALSE) {
+                        inline = FALSE, fill = TRUE) {
```
cpsievert, line 797: This should probably be `F`

R/bootstrap.R
```diff
@@ -1088,11 +1096,11 @@ imageOutput <- function(outputId, width = "100%", height="400px",
 #' @export
 plotOutput <- function(outputId, width = "100%", height="400px",
                        click = NULL, dblclick = NULL, hover = NULL, brush = NULL,
-                       inline = FALSE) {
+                       inline = FALSE, fill = TRUE) {
```
cpsievert, line 1099:
```suggestion
                       inline = FALSE, fill = !inline) {
```

R/bootstrap.R
```diff
@@ -1155,12 +1168,16 @@ dataTableOutput <- function(outputId) {
 #' )
 #' @export
 htmlOutput <- function(outputId, inline = FALSE,
-  container = if (inline) span else div, ...)
+  container = if (inline) span else div, fill = FALSE, ...)
```
cpsievert, line 1171:
```suggestion
  container = if (inline) span else div, fill = FALSE, fillContainer = fill, ...)
```

cpsievert: Maybe there is a better name here?

R/bootstrap.R
cpsievert: Probably, but I'm not going to mess with it since it would break `renderImage()` returning `height="100%"`

NEWS.md
```diff
@@ -7,6 +7,10 @@ shiny 1.7.2.9000
 
 ### New features and improvements
 
+* `plotOutput()` and `imageOutput()` gain a `fill` argument. If `TRUE` (the default for `plotOutput()`), the output container is allowed to grow/shrink to fit a fill container (`htmltools::asFillContainer()`) with an opinionated height. Most importantly, this means `plotOutput()` will grow/shrink by default [inside of `bslib::card_body_fill()`](https://rstudio.github.io/bslib/articles/cards.html#responsive-sizing). (#3715) 
+
+* `uiOutput()` gains `fill` and `fillItem` arguments. Set `fill = TRUE` to treat the `uiOutput()`'s `container` as a fill container (`htmltools::asFillContainer()`), allowing it's contents to grow/shrink when the `container` has an opinionated height. (#3715)
```
jcheng5, line 12:
```suggestion
* `uiOutput()` gains `fill` and `fillItem` arguments. Set `fill = TRUE` to treat the `uiOutput()`'s `container` as a fill container (`htmltools::asFillContainer()`), allowing its contents to grow/shrink when the `container` has an opinionated height. (#3715)
```

jcheng5, line 12: Is it just me or do the `fill` arguments in `plotOutput`/`imageOutput` vs. `uiOutput` have opposite meanings? In the former, it means "use `asFillItem()` on me" and in the latter, it means "use `asFillContainer()` on me"?

R/bootstrap.R
```diff
@@ -1155,12 +1174,19 @@ dataTableOutput <- function(outputId) {
 #' )
 #' @export
 htmlOutput <- function(outputId, inline = FALSE,
-  container = if (inline) span else div, ...)
+  container = if (inline) span else div, fill = FALSE, fillItem = !inline, ...)
```
jcheng5, line 1177: I was envisioning `fillItem = FALSE` and `fill = fillItem`. (Or possibly vice versa.) Like, usually the contents of a `uiOutput` is not going to be a fill item, but if it is (and it matters) then probably the `uiOutput`'s container is a fill container?

NEWS.md
cpsievert: It's not just you...it was intentional (for reasons I explain below)

R/bootstrap.R
cpsievert: I don't like the fact that you'd have to remember to `uiOutput(fillItem = TRUE)` to opt-in to fill behavior, yet do `fill = FALSE` anywhere else to opt-out. I may have a better idea that I'll implement in the next commit.

PR #3724, opened 2022-11-03
Pass error to handler

JohnCoene:
I have asked a few people and in a few places for a way to catch when a shiny application disconnects/crashes along with the error that caused it.

Question on [SO](https://stackoverflow.com/questions/74291807/shiny-disconnect-crash?noredirect=1#comment131172858_74291807) and [RStudio Community](https://community.rstudio.com/t/shiny-capture-crash-disconnect/151688/9)

I cannot find a solution with shiny's existing tools, sorry if this is already possible.

I thought of using the handler set as `shiny.error` option to capture the error/its message.
This allows something like the example below.

```r
library(shiny)

errs <- list()

on_error <- function(e) {
  errs <<- append(errs, e$message)
}

options(shiny.error = on_error)

ui <- fluidPage(
  actionButton("crash", "Crash")
)

  onStop(function(){
    print(errs)
  })
 
  observeEvent(input$crash, {
    print(error)  # no `error` object => disconnect
  })
}

shinyApp(ui, server)
```

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -488,7 +488,7 @@ shinyCallingHandlers <- function(expr) {
         return()
 
       handle <- getOption('shiny.error')
-      if (is.function(handle)) handle()
+      if (is.function(handle)) handle(e)
     }
   )
 }
--- a/R/utils.R
+++ b/R/utils.R
@@ -489,16 +489,11 @@ shinyCallingHandlers <- function(expr) {
 
       handle <- getOption('shiny.error')
       if (is.function(handle)) {
-        # if the handler takes at least one argument
-        # pass error along
-        args <- formalArgs(handle)
-
-        if(length(args) > 0) {
+        if (length(formals(handle)) > 0) {
           handle(e)
-          return()
+        } else {
+          handle()
         }
-        
-        handle()
       } 
     }
   )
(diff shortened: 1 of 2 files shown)


R/utils.R
```diff
@@ -488,7 +488,18 @@ shinyCallingHandlers <- function(expr) {
...
+        # if the handler takes at least one argument
+        # pass error along
+        args <- formalArgs(handle)
+
+        if(length(args) > 0) {
+          handle(e)
+          return()
+        }
+        
+        handle()
```
cpsievert, line 501:
```suggestion
        if (length(formals(handle)) > 0) {
          handle(e)
        } else {
          handle()
        }
```

cpsievert:
Now that I think of it, I'm not sure the potential breaking change here is worth it considering that you can do:

```r
on_error <- function() {
  errs <<- append(errs, rlang::last_error())
}

options(shiny.error = on_error)
```

JohnCoene:
But this does not work for me `Warning: Error in : Can't show last error because no error was recorded yet`.

Shiny seems to be handling errors in `conditions.R` in ways I don't quite understand but it results in traceback, `warnings()` call, etc. to be empty/`NULL`, as far as I can tell at least.

JohnCoene:
Closing in favour of #3726

PR #3729, opened 2022-11-09
insertui: use 'beforeEnd' by default without manually setting it

daattali:
'beforeEnd' is a good default, it's what I've always used as well. However, by looking at the function's signature, because the vector of options starts with "beforeBegin", I always assumed that "beforeBegin" was the default. I do realize now that it's not true and that the documentation for `where` does say this, but I think it will be even more clear -- both documentation-wise and code-wise -- if we just have 'beforeEnd' naturally be the default


jonthegeek:
It looks likely that a `usethis::pr_merge_main()` will take care of the GHA failures (or at least the required one). The change is minimal and non-invasive, since it results in the same behavior.

cpsievert:
Thanks @daattali, maybe it also makes sense to change the order of options in the roxygen docs?

Issue #3735, opened 2022-11-18
NULL width/height causing "segfault" R exit in renderPlot


slodge:
### System details

Browser Version: Not applicable

Output of `sessionInfo()`:

```
R version 4.2.0 (2022-04-22 ucrt)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 19042)

Matrix products: default

locale:
[1] LC_COLLATE=English_United Kingdom.utf8  LC_CTYPE=English_United Kingdom.utf8    LC_MONETARY=English_United Kingdom.utf8 LC_NUMERIC=C                           
[5] LC_TIME=English_United Kingdom.utf8    

attached base packages:
[1] stats     graphics  grDevices datasets  utils     methods   base     

loaded via a namespace (and not attached):
 [1] Rcpp_1.0.9           pillar_1.8.1         compiler_4.2.0       later_1.3.0          urlchecker_1.0.1     prettyunits_1.1.1    profvis_0.3.7       
 [8] remotes_2.4.2        tools_4.2.0          testthat_3.1.5       digest_0.6.30        pkgbuild_1.3.1       pkgload_1.3.2        memoise_2.0.1       
[15] lifecycle_1.0.3      tibble_3.1.8         pkgconfig_2.0.3      rlang_1.0.6          shiny_1.7.3.9001     cli_3.4.1            rstudioapi_0.14     
[22] yaml_2.3.6           curl_4.3.3           fastmap_1.1.0        withr_2.5.0          stringr_1.4.1        desc_1.4.2           vctrs_0.5.1         
[29] fs_1.5.2             htmlwidgets_1.5.4    devtools_2.4.5       rprojroot_2.0.3      glue_1.6.2           R6_2.5.1             processx_3.8.0      
[36] fansi_1.0.3          sessioninfo_1.2.2    callr_3.7.3          purrr_0.3.5          magrittr_2.0.3       ps_1.7.2             promises_1.2.0.1    
[43] ellipsis_0.3.2       htmltools_0.5.3.9002 usethis_2.1.6        mime_0.12            xtable_1.8-4         renv_0.16.0          httpuv_1.6.6        
[50] utf8_1.2.2           stringi_1.7.8        miniUI_0.1.1.1       cachem_1.0.6         crayon_1.5.2         brio_1.1.3    
```

### Example application *or* steps to reproduce the problem

Sorry - **not** got an easily reproducible sample right now.

[...]

PR #3737, opened 2022-11-18
Protect renderPlot against NULL dimensions

slodge:
Somehow our bespoke RMarkdown Shiny app is sending hidden:false for plots, but also not sending any dimensions.

This causes the R process hosting the shiny app to crash - with very little debug output.

This PR protects drawPlot against these NULL values.

- [x] Add an entry to NEWS.md concisely describing what you changed.
- [x] Not Applicable: If appropriate, add unit tests in the tests/ directory.
- [x] Not Applicable: If you made any changes to the JavaScript files in the srcjs/ directory, make sure you build the output JavaScript files. See tools/README.md file for information on using the build system.
- [x] Run Build->Check Package in the RStudio IDE, or devtools::check(), to make sure your change did not add any messages, warnings, or errors.
- [x] If you find a bug in Shiny, you can also [file an issue](https://github.com/rstudio/shiny/issues/new). Please provide as much relevant information as you can, and include a minimal reproducible example if possible.
- [x] Work out how I've messed up Stuart Lodge and slodge in the commits - sorry!

See bug report: https://github.com/rstudio/shiny/issues/3735

### DIFF
--- a/R/render-plot.R
+++ b/R/render-plot.R
@@ -235,6 +235,22 @@ drawPlot <- function(name, session, func, width, height, alt, pixelratio, res, .
   #  9. Return img, value, displaylist, coordmap
   # 10. On error, take width and height dependency
 
+  checkDimsForNulls <- function(width, height) {
+    null_width <- is.null(width)
+    null_height <- is.null(height)
+    if (null_width || null_height) {
+      stop("null ",
+           paste0(c("width", "height")[c(null_width, null_height)], collapse=" and "),
+           " seen for plot output ",
+           name,
+           ". Make sure that the plot container has non-zero size.")
+    }
+  }
+
+  # if any dim is null, this is most likely because the shiny app hasn't sent
+  # any clientData height/width for that output yet.
+  checkDimsForNulls(width, height)
+
   outfile <- tempfile(fileext='.png') # If startPNG throws, this could leak. Shrug.
   device <- startPNG(outfile, width*pixelratio, height*pixelratio, res = res*pixelratio, ...)
   domain <- createGraphicsDevicePromiseDomain(device)
--- a/NEWS.md
+++ b/NEWS.md
@@ -17,6 +17,8 @@ shiny 1.7.3.9000
 
 * Closed #3687: Updated jQuery-UI to v1.13.2. (#3697)
 
+* Closed #3735: Protect renderPlot against NULL height/width (#3736)
+
 
 shiny 1.7.3
 ===========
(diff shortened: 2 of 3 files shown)


R/render-plot.R
```diff
@@ -235,6 +235,22 @@ drawPlot <- function(name, session, func, width, height, alt, pixelratio, res, .
...
+  checkDimsForNulls <- function(width, height) {
+    null_width <- is.null(width)
+    null_height <- is.null(height)
+    if (null_width || null_height) {
+      stop("null ",
+           paste0(c("width", "height")[c(null_width, null_height)], collapse=" and "),
+           " seen for plot output ",
+           name,
+           ". Make sure that the plot container has non-zero size.")
+    }
```
cpsievert, line 247: I wonder if it would be more sensible to `req()` these values?

jcheng5:
Thank you! Can I suggest that startPNG also check for and throw an error on NULL dimensions? Ultimately it’s ragg that we need to prevent from seeing NULL.

slodge:
> Thank you! Can I suggest that startPNG also check for and throw an error on NULL dimensions? Ultimately it’s ragg that we need to prevent from seeing NULL.

:+1:

Have moved code to startPNG and added `res` to the NULL checking

R/render-plot.R
slodge: I think `req` would miss out debug info that is actually quite useful. 

I don't think Shiny authors should ever actually see this error in "normal development" - in normal development there should never by NULL height and widths - I think the Shiny client should always be providing them for visible outputs? 

... although there might be some problem if suspendWhenHidden has been set to `FALSE`? Not sure what should happen in that case - should the JS be sending sizes for those plot outputs?

slodge:
Argh... something funky is happening with my git client - keeps dropping my email address from commits - sorry!

slodge:
OK... a few force pushes later - hopefully the commits are finally resolved to my email!

Problem was to do with Windows; a network based home directory (where my git credentials are stored); and a broken VPN  connection.

R/render-plot.R
```diff
@@ -235,6 +235,22 @@ drawPlot <- function(name, session, func, width, height, alt, pixelratio, res, .
...
+  }
+
+  # if any dim is null, this is most likely because the shiny app hasn't sent
+  # any clientData height/width for that output yet.
+  checkDimsForNulls(width, height)
+
```
cpsievert, line 253: No longer needed since we'll make this check in `startPNG()`

```suggestion
```

cpsievert:
Thanks for the contribution @slodge. Since there's a bit of context/subtlety here, I'll be finishing this off in #3739

slodge:
Thanks...

The reason I wanted to keep the `checkDimsForNulls` in `drawPlot` is because having the name in the error message there was incredibly useful for a user (me!) trying to debug problems.

PR #3739, opened 2022-11-21
Throw informative error in `renderPlot()` early if height/width of a plot aren't yet defined

### LINKED ISSUE #1409: suspendWhenHidden is not sufficient to be able to render elements immediately
`suspendWhenHidden = FALSE` works great _after_ an element has already been rendered before and it should be re-rendered eagerly, even when hidden. This can be useful when, for example, there are several tabs dependent on the same inputs and we want to re-render them all whenever these change (so the user does not notice any lag after the initial time). However, this will not work if the elements have never been rendered before (e.g. on app startup).

Reproducible example:

``` r
library(shiny)
library(ggplot2)

ui <- shinyUI(fluidPage(
tabsetPanel(
tabPanel("Daily",
plotOutput("DailyCandles")
),
tabPanel("Weekly",
plotOutput("WeeklyCandles")
)
))
)

server <- shinyServer(function(input, output) {

Sims = 50000

df <- data.frame(x = 1:Sims, y= rnorm(Sims))

output$DailyCandles <- renderPlot({
ggplot(df,aes(x=x,y=y)) + geom_line()
})

output$WeeklyCandles <- renderPlot({
ggplot(df,aes(x=x,y=y)) + geom_line()
})

outputOptions(output, "WeeklyCandles", suspendWhenHidden = FALSE)

})

# Run the application 
```

We can either make it so that `suspendWhenHidden` also works here, or we can create a new construct for this (e.g. using Joe's `reactiveTrigger`?)

Source: [Mauro Herrada's comment on Disqus](https://disqus.com/home/discussion/shiny-rstudio/shiny_tabsets/#comment-2933317485).

### LINKED ISSUE #3704: removing the priority of the `ragg` package in the `shiny:::startPNG` - crashing the R session
linked to https://github.com/r-lib/ragg/issues/116
very important it seems to not be a problem of arm64 only.

Please consider removal of the priority of the `ragg` package in the `shiny::startPNG`.
The `ragg` package is causing a hard crash of the R session and unfortunately maintainer seems to NOT have time for fixes.

In my opinion the  core R  grDevices seems to be a much better choice, at least regarding stability.

### LINKED ISSUE #3735: NULL width/height causing "segfault" R exit in renderPlot

cpsievert:
Closes #3704. 
Closes #3735. 
Closes #1409. 

## TODO

Unit tests


R/imageutils.R
```diff
@@ -11,7 +11,27 @@ startPNG <- function(filename, width, height, res, ...) {
...
   }
 
-  args <- rlang::list2(filename=filename, width=width, height=height, res=res, ...)
+  # It's possible for width/height to be NULL (e.g., when using
+  # suspendWhenHidden=F w/ tabsetPanel()), which will lead to an error when
+  # attempting to open the device (rstudio/shiny#1409). In this case, ragg will
+  # actually segfault (instead of error), so explicitly throw an error before
+  # opening the device (r-lib/ragg#116, rstudio/shiny#3704)
+  check_empty_png_size <- function(x) {
+    if (length(x) > 0) return(x)
```
jcheng5, line 20: Do you want to check for `x == 0` as well? Or do we not care what ragg does in that case

jcheng5 (approved):

R/imageutils.R
cpsievert: ragg won't seqfault in that case (and I'm pretty sure they'll all, at least eventually, result in `Error  : figure margins too large`, so it didn't seem worth including that case

PR #3744, opened 2022-12-02
fixed the documentation of `sendBinaryMessage` function.

hq9000:
previously, the documentation referred to a non-existent `addBinaryMessageHandler` function to be called on the frontend side. Experimentally, it was found out that the same `addCustomMessageHandler` just works. This is the function used to register handlers of other custom messages.

### DIFF
--- a/man/session.Rd
+++ b/man/session.Rd
@@ -130,11 +130,11 @@ as the page loads; the function you provide to
 \code{sendCustomMessage} is called on the server.
 }
 \item{sendBinaryMessage(type, message)}{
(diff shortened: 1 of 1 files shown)


cpsievert:
Thanks @hq9000! It'd be helpful if you could run `devtools::document()` and push up the changes.

hq9000:
> Thanks @hq9000! It'd be helpful if you could run `devtools::document()` and push up the changes.

hi @cpsievert! It's done. Thanks for pointing out.

PR #3758, opened 2022-12-15
Make `?shiny-package` topic internal in pkgdown

cpsievert:
Before this change, I get the following error when trying to build the reference in `{shiny-dev-center}`

```
-- Building function reference -------------------------------------------------
Error: 
! in callr subprocess.
Caused by error in `check_missing_topics(rows, pkg)`:
! All topics must be included in reference index
✖ Missing topics: shiny-package
ℹ Either add to _pkgdown.yml or use @keywords internal
ℹ See `$stdout` and `$stderr` for standard output and error.
```


schloerke (changes requested):
I'd like to see that topic still in the package index. I think we can alter the pkgdown file and add an internal topic section. 

Related: https://github.com/r-lib/pkgdown/pull/1783/files

cpsievert:
Nice find! Re-running `_scripts/generate-pkgdown-reference.R` with [470c1af](https://github.com/rstudio/shiny/pull/3758/commits/470c1af68b1f9056cd814b73abe1aea095854327) generates the same result :)

Issue #3748, opened 2022-12-06
Session ID/token exposed with downloadHandler


leeroyaus:
Hello

When creating a link to download files using downloadHandler, the session ID/token is exposed in the link. This seems to be generally regarded as a serious security issue - see details at the link below:

https://owasp.org/www-project-web-security-testing-guide/latest/4-Web_Application_Security_Testing/06-Session_Management_Testing/04-Testing_for_Exposed_Session_Variables

Is it possible to use a different token/ID for links using downloadHandler? Or could you provide more information about how this might not be a serious security issue?

jcheng5:
Hi, sorry for the delay in responding.

You're right to point out that leaking session tokens in the URL is generally regarded as serious. However, in the case of Shiny there are a couple of mitigating factors:

1. Contrary to what session tokens are usually used for, Shiny doesn't use them for authentication/authorization; that is, it's not possible to steal a session token and use it to impersonate a logged-in user (on Connect, ShinyApps.io, or any other Shiny hosting server), access the apps they are able to access, etc. Cookies are used for that instead.
2. Shiny's session tokens are tied to the lifetime of a particular page load. Once that page is navigated away from, or a tab is closed, or even on page reload, the session token is no longer valid. (Actually there's approximately a 15 second delay for the session token to become invalid, as it takes that long for Shiny to be convinced that you're not the victim of an unintended, temporary network interruption.)
3. Connect and ShinyApps.io are both https-only, which prevents disclosure via packet sniffing (but doesn't prevent disclosure via log files or malicious proxies).

So the attack window is quite a bit narrower, and the effect of an attack much less harmful, than the general case as described in the OWASP link. This has been enough to allay other customers' concerns in the past (it's far from the first time this has come up).

However, thinking about this one more time made me realize there is one attack vector that we could easily close, which is the scenario where Alice and Bob are both authorized users of App X. If Bob gets ahold of one of Alice's download URLs while Alice is still on the page, he can access the same download file she would get if she clicked the link. In this case, it's easy for Shiny to realize that the owner of the session and the requester of the download are not the same person. I'll file a PR to close this hole.

The upshot is that, with this upcoming change, this should be a total non-issue for any app _that requires users to log in_, since the session ID will be scoped to the logged-in user.

jfunction:
[...]

PR #3766, opened 2023-01-13
Ensure logged-in user matches when handling session-specific routes

jcheng5:
Mitigation for #3748, see that issue for more information.

## Testing notes

1. Install Shiny from this branch.
2. Deploy the [download example app](https://github.com/rstudio/shiny/tree/20f8a181d44f1fdb507e868c5cd425601ddaadf9/inst/examples/10_download) to ShinyApps.io.
3. From the ShinyApps.io dashboard, mark the deployed app as requiring login, and authorize two different users.
4. Access the deployed app via a browser (logging in if necessary).
5. Right-click the Download button and copy the link.
6. From a second browser, access the app again, this time logging in as the other user.
7. Paste the link from step 5 into the second browser's URL bar and press Enter.

Before this PR, the download would succeed. After this PR, the download returns 404.

### DIFF
--- a/R/shiny.R
+++ b/R/shiny.R
@@ -1876,6 +1876,22 @@ ShinySession <- R6Class(
     # Provides a mechanism for handling direct HTTP requests that are posted
     # to the session (rather than going through the websocket)
     handleRequest = function(req) {
+      if (!is.null(self$user)) {
+        if (is.null(req$HTTP_SHINY_SERVER_CREDENTIALS)) {
+          # Session owner is logged in, but this requester is not
+          return(NULL)
+        }
+
+        creds <- NULL
+        try({creds <- safeFromJSON(req$HTTP_SHINY_SERVER_CREDENTIALS)},
+          silent = TRUE
+        )
+        if (!identical(self$user, creds$user)) {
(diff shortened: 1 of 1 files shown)


R/shiny.R
```diff
@@ -1876,6 +1876,22 @@ ShinySession <- R6Class(
...
+        if (is.null(req$HTTP_SHINY_SERVER_CREDENTIALS)) {
+          # Session owner is logged in, but this requester is not
+          return(NULL)
+        }
+
+        creds <- NULL
+        try({creds <- safeFromJSON(req$HTTP_SHINY_SERVER_CREDENTIALS)},
+          silent = TRUE
+        )
+        if (!identical(self$user, creds$user)) {
```
wch, line 1889: This could throw an error if `creds` is not a list or `NULL`. For example, it could be an atomic vector.

```R
x <- shiny:::safeFromJSON("[1,2,3]")
x$user
#> Error in x$user : $ operator is invalid for atomic vectors
```

So maybe this part should also be inside of the `try`.

jcheng5:
Barret said to ignore the failing r-devel check for now

PR #3772, opened 2023-01-20
Fix `stopPropagation` error in ion.rangeSlider

### LINKED ISSUE #3771: JS error: `i.stopPropagation is not a function` in ion.rangeSlider
To reproduce, run this app and open it in the browser:

```R
runExample('01_hello', launch.browser=FALSE, display.mode = "normal")
```

To trigger the error, click on a blank space in the app, then press Tab. (I think it is triggered when a hidden element gets focus.)

image

This is happening in the `pointerClick` method here: 
https://github.com/rstudio/shiny/blob/0c177d30dc11cb2aed1612db456158b31306af9a/inst/www/shared/ionrangeslider/js/ion.rangeSlider.js#L864-L866

It in turn is being called from `pointerFocus` -- but that function, instead of passing in a real Event object, passes in an object that has two properties: `preventDefault` and `pageX`:
https://github.com/rstudio/shiny/blob/0c177d30dc11cb2aed1612db456158b31306af9a/inst/www/shared/ionrangeslider/js/ion.rangeSlider.js#L742-L757

The call to `e.stopPropagation` that is raising the error is something we added in this patch: https://github.com/rstudio/shiny/blob/0c177d30dc11cb2aed1612db456158b31306af9a/inst/www/shared/ionrangeslider/js/ion.rangeSlider.js

So we can simply update the patch to contain an empty function for `stopPropagation.`

wch:
Closes #3771.


cpsievert:
Nice, thanks!

cpsievert (approved):

Issue #2629, opened 2019-09-19
reactiveValues inconsistent/unclear behaviour in regards to order of items


daattali:
The documentation of `reactiveValues` mentions that it functions similarly to a list. Lists, unlike sets, should preserve order based on the order in which items were added.

This does not seem to be the case with reactiveValues. There are a few interrelated issues here:

```
# initializing reactiveValues with b,a and then asking for the names, the names come back sorted
> rv <- reactiveValues(b="b", a="a")
> names(rv)
[1] "a" "b"

# initializing a reactiveValues and only adding the items afterwards has the same behaviour
> rv <- reactiveValues(); rv$b <- "b"; rv$a <- "a"
> names(rv)
[1] "a" "b"

# BUT doing the same with a list, the order is preserved
> li <- list(b="b", a="a")
> names(li)
[1] "b" "a"

# converting to a list also doesn't keep the order
> str(reactiveValuesToList(rv))
List of 2
 $ a: chr "a"
 $ b: chr "b"
```

That's one issue. But it gets worse, because (as far as I can tell) this behaviour doesn't simply sort by alphabet, I can't tell how to predict the result, so to me it seems inconsistent. If instead of "b","a" I use "foo","bar", I get different results:

```
# asking for the names still returns them in alphabetical instead of chronological order
> rv <- reactiveValues(foo="foo", bar="bar")
> names(rv)
[1] "bar" "foo"

# BUT converting to a list now does return the original order (when using b,a in the previous example, it did not)
> str(reactiveValuesToList(rv))
List of 2
 $ foo: chr "foo"
 $ bar: chr "bar"
```

I tried running more complex experiments and couldn't find a pattern - when I add multiple items, they seem to be returned in a random order.

There doesn't seem to be a way to know in what order the items will come back out of a reactiveValues object.

ColinFay:
AFAIK even if the doc says it behaves like a list, the truth is that `reactiveValues` behaves like an `environment`, hence the ordered naming and the impossibility to use `[` and `[[` with a number (environments don't have a sense of order). 

``` r
e <- environment()
e$foo <- "foo"
e$bar <- "bar"
names(e)
#> [1] "env"  "foo"  "e"    "bar"  "data"
e[1]
#> Error in e[1]: object of type 'environment' is not subsettable
e[[1]]
#> Error in e[[1]]: wrong arguments for subsetting an environment

library(shiny)

isolate({
  rv <- reactiveValues(foo="foo", bar="bar")
[...]

PR #3774, opened 2023-01-23
Ensure that `reactiveValues` keys and values are sorted

wch:
This PR also ensures that anywhere that a `reactiveValues` `$names()` or `$values()` method is called, the keys will be sorted in the same order that they were inserted. (Note: a previous version of this PR sorted lexicographically, and some of the comments were for that version.)

In our shinycoreci tests, we found that bookmarked input values were not always in consistent order. For example, for the same app in the same state, both of these URLs might be generated:

* http://127.0.0.1/?_inputs_&reset=0&x=5
* http://127.0.0.1/?_inputs_&x=5&reset=0

### DIFF
--- a/tests/testthat/test-reactivity.r
+++ b/tests/testthat/test-reactivity.r
@@ -127,17 +127,19 @@ test_that("ReactiveValues", {
 })
 
 test_that("reactiveValues keys are sorted", {
-  values <- reactiveValues()
-  values$a <- 1
-  values$A <- 11
+  values <- reactiveValues(b=2, a=0)
+  values$C <- 13
+  values$A <- 0
   values$c <- 3
   values$B <- 12
-  values$C <- 13
-  values$b <- 2
-  expect_identical(isolate(names(values)), c("A", "B", "C", "a", "b", "c"))
+  # Setting an existing value shouldn't change order
+  values$a <- 1
+  values$A <- 11
+
+  expect_identical(isolate(names(values)), c("b", "a", "C", "A", "c", "B"))
   expect_identical(
     isolate(reactiveValuesToList(values)),
-    list(A=11, B=12, C=13, a=1, b=2, c=3)
+    list(b=2, a=1, C=13, A=11, c=3, B=12)
   )
 })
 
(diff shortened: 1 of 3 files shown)


jcheng5:
I'm OK with this but it's a slight shame that they're not in insertion order, like a named list. @rpodcast and I needed this the other day and ended up resorting to tracking the insertion order in a side list.

wch:
Just curious - what did you need insertion order for?

rpodcast:
> Just curious - what did you need insertion order for?

@jcheng5 and I have been working on improving my workflow of dynamic UI insertions into a Shiny app based on some real life examples. The goal is to have a custom R6 class to manage a lot of the details for an improved user experience over my existing methods. In some of my applications, the order of how these dynamic UIs are entered actually matters as they can represent sequential parameter sets that need to follow the order of insertion. I realize that's not a very common use case of dynamic UI, but for my purposes it can be very helpful.

wch:
@jcheng5 I'll update this PR to return keys in insertion order.

tests/testthat/test-reactivity.r
```diff
@@ -126,6 +126,23 @@ test_that("ReactiveValues", {
...
 
+test_that("reactiveValues keys are sorted", {
+  values <- reactiveValues(b=2, a=0)
+  values$C <- 13
+  values$A <- 0
+  values$c <- 3
+  values$B <- 12
+  # Setting an existing value shouldn't change order
+  values$a <- 1
+  values$A <- 11
```
jcheng5, line 137: I guess we don't have the notion of removing a value? That's the only other case I could think of.

jcheng5 (approved):

tests/testthat/test-reactivity.r
wch: Right, we don't have a way to remove a value.

daattali:
I've ran into this issue many times as well, I'll see if my 4 year old issue can be closed https://github.com/rstudio/shiny/issues/2629

daattali:
Yep, I love when you guys fix old issues I've been waiting on for years 🥳

PR #3775, opened 2023-01-23
Initialize `Map` objects at load time instead of build time

wch:
This avoids potential problems that could arise from storing `fastmap` objects into the built Shiny package.

### DIFF
--- a/R/globals.R
+++ b/R/globals.R
@@ -7,16 +7,9 @@
   # the private seed during load.
   withPrivateSeed(set.seed(NULL))
 
-  # Create this at the top level, but since the object is from a different
-  # package, we don't want to bake it into the built binary package.
-  restoreCtxStack <<- fastmap::faststack()
-
-  # Map objects include fastmap objects and we don't want to bake those into a
-  # built package.
-  inputHandlers <<- Map$new()
-  appsByToken <<- Map$new()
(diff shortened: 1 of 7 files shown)


R/globals.R
```diff
@@ -11,6 +11,13 @@
   # package, we don't want to bake it into the built binary package.
   restoreCtxStack <<- fastmap::faststack()
 
+  # Map objects include fastmap objects and we don't want to bake those into a
+  # built package.
+  inputHandlers <<- Map$new()
```
schloerke, line 16: Needs to have the initial value removed

schloerke:
Two more: 
- [ ] https://github.com/rstudio/shiny/blob/9389160af06b2f26d3746fa06c6ac0df8e76c8dd/R/graph.R#L562
- [ ] https://github.com/rstudio/shiny/blob/87c673f2838122ad00f806434d26216594041915/R/server.R#L125
(Didn't see any others)

wch:
I added an `on_load()` function for registering expressions to be executed when the package is loaded. This makes it possible to put the initialization code next to the declaration, instead of having to put the initialization code in the package's `.onLoad` function.

schloerke (approved):

schloerke:
Would be nice to do this with the R6 classes in the future to reduce the built R package size. A PR for another day

PR #3782, opened 2023-02-23
Send initial values for plot interactions

wch:
This PR changes the client-side code so that when plot interactions are used, they send initial values immediately, instead of waiting for the user to start interacting with the plot.

This doesn't change any of the behavior in Shiny for R, because when (for example) the server code accesses `input$plot_click` and that value has not yet been set by the client, that expression simply returns `NULL`. After this change, the client sends an initial value of `null`, so the R expression still returns `NULL`.

However, in Shiny for Python, if you access `input.plot_click()` before the value has been set by the client, it raises a `SilentException`. After this change, the client sends an initial value of `null`, so the Python expression returns `None` instead of throwing the exception.


jcheng5:
I'd like to see an automated test for this--if E2E, it can be either in R or Python (Barret's branch).

jcheng5 (approved):

wch:
In R, the behavior is unchanged, so I'll add a test in Python.

jcheng5:
It's not unchanged if you use `names(input)` 😉

jcheng5:
Interesting... what would you expect to happen here? The plot doesn't render due to `req(FALSE)`, but the plot output is on the page. Should `plot_click` be `NULL` or missing?

```r
library(shiny)

ui <- fluidPage(
  plotOutput("plot", click="plot_click"),
  verbatimTextOutput("out")
)

  output$out <- renderPrint({
    names(input)
  })
  
  output$plot <- renderPlot({
    req(FALSE)
  })
}

shinyApp(ui, server)
```

wch:
Interesting... As a matter of fact, with this PR the `names(input)` returns `character(0)`.

That said, I think it would make a bit more sense if it returned `"plot_click"`, since it is specified in the `plotOutput`. I'm not sure where we'd put the code to make it work that way, though. If it were an input binding, we could put it in the `subscribe` method, but output bindings don't have that method.

PR #3825, opened 2023-05-23
fix typo in the docs for bindCache

### LINKED ISSUE #3824: Typo in the bindCache docs
In the documentation for `bindCache` https://github.com/rstudio/shiny/blob/90539bff2565225f9079193532dc9c3232eabbf6/R/bind-cache.R#L162 the documentation suggests that you should set `self$cache` in your server function, but then the example shown shows that you should set `session$cache`. I suspect that the latter is the correct usage, with `self$cache` is internal to the `{shiny}` package itself.

tomjemmett:
the docs reference setting `self$cache`, but then the example shows `session$cache`. this changes the former to the latter.

This resolves #3824

### DIFF
--- a/R/bind-cache.R
+++ b/R/bind-cache.R
@@ -159,8 +159,8 @@ utils::globalVariables(".GenericCallEnv", add = TRUE)
 #'   ```
 #'
 #'   To use different settings for a session-scoped cache, you can set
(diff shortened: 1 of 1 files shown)


cpsievert:
Thank you!

PR #3837, opened 2023-06-08
test(json digits): Try unsetting option in test to get default behavior

schloerke:
Related to #3819 
Error first seen in #3816


schloerke:
Don't know how it passed in the original PR, but it is fixed now.

wch (approved):

PR #3850, opened 2023-06-30
1.7.4.1 release candidate

### LINKED ISSUE #3849: R-devel raises warnings on package load, because of `as.numeric_version()`
I just got an email from CRAN:

*****

A recent PR (<https://bugs.r-project.org/show_bug.cgi?id=18548>) showed
that whereas version inputs to numeric_version() and package_version()
were always documented to be "suitable character vectors" (or numeric
vesion objects), the implementation also accepted numeric inputs via
as.character() coercion, the results of which however are (at least
currently) dependent on the setting of the OutDec option.  Given that
numeric input is certainly error-prone, imagine

```
  as.numeric_version(1.20)
  as.numeric_version(1.5-1)
```

R Core has decided it would be best to gradually phase out such inputs.

Thus, With c84610 R-devel now *warns* about such inputs, which will
typically not show in the check results (but immediately giving errors
would cause too much disruption).

Now in shiny, you have

```
  register_upgrade_message("htmlwidgets", 1.5)
```

in your .onLoad() in globals.R, which should really be

```
  register_upgrade_message("htmlwidgets", "1.5")
```

Can you please fix asap?  (This causes warnings whenever the shiny
namespace is loaded ...)

wch:
This fixes an issue that CRAN asked us to fix ASAP. Closes #3849.


schloerke:
The "GHA push routine" checks are a false positive. The true check was done for [f5b3954](https://github.com/rstudio/shiny/pull/3850/commits/f5b395485e7da7430400516b2229294dce3ba2e4) above. This commit was tagged and the PR needs to have a "merge" commit. After merging, the "merge" checks pass.

Merging PR

PR #3851, opened 2023-07-03
Confirm bug described in #3849

schloerke:
Related #3850


schloerke:
Abandoning PR as it requires `{htmlwidgets}` v1.5 which was made 4 yrs ago.

Issue #3700, opened 2022-09-12
selectizeInput hides part of the matched input


ismirsehregal:
It seems shiny is using a selectize.js version in which [this issue](https://github.com/selectize/selectize.js/issues/1149) is still present.

Here is an example:

```
library(shiny)

ui <- fluidPage(
  sidebarLayout(
    sidebarPanel(
      selectizeInput(
        inputId = "id_category",
        label = "Category (multiple answers possible)",
        multiple = TRUE,
        choices = c(
          "Producer price oil",
          "Producer price pulp",
          "Producer price oat",
          "Import price oil"
        )
      ),
    ), mainPanel()
  )
)

server <- function(input, output) {}

shinyApp(ui, server)
```
![Animation](https://user-images.githubusercontent.com/36849480/189644501-9d2de70b-6112-4861-a927-325f21b0711c.gif)

PS: Coming from [here](https://stackoverflow.com/q/73688369/9841389)

Issue #2957, opened 2020-07-02
SelectizeInput highlight issue


petervdberg:
If I create a selectizeInput with repeating characters in a choice value (in the example "aa aa"), and search for the entire choice value (by clicking the selectize, hitting backspace and type), the highlight functionality seems to break the dropdown, and shows the choice only partially (in the example "a")

![selectizeInput](https://user-images.githubusercontent.com/1250637/86329349-4e8fc900-bc46-11ea-9327-b6a7aad95f0d.png)

```R


selectInput("x", NULL, choices = c("aa aa"))
```

leonawicz:
Whether single- or multi-select, I can reproduce this with any set of choices. There does not need to be repeating text within a choice.

`selectInput("x", NULL, c("aba", "abb", "abc"), multiple = TRUE)`

Type `ab ab`. Or type `ab a` and hit space again. Or `a ab`. Values don't seem to change, but all choices in client-side display becomes truncated based on the character match when anything is repeated _as you type_. I don't know the source of this or what version of shiny or selectize it may have crept in but it makes choice lists with any somewhat repetitive multi-word phrases unusable.

leonawicz:
Going with the tagged releases, I have to revert back to shiny v1.1.0 to see the problem go away. It occurs from 1.2.0 onward. 

Interestingly, the issue where the yellow highlighting gets stuck on options text after you clear or make selections also goes away prior to v1.2.0.

But perhaps this is related to an included version of selectize.js. I don't have time to dig further. But reverting back to shiny 1.1 is too much of a rollback imo to get around this issue unfortunately.

wch:
A reproducible example:

```R
library(shiny)
shinyApp(
  ui = fluidPage(
    selectInput("x", NULL, c("aba", "abb", "abc"), multiple = TRUE)
  ),
  server = function(input, output) {}
)
```

wch:
I've bisected it to 6b605804d2c8bf2f5299415953432cc2a773ab3a, where we upgraded to Selectize 0.12.4.

That also appears to be the latest official release. There are subsequent versions, 0.12.5 and 0.12.6, but those aren't tagged as releases.

leonawicz:
I can confirm using `inlcudeScript()` in ui.R to load a copy of v0.12.1 after 0.12.4 is a sufficient workaround for the problem if it's needed immediately.

leonawicz:
[...]

PR #3875, opened 2023-08-17
Upgrade to selectize.js 0.15.2

### LINKED ISSUE #3868: Upgrade `selectize.js` to latest version
[The currently embedded version is `0.12.4`](https://github.com/rstudio/shiny/blob/68546c319e465a9cb113ea4499912823e264e75f/inst/www/shared/selectize/js/selectize.min.js).  [The latest version is `0.15`](https://selectize.dev/).

I am in particular interested in the availability of the `clear_button` plugin.

cpsievert:
Closes #3868
Closes https://github.com/rstudio/shiny/issues/3839
Closes https://github.com/rstudio/shiny/issues/3700
Closes https://github.com/rstudio/shiny/issues/2957
(Will close) https://github.com/rstudio/py-shiny/issues/655


wch:
The updateSelectize.R script currently downloads a selectize zip file from Github _and_ it also updates the package.json to add a specific selectize version, and downloads it to node_modules with `yarn`. This results in two separate copies.

I think it would be clearer and safer to only install selectize via package.json, and not do the direct download from Github. This also means that the package.json would be the source of truth for the selectize version, and the README would need to be updated to reflect that.

See #3858 and #3788 for previous discussion of this topic.

gadenbuie:
This might be addressed by this PR already, but I just noticed that select inputs add margin bottom on the `.selectize-control` element, which is nested quite deeply in the component.

image

Do you think it'd be possible to move that margin higher up in the component tree so that we could more easily disable it in a gap-spacing context (in bslib)?

srcts/src/bindings/input/selectInput.ts
```diff
@@ -292,8 +294,22 @@ class SelectInputBinding extends InputBinding {
       control.destroy();
       control = $el.selectize(settings)[0].selectize as SelectizeInfo;
     }
+
+    // (Hopefully temporary) workaround for a v0.15.2 selectize bug where the
+    // dropdown (instead of the <input>) gets focus after an item added via
+    // click. https://github.com/selectize/selectize.js/issues/2032
```
gadenbuie, line 300: The most minor of comments, but I use extensions that find `FIXME` and `TODO` comments that make it easier to find these things again later.

```suggestion
    // (Hopefully temporary) workaround for a v0.15.2 selectize bug where the
    // dropdown (instead of the <input>) gets focus after an item added via
    // click. 
    // FIXME: https://github.com/selectize/selectize.js/issues/2032
```

cpsievert:
Great point @gadenbuie, addressed in https://github.com/rstudio/shiny/pull/3875/commits/b3d32981e9136943974e18051ef6a9b94c3147c3

(I'm guessing that rule was there before `.shiny-input-container {margin-bottom: }` existed, and nobody noticed the slightly extra bit of margin (compared to other inputs)?)

inst/www/shared/selectize/scss/selectize.bootstrap5.scss
```diff
@@ -1,3 +1,200 @@
-$input-line-height-sm: $form-select-line-height !default;
-@import 'selectize.bootstrap4';
-.selectize-control{padding:0;}
+
+$enable-shadows: true !default;
+$select-font-family: inherit !default;
+$select-font-size: inherit !default;
+$select-line-height: $input-btn-line-height !default; //formerly line-height-computed
```
gadenbuie, line 5: Just a note that we should use `$input-btn-*` for `font-family` and `font-size` as well. See https://github.com/rstudio/bslib/pull/763

tools/selectize-patches/006-bs5-dark-mode-support.patch
```diff
@@ -0,0 +1,112 @@
...
+index 44fff4ae4..8364d3123 100644
+--- a/inst/www/shared/selectize/scss/selectize.bootstrap5.scss
++++ b/inst/www/shared/selectize/scss/selectize.bootstrap5.scss
+@@ -4,7 +4,8 @@ $select-font-family: if($input-btn-font-family, $input-btn-font-family, inherit) !default;
+ $select-font-size: $input-btn-font-size !default;
+ $select-line-height: $input-btn-line-height !default; //formerly line-height-computed
+ 
+-$select-color-text: $gray-800 !default; //$gray-800
++$select-color-rgb: var(--bs-emphasis-color-rgb, 0, 0, 0) !default;
++$select-color-text: rgba($select-color-rgb, 0.83) !default;
```
cpsievert, line 49:
```suggestion
+$select-color-text: RGBA($select-color-rgb, 0.83) !default;
```

gadenbuie (approved):
LGTM!

PR #3890, opened 2023-09-07
Fix regression in `updateSelectizeInput()`

### LINKED ISSUE #3889: Regression in `updateSelectizeInput()` introduced by v0.15.2
We recently (in #3875) upgraded selectize.js to v0.15.2. This upgrade appears to have caused essentially any use of `updateSelectizeInput()` to break the input binding in such a way that, once the update occurs, the input value is no longer reflected in the server-side input value. Here's a minimal reprex (notice how `textOutput()` always says "Selected a", when it should reflect the `selectizeInput()`'s value).

```r
library(shiny)

ui <- fluidPage(
  selectizeInput("x", NULL, choices = letters),
  textOutput("selected")
)

  updateSelectizeInput(
    session, "x", choices = c("foo", "bar")
  )
  
  output$selected <- renderText({
    paste("Selected", input$x)
  })
  
}

shinyApp(ui, server)
```

cpsievert:
Closes #3889


gadenbuie (approved):
I've tested this locally with an app that makes heavy use of `updateSelectizeInput()` and didn't notice any problems!

PR #3897, opened 2023-09-14
Update to the development version of selectize.js

### LINKED ISSUE #3896: selectize dropdown isn't dismiss-able after multiple `updateSelectInput()` calls
Introduced by #3875. To reprex, run this app and notice how, after selecting items in the 1st dropdown, the dropdown doesn't go away when clicking elsewhere on the page:

```r
library(shiny)

ui <- fluidPage(
  div(
    style = "display:flex",
    selectizeInput("foo", NULL, letters, multiple = TRUE),
    selectizeInput("bar", NULL, letters, multiple = TRUE)
  )
)

  observe({
    updateSelectizeInput(inputId = "foo", choices = letters[1:5])
    updateSelectizeInput(inputId = "bar", choices = letters[6:10])
  })
  
}

shinyApp(ui, server)
``` 

Here are some really odd things to note about the conditions under which the reprex occurs:

1. More than one input/update is necessary.
2. Only the 1st dropdown is problematic.
3. The order of the updates matters. That is, if you update `"bar"` first, the problem disappears?!

cpsievert:
Fixes #3896

This PR moves us from the v0.15.2 tag to the current `HEAD` of https://github.com/selectize/selectize.js. For reference, here are those changes:

https://github.com/selectize/selectize.js/compare/v0.15.2...e3f2e0b

Turns out those changes include some pretty important fixes to focus handling that not only fixes #3896, but also removes the need for the changes added to our `selectInput` binding as a part of the original v0.15.2 update (#3875) to fix this selectize issue (https://github.com/selectize/selectize.js/issues/2032), so this PR also gets rid of that change (https://github.com/rstudio/shiny/pull/3875/files#diff-38d2e2c823f44474b8eb28e08331b223077004554c47884d287f18224d4f8fdc)


gadenbuie (approved):
LGTM!

I'm assuming we'll hope for a selectize release before our next Shiny release? Or at least we should monitor selectize so we can update to an official release when it arrives.

cpsievert:
Yea, hopefully. FWIW, I do subscribe to GH releases of various projects we care about (selectize, bootstrap, etc)

PR #3915, opened 2023-10-16
Soften visually jarring greyout when autoreloading

jcheng5:
(Note: This mostly doesn't apply to Shiny for R, because autoreloading is pretty quick, but rather for Shiny for Python, which terminates the existing Shiny process before starting a new one.)

This change detects when Shiny is autoreloading, and instead of showing the typical disconnection treatment (dark grey), it initially shows nothing and then quickly fades to very light grey.

The goal is for instant autoreloads to feel seamless, but longer autoreloads to make it clear when the autoreload has completed, and still not feel as jarring as a full disconnect.

## TODO

- [x] Putting computer to sleep with Shiny for Python in VSCode causes the autoreload to be broken. While I'm in this code, I'll have the auto-reload code try to reconnect.

## Testing Notes (softened greyout during autoreload)

This can only be tested when the copies of shiny-autoreload.js and shiny(.min).js and shiny.css from inst/www/shared are all copied to a Shiny for Python installation.

- Launch a Shiny for Python app using the VS Code extension. Make it one that starts quickly.
- Make a trivial change and save it. You should see it restart without a flash of grey.
- Add `import time; time.sleep(3)` to the app.py, and save it. Now you should see it fade to a lighter grey before reloading.
- Remove those lines, and add `raise RuntimeError("boom")` to the app's server function. Now you should see the app turn the darker grey on load. (Autoreload should still work, but shouldn't make the dark grey lighter.)

## Testing Notes (auto-reload after computer sleep)

- Launch a Shiny app with autoreload turned on. (This can either by Shiny for R with `options(shiny.devmode=TRUE)` or Shiny for Python with `shiny run --reload` or by using the Shiny for Python VS Code extension's "Run Shiny App" button.) If using RStudio, pop the Viewer out into an external window and use that browser to test instead of Viewer.
- Put the computer to sleep.
- Wake the computer up. The Shiny apps you left running should be greyed out.
- Make a trivial change to your app code and save it.

With this change, the greyed-out apps should reload.


wch (approved):

cpsievert:
@jcheng5 just checking, did you want to mention this in the NEWS?

Issue #3909, opened 2023-10-02
Can't add items to `selectizeInput(multiple = TRUE)` via mouse after multiple updates


cpsievert:
Introduced by #3875. 

In the example below, items can't be added to the 1st input via mouse click (keyboard interaction works fine). Interestingly, if the `"bar"` update() gets removed and/or happens before the `"foo"` update, then the problem goes away

```r
library(shiny)

ui <- fluidPage(
  div(
    style = "display:flex",
    selectizeInput("foo", NULL, letters, multiple = TRUE),
    selectizeInput("bar", NULL, letters, multiple = TRUE)
  )
)

server <- function(input, output, session) {
  
  observe({
    updateSelectizeInput(inputId = "foo", choices = letters[1:5])
    updateSelectizeInput(inputId = "bar", choices = letters[6:10])
  })
  
}

shinyApp(ui, server)
```

cpsievert:
Turns out this is due to bug in the latest version of selectize https://github.com/selectize/selectize.js/issues/2058

PR #3918, opened 2023-10-19
Add non-minified selectize.js source; and respect shiny.minified option

cpsievert:
This PR adds a non-minified copy of selectize.js' JavaScript, which will be useful for the eventual fix for #3909 (we'll likely need to patch the JS for the fix).

It also makes it so now, with `shiny::devmode()` or `options(shiny.minified = FALSE)`, you'll get the non-minified version


gadenbuie (approved):
Should there be a separator here

PR #3919, opened 2023-10-20
Fix selectize event namespacing issues

### LINKED ISSUE #3909: Can't add items to `selectizeInput(multiple = TRUE)` via mouse after multiple updates

cpsievert:
Closes #3909 

Note that I've proposed the same fix to selectize (https://github.com/selectize/selectize.js/pull/2059), but given there hasn't been activity for 9 months, it might take awhile to get a response.

### TODO

Add an automated test (if we don't already have coverage)


gadenbuie (approved):
Excellent detective work! 🕵️

PR #3923, opened 2023-10-25
Prevent `.selectize()` updates from destroying `.data()` and event listeners

cpsievert:
This PR reverts #3890 (which closed https://github.com/rstudio/shiny/issues/3889) and instead reverts [a recent change made to selectize.js](https://github.com/selectize/selectize.js/commit/d2564a0a9f870e3ee53e9306436907a5df0615b6) to fix that same issue. As [this comment](https://github.com/selectize/selectize.js/commit/d2564a0a9f870e3ee53e9306436907a5df0615b6#r103850773) highlights, that selectize.js change removes `.data()` and event listeners attached to the `<select>` tag (or any of it's children), which is problematic for our `selectInput` binding since bindings are storied/acquired via `.data()` (and it also relies on `change` events to set values back to the server).

#3890 originally worked around the issue by unbinding before and binding after any update occurs (akin to how dynamic UI works), but after #3904, to do that properly would further complicate an already complex `selectInput` binding. Also, once I came back to look more deeply at this, it felt like this destruction action has a pretty good chance of breaking 3rd party code. And, since it seems the selectize.js change is solely about performance, it seems best to just revert it (it feels like a future release of selectize.js might have to do something similar -- or at least change that section in such a way that we'll get alerted by it).


cpsievert:
Talked though this in person with @gadenbuie

PR #3926, opened 2023-10-26
Update selectInput's binding to use selectize.js' `getValue()` method when relevant

### LINKED ISSUE #3925: `selectizeInput()`'s input value can be incorrect after `updateSelectizeInput()`
Introduced by #3875 and discovered via [this test app](https://github.com/rstudio/shinycoreci/blob/main/inst/apps/306-accordion-add-remove/). Below is a minimal reprex of the problem we were seeing there.

To reproduce, run the `app.R` below. Once open, remove "A" and "D" from the 1st selectize (choices). That causes `input$selected` to be `NULL` (when it should be `"B"`!).

If you then open the browser's devtools, and run the following JS `$("select#selected").val()`, you'll see the result is `[]`, which is why the binding is reporting `NULL`. Interestingly, `$("select#selected").selectize().getValue()` returns the expect result (`["B"]`), so selectize.js must've recently moved away from using the DOM as the source of truth for the select's state.

<details>
<summary><code>app.R</code></summary>

```r
library(shiny)

choices <- c("A", "B", "C", "D")

ui <- fluidPage(
  verbatimTextOutput("selected"),
  div(
    style = "display:flex;",
    selectInput(
      "choices", NULL, multiple = TRUE,
      choices = choices, selected = choices
    ),
    selectInput(
      "selected", NULL, multiple = TRUE, 
      choices, selected = c("A", "B", "D")
    )
  )
)

  observeEvent(input$choices, ignoreInit = TRUE, {
    updateSelectInput(
      inputId = "selected", choices = input$choices,
      selected = input$selected
    )
  })

  output$selected <- renderPrint({
    paste("Value of the 2nd selectize (input$selected):", paste(input$selected, collapse = ", "))
  })

}

shinyApp(ui, server)
```

</details>

cpsievert:
Closes #3925


srcts/src/bindings/input/selectInput.ts
```diff
@@ -62,10 +61,14 @@ class SelectInputBinding extends InputBinding {
...
-  ): NotUndefined<ReturnType<JQuery<HTMLElement>["val"]>> {
-    return $(el).val() as NotUndefined<ReturnType<JQuery<HTMLElement>["val"]>>;
+  getValue(el: SelectHTMLElement): any {
+    if (!isSelectize(el)) {
+      return $(el).val();
+    } else {
+      const selectize = this._selectize(el);
+
+      return selectize?.getValue();
+    }
```
cpsievert, line 71: FWIW, this is essentially same logic as `.setValue()` (which is why I've written it this way)

PR #3929, opened 2023-10-27
Revert `Shiny.renderContent()` back to sync instead of async

cpsievert:
Motivated by https://github.com/rstudio/bslib/issues/872

This PR essentially reverts the changes made to [`srcts/src/shiny/render.ts`](https://github.com/rstudio/shiny/pull/3904/files#diff-7aa61053d5a807425314d89085e38603c93b1c63a841d378e279aa3cd70417de) in https://github.com/rstudio/shiny/pull/3904, then does:

1. Keeps the `shinyBindAll(el)` -> `await shinyBindAll(el)` change made to `renderContentAsync()` 
2. Adds a `shinyBindAll(el)` -> `return shinyBindAll(el)` change in `renderContent()`  (so you can await the bind if need be)


wch:
If we were to use the synchronous code path for loading these resources, then it won't work in shinylive on Chrome (because Chrome's service worker implementation doesn't intercept synchronous XHR requests, which is what jQuery users for synchronous loading of JS resources).

I'll comment on the bslib issue about possible ways forward there.

srcts/src/shiny/render.ts
```diff
@@ -79,22 +79,48 @@ async function renderContent(
   }
 }
 
-// This function was introduced in v1.7.5, then deprecated in v1.8.0 once we
```
wch, line 82: For `renderContent`, please add some information about why this won't work in shinylive (and about the synchronous XHR warnings even outside of shinylive, #789).

And for `renderContentAsync`, tell people that this is what they should use going forward.

cpsievert: There's a pretty good (and lengthy) comment just above `renderContentAsync()` on this topic

wch (approved):

PR #3930, opened 2023-10-28
Allow `InputBinding.receiveMessage` to be async

wch:
This is motivated by https://github.com/rstudio/bslib/issues/872.
It allows an input binding's `receiveMessage()` method to be synchronous or asynchronous, similar to [`OutputBinding.renderValue()`](https://github.com/rstudio/shiny/blob/77bc4e9ec72e97a25f64134ba981d8b2e9cb5895/srcts/src/bindings/output/outputBinding.ts#L14).


wch:
It might make sense to take this PR _and_ #3929, which makes `renderContent` synchronous again, and adds back `renderContentAsync`.

Although this PR solves the specific issue we have with bslib (when paired with https://github.com/rstudio/bslib/pull/874), there are potentially cases where people have written JS code which calls `renderContent` and then immediately tries to do something with the new content. Changing `renderContent` to async could break that code. Although those cases could generally be easily fixed to work with async, it would still be a breaking change. And since we have `renderHtml`/`renderHtmlAsync` and `renderDependencies`/`renderDependenciesAsync`, it would make sense to also have `renderContent`/`renderContentAsync`.

cpsievert (approved):

cpsievert:
> It might make sense to take this PR and https://github.com/rstudio/shiny/pull/3929

Yea, I still think we should take https://github.com/rstudio/shiny/pull/3929

PR #3931, opened 2023-10-30
Client error console and duplicate input/output binding errors

nstrayer:
# What?
This PR adds a client-side error console for surfacing errors to the user that previously would go unseen unless they had their browser's devtools open. 

Also included is special detection of duplicate input and output bindings which would previously silently either break your app (in the case of outputs) or cause confusing behavior (only first bound input would work.) 

image

There is support for multiple errors:
image

The following test app will trigger input and output binding errors. (Note you'll need to comment out the second output to get the input error to show up because output binding errors stop the entire Shiny runtime.

```R
library(shiny)

ui <- fluidPage(

    # Application title

    sidebarLayout(
        sidebarPanel(
            sliderInput("bins",
                        "Number of bins:",
                        min = 1,
                        max = 50,
                        value = 30),
            numericInput("bins","Number of bins:",value = 30)
        ),
        mainPanel(
           # Multiple binding collisions will be caught and displayed in a
           # single error message
           plotOutput("distPlot", width = "50%"),
           plotOutput("distPlot"),
           plotOutput("secondPlot", width = "50%"),
           plotOutput("secondPlot")
        )
    )
)

        x    <- faithful[, 2]
        bins <- seq(min(x), max(x), length.out = input$bins + 1)

        hist(x, breaks = bins, col = 'darkgray', border = 'white',
             xlab = 'Waiting time to next eruption (in mins)',
             main = 'Histogram of waiting times')
    })
}

# Run the application
```

# How?
Adds two webcomponents: `shiny-error-console-container` and `shiny-error-console`. A function `showErrorInClientConsole(e)` adds the errors to the page (as children of a `shiny-error-console-container` for layout purposes) as they come in.


nstrayer:
App that shows catching of duplicate IDs across static and dynamic UI

```R
library(shiny)

ui <- fluidPage(
    sidebarLayout(
        sidebarPanel(
            sliderInput("bins",
                        "Number of bins:",
                        min = 1,
                        max = 50,
                        value = 30),
            checkboxInput("showDynamicPlot", "Show dynamic plot")
        ),
        mainPanel(
           plotOutput("distPlot"),
           # Uncomment this line to trigger initialization duplicate error
           # plotOutput("distPlot", width = "50%"),
           uiOutput("dynamicPlot"),
        )
    )
)

        x    <- faithful[, 2]
        bins <- seq(min(x), max(x), length.out = input$bins + 1)

        hist(x, breaks = bins, col = 'darkgray', border = 'white',
             xlab = 'Waiting time to next eruption (in mins)',
             main = 'Histogram of waiting times')
    })

    output$dynamicPlot <- renderUI({
      if (input$showDynamicPlot) {
        sliderInput("bins",
          "Number of bins:",
          min = 1,
          max = 50,
          value = 30)
      }
    })
}

```

nstrayer:
image
The latest commit changes the styles so that the container component holds most of the logic. Also adds a dismiss button and a collapse button.

michaelhogersnplm:
Hello @nstrayer, great addition - it certainly has the potential to reduce confusion occasionally. I'm curious, though: could there be a feature to switch this off, perhaps through a setting, for instances where it might be preferable to keep it concealed in a production environment?

Best,
Michael

srcts/src/components/errorConsole.ts
```diff
@@ -0,0 +1,511 @@
...
+ */
+export class ShinyClientError extends Error {
+  headline: string;
+
+  constructor({ headline, message }: { headline: string; message: string }) {
+    super(message);
+    this.name = "ShinyClientError";
+    this.headline = headline;
+  }
+}
```
wch, line 511: Let's move this to a different file, like src/shiny/error.ts.

srcts/src/shiny/bind.ts
```diff
@@ -85,9 +131,16 @@ function bindInputs(
       if (el.hasAttribute("data-shiny-no-bind-input")) continue;
       const id = binding.getId(el);
 
-      // Check if ID is falsy, or if already bound
-      if (!id || boundInputs[id]) continue;
+      // Check for duplicates in bindingIds array and keep track of them
+      const duplicateId = bindingIdExists(id);
```
wch, line 135:
```suggestion
      const isDuplicateId = bindingIdExists(id);
```

srcts/src/shiny/bind.ts
```diff
@@ -137,6 +190,11 @@ function bindInputs(
     }
   }
 
+  // Send error message to the user if duplicate IDs are found
+  if (inputDuplicateIds.size > 0) {
+    throwDuplicateIdError(inputDuplicateIds);
```
wch, line 195: Should `throw` from here, instead of from inside of the function.

srcts/src/shiny/bind.ts
```diff
@@ -156,13 +217,22 @@ async function bindOutputs(
...
 
       // Check if ID is falsy
       if (!id) continue;
 
+      // Check for duplicates in bindingIds array and keep track of them
+      if (bindingIdExists(id)) {
+        outputDuplicateIds.add(id);
+      }
+
+      bindingIds.outputs.add(id);
```
wch, line 234: Check that this is consistent with how it's done for inputs.

srcts/src/shiny/bind.ts
```diff
@@ -44,6 +45,48 @@ function valueChangeCallback(
...
+  inputs: new Set<string>(),
+};
+
+/**
+ * Check if a binding id already exists in the bindingIds set across both inputs and outputs
+ * @param id Id to check
+ * @returns boolean indicating if that binding Id has already been added to app
+ */
+function bindingIdExists(id: string): boolean {
+  return bindingIds.outputs.has(id) || bindingIds.inputs.has(id);
```
wch, line 71: Check that `id` is not an empty string. In that case, throw.

srcts/src/shiny/bind.ts
```diff
@@ -85,9 +131,16 @@ function bindInputs(
...
+      // Check for duplicates in bindingIds array and keep track of them
+      const duplicateId = bindingIdExists(id);
+      if (duplicateId) {
+        inputDuplicateIds.add(id);
+      }
+      // Check if ID is falsy, or if already bound, or has the same ID as
+      // another input binding and if it is, skip
+      if (!id || boundInputs[id] || duplicateId) continue;
 
+      bindingIds.inputs.add(id);
```
wch, line 143: Abstract this into a function which checks that `id` is not empty string.

srcts/src/shiny/bind.ts
```diff
@@ -85,9 +131,16 @@ function bindInputs(
...
-      // Check if ID is falsy, or if already bound
-      if (!id || boundInputs[id]) continue;
+      // Check for duplicates in bindingIds array and keep track of them
+      const duplicateId = bindingIdExists(id);
+      if (duplicateId) {
+        inputDuplicateIds.add(id);
+      }
+      // Check if ID is falsy, or if already bound, or has the same ID as
+      // another input binding and if it is, skip
+      if (!id || boundInputs[id] || duplicateId) continue;
```
wch, line 141: Move `!id` check above the duplicated ID check.

Remove all of the `boundInputs` code because it's redundant with the new code.

srcts/src/shiny/bind.ts
```diff
@@ -223,6 +298,8 @@ function unbindInputs(
 
     $(el).removeClass("shiny-bound-input");
     delete boundInputs[id];
+
+    bindingIds.inputs.delete(id);
```
wch, line 302: Wrap this into a function.

srcts/src/shiny/index.ts
```diff
@@ -110,9 +111,14 @@ function setShiny(windowShiny_: Shiny): void {
   $(function () {
     // Init Shiny a little later than document ready, so user code can
     // run first (i.e. to register bindings)
-    setTimeout(function () {
+    setTimeout(async function () {
       /* eslint-disable @typescript-eslint/no-floating-promises */
```
wch, line 115:
```suggestion
```

wch (changes requested):
Looks mostly good, just a few comments.

nstrayer:
> Hello @nstrayer, great addition - it certainly has the potential to reduce confusion occasionally. I'm curious, though: could there be a feature to switch this off, perhaps through a setting, for instances where it might be preferable to keep it concealed in a production environment?
> 
> Best, Michael

Hi Michael,

This is a good point. Right now this only should catch errors that would have otherwise broken your app. We're going to try bringing it in where it is always active and then run our large suite of tests on Shiny to see if we're going to start causing spurious error messages to appear. If it becomes a problem then the plan would be to make it only show up when devmode is turned on for Shiny.

nstrayer:
In [34cf27e](https://github.com/rstudio/shiny/pull/3931/commits/34cf27ed46e5c96c9c2cc110004f60b88c6852d4), I switched where the logic for throwing on duplicate ID bindings was. Now we let Shiny run through all the input and output bindings. Then after both `bindInputs()` and `bindOtuputs()` have run, we validate and throw a much more helpful error message that contains all the duplicates. This means we can get error messages like this:

```
The following IDs were repeated: 
"vars": (2 inputs and 1 output), 
"bins": (2 inputs)
```

nstrayer:
This is what the current error messages look like, since the previous screenshots are out-of-date: 
image

srcts/src/shiny/bind.ts
```diff
@@ -85,9 +204,11 @@ function bindInputs(
       if (el.hasAttribute("data-shiny-no-bind-input")) continue;
       const id = binding.getId(el);
 
-      // Check if ID is falsy, or if already bound
-      if (!id || boundInputs[id]) continue;
+      // Check if ID is falsy, or if already bound, or has the same ID as
+      // another input binding and if it is, skip
```
wch, line 208: Fix comment

srcts/src/shiny/bind.ts
```diff
@@ -44,6 +41,128 @@ function valueChangeCallback(
...
+
+  // Main store of bindings.
+  const bindings: IdToBindingTypes = new Map();
+
+  /**
+   * Checks if the bindings registery is valid. Currently this just checks for
+   * duplicate IDs but in the future could be expanded to check more conditions
+   * @returns ShinyClientError if current ID bindings are invalid, otherwise null
+   */
+  function getValidity():
```
wch, line 68: Rename to `checkValidity`.

srcts/src/shiny/bind.ts
```diff
@@ -44,6 +41,128 @@ function valueChangeCallback(
...
+
+    bindings.forEach((inputOrOutput, id) => {
+      if (inputOrOutput.length > 1) {
+        duplicateIds.set(id, inputOrOutput);
+      }
+    });
+
+    if (duplicateIds.size === 0) return { status: "ok" };
+
+    const duplicateIdMsg = [...duplicateIds.entries()]
```
wch, line 81: Use a more efficient copying technique.

wch (approved):

PR #3947, opened 2023-11-29
Make error console sizing constant across base and bslib apps

nstrayer:
Previously, when the error console appeared in an app using bslib it was comically large. This was due to the original console being designed with rem units for sizing and base shiny using 14px as the base font size. The bootstrap that bslib brings in has upped that size to 16px (in keeping with standards) and as a result, the error console got larger accordingly. 
This PR updates this so that we use hard-coded font sizes instead. The main flaw is that the console does not correspond to zooms done by the user on the font size. However, we already break that a bit by hard-coding a font size on the body so it's a reasonable tradeoff.


jcheng5:
I don't think of the error console as part of the UI, nor subject to the app author's preferences in terms of bslib or other CSS customizations--I think it's totally OK for it to be hard-coded to a px size.

jcheng5 (approved):

jcheng5 (approved):

PR #3954, opened 2023-12-06
Allow outputs to stay in progress mode after flush

jcheng5:
Up until now, Shiny bound output enter progress mode (the `.recalculating` class is added) when their respective observer invalidates ([source](https://github.com/rstudio/shiny/blob/59b1c46485a9c922824bfa04a19194a5668a1118/srcts/src/shiny/shinyapp.ts#L1404-L1417)), and exit progress mode when that reactive flush cycle completes ([source](https://github.com/rstudio/shiny/blob/59b1c46485a9c922824bfa04a19194a5668a1118/srcts/src/shiny/shinyapp.ts#L697-L700)).

This PR adds a `req(FALSE, cancelOutput="progress")` which behaves similarly to `cancelOutput=TRUE`, but also keeps the output in `.recalculating` state even across flush cycles. This is called "persistent progress" and an output can leave this state when it is invalidated again and doesn't call `req(FALSE, cancelOutput="progress")` during that flush cycle.

This will be useful for implementing long-running tasks that don't hold up the flush cycle, leaving sessions responsive to do other tasks.

This feature is borderline internal (not that we can't commit to it, but because it's so low-level) so I'm not sure it needs a NEWS.md update--that can wait for the high-level feature we build on top of this, I think.


jcheng5:
The binding state management code is getting pretty creaky, but I don't think this is the straw that breaks the camel's back.

wch (approved):

Issue #3968, opened 2024-01-14
Update jQuery version to 3.7.1


davidmacro:
We are getting some warnings by our automated vulnerability checkers about outdated libraries, one of them being the use of jQuery 3.6.0. 

If possible, we would like to have jQuery upgrade to latest version (currently 3.7.1).

PR #3969, opened 2024-01-14
Update jquery

davidmacro:
I followed the steps as outlined in **tools/updatejQuery.R** to update jQuery to 3.7.1 (current latest stable) and jQuery Types to 3.5.29 (current latest stable). This ran without problems. 

To be sure, I verified that: 

- Correct .js files were included in `inst/www/shared` folders (`jquery.js`, `jquery.min.js`, `jquery.min.js.map)`
- `yarn.lock` and 'package.json'.file were updated accordingly
- 'devtools::check()' ran and introduced no (new) issues. 

_Note: devtools::check() did give me an (unrelated) error in the tabPanel test. But this test also occurs in main/ at the moment. Looking at the diff, it appears to be that the generated html wraps a body tag around the code that is to be asserted._


davidmacro:
Link to issue #3968

cpsievert:
Thanks, if you could also update the `NEWS.md`, that'd be great!

package.json
```diff
@@ -23,7 +23,7 @@
     "@types/bootstrap-datepicker": "0.0.14",
     "@types/datatables.net": "^1.10.19",
     "@types/ion-rangeslider": "2.3.0",
-    "@types/jquery": "3.5.14",
+    "@types/jquery": "3.5.29",
```
cpsievert:
```suggestion
    "@types/jquery": "3.5.32",
```

PR #3978, opened 2024-01-24
Fix duplicate ID check logic

wch:
This PR fixes a bug in the duplicate ID check.


inst/www/shared/shiny.js
```diff
@@ -18577,7 +18577,7 @@
         if (counts.input === 1 && counts.output === 1 && !Shiny.inDevMode()) {
           return;
         }
-        if (counts.input > 0 || counts.output > 0) {
```
nstrayer, line 18580: I must have meant this to be an `&&` instead. I think the change is functionally equivalent.

wch: I think they're still not quite equivalent, for example if `input==0` and `output==2`.

PR #3999, opened 2024-03-16
chore: remove repetitive words

avoidaway:


wch:
Thanks for this PR. However, this PR modifies a file that should not be modified (jquery.js). Can you revert the change to that file?

avoidaway:
> Thanks for this PR. However, this PR modifies a file that should not be modified (jquery.js). Can you revert the change to that file?

Thanks! Reverted

cpsievert:
Thanks!

PR #4001, opened 2024-03-20
Use bs-danger instead of bs-danger-text-emphasis for text color on error messages

### LINKED ISSUE #4000: Error message text color is too subtle with Bootstrap 5
In https://github.com/rstudio/shiny/pull/3914, the text color of error messages changed from `red` to `--bs-danger-text-emphasis`, which I think was designed to be put on a danger bg, and IMO is too close to `black` for error messages. Perhaps `--bs-danger` would be better @gadenbuie?

```r
library(shiny)
library(bslib)

ui <- page_fixed(
  textOutput("foo")
)

  output$foo <- renderText({1+"a"})
}

shinyApp(ui, server)
```

Screenshot 2024-03-19 at 5 29 40 PM

cpsievert:
Closes #4000


gadenbuie (approved):
Thanks!

PR #4008, opened 2024-03-21
v1.8.1 release candidate

cpsievert:


gadenbuie (approved):
Looks great!

Just a note that we might want to retroactively add the extended task blog post/announcement to the NEWS once it's published.

PR #4010, opened 2024-03-22
Revert changes to `shinyDeprecated()` & update `renderDataTables()` tests to always use shiny's datatables implementation

cpsievert:
This PR does two things:

1. Reverts #4007. Unfortunately, this would cause 7 reverse dependencies to start failing, so the change doesn't feel worth our time (or other maintainers time).
2. In looking in these `renderDataTables()` tests again, I realized that the return value of `renderDataTable()` will change when the next version of `{DT}` goes to CRAN (which will break the relevant shiny unit test). Instead of updating that unit test to reflect what `{DT}` gives us, I've set the relevant option to use shiny's implementation (so that future changes to DT doen't break shiny's tests).

### DIFF
--- a/R/deprecated.R
+++ b/R/deprecated.R
@@ -30,13 +30,7 @@ shinyDeprecated <- function(
   }
 
   # lifecycle::deprecate_soft(when, what, with = with, details = details, id = id, env = env)
-  rlang::warn(
-    message = msg,
-    .frequency = "always",
-    .frequency_id = msg,
-    .file = stderr(),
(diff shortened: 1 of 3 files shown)


gadenbuie (approved):
Thanks! I wouldn't want the shiny deprecated message change to block the release. But I do think it's worth considering again in the future. It's hard to track down the source of a message, it's definitely easier to find the source of the deprecation notice if it's a warning. Can you copy the reverse depends failures into a new issue so I can take a look?

PR #4015, opened 2024-03-27
ExtendedTask should not be cloneable

jcheng5:
I was reviewing the docs for ExtendedTask and noticed that the `$clone()` method was documented. I'm not sure what it would mean if an ExtendedTask object was cloned, but I'm pretty sure it wouldn't be good, given that it contains reactive values inside. And I don't know why someone would want to clone it anyway. Better to just prevent it.

### DIFF
--- a/R/extended-task.R
+++ b/R/extended-task.R
@@ -42,7 +42,7 @@
 #'   session to immediately unblock and carry on with other user interactions.
 #'
 #' @export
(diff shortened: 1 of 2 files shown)


wch (approved):
Just another reminder that I should have made `cloneable=FALSE` the default from the beginning.

PR #4025, opened 2024-04-06
chore: remove repetitive word

guqicun:


cpsievert:
The change here is to a comment in jQuery's source, which we don't make changes to. Please consider submitting a fix to jQeury itself https://github.com/jquery/jquery

PR #4041, opened 2024-05-03
Disable `downloadLink()`/`downloadButton()` until their `href` is ready

### LINKED ISSUE #2614: Download button works even without handler
```R
library(shiny)
ui <- fluidPage(
  downloadButton("download")
)
}
shinyApp(ui, server)
```

It downloads the current page — this lead to a few minutes head scratching when I accidentally used the wrong identifier when trying to set a handler.

### LINKED ISSUE #3606: Download handler downloads some "download.html" with the current page's HTML code while app is first loading
(I am not sure this is an issue with Shiny or Shiny Server - it might be the latter as I cannot reproduce it locally.)

### System details

Browser Version: Windows and Linux, Chrome 100 and Firefox

### Steps to reproduce the problem

1. Visit https://shiny.rstudio.com/gallery/file-download.html
2. Reload page
3. **Quickly** click "Download"

chrome_f0bP4kXCd3

(In the video, I clicked twice - the first download is incorrect, the second download is correct.)

Some "download.html" file is downloaded, see [download.html.txt](https://github.com/rstudio/shiny/files/8383553/download.html.txt).

I see this problem a lot when automating a Shiny app with Selenium, and have also come across it in manual app use. Apparently, it is not enough to wait for the download button to become clickable. Can I modify the Shiny app to prevent this issue? Is there any other state change that I can reliably wait for in Selenium to ensure that the download will succeed?

cpsievert:
Closes #2614
Closes #3606


gadenbuie (approved):
I haven't run the code, but I have read it carefully. It looks like very elegant fix to the problem! I'm glad to see this one getting fixed :)

PR #4047, opened 2024-05-07
Allow for progress binding to move from Idle to Invalidated status

### LINKED ISSUE #4046: Updating reactive value on a timer results on JS error
Introduced by #4039. This app now produces a JS error: `Shiny server has sent a progress message for number, but the output is in an unexpected state of: idle`.

Seems it's expected that a progress message can be sent to move output status from Idle to Invalidated (i.e., skip the value stage)

```r
library(shiny)

devmode()

ui <- fluidPage(uiOutput("number"))

    output$number <- renderUI({ val() })
    val <- reactiveVal(0)
    observe({
        invalidateLater(3000)
        val(runif(1))
    })
}

shinyApp(ui, server)
```

cpsievert:
Closes #4046


cpsievert:
Merging now to unblock other PRs (also went through this fix together with @wch)

PR #4061, opened 2024-05-22
Hide the busy pulse when disconnected

cpsievert:
Currently, when `useBusyIndicators()` is activated, and the app disconnects while shiny is busy (e.g., an error in an observer occurs), you may get a pulse at the top of the page. That isn't an ideal experience because the server isn't actually busy -- it's been disconnected.

Here's a testing app for the change. Before this change, the pulse would persist after the error occurs. After this change, the pulse goes away after the error occurs.

https://github.com/rstudio/shiny/assets/1365941/337b474e-4743-4bc1-bf9a-a6ecaff2bda8

```r
library(shiny)

ui <- fluidPage(
  useBusyIndicators(),
  actionButton("btn", "Generate error"),
  plotOutput("p"),
)

  observeEvent(input$btn, {
    Sys.sleep(2)
    stop("boom")
  })

  output$p <- renderPlot({
    Sys.sleep(2)
    plot(1)
  })

}

shinyApp(ui, server)

```


srcts/extras/busy-indicators/busy-indicators.scss
```diff
@@ -90,16 +90,25 @@
...
 */
 [data-shiny-busy-spinners][data-shiny-busy-pulse] {
-  &.shiny-busy:not(:has(.recalculating))::after {
+  &.shiny-busy::after {
     @include shiny-page-busy;
   }
+  &.shiny-busy:has(.recalculating)::after {
+    display: none;
+  }
+  &.shiny-busy:has(#shiny-disconnected-overlay)::after {
```
jcheng5, line 99: Someday might be worth adding a class or attribute somewhere (`<html>`? `<body>`?) for disconnected status.

jcheng5 (approved):

PR #4062, opened 2024-05-23
Delay longer before dimming opacity when spinners are active

cpsievert:


gadenbuie (approved):
LGTM! (paired in writing this and reviewed together)

PR #4063, opened 2024-05-23
Add `Shiny.initializedPromise`

wch:
This PR adds `Shiny.initializedPromise`, which is a Promise-like object.

- `Shiny.initializedPromise` is resolved when the `shiny:sessionintialized` event is triggered.

The problem with the previous way of doing things, is that there needs to be two paths in the code: If your code executes before the event is triggered, then you need to create a jQuery event listener for the `shiny:sessionintialized` event. If your code executes after the event, then you need to just run your function. And detecting whether the event has happened is not straightforward.

With the new method, you can simply write this code, and it will work in both situations:

```js
Shiny.initializedPromise.then(() => {
  // Do something here
});
```

See below for an example.

This pull request also:

- Converts `window.Shiny` to a normal JS class object.
- Adds a prettier plugin which sorts imports
- Simplifies some types in the `Shiny` class.

*****

The following example app demonstrates how the promises work and compares them to the event listeners. The promises chains and event listeners are set up in both the static UI and in dynamic UI.

For the static UI, all of the callbacks will execute. But for the dynamic UI, only the promise callbacks will execute; the event listeners will not. This is because the dynamic UI code runs _after_ those events have already been triggered.

```R
shinyApp(
  ui = fluidPage(
    "Initialization test app",
    uiOutput('dynui'),
    pre(id="logs"),
    tags$script(HTML('
      function log(s) {
        document.getElementById("logs").innerText += s + "\\n";
      };
    
      Shiny.initializedPromise.then(() => {
        log("static content: initializedPromise resolved!")
      });
      $(document).on("shiny:sessioninitialized", () => {
        log("static content: shiny:sessioninitialized event!")
      });      
    '))
  ),
    output$dynui <- renderUI({
      tags$script(HTML('

        Shiny.initializedPromise.then(() => {
          log("dynamic content: initializedPromise resolved!")
        });
        $(document).on("shiny:sessioninitialized", () => {
          log("dynamic content: shiny:sessioninitialized event!")
        });      
      '))
    })
  }
)
```

If you open this app in a browser, this is what you will see:

image


.eslintrc.yml
```diff
@@ -35,10 +35,6 @@ rules:
 
   default-case:
     - error
-  indent:
-    - error
-    - 2
-    - SwitchCase: 1
```
wch, line 41: This was causing problems with functions where the return type was an object, like:

```js
export function promiseWithResolvers<T>(): {
  promise: Promise<T>;
  resolve: (value: PromiseLike<T> | T) => void;
  reject: (reason?: any) => void;
} {   // <- eslint said there was an error here, but that's how prettier auto-formats it
  // Code here
}
```

srcts/extras/globalShiny.ts
```diff
@@ -2,19 +2,16 @@
...
 
   // Tell window.Shiny exists
   interface Window {
     // eslint-disable-next-line @typescript-eslint/naming-convention
-    Shiny: RStudioShiny;
+    Shiny: ShinyClass;
   }
-
-  // Make `Shiny` a globally available type definition. (No need to import the type)
-  type Shiny = RStudioShiny;
```
wch, line 19: This type didn't seem to be used anywhere, and it was confusing that the type and the variable had the same name.

srcts/src/initialize/index.ts
```diff
@@ -2,15 +2,14 @@ import { disableFormSubmission } from "./disableForm";
...
-import { setShiny } from "../shiny";
+import { Shiny } from "../shiny";
 import { setUserAgent } from "../utils/userAgent";
 import { windowUserAgent } from "../window/userAgent";
 
 import { initReactlog } from "../shiny/reactlog";
 
 function init(): void {
-  setShiny(windowShiny());
+  window.Shiny = window.Shiny || new Shiny();
```
wch, line 12: I think this captures what the previous `setShiny()` function was doing, but I am not 100% sure.

srcts/src/shiny/index.ts
```diff
@@ -54,7 +56,7 @@ interface Shiny {
   renderContent: typeof renderContent;
   renderHtmlAsync: typeof renderHtmlAsync;
   renderHtml: typeof renderHtml;
-  user: string;
+  user?: string;
```
wch, line 59: I added the `?` here, which reflects how the code actually is written. (The conversion from interface to class flagged this issue.)

srcts/src/shiny/init.ts
```diff
@@ -517,8 +517,15 @@ async function initShiny(windowShiny: Shiny): Promise<void> {
...
-  $(document).one("shiny:connected", function () {
+  $(document).one("shiny:connected", (event) => {
     initDeferredIframes();
+    // @ts-expect-error; .socket property isn't a known property of
+    // JQuery.TriggeredEvent, but it was added on there anyway.
+    windowShiny._resolveConnectedPromise(event.socket);
+  });
+
+  $(document).one("shiny:sessioninitialized", () => {
+    windowShiny._resolveSessionInitPromise();
```
wch, line 528: The promises are resolved via event listeners, but one could argue that they should be resolved directly from the initialization code. This would be more direct, and let us avoid going through the jQuery event system.

srcts/src/initialize/index.ts
wch: One reason that we previously didn't want to simply overwrite the `window.Shiny` object was because some other library might create a `window.Shiny = {}` object and then populate some of the properties.

This is the old code in this repo that created the `window.Shiny` object this way. https://github.com/rstudio/shiny/blob/v1.4.0/srcjs/_start.js 

```js
  var exports = window.Shiny = window.Shiny || {};
```


However, in practice it looks like, at least for shiny-server-client, its JS must get loaded after shiny.js, so the change in the PR shouldn't cause any behavioral changes:
https://github.com/rstudio/shiny-server-client/blob/92cbc96583f958f4e379120323fb3a48c2e61fd5/lib/main.js#L167-L187

wch: From a conversation with @schloerke: It would be better to just print a console error message if `window.Shiny` already exists at this point.

srcts/src/shiny/index.ts
```diff
@@ -1,42 +1,44 @@
...
+  renderHtmlAsync,
+} from "./render";
+import type { Handler, ShinyApp, ShinyWebSocket } from "./shinyapp";
 import { addCustomMessageHandler } from "./shinyapp";
-import { initInputBindings } from "../bindings/input";
-import { initOutputBindings } from "../bindings/output";
-import { showErrorInClientConsole } from "../components/errorConsole";
 
-interface Shiny {
+class Shiny {
```
wch, line 34:
```suggestion
class ShinyClass {
```

srcts/extras/globalShiny.ts
wch: Should check if bslib complains about this change.

srcts/types/extras/globalShiny.d.ts
```diff
@@ -1,8 +1,7 @@
-import type { Shiny as RStudioShiny } from "../src/shiny/index";
+import type { Shiny as ShinyClass } from "../src/shiny/index";
 declare global {
-    const Shiny: RStudioShiny;
+    const Shiny: ShinyClass;
     interface Window {
-        Shiny: RStudioShiny;
+        Shiny: ShinyClass;
     }
-    type Shiny = RStudioShiny;
```
wch, line 7: The removal of this type leads to some required changes in bslib.
Note that the change is only with types; the JS code does not need to change.
https://github.com/rstudio/bslib/pull/1054

srcts/src/initialize/index.ts
```diff
@@ -2,15 +2,17 @@ import { determineBrowserInfo } from "./browser";
...
 import { windowUserAgent } from "../window/userAgent";
 
 import { initReactlog } from "../shiny/reactlog";
 
+// eslint-disable-next-line @typescript-eslint/naming-convention
+let Shiny: ShinyClass;
+
 function init(): void {
-  setShiny(windowShiny());
+  Shiny = window.Shiny = window.Shiny || new ShinyClass();
```
jcheng5, line 15: Either merge `window.Shiny`, or, throw an error if `window.Shiny` is already defined.

jcheng5:
I'm not a fan of the names `isConnected` and `isInitialized`, that would make me assume it's going to return a boolean. @gadenbuie and I had a similar discussion yesterday, he ended up with the equivalent of `getConnectedPromise`, I think my other suggestions were `waitForConnect` and `joinConnected`.

jcheng5:
What does `isConnected` do if you call it after the app has been disconnected? You just get the resolved promise, I assume... What about on reconnection?

I think I understand what the initialized one is for, I maybe am less sure what the connected one is for. The thing where `Shiny.setInputValue` and other Shiny methods aren't available, is that before initialization or before connected?

wch:
- I'm fine with dropping `isConnected` and just keeping `isInitialized`. This avoids the issue of what to do on a reconnection. And the `shiny:connected` event is not really useful for most cases anyway.

- I don't have an attachment to `isInitialized`. The name `getInitializedPromise` implies that it's a method which returns a promise, but that's not what it is; it's the promise itself. I think the name `initializedPromise` would work, and is a fairly conventional name.

jcheng5 (approved):

PR #4065, opened 2024-05-24
Update formatting settings

wch:
This PR makes the following changes:

- Updates the VS Code settings to format .json files on save.
- Adds `prettier-plugin-organize-imports` to sort JS/TS imports on save.


tools/updatePackageJsonVersion.R
```diff
@@ -1,5 +1,5 @@
 # Updates the package.json `version` to match the DESCRIPTION `Version`
-pkg <- jsonlite::read_json("package.json", simplifyVector = TRUE)
```
wch, line 2: The `simplifyVector = TRUE` here was causing the package.json file to be changed in a way that broke it.

PR #4132, opened 2024-09-24
Fix race condition in throttle/debounce tests

### LINKED ISSUE #4133: Throttle/debounce tests contain a race condition that's causing tests to fail
### System details

This manifests as a difference between the [ubuntu-20.04 (4.1.3)](https://github.com/rstudio/shiny/actions/runs/11020045428/job/30603949400#logs) environment on your task runners, and every other environment I've tried.

Look at any of the recent test runs on PRs other than #4132 (which is my attempt to fix this issue)

There's a race condition in test-reactivity.R between this `invalidateLater(100)`:

https://github.com/rstudio/shiny/blob/d84aa94762b4ffaf7533a007b6cb92c40f4f29af/tests/testthat/test-reactivity.r#L1245-L1248

and this `throttle(..., 500)`:

https://github.com/rstudio/shiny/blob/d84aa94762b4ffaf7533a007b6cb92c40f4f29af/tests/testthat/test-reactivity.r#L1254-L1255

which both fire at 1000ms, creating a race condition that makes the results of this test implementation-dependent.

dvg-p4:
This should resolve the spurious failure that's been showing up in all recent PRs to this repo.

This test updates a reactive value every 100 ms, and tests the behavior of a throttle and debounce with delays set to 500 ms. Since 500 is a multiple of 100, this makes the test sensitive to the precise specifics of CPU scheduling and/or shiny queueing, since the throttle timer expiring (at 2 * 500ms) and the invalidateLater timer triggering  (at 10 * 100ms) both occur at ~1 second. This makes the test sensitive to implementation specifics, since it's not clear which one will happen first. (Empirically, the behavior appears to depend on the particular environment, with the throttle only catching the new value on ubuntu-20.04/R4.1.3.)

This PR fixes the issue by adding an additional 20ms delay to the first `invalidateLater()` loop, offsetting it from the throttle enough to consistently avert the race condition.

Closes #4133 

(dunno why it took me so long to remember the term "race condition" lol)

### DIFF
--- a/tests/testthat/test-reactivity.r
+++ b/tests/testthat/test-reactivity.r
@@ -1243,7 +1243,10 @@ test_that("debounce/throttle work properly (with priming)", {
 
   # This observer will be what changes rv$a.
   src <- observe({
-    invalidateLater(100)
+    if (isolate(rv$a) == 0)
+      invalidateLater(120) # delay 20ms extra once to avoid threshold errors
+    else
+      invalidateLater(100)
     rv$a <- isolate(rv$a) + 1
   })
   on.exit(src$destroy(), add = TRUE)
@@ -1322,7 +1325,10 @@ test_that("debounce/throttle work properly (without priming)", {
 
   # This observer will be what changes rv$a.
   src <- observe({
-    invalidateLater(100)
+    if (isolate(rv$a) == 0)
+      invalidateLater(120) # delay 20ms extra once to avoid threshold errors
+    else
+      invalidateLater(100)
     rv$a <- isolate(rv$a) + 1
   })
   on.exit(src$destroy(), add = TRUE)
--- a/tests/testthat/test-reactivity.r
+++ b/tests/testthat/test-reactivity.r
@@ -1244,7 +1244,7 @@ test_that("debounce/throttle work properly (with priming)", {
   # This observer will be what changes rv$a.
   src <- observe({
     if (isolate(rv$a) == 0)
-      invalidateLater(120) # delay 20ms extra once to avoid threshold errors
+      invalidateLater(120) # delay 20ms extra once to avoid race condition
     else
       invalidateLater(100)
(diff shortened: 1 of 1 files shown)


dvg-p4:
Hah, now that's slow enough for ubuntu, but too fast for MacOS

dvg-p4:
Even if 104 seconds hits a sweet spot and works, this is probably an indication that this particular test is not very robust and should be more fundamentally reworked.

dvg-p4:
oh sweet that worked!

dvg-p4:
Squashing all this nonsense into a single commit...

dvg-p4:
Still dunno how to get that license/cla thing.

tests/testthat/test-reactivity.r
```diff
@@ -1243,7 +1243,10 @@ test_that("debounce/throttle work properly (with priming)", {
 
   # This observer will be what changes rv$a.
   src <- observe({
-    invalidateLater(100)
+    if (isolate(rv$a) == 0)
+      invalidateLater(120) # delay 20ms extra once to avoid threshold errors
```
dvg-p4, line 1247:
```suggestion
      invalidateLater(120) # delay 20ms extra once to avoid race condition
```

tests/testthat/test-reactivity.r
```diff
@@ -1322,7 +1325,10 @@ test_that("debounce/throttle work properly (without priming)", {
...
```
dvg-p4, line 1329:
```suggestion
      invalidateLater(120) # delay 20ms extra once to avoid race condition
```

dvg-p4:
...dammit, it looks like there's more than 20ms of slop here, may need to just make these delays generally longer as well.

dvg-p4:
Superseded by #4134, _there's_ the more robust solution (simpler too!)

PR #4143, opened 2024-10-10
Set `typescript.tsdk` so VSCode's TS language features work properly

dvg-p4:
EDIT: The real solution to the problem I was having was simply to configure the TypeScript version used, via VSCode UI. Hit the braces on the bottom right:

image

Then "Select Version":

image

Then "Use Workspace Version".

------

This prevents spurious linting warnings, and makes type detection work properly with the complex constructions used in [some places](https://github.com/rstudio/shiny/blob/abf71389becd29f7999765da93e5aee152b271f7/srcts/src/utils/object.ts#L17-L27).

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -20,5 +20,6 @@
     "editor.defaultFormatter": "esbenp.prettier-vscode",
     "files.trimTrailingWhitespace": true,
     "files.insertFinalNewline": true,
-},
+  },
+  "typescript.tsdk": "./node_modules/typescript/lib",
 }


gadenbuie:
> This prevents spurious linting warnings, and makes type detection work properly

I think there might be something particular to your local setup that's involved here. I don't see any lint errors in VS Code currently.

dvg-p4:
Do you have `TypeScript > Validate` enabled or disabled?

dvg-p4:
Also, if you open up this line in VSCode:

https://github.com/rstudio/shiny/blob/abf71389becd29f7999765da93e5aee152b271f7/srcts/src/bindings/input/selectInput.ts#L204

and hover over `data`, does it show 

image

or....never mind, I can't seem to reproduce the old behavior anymore, even after switching back to main. But it used to say `data: never` before I started messing with the settings.

dvg-p4:
Messing with the settings appears to have kicked VS Code into the "use workspace version", but is not generally needed.

dvg-p4:
...actually it turns out, selecting the setting in the UI just automatically adds this line to to the project's settings.json file.

dvg-p4:
...but if I revert the change to settings.json and reload VSCode then it keeps using the right version anyways???

PR #4147, opened 2024-10-22
feat(clientdata): Add viewport and scrollable size values to clientdata

karangattu:
### What is this PR about?

Add viewport height & width along with scrollable area height and width to the clientdata.
Update those values in clientdata anytime the user resizes the browser or makes changes to the DOM structure by removing or adding elements.

### DEMO
viewport_and_scroll_size_clientdata

### Example App
```R
 library(shiny)
library(bslib)

ui <- page_fluid(
  h3("URL components"),
  verbatimTextOutput("urlText"),

  h3("Parsed query string"),
  verbatimTextOutput("queryText"),

  h3("Added Text"),
  uiOutput("dynamicUI")
)

  # Return the components of the URL in a string:
  output$urlText <- renderText({
    paste(sep = "",
          "protocol: ", session$clientData$url_protocol, "\n",
          "hostname: ", session$clientData$url_hostname, "\n",
          "pathname: ", session$clientData$url_pathname, "\n",
          "port: ",     session$clientData$url_port,     "\n",
          "search: ",   session$clientData$url_search,   "\n",
          "width: ",    session$clientData$window_width, "\n",
          "height: ",   session$clientData$window_height, "\n",
          "scroll_width: ",  session$clientData$scroll_width, "\n",
          "scroll_height: ", session$clientData$scroll_height, "\n"
    )
  })

  # Parse the GET query string
  output$queryText <- renderText({
    query <- parseQueryString(session$clientData$url_search)

    # Return a string with key-value pairs
    paste(names(query), query, sep = "=", collapse=", ")
  })

  # Reactive value to store added text
  addedLines <- reactiveVal(character(0))

  # Reactive value to control visibility
  showAddedText <- reactiveVal(TRUE)

  # Dynamic UI for added text section
  output$dynamicUI <- renderUI({
    if (showAddedText()) {
      tagList(
        actionButton("addText", "Add Text"),
        actionButton("deleteText", "Delete Text", class = "btn-danger"),
        verbatimTextOutput("addedText")
      )
    }
  })

  # Add new line of text when button is clicked
  observeEvent(input$addText, {
    newLine <- paste("Line", length(addedLines()) + 1, "added at", Sys.time())
    addedLines(c(addedLines(), newLine))
  })

  # Display added text
  output$addedText <- renderText({
    paste(addedLines(), collapse = "\n")
  })

  # Delete text and hide the section when delete button is clicked
  observeEvent(input$deleteText, {
    addedLines(character(0))
    showAddedText(FALSE)
  })
}

shinyApp(ui, server)
```

### Testing done
Verified that the values in clientdata are updated anytime the browser is resized, or elements are added or removed from the DOM. Tested it on Chrome, Firefox and Safari browsers on MacOS.


srcts/src/shiny/index.ts
```diff
@@ -606,6 +606,24 @@ class ShinyClass {
...
+        document.documentElement.scrollWidth
+      );
+      inputs.setInput(
+        ".clientdata_scroll_height",
+        document.documentElement.scrollHeight
+      );
+    }
+
+    sendImageSizeFns.setImageSend(inputBatchSender, doSendWindowSize);
+    sendImageSizeFns.regular(); // Initial send
```
cpsievert, line 623: I don't think we should be immediately invoking `.regular()` (or `doSendWindowSize()`) here, but instead, we should be adding to `initialValues` similar to how the other initial client data values are populated below.

srcts/src/shiny/index.ts
```diff
@@ -606,6 +606,24 @@ class ShinyClass {
...
+        ".clientdata_scroll_width",
+        document.documentElement.scrollWidth
+      );
+      inputs.setInput(
+        ".clientdata_scroll_height",
+        document.documentElement.scrollHeight
+      );
+    }
+
+    sendImageSizeFns.setImageSend(inputBatchSender, doSendWindowSize);
```
cpsievert, line 622: Unfortunately, I don't think `.setImageSend()` was designed to be used this way (i.e., it's only designed to have one callback, not multiple). 

I have a couple ideas on how we could proceed -- probably best if we talk through it in person

srcts/src/shiny/index.ts
```diff
@@ -618,6 +636,20 @@ class ShinyClass {
...
+    // Send whether the client is mobile
+    initialValues[".clientdata_is_mobile"] = /Mobi|Android/i.test(
+      navigator.userAgent
+    );
+
+    // Send whether the client is desktop
+    initialValues[".clientdata_is_desktop"] =
+      !/Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(
+        navigator.userAgent
+      );
```
cpsievert, line 651: After looking into this, I regret suggesting that we attempt this :)

The definition of a mobile device doesn't seem very well-defined, especially if basing that off of `userAgent`.

I would be more comfortable basing it off of the `window.innerWidth` (since that's how we determine it for collapsing layouts), but if you don't have an immediate need for it, I'd say we leave it out of this PR

karangattu: Agreed, removed from this PR

PR #4186, opened 2025-02-04
Reduce stack trace length

olivroy:
Basically, the stack trace isn't truncated well if a package uses cli or rlang to throw errors. This is common in many tidyverse packages for example.

This PR adds some calls to allow truncating the top of the stack trace a bit more.

Opening the PR from discussion, because it doesn't really work since I can't seem to get rid of the "<Anonymous>" call..


tests/testthat/_snaps/stacks.md
```diff
@@ -88,3 +88,100 @@
...
+---
+
+    Code
+      df
+    Output
+         num                call                 loc
+      1   69         <Anonymous>                    
+      2   68     signalCondition                    
+      3   67        signal_abort                    
+      4   66        rlang::abort
```
olivroy, line 100: The goal is to get rid of these, but my approach doesn't work currently. Let me know if you have an idea on how to do this.

PR #4199, opened 2025-03-10
Implements `$cancel()` method for ExtendedTask

### LINKED ISSUE #4093: ExtendedTasks: how to kill a task?
Thanks ExtendedTasks implementation, it has a promissing future!

As of shiny 1.8.1.1, the methods available for ExtendedTasks are:

         • ‘ExtendedTask$new()’

         • ‘ExtendedTask$invoke()’

         • ‘ExtendedTask$status()’

         • ‘ExtendedTask$result()’

It would be quite useful to have a `ExtendedTask$Kill()` method that would kill the respective process associated with the task. Such method would better handle cases of expensive tasks where user decides to change inputs half-way. Such actions would create multiple processes running expensive tasks, which is something we want to avoid.

Concept:

```r
ExtendedTask$kill() # kill all previous task
ExtendedTask$invoke() #start a new task based on new input
```

shikokuchuo:
Closes #4093.

Adds a `$cancel()` method to `ExtendedTask`, analogous to the [python implementation](https://github.com/posit-dev/py-shiny/blob/91e86e17b8053d45666ff6df152ca6cb4f2d3ead/shiny/reactive/_extended_task.py#L83).

This keeps it simple, and is straightforward as the `do.call()` in the code creates the task object anyway, and we just need to store it in a private field.

Previously I'd thought that the promise itself might have a `$cancel` method, but looking at javascript promises, this actually doesn't exist. It's also not natural to implement it that way, as `promises::promise()` takes functions as arguments, rather than the task object itself. We'd most likely need to rely on each `as.promise()` method to create this.

We could make cancellation extendable via a S3 generic for the task object for example, but as mirai is the only known implementation with cancellation capability thus far, we might defer this.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -108,6 +108,7 @@ Suggests:
     magrittr,
     yaml,
     future,
+    mirai,
     dygraphs,
     ragg,
     showtext,
--- a/R/extended-task.R
+++ b/R/extended-task.R
@@ -41,12 +41,13 @@
 #'   is, a function that quickly returns a promise) and allows even that very
 #'   session to immediately unblock and carry on with other user interactions.
 #'
-#' @examplesIf rlang::is_interactive() && rlang::is_installed("future")
-#'
+#' @examplesIf rlang::is_interactive() && rlang::is_installed("mirai")
 #' library(shiny)
 #' library(bslib)
-#' library(future)
-#' plan(multisession)
+#' library(mirai)
+#'
+#' daemons(1)
+#' onStop(function() daemons(0))
 #'
 #' ui <- page_fluid(
 #'   titlePanel("Extended Task Demo"),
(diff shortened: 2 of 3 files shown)


shikokuchuo:
Adding to the documentation example - you can see that the current cancellation behaviour (at mirai) is very explicit in causing the promise to be rejected. This was thought to be the 'safest' behaviour, in case it needs to be logged etc.

An alternative would be for cancellation to fail silently, as in the interrupt case.

shikokuchuo:
> This keeps it simple, and is straightforward as the `do.call()` in the code creates the task object anyway, and we just need to store it in a private field.

Need to think more on the implementation, as this task object could already be a promise e.g. if the user adds their own `$then()` logic within the `ExtendedTask` initializer. Perhaps there's no easy solution to this.

shikokuchuo:
I'm fairly certain that this implementation, if desired, would belong at promises rather than be a special case within ExtendedTask.

This would cause our definition of promises to deviate from javascript promises. I see that there are also other asks of the interface at promises, so I'll defer this to a point where we make an overall consideration on what the UI should be.

I'm closing this PR in any case.

PR #4209, opened 2025-04-25
feat!(submitButton): don't treat any HTML `type='submit'` button/input like a `submitButton()`

cpsievert:
This change is needed for https://github.com/rstudio/bslib/pull/1204, which proposes a more fully featured `input_submit_button()`, as well as a `input_submit_textarea()`. In both cases, it will be nice (if not necessary) to include `type="submit"` on the buttons. Unfortunately, that means that shiny.js will suspend _all_ inputs until submit, which is not what we want.

Technically this is a breaking change, but there is guidance in the NEWS for back-compat if someone happens to be relying on the current behavior.

Closes https://github.com/posit-dev/py-shiny/issues/1770


NEWS.md
```diff
@@ -1,6 +1,16 @@
 # shiny (development version)
 
-## New features and improvements
+## New features
+
+* `textInput()`, `textAreaInput()`, `numericInput()` and `passwordInput()` all gain an `updateOn` option. `updateOn = "change"` is the default and previous behavior, where the input value updates immediately whenever the value changes. With `updateOn = "blur"`, the input value will update only when the text input loses focus or when the user presses Enter (or Cmd/Ctrl + Enter for `textAreaInput()`). (#4183)
+
+## Changes
+
+* Shiny no longer suspends input changes when _any_ `<input type="submit">` or `<button type="submit">` is on the page. Instead, it now only suspends when a `submitButton()` is present. If you have reason for creating a submit button from custom HTML, add a CSS class of `shiny-submit-button` to the button.
```
cpsievert, line 9:
```suggestion
* Shiny no longer suspends input changes when _any_ `<input type="submit">` or `<button type="submit">` is on the page. Instead, it now only suspends when a `submitButton()` is present. If you have reason for creating a submit button from custom HTML, add a CSS class of `shiny-submit-button` to the button. (#4209)
```

srcts/src/shiny/index.ts
```diff
@@ -177,12 +177,12 @@ class ShinyClass {
...
 
-    if ($('input[type="submit"], button[type="submit"]').length > 0) {
+    if (document.querySelector(".shiny-submit-button")) {
       // If there is a submit button on the page, use defer decorator
       target = inputsDefer;
 
-      $('input[type="submit"], button[type="submit"]').each(function () {
-        $(this).click(function (event) {
+      document.querySelectorAll(".shiny-submit-button").forEach(function (x) {
+        x.addEventListener("click", function (event) {
```
schloerke, line 185: I'm up for the organic transfer away from jQuery

schloerke (approved):

PR #4210, opened 2025-04-25
feat(textAreaInput): Add an `autoresize` option

cpsievert:
Adds an `autoresize` option to `textAreaInput()`. When `TRUE`, the input will automatically resize the height to the contents of the input. 

Note this effectively backports this feature from Python (https://github.com/posit-dev/py-shiny/pull/463) to R. 

In this PR, I decided to bundle the autoresize logic with "core" Shiny assets, removing the need for another dependency. This is handy since I'd also like re-use this logic for another component in https://github.com/rstudio/bslib/pull/1204. 

**IMPORTANT**: if we do end up bundling the autoresize logic with shiny.js, py-shiny should get updated to gut the autoresize dependency.


NEWS.md
```diff
@@ -1,6 +1,14 @@
 # shiny (development version)
 
-## New features and improvements
+## New features
+
+* `textInput()`, `textAreaInput()`, `numericInput()` and `passwordInput()` all gain an `updateOn` option. `updateOn = "change"` is the default and previous behavior, where the input value updates immediately whenever the value changes. With `updateOn = "blur"`, the input value will update only when the text input loses focus or when the user presses Enter (or Cmd/Ctrl + Enter for `textAreaInput()`). (#4183)
+
+* `textAreaInput()` gains a `autoresize` option, which automatically resizes the text area to fit its content.
```
```suggestion
* `textAreaInput()` gains a `autoresize` option, which automatically resizes the text area to fit its content. (#4210)
```

srcts/src/bindings/input/textarea.ts
```diff
@@ -2,11 +2,96 @@ import $ from "jquery";
...
+  }
+
+  // document.readyState in ["interactive", "complete"];\
+  const textAreas = document.querySelectorAll(
+    "textarea.textarea-autoresize"
+  ) as NodeListOf<HTMLTextAreaElement>;
+  textAreas.forEach(updateHeight);
+}
+
+updateOnLoad();
```
wch, line 97: This polling feels kind of hacky to me. It looks like this code was added here, but I don't know the exact cases where this is an issue: https://github.com/posit-dev/py-shiny/commit/109556677b298c098516082b71e54d0444b10dbb#diff-88765ff358667f52c1857ada58c9cac78d8931753ae12672c3a5bb5d438eb097

I believe something like this would be equivalent, except it would be slightly faster because it doesn't poll:

```ts
function update {
  // document.readyState in ["interactive", "complete"];\
  const textAreas = document.querySelectorAll(
    "textarea.textarea-autoresize"
  ) as NodeListOf<HTMLTextAreaElement>;
  textAreas.forEach(updateHeight);
}

if (document.readyState === 'loading') {
  document.addEventListener("DOMContentLoaded", function () {
    update();
    // If we're being really fastidious, we could also unregister this callback here
  });
} else {
  // If we get here, the readyState is "interactive" or "complete"
  update()
}
```

Another, easier way is to use `Shiny.initializedPromise`, although this would be later than the previous version, because this promise resolves only after all the content has been loaded AND Shiny has initialized.

```ts
Shiny.initializedPromise.then(() => {
  const textAreas = document.querySelectorAll(
    "textarea.textarea-autoresize"
  ) as NodeListOf<HTMLTextAreaElement>;
  textAreas.forEach(updateHeight);
});
```

cpsievert: Yea, good point. Now that I look a little closer at the code, I think it could incorporate some/all of it into the binding itself (i.e., update height on `initialize`). Will look into that next week

cpsievert: Addressed in [70aae7c](https://github.com/rstudio/shiny/pull/4210/commits/70aae7c455a19e15e6485427e28cde32287fa910)

schloerke: The goal was to get the height set as soon as possible.

I believe your solution should work. Thank you!

cpsievert:
@wch ok to merge?

wch (approved):

Issue #2196, opened 2018-09-26
Parallel + Progress Bars


ajaypillarisetti:
First: thanks, as always, for the excellent work and for folks' time in advance. I've got a few apps that have used progress bars for fairly large computations (MC simulations of emissions related to simple combustion devices). These progress bars worked until recently using parallel ldply. Below is a basic example adapted from the Progress documentation that works:

```
library(shiny)
library(parallel)

## Only run examples in interactive R sessions
if (interactive()) {

ui <- fluidPage(
  plotOutput("plot")
)

server <- function(input, output, session) {
  output$plot <- renderPlot({
    progress <- Progress$new(session)
    on.exit(progress$close())

    progress$set(message = 'Calculation in progress.')

    mclapply(1:5000, function(x){
        progress$inc(1/5000, detail = paste('Computating...'))
    }, mc.cores=4)

    plot(cars)
  })
}

shinyApp(ui, server)
}
```

Anything over 1 core, however, fails:

```
library(shiny)
library(parallel)

## Only run examples in interactive R sessions
if (interactive()) {

ui <- fluidPage(
  plotOutput("plot")
)

server <- function(input, output, session) {
  output$plot <- renderPlot({
    progress <- Progress$new(session)
    on.exit(progress$close())

    progress$set(message = 'Calculation in progress.')

    mclapply(1:5000, function(x){
        progress$inc(1/5000, detail = paste('Computating...'))
    }, mc.cores=4)

    plot(cars)
  })
}

shinyApp(ui, server)
}
```

This is the same behavior I'm seeing in my app. Any ideas? I can guess that drawing progress bars during parallel processes is a bit fraught. Thanks!

vj-09:
same issue here.

Bogaso:
Facing the same problem. Appreciate if this can be addressed earliest possible

jcheng5:
I'm surprised anything like this would ever work? The default progress bar in Shiny is implemented in memory, and mclapply launches tasks in sub-processes. If anyone can reproduce something like this working in a previous version of Shiny I'd be very interested to see it.

You _might_ have better luck with the ipc package: https://github.com/fellstat/ipc I'm uncertain if it could work with mclapply but it'd be a more promising avenue to pursue than Shiny's own progress bars.

schloerke:
Rule to follow: `session` (or any reactive value) will not work as expected when executed in a different R session. 

[...]

PR #4214, opened 2025-05-02
Fix performance regression related to limiting deep call stack growth

### LINKED ISSUE #4213: Performance regression launching promises in Shiny 1.10
### Symptom

In Shiny 1.10, launching many promises at once e.g. by `mirai::mirai()` or `mirai::mirai_map()` is now noticeably slower than in previous versions.

Using this [mirai Shiny vignette example](https://mirai.r-lib.org/articles/shiny.html#advanced-promises-coin-flips), this blocked 1-2s before the counter started to update in v1.9.1, whereas in v1.10 it's taking up to 14s.

### Cause

The cause has been isolated to this PR #4156 related to limiting deep call stack growth.

Profiling using profvis shows recursive calls to the hashing for stack trace deduplication taking up the most of the time. In fact most of the time is spent in this [getCallNames()](https://github.com/rstudio/shiny/blob/ca41c0831befcf8feea17980a91983ce21d1b8a2/R/conditions.R#L64) function, with the `format()` call within that being a particularly egregious offender.

### Solution

@cpsievert looked over the evidence I provided and agrees on the cause. He also suggested a fix that gets us most of the way back (now blocking 3-4s).

As hashing uses `rlang::hash()`, this can take any arbitrary R object not just a string. Hence we can use another version of `getCallNames()` that completely avoids the expensive formatting calls, as long as they keep the structure of the calls and produce unique hashes.

@jcheng5 this is the issue I mentioned in passing a while ago.

shikokuchuo:
Closes #4213.

Credit to @cpsievert for the suggestion.

This PR introduces a version of `getCallNames()` just for hashing that avoids the expensive formatting calls needed for nice output. We only need this function to create unique output to be hashed.

Notably the deep stack trace [tests](https://github.com/rstudio/shiny/blob/main/tests/testthat/test-stacks-deep.R) continue to pass.

I shouldn't think this would cause any issues, but would just like to draw your attention to this @jcheng5 in case you can think of a better solution.

### DIFF
--- a/R/conditions.R
+++ b/R/conditions.R
@@ -75,6 +75,16 @@ getCallNames <- function(calls) {
   })
 }
 
+# Do less to speed things up, just need to keep call stack unique for hashing
+getCallNamesForHash <- function(calls) {
+  lapply(calls, function(call) {
+    name <- call[[1L]]
+    if (is.function(name)) return("<Anonymous>")
+    if (typeof(name) == "promise") return("<Promise>")
+    name
+  })
+}
+
 getLocs <- function(calls) {
   vapply(calls, function(call) {
     srcref <- attr(call, "srcref", exact = TRUE)
@@ -144,7 +154,7 @@ getCallStackDigest <- function(callStack, warn = FALSE) {
     )
   }
 
-  rlang::hash(getCallNames(callStack))
+  rlang::hash(getCallNamesForHash(callStack))
 }
 
 saveCallStackDigest <- function(callStack) {


R/conditions.R
```diff
@@ -75,6 +75,16 @@ getCallNames <- function(calls) {
   })
 }
 
+# Do less to speed things up, just need to keep call stack unique for hashing
```
cpsievert, line 78: I'd also move the implementation of this function right next to `getCallNames()` so that it's a bit more obvious that they are closely related

```suggestion
# A stripped down version of getCallNames() that intentionally avoids deparsing expressions.
# Instead, it leaves expressions to be directly `rlang::hash()` (for de-duplication), which
# is much faster than deparsing then hashing.
```

shikokuchuo: Thanks!

shikokuchuo: This new function is already right after `getCallNames()`, unless I'm missing what you mean.

shikokuchuo:
FYI the current Windows check failures look unrelated.

schloerke:
From https://github.com/rstudio/shiny/issues/2196 : 

<details>

<summary> Reprex app: </summary>

```r
library(shiny)
library(future)
# Use `promises::future_promise()` to allow other shiny users to be able to compute.
# Recommended to always use `promises::future_promise()` instead of `future::future()`
library(promises)
# Pick a plan that makes sense for your use case.
# For this example, using `multisession` and max workers = 3
future::plan("multisession", workers = 3)

ui <- fluidPage(
  sliderInput("slider", "Number of items", min = 10, max = 500, value = 250),
  plotOutput("plot")
)

  output$plot <- renderPlot({
    req(input$slider)

    progress <- Progress$new(session)
    progress$set(message = 'Calculation in progress.')

    # Retrieve `input$slider` outside of `future`
    n <- input$slider

    proms <- lapply(seq_len(n), function(i) {
      fp <- promises::future_promise({
        message("Working...", i)
        # Do heavy computations in a `future` worker
        data.frame(i = i, pid = Sys.getpid())
      })

      fp$finally(function() {
        message("Computing...", i)
        # # Update progress bar in the main R session
        progress$set(value = i / n, message = paste('Computing...', i))
      })

      fp
    })
    # Wait for all promises to finish. Return single promise.
    prom_all <- promises::promise_all(.list = proms)

    prom_all %...>%
      {
        # Extract result and plot in main R session
        plot(do.call(rbind, .))
      }

    promises::finally(prom_all, function() {
      # This will be called when the promise is resolved or rejected
      progress$close()
    })

    prom_all
  })
}

shinyApp(ui, server)
```

</details>

Time to set up promises in app:

* With PR: (< 5s) Screenshot 2025-05-12 at 10 03 32 AM
* On `main`: (20s) Screenshot 2025-05-12 at 10 03 28 AM

---------

(Doesn't test coverage, but speed is reverted back to expected. Thank you!)

jcheng5:
Impressive work, @shikokuchuo! LGTM!

jcheng5 (approved):

schloerke:
@jcheng5 I'm happy to merge if you approve #4218 changes

shikokuchuo:
Thanks all!

PR #4222, opened 2025-05-28
family->given for R Core authorship

MichaelChirico:
This is the more typical and correct way to cite this contributor.

c.f.

```r
a = tools::CRAN_authors_db()
sum(grepl("(?i)r core team", a$family))
# [1] 8
sum(grepl("(?i)r core team", a$given))
# [1] 61
```

From `?person`:

> For persons which are not natural persons (e.g., institutions, companies, etc.) it is appropriate to use `given` (but not `family`) for the name, e.g., `person("R Core Team", role = "aut")`.


cpsievert:
Thanks!

PR #4245, opened 2025-06-26
v1.11.1 release candidate

cpsievert:
Patch fix release to address #4240 and #4239


averissimo:
#4240 is breaking the `{teal}` framework. Is there any ETA for this patch release?

PR #4248, opened 2025-06-27
Revert the addition of spacing between `icon` and `label` in `actionButton()`

cpsievert:
After some discussion in #4247, we've decided to revert the recent addition of spacing between the icon and label in `actionButton()` (first introduced in #4242 and later revised in #4247). The reasoning being that the current markup design isn't ideal for the long-term -- we're going with this separator approach (rather than giving the icon and label their own containers) to maximize back-compat (and help us get a patch fix out the door). Also, if and when we decide to add spacing (via containers), we'll want to deliberate whether we do the same for `downloadButton()` and `modalButton()` (for consistency).


gadenbuie (approved):

vnijs:
Just found this PR discussion after adding a comment to the (closed) issue below. I noticed that the spacing between icon and text for action buttons has been reduced. The icon and text are now very close together and are closer than in any previous versions of shiny, at least that I can recall. Is there an easy way to add the spacing back in?

https://github.com/rstudio/shiny/issues/4239#issuecomment-3026216007

PR #4275, opened 2025-08-21
fix: add CSS workaround to hide all but the first selectize remove button

### LINKED ISSUE #4274: Updating selectize.js options when `remove_button` plugin is present leads to multiple remove buttons
This is more of an issue for py-shiny (since there we default to adding the `"remove_button"` plugin), but it's also a general issue that can be replicated with r-shiny. Note how this leads to two remove buttons on a single item:

```python
library(shiny)
library(bslib)

ui <- page_fluid(
  selectizeInput(
    "x",
    NULL,
    choices = NULL,
    multiple = TRUE,
    options = list(plugins = list("remove_button"))
  )
)

  observe({
    updateSelectizeInput(
      session,
      "x",
      choices = paste("Foo", 0:2),
      server = TRUE,
      options = list(placeholder = "Search")
    )
  })
}

```

Image

cpsievert:
Closes #4274


NEWS.md
```diff
@@ -7,6 +7,7 @@
 ## Bug fixes
 
 * `updateActionButton()`/`updateActionLink()` now correctly renders HTML content passed to the `label` argument. (#4249)
+* Fixed an issue where `updateSelectizeInput(options = list(plugins="remove_button"))` could lead to multiple remove buttons. (#4275)
```
elnelson575, line 11: Maybe add a line above this for consistency's sake?

cpsievert, line 11:
```suggestion

* Fixed an issue where `updateSelectizeInput(options = list(plugins="remove_button"))` could lead to multiple remove buttons. (#4275)
```

PR #4276, opened 2025-08-21
feat: add `shinyRemoveButton` selectize.js option (for py-shiny)

cpsievert:
This PR is to support https://github.com/posit-dev/py-shiny/pull/2064. See there for more details.

Note that this changes are currently irrelevant (i.e., a no-op) for r-shiny since it doesn't currently make use of this option, but it could in the future?


srcts/src/bindings/input/selectInput.ts
```diff
@@ -305,6 +314,44 @@ class SelectInputBinding extends InputBinding {
...
+    const optionPlugins = options.plugins || [];
+
+    plugins.forEach((plugin) => {
+      if (!optionPlugins.includes(plugin)) {
+        optionPlugins.push(plugin);
+      }
+    });
+
+    options.plugins = optionPlugins;
+    delete options.shinyRemoveButton;
```
cpsievert: Maybe we copy instead of mutate?

elnelson575: Yeah, could be nice to leave the options as a record

PR #4269, opened 2025-08-04
feat(otel): Add `bindOtel()`

schloerke:
Related: #4258

Fixes #4265

PR #4279, opened 2025-08-25
feat: `reactive()` and `bindCache()` default `label=` are now namespaced

schloerke:
For example, if the reactive `r <- reactive({force(1 + 1)})` is created within the module, `mymod`. Then the default name will be `"mymod-r"`.

If the reactive is created at the top level domain, then it will remain as `"r"`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -12,7 +12,9 @@
 
 ## Changes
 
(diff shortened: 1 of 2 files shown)


schloerke:
Closing as taking a different approach in #4269 . Something closer to `[MOD_NAME] reactive: NAME`

PR #4284, opened 2025-09-03
Update package.json for publishing

wch:
This PR updates package.json for publishing to npmjs.com.


package.json
```diff
@@ -74,6 +71,8 @@
   },
   "scripts": {
     "prepare": "node --eval \"1+1; // help yarn users not install the whole repo; https://github.com/yarnpkg/yarn/issues/2822#issuecomment-847360267\"",
+    "prepack": "cp README.md README-orig.md && cp npm-README.md README.md",
+    "postpack": "test -f README-orig.md && cp README-orig.md README.md && rm README-orig.md || true",
```
schloerke: Is this the logic you're looking for?

```suggestion
    "postpack": "test -f README-orig.md && cp README-orig.md README.md && (rm README-orig.md || true)",
```

schloerke (approved):

package.json
wch: Actually, I think I can take the `true` out.

wch:
I've updated it so that:

- It publishes to `@posit/shiny`
- It only publishes types

The current build system puts the .d.ts files into `srcts/types/`, and the built JS files into `inst/www/shared`. There is also a lot of other stuff in the latter directory.

In the future, I think we'll want to publish the JS and types together in a single package. To do that, I think we should:

- Separate Shiny's JS code from the R code, into a separate repository (or a monorepo with R, Python, and JS subdirs).
- Put all the built JS and .d.ts files into a directory like `dist/`
- Have the R and Python packages copy the built JS files, or bundle these JS files with their own JS code.

package.json
```diff
@@ -69,7 +69,9 @@
     "build_types": "tsc -p tsconfig.json",
     "coverage_detailed": "npx --yes type-check --detail",
     "coverage": "type-coverage -p tsconfig.json --at-least 90",
-    "circular": "npx --yes dpdm --transform ./srcts/src/index.ts"
+    "circular": "npx --yes dpdm --transform ./srcts/src/index.ts",
+    "prepack": "cp README.md README-orig.md && cp npm-README.md README.md",
```
wch, line 73: Note: It looks like `yarn pack` does *not* ignore README-orig.md (so it gets included in the package), but `npm pack` *does* ignore the file, which is the behavior that we want.

From what I can find, it's going to require awkward stuff to make `yarn pack` ignore that file, like moving the temporary file to /tmp. Since we're planning on switching to `npm` anyway, I suggest we just leave this as it is. We can publish it with `npm publish` just fine, even if we're using yarn for other steps.

PR #4314, opened 2025-11-07
feat(otel): Add metrics for up/down busy count

schloerke:
Related: https://github.com/thomasp85/reqres/pull/16/files#diff-0cc4a76032b57fcc125d41dfa3fb0f0c39976bb00a1d84bb56a0b77c331ce2d1R65-R76


schloerke:
While a fun idea, it is typically batched over a few seconds... which the metric will typically sum to 0.
Also, I could not get langfuse, logfire, jaegar to display any metrics. ONLY `otel-tui` in the terminal 😒 

Therefore, I'm closing the PR.

PR #4338, opened 2025-12-12
Catch RenderXXX functions in UI

### LINKED ISSUE #4189: Catch render functions included in UI
I sometimes find myself doing this where I mistakenly use a `renderX()` in the UI:

```
library(shiny)

ui <- fluidPage(
  renderTable("table")
)

  output$table <- renderTable("table")
}

shinyApp(ui, server)
```

Surprisingly, this doesn't produce any errors, unlike if you mix it up the other way round and put `output$table <- tableOutput("table")` in the server. Is there anything that could be done to catch it and issue a warning or an error? If not normally, then how about when `devmode()` is active?

simon-smart88:
Finally adding a suggestion to fix #4189 after doing this for the umpteenth time.

It would be nice if we could specify the file or module containing the problematic render function, but I am not sure how to do it.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # shiny (development version)
 
+## Bug fixes and minor improvements
+
+* Using `renderXXX` functions in the UI will now return an error (#4189).
+
 # shiny 1.12.1
 
 ## New features


R/shinywrappers.R
```diff
@@ -328,6 +328,12 @@ useRenderFunction <- function(renderFunc, inline = FALSE) {
 #' @export
 #' @method as.tags shiny.render.function
 as.tags.shiny.render.function <- function(x, ..., inline = FALSE) {
+  # return an error when render functions are used in UI
+  if (is.null(getDefaultReactiveDomain())) {
+    stop("There is a renderXXX function in the UI, these should only be used in the server function.",
+         "Use the corresponding XXXOutput function in the UI instead.")
+  }
```
cpsievert, line 335: Shoot, sorry. I gave bad advice in https://github.com/rstudio/shiny/issues/4189. We can't throw an error in this case -- if we do, we'll break `renderXXX()` usage inside of R Markdown. Unfortunately, I'm not sure there's sensible approach to throwing this error 😞

PR #4345, opened 2026-01-12
fix: Provide context around SHINY_PORT triggering Shiny Server warning

### LINKED ISSUE #4182: feat: override port via environment variable
My understanding is that the environment variable `SHINY_PORT` can be used to override the port used by a shiny app. However this doesn't seem to be the case and, instead, leads to a spurious warning when used. 

See the following repro. It results in the following warning message: 

```
Warning in runApp(x) :
  Shiny Server v0.3.4 or later is required; please upgrade!

Listening on http://127.0.0.1:9001
```

```r
library(shiny)

# Define UI for application that draws a histogram
ui <- fluidPage(

  # Application title
  titlePanel("Old Faithful"),

  sidebarLayout(
    sidebarPanel(
      sliderInput("bins",
        "Number of bins:",
        min = 1,
        max = 50,
        value = 30
      )
    ),

    # Show a plot of the generated distribution
    mainPanel(
      plotOutput("distPlot")
    )
  )
)

    x <- faithful[, 2]
    bins <- seq(min(x), max(x), length.out = input$bins + 1)

    hist(x,
      breaks = bins, col = "darkgray", border = "white",
      xlab = "Waiting time to next eruption (in mins)",
      main = "Histogram of waiting times"
    )
  })
}

Sys.setenv("SHINY_PORT" = "3000")
# Run the application
shinyApp(ui = ui, server = server, options = list(port = 9001))
```

gadenbuie:
Fixes #4182

When the `SHINY_PORT` environment variable is set and apps are run locally or not in a hosted environment, Shiny emits a warning about using an incompatible version of Shiny Server.

This PR adds additional context to the warning message to recommend either using another environment variable or making it clear that you can ignore the message.

### DIFF
--- a/R/runapp.R
+++ b/R/runapp.R
@@ -221,7 +221,7 @@ runApp <- function(
           "Shiny Server v%s or later is required; please upgrade.",
           .shinyServerMinVersion
         ),
-        "i" = "If you are not using Shiny Server you are likely seeing this message because the `SHINY_PORT` environment variable is set in your environment.",
+        "i" = "If you are not using Shiny Server, you are likely seeing this message because the `SHINY_PORT` environment variable is set in your environment.",
         "i" = "Avoid using `SHINY_PORT` to prevent this warning."
       ))
     }


R/runapp.R
```diff
@@ -216,8 +216,14 @@ runApp <- function(
...
     ver <- Sys.getenv('SHINY_SERVER_VERSION')
     if (utils::compareVersion(ver, .shinyServerMinVersion) < 0) {
-      warning('Shiny Server v', .shinyServerMinVersion,
-              ' or later is required; please upgrade!')
+      rlang::warn(c(
+        sprintf(
+          "Shiny Server v%s or later is required; please upgrade.",
+          .shinyServerMinVersion
+        ),
+        "i" = "If you are not using Shiny Server you are likely seeing this message because the `SHINY_PORT` environment variable is set in your environment.",
```
schloerke:
```suggestion
        "i" = "If you are not using Shiny Server, you are likely seeing this message because the `SHINY_PORT` environment variable is set in your environment.",
```

PR #4349, opened 2026-02-04
Add `startApp()` for non-blocking app execution

shikokuchuo:
## Motivation

Shiny, being an httpuv server, inherently does not need to block — it uses the `later` event loop, which is designed for cooperative concurrency. Adding a non-blocking mode:
- Allows AI coding agents to programmatically start, test, and stop Shiny apps without blocking the R console and run multiple apps sequentially in the same session
- Opens up novel uses of Shiny as a UI for objects that live on in the user's interactive R session

## Changes

Adds `startApp()`, a non-blocking counterpart to `runApp()`:

```r
# Start app in the background — returns immediately
handle <- startApp("myapp")

# Check status and URL
handle$status() # "running", "success" or "error"
handle$url()

# Stop when done
handle$stop()

# Access return value from stopApp()
handle$result() # throws if still running, re-throws errors
```

## Implementation

- `ShinyAppHandle` (R6 class): Returned by `startApp()`. Provides methods for lifecycle management (`stop()`, `status()`, `url()`) and accessing the app's return value (`result()`).
- `.setupShinyApp()`: Shared initialization extracted from `runApp()`, used by both `runApp()` and `startApp()`.
- `serviceNonBlocking()`: Runs the httpuv event loop via `later` callbacks instead of a blocking while loop.
- `.createCleanup()`: Consolidated cleanup logic shared between blocking and non-blocking modes.

## Safeguards

- Only one app can run at a time (same as blocking mode)
- When a new app starts, any previously running non-blocking app is automatically stopped
- Finalizer ensures cleanup if handle is garbage collected
- Cleanup is idempotent (safe to call multiple times)


shikokuchuo (commented):
From a conversations with @hadley, we could tweak the UI further so that:
1. Starting a new app automatically stops the old one
2. Have it automatically run non-blocking if used by an agent (via env vars)

R/runapp.R
```diff
@@ -45,6 +45,18 @@
...
+#'   Defaults to `TRUE`, except when an LLM agent is detected, where it
+#'   defaults to `FALSE`. Can be set globally via `options(shiny.blocking)`.
+#'   Non-blocking mode requires the `later` event loop to be running.
+#'
+#' @return If `blocking = TRUE`, returns the value passed to [stopApp()], or
+#'   throws an error if the app was stopped with an error. If `blocking = FALSE`,
+#'   returns a `ShinyAppHandle` object with methods `stop()`, `status()`,
+#'   `url()`, and `result()`. The `status()` method returns `"running"`,
+#'   `"success"`, or `"error"`. The `result()` method throws an error if called
+#'   while running, or re-throws the error if the app stopped with an error.
```
schloerke: I am worried about the goals of the future desired behavior.

I believe we need a different function for non-blocking runApp. Returning vastly different objects given a new parameter isn't right.

1. Non-blocking for LLMs only
	* On CRAN, LLMS already know to run in the background. If this was switched to be non-blocking, the app won't run as expected. To mitigate this, we'd need to tell the LLM to read the docs... which it could read about the new method instead.
2. If the future goal is to make non-blocking the default behavior, many programs (Posit Connect included) use `runApp()` as is. To add a parameter to make it non-blocking will cause pain for all existing Connect users who do not have the bleeding edge version. This will take _years_ to recover from as updating Connect versions is a non-trivial process for admins.

R/runapp.R
```diff
@@ -388,33 +405,67 @@ runApp <- function(
...
+
+  } else {
+    # NON-BLOCKING MODE: return handle immediately, app runs via later callbacks
+    # Disable on.exit handlers; cleanup is handled by handle$stop()
+    cleanupOnExit <- FALSE
+    cleanup <- createCleanup(server, appParts, appUrl, ops)
+    handle <- ShinyAppHandle$new(appUrl, cleanup)
+    .globals$runningHandle <- handle
+
+    serviceAsync(handle)
```
schloerke: Cosmetic:

I'd like to have this be something towards non-blocking, rather than an `Async` suffix... as that typically deals with `{promises}`.

R/runapp.R
```diff
@@ -495,7 +546,8 @@ runExample <- function(
   launch.browser = getOption("shiny.launch.browser", interactive()),
   host = getOption("shiny.host", "127.0.0.1"),
   display.mode = c("auto", "normal", "showcase"),
-  package = "shiny"
+  package = "shiny",
+  blocking = getOption("shiny.blocking", !is_llm())
```
schloerke: For these extension runApp methods, I would not add the `blocking` parameter. 

If needed, provide an example on how the user can achieve it for runExample only using system.file.

R/runapp.R
```diff
@@ -100,13 +113,20 @@ runApp <- function(
...
+  # Set to FALSE in non-blocking mode, where cleanup is handled by handle$stop().
+  cleanupOnExit <- TRUE
+
+  on.exit(if (cleanupOnExit) handlerManager$clear(), add = TRUE)
 
   if (isRunning()) {
-    stop("Can't call `runApp()` from within `runApp()`. If your ",
-         "application code contains `runApp()`, please remove it.")
+    if (!is.null(.globals$runningHandle)) {
+      .globals$runningHandle$stop()
```
schloerke: If something is running, add message declaring the app is being stopped

R/runapp.R
schloerke: If the goal is to expose state to AI, maybe `startApp()` as it follows `{httpuv}`'s `startServer()` ?

It would need doc updates in `stopApp()` in that they are unrelated.

schloerke (changes requested):

R/runapp.R
shikokuchuo: Changed to `serviceNonBlocking()`.

shikokuchuo:
@cpsievert just to update that I tested the Rstudio `Run App` button with `options(shiny.blocking = FALSE)`, and the different options run in the same way as when blocking - nothing surprising in that regard.

shikokuchuo:
Let's move this forward on the following basis (as discussed in today's team meeting):

- Add `blocking = getOption("shiny.blocking", TRUE)` to `runApp()`
- Make this NON-default for LLMs or otherwise

This has the following advantages:

- Strictly opt-in - no surprising behaviour
- We can still teach an LLM how to use non-blocking mode correctly via docs or a skill
- All current infra (incl. RStudio button) works in BOTH modes
- Single change location (no need to modify `shinyApp` print method etc.)

Noting that @schloerke continues to prefer 2 separate functions.

@cpsievert for you to approve this PR. Thanks!

R/app-handle.R
```diff
@@ -0,0 +1,74 @@
...
+    cleanupFn = NULL,
+    # Whether this handle has captured the result. Distinct from .globals$stopped
+    # which tracks whether a stop was requested (set by stopApp() or stop()).
+    stopped = FALSE,
+    resultValue = NULL,
+    resultError = NULL,
+
+    captureResult = function() {
+      if (private$stopped) return()
+      private$stopped <- TRUE
```
cpsievert: nitpick: I find it slightly surprising that the stop method doesn't manage the stopped flag.

R/runapp.R
```diff
@@ -388,33 +404,72 @@ runApp <- function(
...
+      })
+    )
+
+    if (isTRUE(.globals$reterror)) {
+      stop(.globals$retval)
+    }
+    else if (.globals$retval$visible)
+      .globals$retval$value
+    else
+      invisible(.globals$retval$value)
```
cpsievert:
```suggestion
    if (isTRUE(.globals$reterror)) {
      stop(.globals$retval)
    } else if (.globals$retval$visible) {
      .globals$retval$value
    } else {
      invisible(.globals$retval$value)
    }
```

R/runapp.R
```diff
@@ -91,22 +102,30 @@ runApp <- function(
...
   if (isRunning()) {
-    stop("Can't call `runApp()` from within `runApp()`. If your ",
-         "application code contains `runApp()`, please remove it.")
+    if (!is.null(.globals$runningHandle)) {
+      message("Stopping running Shiny app.")
+      .globals$runningHandle$stop()
+    } else {
+      stop("Can't start a new app while another is running. ",
+           "If your application code contains `runApp()`, remove it. ",
+           "Otherwise, stop the current app first with handle$stop() or stopApp().")
```
cpsievert: Is `handle` actually relevant in this case? Maybe we should just be mentioning `stopApp()`?

R/app-handle.R
shikokuchuo: Agree, fixed in [3a0e862](https://github.com/rstudio/shiny/pull/4349/commits/3a0e8627a4009360b016ff391cc76f68f6e8ee91).

R/runapp.R
shikokuchuo: You're right, I've amended in [bd25096](https://github.com/rstudio/shiny/pull/4349/commits/bd250962e4dd6fd16ba214ec7e326b14e17ac53a).

R/runapp.R
shikokuchuo: Non-blocking is now separated out into `startApp()` in [02a5e0b](https://github.com/rstudio/shiny/pull/4349/commits/02a5e0b40f356c76ec75402cc5154af90e24ffeb). The doc updates you requested in [72636ef](https://github.com/rstudio/shiny/pull/4349/commits/72636ef4a07e5c1f4f32d7b451828f88794d81e0).

shikokuchuo:
Following discussion at Asilomar, non-blocking behaviour is now separated out into a `startApp()` function so it doesn't interfere at all with existing `runApp()` behaviour.

cpsievert:
@shikokuchuo Could you please update the PR description?

shikokuchuo:
> @shikokuchuo Could you please update the PR description?

Ah, of course - all done now!

cpsievert (commented):
A few issues spotted during review — details in inline comments.

R/runapp.R
```diff
@@ -202,6 +176,191 @@ runApp <- function(
...
+  appDir = getwd(),
+  port = getOption("shiny.port"),
+  launch.browser = getOption("shiny.launch.browser", interactive()),
+  host = getOption("shiny.host", "127.0.0.1"),
+  workerId = "",
+  quiet = FALSE,
+  display.mode = c("auto", "normal", "showcase"),
+  test.mode = getOption("shiny.testmode", FALSE)
+) {
+  promises::local_otel_promise_domain()
```
cpsievert: `local_otel_promise_domain()` registers an `on.exit` handler on its caller's frame (via `defer()` → `do.call(on.exit, ..., envir = envir)`). In `runApp()` this works because the frame stays alive for the duration of the blocking event loop. In `startApp()`, the frame exits immediately after the handle is returned, so the OTEL domain is torn down before the app ever serves a request.

The non-blocking app will run without OTEL instrumentation.

Suggested fix: either (a) remove this call from `startApp()` and add a comment explaining why OTEL isn't supported in non-blocking mode yet, or (b) manage the domain's lifetime through the handle — activate it in `startApp()` without `local_`, store the old domain, and restore it in `.createCleanup()`.

R/runapp.R
```diff
@@ -202,6 +176,191 @@ runApp <- function(
...
+  # See matching guard in runApp().
+  setupComplete <- FALSE
+  on.exit(if (!setupComplete) {
+    .globals$onStopCallbacks$invoke()
+    .globals$onStopCallbacks <- Callbacks$new()
+  }, add = TRUE)
+
+  require(shiny)
+
+  appParts <- as.shiny.appobj(appDir)
```
cpsievert: In `runApp()`, `options(warn = max(1, ...), pool.scheduler = ...)` is set at line 116, *before* `as.shiny.appobj()` at line 137. Here in `startApp()`, `as.shiny.appobj()` runs first, and options aren't set until `.setupShinyApp()` (lines 320–326).

This means during app object conversion in `startApp()`, warnings below level 1 won't be printed immediately and `pool.scheduler` isn't active. Minor behavioral inconsistency with `runApp()` — could be aligned by moving the `options()` call above this line (and passing `ops` to `.setupShinyApp()` the same way `runApp()` does).

R/runapp.R
```diff
@@ -202,6 +176,191 @@ runApp <- function(
...
+
+  # Resolve arguments that may also be specified via shinyApp(options = ...)
+  # See runApp() for the full precedence matrix.
+  appOps <- appParts$options
+  findVal <- function(arg, default) {
```
cpsievert: Nit: the `findVal` helper and this 15-line option resolution block are duplicated verbatim from `runApp()` (lines 162–177). Since both callers immediately pass the resolved values to `.setupShinyApp()`, this logic could be folded into `.setupShinyApp()` itself (e.g., accept the raw formals + `appParts$options` and do the merge there). Not blocking, but worth considering to prevent these two copies from drifting apart over time.

tests/testthat/test-non-blocking.R
```diff
@@ -0,0 +1,275 @@
...
+  on.exit(handle$stop(), add = TRUE)
+
+  expect_true(isRunning())
+
+  handle$stop()
+  expect_false(isRunning())
+})
+
+test_that("startup failure clears app state (regression test)", {
+  # If startup fails after initCurrentAppState() but before earlyCleanup <- FALSE,
```
cpsievert: Stale variable name in the comment: `earlyCleanup` was renamed to `cleanupOnExit` in the implementation. Should read:

```r
# If startup fails after initCurrentAppState() but before cleanupOnExit <- FALSE,
```

R/runapp.R
shikokuchuo: Moved up in [88b4fac](https://github.com/rstudio/shiny/pull/4349/commits/88b4facb8ffd4c26001a707cae5a34af41c38ac3).

R/runapp.R
shikokuchuo: Fixed in [88b4fac](https://github.com/rstudio/shiny/pull/4349/commits/88b4facb8ffd4c26001a707cae5a34af41c38ac3) - we move this logic down to the helper but evaluate missingness in the caller environment, which is passed down - as missingness is not transitive when there are default arguments.

R/runapp.R
shikokuchuo: Thanks — fixed in c4c91f62a. Went with scoped wrapping rather than (a) or (b):        
                                                                                             
  - `startApp()` wraps the synchronous setup phase (covers `onStart`).                           
  - `serviceNonBlocking()` wraps each `serviceApp()` iteration.                                  
                                                                                             
Span wrapping is attached at promise-registration time, so promises registered inside any of those stay instrumented when they eventually fulfil. 

Avoided (b) because leaving the domain on `globals$domain` works in `runApp()` (console is blocked) but in `startApp()` would leak Shiny's OTEL context into unrelated user promises created at the console between ticks. Scoped wrapping keeps OTEL active only when Shiny is processing its own work.

R/runapp.R
```diff
@@ -388,39 +524,52 @@ runApp <- function(
...
-  }, add = TRUE)
+  on.exit(if (cleanupOnExit) callAppHook("onAppStop", appUrl), add = TRUE)
 
-  # ============================================================================
-  # Run event loop via httpuv
-  # ============================================================================
+  # Initialize globals used by the event loop and stopApp()
   .globals$reterror <- NULL
   .globals$retval <- NULL
   .globals$stopped <- FALSE
```
cpsievert: This resets a stop request made in `onStart()`. If `onStart()` calls `stopApp(...)`, `startApp()` still returns a handle that stays in `"running"` because `.globals$stopped`, `.globals$retval`, and `.globals$reterror` are cleared here before the first service tick. I think this setup path needs to preserve an early stop request instead of clobbering it. One straightforward fix would be to initialize these globals before `onStart()` runs and then avoid overwriting them afterward if startup already requested a stop.

R/server.R
```diff
@@ -491,13 +494,58 @@ serviceApp <- function() {
...
+  serviceLoop <- function() {
+    if (!identical(.globals$serviceGeneration, generation)) {
+      return(invisible())
+    }
+    if (!.globals$stopped) {
+      promises::with_otel_promise_domain(
+        ..stacktraceoff..(
+          captureStackTraces(
+            tryCatch(
+              ..stacktracefloor..(serviceApp(.shinyServiceDelaySecs * 1000)),
```
cpsievert: This hard-codes a 1 ms polling loop for non-blocking apps. In practice that means the process wakes up hundreds of times per second even when the app is idle, which looks like an avoidable CPU regression. I think it would be better to let `serviceApp()` keep its adaptive timeout logic, or compute the next `later()` delay from the next scheduled event instead of polling at a fixed 1 ms cadence.

R/runapp.R
```diff
@@ -135,32 +136,140 @@ runApp <- function(
...
+#'
+#' # Check status
+#' handle$status()
+#' handle$url()
+#'
+#' # Stop the app
+#' handle$stop()
+#' }
+#'
+#' @seealso [runApp()] for blocking mode, [stopApp()] to stop a running app.
```
cpsievert: The docs here are directionally right, but I think they can be more explicit about the intended API. Users will naturally reach for `stopApp()` when they see `startApp()`. It would help to say plainly in the `startApp()` docs that `handle$stop()` is the intended way to stop a non-blocking app, and that `stopApp()` is primarily for use from inside app code and has different semantics.

cpsievert (commented):
A few targeted follow-ups from local verification. The main correctness issue is the startup-time stop path in `startApp()`. I also think the fixed 1 ms polling loop is worth revisiting before this lands. Details are inline.

R/runapp.R
shikokuchuo: Good catch — and the same bug exists in `runApp()` on main, just hidden by the blocking loop. Fixed it generally: the stop globals are now initialized once at the start of a run (right after any previous app is shut down), before `onStart()` or any setup hook executes. Nothing writes them again during setup, so a stop request can't be clobbered.

In `startApp()`, if setup leaves the app already stopped, I skip the service loop and put the handle straight into its terminal state, so the returned handle reflects reality immediately.

Added a regression test covering blocking return, non-blocking success, and non-blocking error paths.

R/runapp.R
shikokuchuo: Agreed that `handle$stop()` should be the obvious answer. I've added a paragraph to the `startApp()` function docs to make this explicit (and characterizing `stopApp()` as the in-app counterpart). This allows the potentially misleading `@seealso` to `stopApp()` to be dropped.

R/server.R
shikokuchuo: Replaced the 1 ms polling with adaptive scheduling in dce11a8:

- Inner call is now `serviceApp(NA)` — drains queued httpuv I/O and `later` callbacks via `later::run_now()` and returns immediately, never blocking R while waiting
- Next iteration is scheduled off `min(.shinyServiceMaxDelaySecs, timerCallbacks$timeToNextEvent() / 1000, later::next_op_secs())`, so we sleep until the next reactive timer or `later` op (httpuv-scheduled I/O included)
- Idle wakes drop from ~1000/s to ~20/s

Kept a 50 ms ceiling as a defensive backstop for reactive-state changes that don't show up in `next_op_secs()` — e.g. a `reactiveVal` set from a promise resolution. Removing it entirely would need hooks into every reactive mutation path.

R/runapp.R
```diff
@@ -135,32 +136,163 @@ runApp <- function(
...
+  # On success, cleanup responsibility is handed to the caller via the
+  # returned cleanup function.
+  cleanupOnExit <- TRUE
+
+  on.exit(if (cleanupOnExit) handlerManager$clear(), add = TRUE)
+
+  if (isRunning()) {
+    if (!is.null(.globals$runningHandle)) {
+      message("Stopping running Shiny app.")
+      .globals$runningHandle$stop()
```
cpsievert: I think this auto-replace branch may be slightly too broad. The intended behavior seems to be that a new top-level `startApp()` call can replace an existing non-blocking app, but this path also appears reachable from code already running inside an app, which means a nested app launch can stop the outer app instead of hitting the usual protective error. I verified that with an outer app started via `startApp()` that later triggers another `startApp()`. Would it make sense to narrow this branch so auto-replace only applies to true top-level calls, while nested launches from within app execution still error?

R/runapp.R
```diff
@@ -135,32 +136,163 @@ runApp <- function(
...
+    appParts <- as.shiny.appobj(appDir)
+    .setupShinyApp(
+      appDir, appParts, port, launch.browser, host,
+      workerId, quiet, display.mode, test.mode, ops = ops
+    )
+  })
+  setupComplete <- TRUE
+
+  handle <- ShinyAppHandle$new(result$appUrl, result$cleanup)
+  .globals$runningHandle <- handle
```
cpsievert, line 258: I may be missing something, but I don’t think the finalizer/GC safeguard is actually reachable with the current ownership model. `ShinyAppHandle` registers a finalizer, but we also keep a strong reference to the handle here in `.globals$runningHandle`, so dropping the user-visible handle does not seem sufficient to make it collectible while the app is still running. I verified this at a high level with `startApp(app, ...)`, then dropping the return value and forcing `gc()`: `.globals$runningHandle` was still non-`NULL` and `isRunning()` was still `TRUE`. A local follow-up might be either to avoid describing GC-based cleanup as a real safeguard for now, or to change this global state so it doesn’t keep the handle alive directly.

R/runapp.R
shikokuchuo: Good catch! I added an additional check to see if we're already within an app, along the cold path in [4a4e128](https://github.com/rstudio/shiny/pull/4349/commits/4a4e1283e6e7a3037b913c589b149db77a2385e7).

R/runapp.R
shikokuchuo: I've added a comment to make this clear - the finalizer is kept for session exit, and defensively in case the global handle is replaced outside of our code.

R/runapp.R
```diff
@@ -135,32 +136,164 @@ runApp <- function(
...
+  if (isRunning()) {
+    # Auto-replace only at top level; a nested launch from inside a tick
+    # (server, observer, promise callback) must error, not tear down its host.
+    if (is.null(.globals$runningHandle) || .isInAppTick()) {
+      stop("Can't start a new app while another is running. ",
+           "If your application code contains `runApp()` or `startApp()`, remove it. ",
+           "Otherwise, stop the current app first with stopApp().")
+    }
+    message("Stopping running Shiny app.")
+    .globals$runningHandle$stop()
```
cpsievert, line 292: I think there’s still one important behavioral question to resolve around auto-replacement.

Right now, if `startApp(app1)` is running and we call `startApp(app2)`, the code stops `app1` before it knows whether `app2` will start successfully. If `app2` then fails during startup, we end up with no app running. That feels like a potentially surprising contract for callers.

I don’t think this necessarily means the current implementation is wrong, but I do think we should be explicit about which behavior we want. As I see it, there are a few options:

1. Keep auto-replacement, but accept the current semantics.
That would mean documenting clearly that starting a new app tears down the old one immediately, even if the new app later fails to initialize.

2. Keep auto-replacement, but make replacement atomic.
In that model, `app1` would remain running unless `app2` successfully completes startup. This feels like the cleanest caller contract, but it may require more restructuring because the startup path currently leans on shared global state.

3. Drop implicit auto-replacement.
In that model, `startApp()` would error if another app is already running, and callers would do an explicit `handle$stop()` first. That would avoid the failed-replacement issue entirely and would likely simplify some of the replacement/reentrancy logic.

I’m not strongly pushing one direction here, but I do think we should choose deliberately, since the current behavior is observable API behavior and not just an internal implementation detail.

R/server.R
```diff
@@ -491,15 +494,94 @@ serviceApp <- function() {
...
+  }
+  later::later(serviceLoop)
+}
+
+.shinyServiceMaxDelaySecs <- 0.05
 .shinyServerMinVersion <- '0.3.4'
 
+# TRUE if `serviceApp` is on the call stack — i.e. we're inside an app tick.
+# Walked on demand so the service hot path stays untouched.
+.isInAppTick <- function() {
```
cpsievert, line 573: This does fix the nested-launch case when the call happens during `serviceApp()`, but I can still reproduce a replacement from an app-scheduled `later` callback if that callback is pumped from the console with `later::run_now()`. In that path `serviceApp()` is not on the stack, so `.isInAppTick()` returns `FALSE` and the nested `startApp()`/`runApp()` still succeeds. If the intended rule is "no nested launches while another app is running", I think we still need to close that hole or document that console-pumped callbacks are treated differently.

<details>
<summary>Minimal repro</summary>

```r
library(shiny)

inner <- shinyApp(
  ui = fluidPage(),
  server = function(input, output, session) {},
  onStart = function() stopApp("inner_done")
)

outer <- shinyApp(
  ui = fluidPage(),
  server = function(input, output, session) {},
  onStart = function() {
    later::later(function() {
      assign(
        "nested_result",
        try(startApp(inner, launch.browser = FALSE, quiet = TRUE), silent = TRUE),
        envir = .GlobalEnv
      )
    }, 0)
  }
)

handle <- startApp(outer, launch.browser = FALSE, quiet = TRUE)
later::run_now(timeoutSecs = 1)

handle$status()
# currently becomes "success" because the nested start replaces `outer`

inherits(nested_result, "try-error")
# currently FALSE
```
</details>

cpsievert: FWIW, this edge case feels like a sign that the auto-replacement feature, while nice from a UX perspective, is tricky to get completely right. Again, I'm open to you ultimately deciding on a direction, but I think it's also worth calling out. 

I'm also willing to admit that this edge case is contrived and not a realistic thing to be concerned about, but I want to make sure we've considered counter examples

R/runapp.R
shikokuchuo: I think 1, what we currently have makes sense. Starting an app is an explicit signal that you want a new app to run (and hence a previous one to stop). Whether it fails or not, I don't see a situation where the user would want the old app to continue. The expectation is surely that the new app start succeeds.

3 is clean, but I think potentially surprising if the user failed to keep the original app handle for instance.

2 doesn't make sense for me: "ensure an app is running, but I don't care which one".

R/server.R
shikokuchuo: Yeah, this needs all of:

1. A setup-time hook (e.g. `onStart`) that schedules a later callback,
2. That callback re-entering `startApp()` on a different app, and
3. The later queue being drained manually — via a user-driven `later::run_now()` — before serviceLoop has had a chance to run.

Under normal use the very first thing to fire after `startApp()` returns is `serviceLoop`, which drains pending callbacks through `serviceApp()`. In that path `.isInAppTick()` works correctly and the nested launch is rejected. The repro only escapes that path because nothing yields to the event loop in the normal way.

Given the narrowness of this case, I don't think it says anything about the validity of our design.

cpsievert: Sounds good. Thanks for thinking through it.

R/runapp.R
cpsievert: Okay, I can get behind your reasoning. Maybe it's worth explicitly calling that out as known behavior somewhere in the docs?

shikokuchuo: Sure, I've added a section to document this behaviour.

cpsievert (approved):
Great work, thanks @shikokuchuo!

NEWS.md
```diff
@@ -2,6 +2,11 @@
 
 ## New features
 
+* New `startApp()` runs a Shiny app in non-blocking mode, returning a
+  `ShinyAppHandle` object with `stop()`, `status()`, `url()`, and `result()`
+  methods. When a new app is started, any previously running non-blocking app
+  is automatically stopped.
```
```suggestion
  is automatically stopped. (#4349)
```

PR #4352, opened 2026-02-10
feat: Maybe annotate source for Ark

gadenbuie:
This PR introduces `maybeAnnotateSourceForArk()`, which annotates source files for interactive debugging with Ark, primarily in Positron.

Thanks to @lionel-, we also cleaned up some old Windows-only code that worked around a bug in R 3.1.2 (now fixed).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # shiny (development version)
 
+## New features
+
+* Shiny now supports interactive breakpoints when used with Ark (e.g. in Positron). (#4352)
+
 ## Bug fixes and minor improvements
 
 * Fixed an issue with `actionLink()` that extended the link underline to whitespace around the text. (#4348)


R/utils.R
```diff
@@ -1366,31 +1366,66 @@ tryNativeEncoding <- function(string) {
...
+  lines_str <- paste(lines, collapse = "\n")
+  tryCatch(
+    {
+      cli::cli_inform("Annotating source for Ark: {.val {uri}}") # FIXME
+      annotated <- ark_annotate_source(lines_str, uri)
+      if (!is.null(annotated)) {
+        lines <- strsplit(annotated, "\n", fixed = TRUE)[[1]]
+      } else {
+        cli::cli_warn("Ark annotation returned NULL") # FIXME
+      }
```
cpsievert: Remember to drop the `FIXME`s before merging

R/utils.R
```diff
@@ -1366,31 +1366,66 @@ tryNativeEncoding <- function(string) {
...
 }
 
+maybeAnnotateSourceForArk <- function(file, lines) {
+  ark_annotate_source <- get0(".ark_annotate_source", baseenv())
+
+  if (is.null(ark_annotate_source)) {
+    return(lines)
+  }
+
+  file <- normalizePath(file, mustWork = TRUE) # Just to be safe
```
cpsievert: Could drop this now that `sourceUTF8()` does this at the beginning? And if we end up keeping it We probably need the same `winslash = "/"`?

R/utils.R
```diff
@@ -1366,31 +1366,66 @@ tryNativeEncoding <- function(string) {
...
   enc <- if (any(Encoding(lines) == 'UTF-8')) 'UTF-8' else 'unknown'
-  src <- srcfilecopy(file, lines, isFile = TRUE)  # source reference info
-  # oddly, parse(file) does not work when file contains multibyte chars that
-  # **can** be encoded natively on Windows (might be a bug in base R); we
-  # rewrite the source code in a natively encoded temp file and parse it in this
-  # case (the source reference is still pointed to the original file, though)
-  if (isWindows() && enc == 'unknown') {
-    file <- tempfile(); on.exit(unlink(file), add = TRUE)
-    writeLines(lines, file)
-  }
```
cpsievert, line 1381: The fix for this went into R 3.1.2. Let's update `Depends` to `R (>= 3.1.2)`?

PR #4355, opened 2026-02-13
Multi-app non-blocking mode for `runApp()`

shikokuchuo:
This is an alternative to #4349 which also enables running multiple Shiny apps at the same time.

## Summary

- `runApp()`, `runExample()`, and `runGadget()` gain a `blocking` parameter. When `blocking = FALSE`, the app runs in the background via `later` callbacks and returns a `ShinyAppHandle` object immediately, freeing the R console. Multiple apps can run concurrently.
- Internal app state is refactored from global singletons to per-app environments, enabling safe concurrent execution.

### Non-blocking mode

```r
app <- runApp("myapp", blocking = FALSE)
app$url()     # "http://127.0.0.1:7890"
app$status()  # "running"
app$stop()
app$result()  # return value from stopApp()
```

The default can be changed globally with `options(shiny.blocking = FALSE)`.

### Architecture changes

- **Per-app state**: Each app gets its own `appState` environment (keyed by token in `.globals$appStates`), replacing the single `.globals$appState`. Fields like `retval`, `reterror`, `stopped`, `handlerManager`, `onStopCallbacks`, and `onUnhandledErrorCallbacks` are all per-app.
- **Shared service loop**: A single `later`-based `serviceLoop()` services all non-blocking apps, calling `serviceApp()` once per cycle. Per-app lifecycle (capture result, cleanup) is handled by callbacks stored on each `appState`.
- **`ShinyAppHandle`** (R6): Returned by non-blocking `runApp()`. Provides `stop()`, `status()`, `url()`, `result()`. A finalizer auto-stops the app if the handle is garbage collected.
- **Cleanup consolidation**: A single `createCleanup()` closure handles server stop, callback invocation, handler manager clearing, and state removal. Global state (options, global `onStop` callbacks) is only restored when the last app stops.
- **Per-request app context**: HTTP and WebSocket handlers set `.globals$currentAppState` at the start of each request, so that `getCurrentAppState()` always returns the correct app even when httpuv interleaves requests from multiple concurrent apps.


shikokuchuo:
The non-blocking branch has now been much simplified. Closing as this branch has diverged significantly, and we'd need to rebase before proceeding again.

PR #4375, opened 2026-04-28
Remove dead `InputBinding.dispose()` method

### LINKED ISSUE #2875: InputBinding.dispose method never called
As I see in `input_binding.js` file, there is a method in `InputBinding` class that is called `dispose` and it's supposed to be called after unbinding the input.

However it is never called. As a workaround I can listen to `shiny:unbound` event, but it would be good if that method was called as intended (as initialize).

I could make a patch/PR if you're interested.

cpsievert:
Closes #2875.

## Summary

- Removes `InputBinding.dispose()` rather than wiring it up, since it was defined but never called for 6+ years, is not documented in the [custom input bindings article](https://shiny.posit.co/r/articles/build/js-custom-input/), and `unsubscribe()` already serves as the teardown hook.
- Also removes the generated type declaration for `dispose`.

## Test plan

- [x] TypeScript checks, type coverage, linting, and tests all pass


schloerke (approved):
Thank you!

schloerke:
(Assets will need to possibly be built. Please wait for `routine` GHA job to finish)

NEWS.md
```diff
@@ -15,6 +15,10 @@
 * `conditionalPanel()` no longer briefly flashes its contents on app start
   when the condition is initially `FALSE`. (#3505)
 
+* Removed `InputBinding.dispose()` from the JavaScript `InputBinding` class.
+  This method was never called by Shiny's runtime, so any overrides were dead
+  code. Use `unsubscribe()` for cleanup logic instead. (#2875)
```
```suggestion
  code. Use `unsubscribe()` for cleanup logic instead. (#4375)
```

PR #4386, opened 2026-05-27
Fix: helpful error from `need()` when `message` and `label` are missing

### LINKED ISSUE #2509: Add helpful error message
Sorry not to issue a PR, but I'm currently super busy. 

If label is left out of `shiny::need()`, the error message is not easy to figure out. 
It would be nice to include something like "label is missing in `need()`"

```r
> shiny::need(1+1)
Error in paste(label, "must be provided") : 
  argument "label" is missing, with no default
```

Relevant line is here:
https://github.com/rstudio/shiny/blob/e6c2133520e3f194e46aec2d545713ebde3a5fe9/R/utils.R#L1199

samuelbharti:
Closes #2509

## Summary

`shiny::need(expr)` called without either a `message` or `label` argument previously failed with the cryptic `argument "label" is missing, with no default`, raised from inside the default-argument expression `paste(label, "must be provided")`. This PR adds an explicit `missing(message) && missing(label)` check at the top of `need()` so the failure now says `` `need()` requires either a `message` or `label` argument. ``

## Verification

```r
shiny::need(1 + 1)
#> Error: `need()` requires either a `message` or `label` argument.

shiny::need(NULL, label = "input$x")
#> [1] "input$x must be provided"

shiny::need(NULL, "custom message")
#> [1] "custom message"
```

Tests in `tests/testthat/test-utils.R` exercise the new error path alongside the existing `label`-only, `message`-only, `message = FALSE`, and both-provided cases.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -1064,7 +1064,9 @@ validate <- function(..., errorClass = character(0)) {
 need <- function(expr, message = paste(label, "must be provided"), label) {
 
   if (missing(message) && missing(label)) {
-    stop("`need()` requires either a `message` or `label` argument.", call. = FALSE)
+    cli::cli_abort(
+      "{.fn need} requires either a {.arg message} or {.arg label} argument."
+    )
   }
   force(message) # Fail fast in case `label` is missing but referenced via `message`
 
--- a/R/utils.R
+++ b/R/utils.R
@@ -1063,7 +1063,10 @@ validate <- function(..., errorClass = character(0)) {
 #' @rdname validate
 need <- function(expr, message = paste(label, "must be provided"), label) {
 
-  force(message) # Fail fast on message/label both being missing
+  if (missing(message) && missing(label)) {
+    stop("`need()` requires either a `message` or `label` argument.", call. = FALSE)
+  }
+  force(message) # Fail fast in case `label` is missing but referenced via `message`
 
   if (!isTruthy(expr))
     return(message)
--- a/tests/testthat/test-utils.R
+++ b/tests/testthat/test-utils.R
@@ -106,7 +106,6 @@ test_that("need() works as expected", {
   expect_null(need(TRUE, FALSE))
   expect_null(need(c(NA, NA, TRUE), FALSE))
   expect_null(need(c(FALSE, FALSE, TRUE), FALSE))
-
 })
 
 test_that("need() message/label handling (#2509)", {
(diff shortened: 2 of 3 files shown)


R/utils.R
```diff
@@ -1063,7 +1063,10 @@ validate <- function(..., errorClass = character(0)) {
 #' @rdname validate
 need <- function(expr, message = paste(label, "must be provided"), label) {
 
-  force(message) # Fail fast on message/label both being missing
+  if (missing(message) && missing(label)) {
```
schloerke, line 1066: Optional / stylistic: consider matching the newer `cli::cli_abort()` pattern used elsewhere in this dev cycle (e.g. `R/bootstrap.R:1295-1298` in `downloadButton()`). The inline markup styles the function and argument names automatically:

```r
cli::cli_abort(
  "{.fn need} requires either a {.arg message} or {.arg label} argument."
)
```

That said, `stop(..., call. = FALSE)` matches the surrounding style in `R/utils.R`, so this is purely a judgment call.

tests/testthat/test-utils.R
```diff
@@ -106,6 +106,38 @@ test_that("need() works as expected", {
   expect_null(need(TRUE, FALSE))
   expect_null(need(c(NA, NA, TRUE), FALSE))
   expect_null(need(c(FALSE, FALSE, TRUE), FALSE))
+
```
schloerke: Nit: this extra blank line was added inside the existing `test_that("need() works as expected", { ... })` block, right before its closing `})`. Looks accidental — worth removing to keep the diff minimal.

tests/testthat/test-utils.R
```diff
@@ -106,6 +106,38 @@ test_that("need() works as expected", {
...
+})
+
+test_that("need() message/label handling (#2509)", {
+  # Errors clearly when both `message` and `label` are missing, regardless of
+  # whether `expr` is truthy or falsy (the message default is evaluated eagerly).
+  expect_error(
+    need(1 + 1),
+    "`need\\(\\)` requires either a `message` or `label` argument"
+  )
```
schloerke, line 118: Minor: `fixed = TRUE` reads more cleanly than escaping the parens in a regex, and avoids future regex foot-guns:

```r
expect_error(
  need(1 + 1),
  "`need()` requires either a `message` or `label` argument",
  fixed = TRUE
)
```

Same applies to the second `expect_error()` call just below.

schloerke (commented):
A few small suggestions on the implementation and tests — none blocking.

R/utils.R
samuelbharti: Switched to `cli::cli_abort()` with the suggested inline markup in 15073686b. Confirmed the rendered message is identical to before: `` `need()` requires either a `message` or `label` argument. ``

tests/testthat/test-utils.R
samuelbharti: Good catch — that was accidental. Removed in 15073686b.

tests/testthat/test-utils.R
samuelbharti: Switched both `expect_error()` calls to `fixed = TRUE` in 15073686b.

schloerke (approved):

PR #4391, opened 2026-05-28
Docs: clarify `varSelectInput()` purpose and fix typo

### LINKED ISSUE #2334: varSelectInput: typo in documentation, and would also help to make it clear why it's needed
I just looked over the docs for the new varSelectInput and ran into trouble trying to parse this sentence: https://github.com/rstudio/shiny/blob/c79034649077f5fa6f8d615ce10c457eb33a2d2a/R/input-select.R#L240-L244

Another thing I wasn't able to pick up from the documentation is why this input is even necessary and made it into the very elite list of inputs supported by shiny. It feels very esoteric, and I'm having a hard time rationalizing why shiny added it.

The documentation didn't offer much insight other than saying it returns symbols, but it would be helpful if the docs were more explicit on why exactly this is useful. I also looked at @schloerke 's [example app from the PR](https://github.com/rstudio/shiny/pull/2091#issue-193434802) and it still seems to me that the only reason this exists is to return a symbol.

If that is the reason for this input, then would it make sense to instead offer a `symbol = TRUE/FALSE` parameter for select inputs, so that this would be more generic rather than only work for dataframes? The specificity of this input feels a bit forced to suit a very particular need that isn't a general shiny need. Adding more clear documentation about its usage would clear this up :)

samuelbharti:
Fixes #2334

## Summary

`varSelectInput()`'s documentation didn't explain *why* the input exists — only that it returns a symbol. Per @jcheng5 in the issue thread, its sole purpose is to make `selectInput()` ergonomic for tidy-evaluation contexts (e.g. `ggplot2::aes(!!input$var)`) by returning a symbol directly. This PR adds a short paragraph stating that purpose up front and pointing readers to the existing "Server value" section for the mechanics, and recommends plain `selectInput()` when a symbol is not needed.

It also fixes a grammatical typo ("to instead of" → "instead of") in the same docstring's description of `selectize.js`.

No behavior changes; documentation only.

## Verification

- `?varSelectInput` (or the pkgdown reference page) shows the new explanatory paragraph after the title block and before the existing `selectize.js` paragraph.
- The "to instead of the basic select input element" sentence now reads "instead of the basic select input element."
- `devtools::check()` passes.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -36,6 +36,10 @@
 
 * `need()` now gives a clearer error when called without either a `message` or `label` argument, instead of the cryptic "argument \"label\" is missing, with no default". (thanks @chasemc and @sundrelingam, #2509)
 
+* Clarified `varSelectInput()` documentation to explain that the input
+  returns a symbol for use with tidy evaluation, and fixed a grammatical
+  typo. (#2334)
+
 # shiny 1.13.0
 
 ## New features
--- a/R/input-select.R
+++ b/R/input-select.R
@@ -299,9 +299,15 @@ selectizeScripts <- function() {
 #' Create a select list that can be used to choose a single or multiple items
 #' from the column names of a data frame.
 #'
+#' `varSelectInput()` is a convenience wrapper around [selectInput()] for use
+#' with tidy evaluation: it returns the selected column name as a symbol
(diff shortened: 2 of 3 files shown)


cpsievert:
Thanks!

PR #4394, opened 2026-06-01
fix: support running a module in the parent namespace via an empty id

cpsievert:
After the new `session$destroy()` work (#4372), the root-scope detection (`if (!nzchar(namespace))`) errors with "argument is of length zero" when a module is run with an empty id, e.g. `callModule(server, character(0))`.

That empty-id idiom is how packages such as **teal** run a module server in the parent namespace, and it is the documented behavior of `NS()`:

> Length 0 will cause the `id` to be returned without a namespace

This makes the new destroy machinery handle that case. A length-0 namespace is treated as the root scope in the destroy-callback bookkeeping only; the namespace itself is left untouched so `NS(character(0))` keeps returning ids unchanged (normalizing to `""` would instead prepend a stray `-` separator).

Caught by reverse-dependency checks against teal.

### DIFF
--- a/R/mock-session.R
+++ b/R/mock-session.R
@@ -763,7 +763,8 @@ MockShinySession <- R6Class(
     # @param ns The namespace key.
     # @return A Callbacks object.
     getOrCreateDestroyCallbacks = function(ns) {
-      if (!nzchar(ns)) ns <- "..root"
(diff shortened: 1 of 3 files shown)


cpsievert:
Closing in favor of #4395

PR #4395, opened 2026-06-01
fix: support `NULL` as the canonical root namespace for `session$destroy()`, etc

cpsievert:
This refines the unreleased `session$destroy()` work (#4372) so that an empty module id — `callModule(server, character(0))`, the documented `NS()` length-0 idiom that packages like **teal** use to run a module server in the parent namespace — is treated as the root scope throughout the destroy machinery.

The first commit records a pairing-session refactor (making `character(0)` the canonical root value instead of `""`, renaming the destroy `id` argument to `namespace`). The second commit fixes several issues found while reviewing it:

- **Real `ShinySession$destroy()` no longer recurses infinitely.** The renamed `namespace` parameter shadowed the scope’s own captured namespace and the self-destroy branch had been replaced by a recursive `makeScope()` call that never reached `invokeDestroyCallbacks()`. The self-vs-child branch is restored, with the scope’s namespace captured as `selfNamespace`.

- **`MockShinySession` now matches the real session.** It had been left on the old `nzchar()`/`""` logic, so it still errored with "argument is of length zero" on `callModule(server, character(0))` (the teal failure path via `testServer()`). The `character(0)` handling is mirrored so mock and real behave identically.

- **Destroy `namespace` is validated at the public boundary**, before `makeScope()` runs, via a small `validateDestroyNamespace()` helper.

- **Restored safeguards:** the reserved `"..root"` namespace guard in `makeScope()`, and mapping `""` to root in the destroy-key logic (fastmap can’t key on an empty string).

## Tests
Full suite passes (0 failures). New coverage: empty-namespace `makeScope()`/`callModule()` acceptance, un-prefixed ids (`scope$ns("x") == "x"`), and non-recursive teardown.

Caveat for reviewers: the recursion fix is verified through `MockShinySession` (now line-for-line equivalent to the real session) plus static reasoning — the test harness can’t instantiate a live `ShinySession`.

Caught by reverse-dependency checks against teal. Supersedes the minimal-guard approach in #4394 (one of the two should be chosen).

### DIFF
--- a/R/shiny.R
+++ b/R/shiny.R
@@ -1031,6 +1031,14 @@ ShinySession <- R6Class(
           call. = FALSE
         )
       }
+      # `character(0)` is the root; an empty string is not a valid namespace
+      # (it can't be used as a fastmap key, and `NS(\"\")` yields a stray `-`).
+      if (length(namespace) == 1L && !nzchar(namespace)) {
+        stop(
+          "A module namespace must be a non-empty string; use `character(0)` for the root scope.",
+          call. = FALSE
+        )
+      }
       ns <- NS(namespace)
       # The scope's own namespace, captured because the proxy `destroy()` below
       # has a `namespace` parameter that would otherwise shadow it.
--- a/R/shiny.R
+++ b/R/shiny.R
@@ -1031,11 +1031,11 @@ ShinySession <- R6Class(
           call. = FALSE
         )
       }
-      # `character(0)` is the root; an empty string is not a valid namespace
-      # (it can't be used as a fastmap key, and `NS(\"\")` yields a stray `-`).
-      if (length(namespace) == 1L && !nzchar(namespace)) {
+      # `character(0)` is the root; "" / NA are not valid namespaces (they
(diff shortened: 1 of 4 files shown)


R/shiny.R
```diff
@@ -1069,8 +1090,7 @@ ShinySession <- R6Class(
             dots <- eval(substitute(alist(...)))
           }
 
-          if (any_unnamed(dots))
-            stop("exportTestValues: all arguments must be named.")
+          if (any_unnamed(dots)) stop("exportTestValues: all arguments must be named.")
```
cpsievert:
```suggestion
          if (any_unnamed(dots)) 
            stop("exportTestValues: all arguments must be named.")
```

R/shiny.R
```diff
@@ -1069,7 +1090,7 @@ ShinySession <- R6Class(
             dots <- eval(substitute(alist(...)))
           }
 
-          if (any_unnamed(dots))
+          if (any_unnamed(dots))
```
cpsievert:
```suggestion
          if (any_unnamed(dots))
```

PR #4403, opened 2026-06-23
Make `{watcher}` a required dependency for autoreload

shikokuchuo:
## Summary

Moves `{watcher}` from `Suggests` to `Imports` and uses it unconditionally for autoreload. Removes the legacy polling-based file watcher.

## Changes

- **`DESCRIPTION`**: `watcher` → `Imports`, with `(>= 0.2.0)`.
- **`R/shinyapp.R`** (`initAutoReloadMonitor()`): drop the `is_installed("watcher")` branch, the legacy `observe()`/`invalidateLater()` poller, and the `shiny.autoreload.legacy_warning` option. Return `invisible(watcher$stop)` so cleanup flows through the existing `onStop → monitorHandle()` path.
- **`R/shiny-options.R`** / **`man/shinyOptions.Rd`**: simplify the `shiny.autoreload` option docs now that watcher is the sole path (drop the historical "no longer polls"/"no longer expensive" hedging).
- **`NEWS.md`**: note the dependency change.

## Compatibility

- `shiny.autoreload.interval` / `.pattern` unchanged (interval still passed as `latency`, default 250ms).
- `shiny.autoreload.legacy_warning` removed; silently ignored if set.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -97,6 +97,7 @@ Imports:
     sourcetools,
     tools,
     utils,
+    watcher (>= 0.2.0),
     withr,
     xtable
 Suggests:
@@ -119,7 +120,6 @@ Suggests:
     shinytest2,
     showtext,
     testthat (>= 3.2.1),
-    watcher,
     yaml
 Config/Needs/check: shinyjs
 Config/roxygen2/version: 8.0.0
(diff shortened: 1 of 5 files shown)


shikokuchuo:
@cpsievert The original intention of https://github.com/rstudio/shiny/pull/4185 was to enable this by default, and as watcher now uses R's standard build toolchain, this should address any lingering source installation concerns. End of the day, I think this is a good affordance for Shiny to provide.

cpsievert:
Thanks!

Issue #4415, opened 2026-07-16
MCP Apps: tool to update an already-running app in place (auto-registered update_<appId>_app)


schloerke:
## Summary

Add a way for an MCP host/model to **update an already-running Shiny MCP app in place**, rather than only re-opening it. Concretely: when an app declares `arguments` in `mcpConfigure()`, **auto-register a companion `update_<appId>_app` tool** that pushes new arguments into the running session's `mcpUpdates()` reactive channel — no re-render.

Follow-up to #4414 (the `mcpConfigure()` / init-args-via-RestoreContext work).

## Motivation — what we observed with a real host

In a live Claude session driving a clock app (`open_clock_app` with `label` / `paused` args), asking Claude to "change the label" caused it to **re-invoke `open_clock_app`, rendering a fresh widget each time** (the user saw multiple app instances in the transcript). Claude's own explanation:

> "I can't send additional arguments to an already-running instance. The only way to modify the app is by calling the tool again with new parameters."

Root cause: **the model's only lever on the app is calling a tool, and the only tool it has is `open_*`.** Whether a repeat `open_*` call becomes an in-place update or a new render is up to the host — and Claude renders a new instance per call. So the runtime-update path we built (`mcpUpdates()`, fed by a host `ui/notifications/tool-input` to the live iframe) is **effectively never triggered by Claude**. In practice the feature behaves as init-args-only.

Giving the model an explicit **`update` verb** (a separate tool) fixes this: the model can choose to update in place instead of re-opening, and the call reaches the running session.

### The update tool
- When `mcpConfigure(arguments = ...)` is set, auto-register a companion tool, e.g. `update_<appId>_app`, whose `inputSchema` is the **same declared `arguments` allow-list**, with a description like: *"Change parameters of the already-open app in place. Do NOT re-open the app; use this to update the running instance."*
- Unlike the app-opening tool (whose args are delivered to the *iframe* via `ontoolinput`), a plain tool call runs in the **R server process** (like existing `registerMcpTool()` tools). The handler locates the running app session(s) and **pushes the (allow-list-filtered) args into the same `mcpUpdates()` channel** the app already observes:
  ```r
  observe({
    args <- mcpUpdates()          # now also fires from a server-side update push
    if (!is.null(args$label)) ...
  })
  ```
- This **revives `mcpUpdates()` as a real runtime path** — fired by a server-side push instead of a host tool-input notification — and reuses `arguments` end-to-end. No `resources/read`, no new render.

[...]

PR #4418, opened 2026-07-17
feat(mcp): update a running Shiny MCP App in place via update_<appId>_app (#4415)

schloerke:
## Summary

Follow-up to #4414. Gives an MCP host/model an explicit **update verb** so it can change an already-open Shiny MCP App **in place** instead of re-opening it (which today renders a fresh instance per `open_*` call).

When an app declares `mcpConfigure(arguments = ...)`, Shiny now auto-registers a companion **`update_<appId>_app`** tool. A model call to it runs server-side, targets a specific running session, and pushes new arguments through the existing `mcpUpdates()` reactive channel — no re-render, no author code beyond the single `observe(mcpUpdates())` apps already write.

## Design (5 decisions)

1. **Explicit session targeting** — `update_*` takes a required `session` token; no broadcast/most-recent guessing. To update several instances the model loops.
2. **Server-side `reactiveVal` overlay** — each MCP session holds a server-push `reactiveVal`; `mcpUpdates()` overlays it on the client-delivered init args (per-key, latest wins) then allow-list filters. Successive pushes **accumulate**.
3. **Registry via existing `appsByToken`** — keyed by `session$token`; unknown/stale/non-MCP token → tool error (never silent success).
4. **Auto tool, only when `arguments` declared** — handler pushes inside `withReactiveDomain(session, …)` + `requestFlush()` so the app's observer fires in the right context.
5. **Token delivery** — on connect the app announces its `session$token` to the model (reserved-key envelope through `mcpUpdateModelContext`, + a text instruction); `mcpUpdateModelContext()` payload now stamps `session` into `structuredContent`.

Entirely R-side — the JS bridge already forwards these messages.

## Known limitations (see `mcp/limitations.md`)

- **Host-passthrough spike:** needs end-to-end confirmation that a real host (claude.ai) forwards the `{session, state}` structured content + instruction so the model can echo the id. (Verify in Claude Desktop.)
- **Multi-instance:** with several open instances the model must pick the right token; no server-side guarantee.

## Docs & tests

- Spec: `mcp/design-mcp-update-running-app.md`; plan: `mcp/implementation-plan-update-running-app.md`.
- `?mcp-session` roxygen, `mcp/design-mcpConfigure.md`, `mcp/limitations.md`, `NEWS.md` updated; demo apps (clock/demo/cars) get explanatory comments — the update path needs **zero** app-code change.
- Full MCP test suite: 279 passing (registry lifecycle, overlay/accumulate semantics, dispatch incl. error paths, tool visibility, connect-time announcement, envelope shape).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,11 @@
 # shiny (development version)
 
+* Shiny MCP Apps configured with `mcpConfigure(arguments = ...)` now
+  auto-register a companion `update_<appId>_app` tool. A model can call it to
+  update an already-open app instance in place (via `mcpUpdates()`) instead of
+  re-opening the app, targeting a specific instance by the `session` id the app
+  announces on connect. (#4415)
+
 * Experimental support for serving a Shiny app as an MCP App (the Model
   Context Protocol Apps extension, SEP-1865). Calling `mcpConfigure()`
   before the app starts mounts an MCP endpoint at `/mcp` on the app's
--- a/R/mcp-config.R
+++ b/R/mcp-config.R
@@ -132,6 +132,14 @@ mcpToolName <- function() {
   if (is.null(id)) "open_shiny_app" else paste0("open_", id, "_app")
 }
 
+# update_shiny_app, or update_<appId>_app when an appId is configured. The
+# companion to mcpToolName(): the model calls this to change a running
+# instance in place instead of re-opening the app.
+mcpUpdateToolName <- function() {
+  id <- .globals$mcp$appId
+  if (is.null(id)) "update_shiny_app" else paste0("update_", id, "_app")
+}
+
 # Convert the configured `arguments` (named list of ellmer types) into the
 # JSON Schema published as the tool's inputSchema. Mirrors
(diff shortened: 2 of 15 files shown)


schloerke:
### ✅ Verified end-to-end in Claude Desktop (2026-07-17)

Both open items from `mcp/limitations.md` now confirmed working against a real host (Claude Desktop, stdio transport):

- **Host-passthrough spike (§4):** the host forwards the announced `session` token into the model's context, and the model calls `update_<appId>_app` **in place** to change the running instance — it no longer re-opens a fresh instance.
- **Multi-instance targeting (§5):** with **two** instances of the app open, updating the **first** instance's label targeted the correct session token.

The multi-instance case remains a *model-judgment* path (no server-side guarantee), so the caveat stays documented — but it works in practice for the common case. Pushed `c44eac267` marking both as confirmed. Not yet spot-checked on claude.ai (web); the tunnel transport is expected to carry it.

PR #4421, opened 2026-08-06
fix: soft teardown of reactives when a session closes

### LINKED ISSUE #4420: Refreshing during async work errors: Can't access reactive `Theme Counter`; its module session has been destroyed
Found by @gadenbuie (Garrick Aden-Buie) while using shinychat: refreshing the page during a streaming chat response produces

```
Can't access reactive `Theme Counter`; its module session has been destroyed
```

## Summary

Since the `session$destroy()` / `session$onDestroy()` work (#4372, #4395), closing a session tears down every reactive created in the session's domain. `ShinySession$wsClosed()` calls `private$invokeDestroyCallbacks(allowRoot = TRUE)`, which fires the weak destroy wrappers that each `reactiveVal()` / `reactiveValues()` / `reactive()` registers at construction time.

The result: any async continuation that outlives the websocket close now errors on its first reactive access. Refreshing the page while async work is in flight is routine, so this turns a previously benign race into a user-visible error.

Affects shiny 1.14.0 and later, including current `main`. (An earlier version of this issue incorrectly said this was unreleased; the `session$destroy()` work shipped in 1.14.0.)

## Reprex

No shinychat, no `ExtendedTask`, no promises — a plain `later()` callback is enough.

```r
library(shiny)

ui <- fluidPage(
  actionButton("go", "Start"),
  verbatimTextOutput("out")
)

  rv <- reactiveVal(0, label = "my_counter")

  observeEvent(input$go, {
    # Stand-in for any async continuation: later(), a promise, an httr2
    # callback, an ExtendedTask. Reload the page before this fires.
    later::later(function() {
      cat("--- deferred callback; session closed:", session$isClosed(), "---\n")
      show <- function(what, expr) {
        tryCatch(
          { isolate(expr); cat("  OK   ", what, "\n") },
          error = function(e) cat("  ERROR", what, "->", conditionMessage(e), "\n")
        )
      }
      show("session$getCurrentTheme()", session$getCurrentTheme())
      show("read  reactiveVal", rv())
      show("write reactiveVal", rv(1))
    }, delay = 6)
  })

  output$out <- renderText("Click the button, then reload the page within 6 seconds")
}

shinyApp(ui, server)
```

**To reproduce:** click the button, then reload the page within 6 seconds.

Actual (shiny 1.14.0.9000, `main`):

```
--- deferred callback; session closed: TRUE ---
  ERROR session$getCurrentTheme() -> Can't access reactive `Theme Counter`; its module session has been destroyed
  ERROR read  reactiveVal -> Can't access reactive `my_counter`; its module session has been destroyed
  ERROR write reactiveVal -> Can't access reactive `my_counter`; its module session has been destroyed
```

Expected (and what happens if you do *not* reload):

```
--- deferred callback; session closed: FALSE ---
  OK    session$getCurrentTheme()
  OK    read  reactiveVal
  OK    write reactiveVal
```

## How it shows up in shinychat

`Theme Counter` is just the first reactive shinychat happens to touch; user-created reactives fail identically. The chain:

1. `shinychat:::chat_mod_server()` streams the response in a `shiny::ExtendedTask`.
2. The user refreshes; the socket closes and every session-scoped reactive is destroyed.
3. The promise settles and calls `chat_append()` -> `shinychat:::with_current_theme()` -> `bslib::bs_current_theme()`.
4. `bs_current_theme()` calls `session$getCurrentTheme()`, which reads the session-scoped `reactiveVal(0, label = "Theme Counter")` (`R/shiny.R`) -> destroyed -> error.

`ExtendedTask` then fails *again* setting its own state, surfacing as a second error:

```
Unhandled promise error: Can't access reactive `ExtendedTask$private$status`; its module session has been destroyed
```

so the task ends in neither `"success"` nor `"error"`. Erroring during teardown makes cleanup less correct, not more.

## Why erroring here seems wrong

- **Shiny's own code depends on post-close writes.** `ExtendedTask` sets `rv_status` / `rv_error` when its promise settles, which routinely lands after a refresh. Under this rule, `ExtendedTask` itself is in violation.
- **App authors cannot comply.** The offending access is usually inside shinychat/bslib/`ExtendedTask`, not user code. Even in user code there is no correct guard: `if (!session$isClosed())` is TOCTOU, since the socket can close between the check and the access.
- **The write was already inert.** Observers are destroyed when the domain ends, the session is removed from `appsNeedingFlush` on close (`R/server.R`), and `flushOutput()` early-returns when closed (`R/shiny.R`). A post-close set mutates an object graph whose consumers are all gone and which is about to be GC'd — it has no observable effect. Confirmed by experiment: an observer depending on such a reactiveVal does not re-run.
- **It is a race**, so it appears nondeterministically in production logs.

`destroy` and `close` are genuinely different contracts. An explicit `session$destroy(ns)` on a live session means a later access is a leaked handle — a real bug worth erroring on. A socket close means the whole graph is transitively dead and there is no surviving consumer to confuse.

## Proposed fix

Make whole-session close a *soft* teardown while keeping `session$destroy(ns)` hard:

- `onDestroy()` callbacks still fire on close (module cleanup, `invalidateLater` timer cancellation, etc. all keep working).
- Reactive values/expressions are left readable at their last value on close, and reclaimed by ordinary garbage collection.
- Explicit `session$destroy(namespace)` on a live session continues to hard-destroy and raise `shiny.destroyed.error`.

Fix proposed in #4421.

schloerke:
Fixes #4420. Found by @gadenbuie (Garrick Aden-Buie) while using shinychat.

## Problem

Since #4372 / #4395, closing a session tears down every reactive created in the session's domain: `ShinySession$wsClosed()` calls `invokeDestroyCallbacks(allowRoot = TRUE)`, which fires the weak destroy wrapper that each `reactiveVal()` / `reactiveValues()` / `reactive()` registers at construction.

So any async continuation that outlives the websocket close errors on its first reactive access. Refreshing the page during in-flight async work is routine, which turned a previously benign race into a user-visible error:

```
Can't access reactive `Theme Counter`; its module session has been destroyed
```

`Theme Counter` is just the first reactive shinychat happens to touch (`chat_append()` -> `with_current_theme()` -> `bslib::bs_current_theme()` -> `session$getCurrentTheme()`). User-created reactives failed identically, and `ExtendedTask` then failed *again* setting its own state, leaving the task in neither `"success"` nor `"error"`.

Affects shiny 1.14.0 and later, including current `main` -- the `session$destroy()` work shipped in 1.14.0.

## Fix

A close is not a destroy, so the two are now distinguished:

- **On close**, reactive values and expressions are left intact, readable at their last value, and reclaimed by ordinary garbage collection. `onDestroy()` callbacks still fire, so module cleanup and `invalidateLater` timer cancellation are unaffected.
- **On explicit `session$destroy(namespace)`** against a live session, teardown is unchanged and a later access still raises `shiny.destroyed.error` -- there, a stale access is a genuine bug worth surfacing.

`make_weak_destroy_wrapper()` takes the domain and no-ops when that domain is closed. No new state, so `get()` returning the last value and `set()` staying legal-but-inert both fall out naturally.

`MockShinySession$close()` now marks the session closed *before* running teardown, matching `ShinySession$wsClosed()`.

## Why not keep erroring on post-close access

- **Shiny's own code depends on it.** `ExtendedTask` sets `rv_status` / `rv_error` when its promise settles, which routinely lands after a refresh. Under the old rule `ExtendedTask` itself was in violation.
- **App authors cannot comply.** The access is usually inside shinychat/bslib/`ExtendedTask`, not user code, and `if (!session$isClosed())` is TOCTOU.
- **The write was already inert.** Observers are destroyed when the domain ends, the session is dropped from `appsNeedingFlush`, and `flushOutput()` early-returns when closed. Verified experimentally: an observer depending on a post-close write does not re-run.
- **Erroring made teardown worse**, not better -- the first error aborted the promise chain and left `ExtendedTask` in a bogus state.

## Tests

12 new tests in `test-destroy.R`, all of which fail without the source change and pass with it:

- reactiveVal / reactiveValues / reactive expressions readable and writable after close, and module reactives surviving a root close
- an observer does **not** re-run on a post-close write (soft teardown must not resurrect the session)
- explicit module `destroy()` still raises `shiny.destroyed.error`

The theme tests drive a **real `ShinySession`** via a websocket stub rather than `MockShinySession`, whose `getCurrentTheme()` is a `makeWarnNoops` stub that never touches the `Theme Counter` reactiveVal:

- `session$getCurrentTheme()` readable after a real `wsClosed()`
- the `Theme Counter` reactiveVal holds its last value across close (bumped first, so the assertion distinguishes "kept its value" from "reset")
- a real `ExtendedTask` that settles after close reaches `"success"` with its result
- an `ExtendedTask` whose continuation reads the session theme after close, i.e. the full reported chain in one test

Verified by reverting only `R/reactives.R` and `R/mock-session.R`: all 12 fail with the exact errors from the report.

Also confirmed end-to-end in a browser -- click, refresh mid-task, previously two errors, now clean.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -122,7 +122,7 @@ Suggests:
     testthat (>= 3.2.1),
     yaml
 Config/Needs/check: shinyjs
-Config/roxygen2/version: 8.0.0
+Config/roxygen2/version: 8.1.0
 Config/testthat/edition: 3
 Encoding: UTF-8
 Roxygen: list(markdown = TRUE)
--- a/NAMESPACE
+++ b/NAMESPACE
@@ -342,92 +342,104 @@ import(httpuv)
 import(methods)
 import(mime)
 import(xtable)
-importFrom(fastmap,fastmap)
-importFrom(fastmap,is.key_missing)
-importFrom(fastmap,key_missing)
-importFrom(grDevices,dev.cur)
-importFrom(grDevices,dev.set)
-importFrom(htmltools,HTML)
-importFrom(htmltools,a)
-importFrom(htmltools,br)
-importFrom(htmltools,code)
-importFrom(htmltools,div)
-importFrom(htmltools,em)
-importFrom(htmltools,h1)
-importFrom(htmltools,h2)
-importFrom(htmltools,h3)
-importFrom(htmltools,h4)
-importFrom(htmltools,h5)
-importFrom(htmltools,h6)
-importFrom(htmltools,hr)
-importFrom(htmltools,htmlTemplate)
-importFrom(htmltools,img)
-importFrom(htmltools,includeCSS)
-importFrom(htmltools,includeHTML)
-importFrom(htmltools,includeMarkdown)
-importFrom(htmltools,includeScript)
-importFrom(htmltools,includeText)
-importFrom(htmltools,is.singleton)
-importFrom(htmltools,p)
-importFrom(htmltools,pre)
-importFrom(htmltools,singleton)
-importFrom(htmltools,span)
-importFrom(htmltools,strong)
-importFrom(htmltools,suppressDependencies)
-importFrom(htmltools,tag)
-importFrom(htmltools,tagAppendAttributes)
-importFrom(htmltools,tagAppendChild)
-importFrom(htmltools,tagAppendChildren)
-importFrom(htmltools,tagGetAttribute)
-importFrom(htmltools,tagHasAttribute)
-importFrom(htmltools,tagList)
-importFrom(htmltools,tagSetChildren)
-importFrom(htmltools,tags)
-importFrom(htmltools,validateCssUnit)
-importFrom(htmltools,withTags)
-importFrom(lifecycle,deprecated)
-importFrom(lifecycle,is_present)
-importFrom(promises,"%...!%")
-importFrom(promises,"%...>%")
-importFrom(promises,as.promise)
-importFrom(promises,hybrid_then)
-importFrom(promises,is.promise)
-importFrom(promises,is.promising)
-importFrom(promises,new_promise_domain)
-importFrom(promises,promise_reject)
-importFrom(promises,promise_resolve)
-importFrom(promises,with_promise_domain)
-importFrom(rlang,"%||%")
-importFrom(rlang,"fn_body<-")
-importFrom(rlang,"fn_fmls<-")
-importFrom(rlang,as_function)
-importFrom(rlang,as_quosure)
-importFrom(rlang,check_dots_empty)
-importFrom(rlang,check_dots_unnamed)
-importFrom(rlang,enexpr)
-importFrom(rlang,enquo)
-importFrom(rlang,enquo0)
-importFrom(rlang,enquos)
-importFrom(rlang,enquos0)
-importFrom(rlang,eval_tidy)
-importFrom(rlang,expr)
-importFrom(rlang,fn_body)
-importFrom(rlang,get_env)
-importFrom(rlang,get_expr)
-importFrom(rlang,inject)
-importFrom(rlang,is_false)
-importFrom(rlang,is_missing)
-importFrom(rlang,is_na)
-importFrom(rlang,is_quosure)
-importFrom(rlang,list2)
-importFrom(rlang,maybe_missing)
-importFrom(rlang,missing_arg)
-importFrom(rlang,new_function)
-importFrom(rlang,new_quosure)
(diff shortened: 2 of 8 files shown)


schloerke:
Correction: I originally described this as unreleased. It is not -- the `session$destroy()` / `session$onDestroy()` work shipped in **shiny 1.14.0**, so this affects 1.14.0 and later. Issue and PR descriptions updated.

NEWS now adds a new bullet under `# shiny (development version)` rather than amending the released 1.14.0 entry.

PR #4423, opened 2026-08-10
docs: Stop implying that closing a session destroys its reactives

schloerke:
Docs follow-up to #4421. No behavior change.

## Problem

#4421 made a whole-session close leave reactive values and expressions readable, and documented it — but only on the `onDestroy(callback)` entry of `?session`. The `destroy(namespace = NULL)` entry directly below it was left as-is, and still reads the old way. Its first sentence lists what teardown cleans up:

> Destroys a module session scope (and any descendant scopes) by invoking all registered `onDestroy()` callbacks. This includes cleaning up all **reactive values, observers, and reactive expressions** created within that scope (and any descendant scopes).

and its last sentence says:

> On the root session a `namespace` is required; `session$destroy()` with no `namespace` is an error (**the root session is torn down via `close()`**).

Read in sequence within the one `\item`: destroy tears down reactive values and expressions → the root scope is torn down via `close()` → closing the session tears down its reactive values and expressions. That is precisely the inference #4421 set out to remove, and the corrective paragraph sits in the *previous* entry where a reader of this one won't see it.

Worse inside `ShinySession`: the `onDestroy` R6 docstring got the new paragraph in #4421 while the `destroy` docstring immediately below kept "since the root session is torn down via `close()`", so the two adjacent docstrings now contradict each other.

## Changes

- **`destroy(namespace)`, `?session` + the `ShinySession$destroy` R6 docstring** — scope the teardown list to an explicit `destroy()`, and say what `close()` actually does: observers are torn down, reactive values and expressions are left readable at their last value.

- **The root-destroy error message** (4 copies: `R/shiny.R` × 2, `R/mock-session.R` × 2) — it offered `close()` as the way "to tear down the whole session". This fires exactly when someone is reaching for whole-session teardown, so it is the one place a confused user actually reads; it should not present `close()` as the destroy-everything equivalent. Now: "or call `close()` to end the whole session (which is not a destroy: reactive values and expressions stay readable)."

- **"Module data ownership" section of `?session`** — "A reactive value created inside a module is destroyed with the module" had no anchor to an explicit `destroy()` call, and sits two paragraphs above the sentence that supplies the scope. "With the module" invites the session-close reading, so it is now anchored to `session$destroy()`.

- **`MockShinySession$onDestroy`** — never got #4421's clarification, even though `MockShinySession$close()` calls `invokeDestroyCallbacks(allowRoot = TRUE)` and so fires root destroy callbacks. Its docstring framed close as "when the session scope is destroyed" — the exact conflation #4421 broke — while its `ShinySession` counterpart was fixed. Added, matching that wording.

- Two internal comments that used the same "the root scope can only be torn down via `close()`" framing now say that root destroy callbacks are only *dispatched* from `close()`, where the reactives themselves are left intact.

## Verification

- `roxygen2::roxygenise()` run to regenerate `man/session.Rd` and `man/MockShinySession.Rd`. `NAMESPACE` was reverted: the roxygen2 available here is 8.0.0 against the 8.1.0 this package expects, and it rewrote the whole `importFrom` block — pure version churn, unrelated to this change. Worth a `devtools::document()` on CI to confirm.
- Full `testthat::test_local()` suite passes. The two `session$onDestroy: attempt to apply non-function` warnings in `test-reactivity.r` are pre-existing — confirmed by stashing this branch's changes and re-running.
- Grepped `R/` and `tests/` for the old wording; no remaining occurrences, and no test asserts the changed error string.

The same audit against py-shiny turned up the identical structural problem in its ported `Session.destroy()` docstring, fixed there in posit-dev/py-shiny#2428.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # shiny (development version)
 
+* The `?session` docs no longer imply that closing a session destroys its reactive values and expressions. The `destroy(namespace)` entry lists the reactive state that teardown cleans up and then said the root session "is torn down via `close()`" -- i.e. the behavior #4421 removed. The corrective paragraph added in #4421 only landed on the neighboring `onDestroy()` entry, so the two contradicted each other. (#4422)
+
 * Fixed #4420: closing a session no longer destroys the reactive values and expressions created in it, so async work that outlives the connection does not error. Since 1.14.0, refreshing the page while an `ExtendedTask` (or any promise or `later()` callback) was in flight could raise ``Can't access reactive `<NAME>`; its module session has been destroyed`` once it settled. Reactives are now left readable at their last value on close and reclaimed by garbage collection, while an explicit `session$destroy(namespace)` on a live session still tears them down. (#4421)
 
 * The `session$destroy()` parameter is now documented as `namespace` to match the implemented argument name. (#4419)
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # shiny (development version)
 
-* The `?session` docs no longer imply that closing a session destroys its reactive values and expressions. The `destroy(namespace)` entry lists the reactive state that teardown cleans up and then said the root session "is torn down via `close()`" -- i.e. the behavior #4421 removed. The corrective paragraph added in #4421 only landed on the neighboring `onDestroy()` entry, so the two contradicted each other. (#4422)
+* The `?session` docs no longer imply that closing a session destroys its reactive values and expressions. The `destroy(namespace)` entry lists the reactive state that teardown cleans up and then said the root session "is torn down via `close()`" -- i.e. the behavior #4421 removed. The corrective paragraph added in #4421 only landed on the neighboring `onDestroy()` entry, so the two contradicted each other. (#4423)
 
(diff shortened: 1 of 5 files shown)


NEWS.md
```diff
@@ -1,5 +1,7 @@
 # shiny (development version)
 
+* The `?session` docs no longer imply that closing a session destroys its reactive values and expressions. The `destroy(namespace)` entry lists the reactive state that teardown cleans up and then said the root session "is torn down via `close()`" -- i.e. the behavior #4421 removed. The corrective paragraph added in #4421 only landed on the neighboring `onDestroy()` entry, so the two contradicted each other. (#4423)
+
 * Fixed #4420: closing a session no longer destroys the reactive values and expressions created in it, so async work that outlives the connection does not error. Since 1.14.0, refreshing the page while an `ExtendedTask` (or any promise or `later()` callback) was in flight could raise ``Can't access reactive `<NAME>`; its module session has been destroyed`` once it settled. Reactives are now left readable at their last value on close and reclaimed by garbage collection, while an explicit `session$destroy(namespace)` on a live session still tears them down. (#4421)
```
schloerke:
```suggestion
* Fixed #4420: closing a session no longer destroys the reactive values and expressions created in it, so async work that outlives the connection does not error. Since 1.14.0, refreshing the page while an `ExtendedTask` (or any promise or `later()` callback) was in flight could raise ``Can't access reactive `<NAME>`; its module session has been destroyed`` once it settled. Reactives are now left readable at their last value on close and reclaimed by garbage collection, while an explicit `session$destroy(namespace)` on a live session still tears them down. (#4421, #4423)
```

PR #4426, opened 2026-08-14
Adopt shiny-workflows license-note; attribute bundled lit

### LINKED ISSUE #28: Helper to update `LICENSE.note`
Find updated licenses and concatenate into `LICENSE.note`

schloerke:
Adopts the `license-note` helper added in rstudio/shiny-workflows#69 (which closes rstudio/shiny-workflows#28), so the bundled-JavaScript list in `LICENSE.note` is regenerated by the `routine` workflow instead of maintained by hand.

## What changed

- `LICENSE.note` gains a marker pair. The `routine` workflow rewrites only the region between the markers; the hand-written entries above and all the full license texts below are never touched.
- `package.json` gains `licenseNote.bundled`, naming what actually gets bundled from npm.
- `DESCRIPTION` and `LICENSE.note` gain the missing `lit` attribution — see below.

## Why only two packages are listed

Only `@selectize/selectize` and `lit` reach `inst/` from `node_modules`:

- `tools/updateSelectize.R` copies `node_modules/@selectize/selectize/dist` into `inst/www/shared/selectize`.
- esbuild inlines `lit` into `inst/www/shared/shiny.js`, via `srcts/src/components/errorConsole.ts`. Confirmed by finding `lit-html/lit-html` and `lit-html/is-server` in the built `shiny.js`.

Everything else currently in `LICENSE.note` — jQuery, jQuery UI, Bootstrap, bootstrap-datepicker, ion.rangeSlider, DataTables, strftime, showdown.js, highlight.js, the tar implementation from R — is fetched from a CDN or a GitHub archive by a `tools/update*.R` script rather than from npm, so none of it is discoverable this way. Those entries stay hand-listed outside the markers, which is exactly why the helper only owns a delimited block.

Note that `jquery` and `bootstrap-datepicker` do appear in `devDependencies` but are *not* what ships: `updatejQuery.R` downloads from `code.jquery.com` and `updateBootstrapDatepicker.R` from the upstream GitHub archive, and `jquery` is mapped to `window.jQuery` by `esbuild-plugin-globals` rather than bundled.

## The `lit` gap — worth a careful look

Adopting the helper surfaced that **`lit` is redistributed in `inst/www/shared/shiny.js` but was documented nowhere**: absent from the `LICENSE.note` component list, absent from the full license texts (which the file's own preamble promises are "included below"), and absent from `Authors@R`, where every other bundled library has a `cph` entry.

This PR adds lit's BSD-3-Clause text and a `person(, "Google LLC", role = c("ctb", "cph"), comment = "lit library")` entry, following the pattern the file already uses. **This is a licensing judgement and I would like a maintainer to confirm it** rather than take it on my say-so — both that the attribution is correct and that nothing else bundled is similarly missing. `Authors@R` was checked to still parse (35 people).

Also worth a second opinion: the generator reads `@selectize/selectize` as **Apache-2.0** from its `package.json`, which matches the Apache text already under the `selectize.js` heading here.

## Verification

- The generator is idempotent against this branch: `--check` exits 0.
- `npm ci` (231 packages) then generate reproduces the committed block exactly.

## Sequencing

**This is inert until `shiny-workflows`' `v1` tag moves.** Until then `routine` runs the released `@v1`, which has no `license-note` step, so the markers just sit there as ordinary text. Nothing here breaks in the meantime. To exercise it before merge, point `.github/workflows/R-CMD-check.yaml`'s `routine` job at `@main` temporarily (rstudio/shiny-workflows#69 is merged; only the tag move is outstanding).

Merge order: rstudio/shiny-workflows#69, move `v1`, then this.


schloerke:
Re-checked this against the [best practices linked in the original issue](https://r-pkgs.org/license.html#sec-code-you-bundle) and found one more gap in this file, now fixed.

r-pkgs asks that `LICENSE.note` describe "the overall license of the package, **and** the specific licenses of each individual component". This file only ever did the second half — it opened straight into "The shiny package includes other open source software components" without saying what shiny itself is licensed under. Both `{diffviewer}` (the package r-pkgs holds up as the model) and `{bslib}` open with that statement. Added:

> The shiny package as a whole is distributed under MIT (see the LICENSE file).

Also worth noting for reviewers, since it is the reason a file like this exists: shiny is MIT and bundles Apache-2.0 (`@selectize/selectize`), which carries restrictions MIT does not. That is a normal and disclosable situation rather than a problem, but it does make the overall-license sentence load-bearing rather than decorative.

The `cph` half of the guidance is what the `lit` commit already addresses.

schloerke:
It's not automated well enough. Pausing for now

